Issue #32, opened 2026-08-13
continuum replay --upto N crashes with an unhelpful ProjectionError when the prefix excludes RUN_STARTED


Cyrax321:
- [x] I searched existing issues for a duplicate
- [x] I am reporting this against the correct repository
- [x] I am using a supported Python version (3.11 or later)
- [x] I can reproduce this on the latest commit on `main`

## System information
OS: macOS 15.x (darwin)\nPython: 3.13.5\ncontinuum-agent: 0.1.0 (editable install from source)\nSource commit: a539948 (== origin/main)

## Command or code to reproduce
```python
from continuum.storage.sqlite import SQLiteStorage
from continuum import models, EventType, project
from continuum.models import Run, RunStatus
store = SQLiteStorage(':memory:')
store.create_run(Run(run_id='r1', goal='g', status=RunStatus.RUNNING))
store.append_event('r1', EventType.RUN_STARTED, {'goal':'g'})
store.append_event('r1', EventType.WORK_COMPLETED, {'doc':1})
try:
    project('r1', store.read_events('r1', upto=0))
except Exception as e:
    print(type(e).__name__, e)
```

## Current behaviour
`continuum replay r1 --upto 0` (or any `--upto` that yields a prefix without the `RUN_STARTED` event) calls `project` on a prefix that has no goal event, which raises `ProjectionError: run 'r1' has no goal: the log never recorded RUN_STARTED`. The CLI surfaces this raw error instead of explaining that `--upto` excluded the run's start event.

## Expected behaviour
Either reject `--upto` values that cannot include `RUN_STARTED` with a clear message, or replay relative to the run start so a prefix always contains the goal.

## Proof
Reproduction prints:
```
ProjectionError run 'r1' has no goal: the log never recorded RUN_STARTED
```

PR #50, opened 2026-08-14
fix(cli): make replay actually confirm the stored version (closes #31)

### LINKED ISSUE #31: continuum replay claims to confirm the state matches the stored version but never compares

## System information
OS: macOS 15.x (darwin)\nPython: 3.13.5\ncontinuum-agent: 0.1.0 (editable install from source)\nSource commit: a539948 (== origin/main)

```python
import inspect
from continuum.cli import main
src = inspect.getsource(main.cmd_replay)
print('reads latest_version:' , 'latest_version' in src)
print('compares to stored:' , 'stored' in src and 'latest_version' in src)
```

## Current behaviour
`cmd_replay` (the `continuum replay` command) projects the event log and prints the derived progress. Its docstring says it re-derives state `and confirm it matches the stored version`, but the implementation never reads `storage.latest_version` nor compares the replayed state to anything. The claimed verification does not happen.

## Expected behaviour
Either (a) actually compare the replayed state against `storage.latest_version(run_id)` and report a mismatch, or (b) drop the unimplemented claim from the docstring/help. As written it is a silent no-op that looks like a check.

## Proof
Reproduction prints `reads latest_version: False`, confirming `cmd_replay` never touches the stored version. The docstring at `src/continuum/cli/main.py` (`cmd_replay`) reads: 'Re-derive state from events and confirm it matches the stored version.'

dchaudhari7177:
Closes #31.

`cmd_replay`'s docstring promised to "re-derive state from events and confirm it matches the stored version", but it never read `storage.latest_version` and never compared anything.

I took option (a) — implement the check — rather than dropping the claim, because `continuum replay` is the kind of thing that ends up in a pipeline, and the exit-code contract in this repo is treated as a safety guarantee.

## What it does now

Reports the verdict in text and JSON (new `verified` / `verification` keys) and exits `CORRUPTED` (3) on a mismatch — the code already defined for "stored data failed an integrity check".

```
Replayed 8 events -> 6 completed, 0 decision(s), 0 finding(s)
Verification: matches stored version 0 at sequence 4
```

## The part worth reviewing: which prefix gets compared

My first version compared the replayed state against `latest_version` directly. That reports corruption for **any healthy run that did work after its last checkpoint** — I hit it immediately:

```
stored version 0 source_sequence 4 completed 3
(3, '... 6 completed ...\nVerification: DOES NOT match stored version 0', 'replayed state does not match ...')
```

A stored version is the projection of events *up to its own `source_sequence`*, and the log keeps growing. So verification re-folds **that same prefix**:

```python
prefix = storage.read_events(run_id, upto=stored.source_sequence)
replayed = project(run_id, prefix)
```

That's the invariant `SemanticState` already documents — "folding the same prefix again must yield an equal state" — and `state_fingerprint` already excludes `version`, `created_at`, `updated_at` and `source_sequence`, so it compares meaning rather than bookkeeping.

A consequence worth stating: **verification is independent of `--upto`**, because the prefix comes from the stored version, not from what the caller asked to display. `--upto` narrows the report, not the check. (This also means there's no interaction with #32 / PR #39, which changes `--upto` handling — my diff doesn't touch that path.)

When there is no stored version, that's reported as `skipped (no stored version to compare against)` rather than quietly counted as a pass. Reporting "not checked" honestly is the whole point of the issue.

## Tests

7 added to `tests/test_cli.py`, all failing against pristine `main.py`:

- a consistent run reports a match, in text and in JSON
- **a run with 10 events appended after the last version still reports a match** — the regression that would make this feature worse than useless
- a tampered stored version is detected, in text and in JSON, with the run id on stderr
- no stored version is reported as skipped
- verification is unaffected by `--upto`

The tamper injects a phantom `Finding` rather than nudging `progress`. `Progress` has a cross-field validator (`completed + pending + failed exceeds total`), so a bad number is rejected by `SemanticState` on load and the test would be exercising the model's validator instead of this change — the CLI already exits 3 for that via `CorruptedRecord`, on a different path.

## Gate

Ran all three CI checks from CONTRIBUTING on Windows / Python 3.12:

- `pytest` — 684 passed, 4 skipped (677 before this branch)
- `ruff check src/ tests/` and `ruff format --check src/ tests/` — clean

### DIFF
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -37,6 +37,7 @@ from continuum.models import ActionStatus, EnvironmentSnapshot, EnvResource, Ori
 from continuum.recovery import RecoveryEngine, render_contract
 from continuum.state.diff import diff_states, render_diff
 from continuum.state.semantic import ProjectionError, project
+from continuum.state.versioning import state_fingerprint
 from continuum.storage import (
     CheckpointNotFound,
     CorruptedRecord,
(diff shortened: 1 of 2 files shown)


Cyrax321:
Thank you for this PR, and for making `replay` actually verify the stored version!

The fix is correct: `cmd_replay` now re-derives the stored version's own prefix (`storage.read_events(upto=stored.source_sequence)`) and compares `state_fingerprint`s, exiting `CORRUPTED` on mismatch and reporting verification in both text and JSON output. and the verification is intentionally independent of `--upto`.

- pytest: 704 passed, 9 skipped, **0 failed**. Adds 6 tests (match, JSON, post-checkpoint growth, tamper detection, no-version skip, upto independence).
- ruff format --check src/ tests/: FAILS on 3 files. but all 3 (`tests/test_integration_langchain.py`, `tests/test_integration_langchain_agent.py`, `tests/test_integration_langgraph.py`) are stale copies from this PR's old base (current `main` is format-clean). The PR's own `cli/main.py` / `test_cli.py` are formatted. This is base drift, not a defect in the PR's changes.
- mypy src/continuum --strict: 5 errors in 3 files, all `import-not-found` for uninstalled optional adapters; no new strict errors. (This also confirms the `state_fingerprint` import from `continuum.state.versioning` resolves.)

Code & Test Review Findings
- `cmd_replay` / `_verify_against_stored` (src/continuum/cli/main.py): verified prefix selection from `stored.source_sequence` (not the whole log), the `state_fingerprint` equality check, and that `verified is False` maps to `ExitCode.CORRUPTED`. Tests include a tampered stored version (phantom finding) that must not exit 0, and a post-checkpoint growth case that must still verify. Solid and well-targeted.

Concern (not a code defect): branch is stale; rebasing onto current `main` will also clear the unrelated ruff-format drift. Also note PR #39 also touches `cmd_replay` (adds the `--upto` RUN_STARTED guard), so merging both needs a rebase/ordering pass.

PR #51, opened 2026-08-14
fix(mcp): reject negative progress counters even when total is omitted

### LINKED ISSUE #38: continuum_record_progress: negative completed/failed not rejected before write when total is omitted (poisons event log)

## System information

OS: macOS 15.x (darwin)
Python: 3.13.5
continuum-agent: 0.1.0 (editable install from source)
Source commit: a539948 (== origin/main)

The bug is in `continuum_record_progress` in `src/continuum/mcp/server.py`
(lines ~280-299). The negative-counter guard is nested inside `if total is not None`,
so it is skipped when no total is supplied. Minimal reproduction (faithful copy of
the handler body, using the same `AGENT_SOURCE = Origin.EXTERNAL_AGENT`):

```python
from continuum.models import Run, RunStatus, Origin

store = SQLiteStorage(":memory:")
run_id = "run_neg"
store.create_run(Run(run_id=run_id, goal="g", status=RunStatus.STARTED))

def record_progress(completed, total=None, failed=0):
    payload = {"completed": completed, "failed": failed}
    if total is not None:
        payload["total"] = total
        payload["pending"] = max(total - completed - failed, 0)
    # --- exact copy of the guard in mcp/server.py:290-296 ---
    if total is not None:
        if completed < 0 or failed < 0:
            raise ValueError("progress counters must be non-negative")
        if completed + failed > total:
            raise ValueError("over total")
    # --- end copy ---
    store.append_event(run_id, EventType.TASK_UPDATED, payload,
                       source=Origin.EXTERNAL_AGENT)
    return project(run_id, store.read_events(run_id))

# total given -> correctly rejected before write
try:
    record_progress(completed=-5, total=10)
except ValueError as e:
    print("total=10 rejected before write:", e)

# total omitted -> negative NOT rejected here; bad event written, project fails
try:
    record_progress(completed=-5, total=None)
except Exception as e:
    print("total=None raised at projection:", type(e).__name__)

print("events in log:", [(e.sequence, e.payload) for e in store.read_events(run_id)])
try:
    project(run_id, store.read_events(run_id))
except Exception as e:
    print("log is permanently poisoned:", type(e).__name__)
```

## Current behaviour

With `total=None` and a negative `completed` (or `failed`), the handler does NOT
reject the value before writing. It appends the event `{'completed': -5, 'failed': 0}`
to the event log, then `project()` raises `ProjectionError` (because
`Progress._non_negative` rejects negative counters at projection time). The bad
event is now persisted, so every subsequent `project` / `continuum_checkpoint` /
`continuum_validate` / `continuum_resume` against that run fails permanently.

Output of the reproduction:

```
total=10 rejected before write: progress counters must be non-negative
total=None raised at projection: ProjectionError
events in log: [(1, {'completed': -5, 'failed': 0})]
log is permanently poisoned: ProjectionError
```

## Expected behaviour

The negative-counter check (`completed < 0 or failed < 0`) must run regardless of
whether `total` was supplied, so the bad event is rejected *before* it is written
to the log. The handler's own comment states the intent: "Reject impossible
counters before anything is written ... checking here keeps the bad value out of
the event log in the first place." That intent is only met when `total` is given.

The fix is to move the `completed < 0 or failed < 0` validation out of the
`if total is not None:` block (it does not depend on `total`); only the
`completed + failed > total` check needs `total`.

## Logs or CI run

Reproduction output shown above, run directly against `a539948`. The existing
MCP tests (`tests/test_mcp_server.py`) exercise `continuum_record_progress` but
pass `total` in the negative-case tests, so the `total=None` gap is not covered.

lesbass:
## Summary

Fixes #38 — negative progress counters were not rejected before write when `total` was omitted, which let a `{'completed': -5, 'failed': 0}` event poison the run's event log and permanently break every later projection/checkpoint/resume.

## Change

In `continuum_record_progress`, the negative-counter guard (`completed < 0 or failed < 0`) was nested inside `if total is not None:`, so it was skipped when `total` was omitted. The check does not depend on `total`, so it is hoisted out; only the `completed + failed > total` check remains gated on `total`.

## Tests

- Added `test_negative_progress_is_rejected_before_being_written_without_total` in `tests/test_mcp_server.py`, parametrized over `completed=-5` and `failed=-1` with `total` omitted, asserting the tool raises and no `TASK_UPDATED` event is written.

## Verification

- `pytest tests/test_mcp_server.py` — 38 passed
- `pytest --no-cov` (full suite) — 679 passed, 4 skipped

### DIFF
--- a/src/continuum/mcp/server.py
+++ b/src/continuum/mcp/server.py
@@ -287,13 +287,10 @@ def build_server(
         # is intact yet unprojectable would be poisoned permanently. The
         # `Progress` model enforces this at projection time; checking here keeps
         # the bad value out of the event log in the first place.
-        if total is not None:
-            if completed < 0 or failed < 0:
-                raise ValueError("progress counters must be non-negative")
-            if completed + failed > total:
-                raise ValueError(
-                    f"completed ({completed}) + failed ({failed}) exceeds total ({total})"
-                )
+        if completed < 0 or failed < 0:
+            raise ValueError("progress counters must be non-negative")
+        if total is not None and completed + failed > total:
+            raise ValueError(f"completed ({completed}) + failed ({failed}) exceeds total ({total})")
         ctx.storage.append_event(run_id, EventType.TASK_UPDATED, payload, source=AGENT_SOURCE)
 
         state = project(run_id, ctx.storage.read_events(run_id))
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -185,6 +185,36 @@ async def test_over_total_progress_is_rejected_before_being_written(
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "negative_update",
+    [
+        {"completed": -5},
+        {"completed": 0, "failed": -1},
+    ],
+)
+async def test_negative_progress_is_rejected_before_being_written_without_total(
+    server_ctx: tuple[Any, Any],
+    negative_update: dict[str, int],
+) -> None:
+    """Negative counters must be rejected even when `total` is omitted, or the
+    bad event is written and every later projection/checkpoint/resume fails
+    permanently (issue #38)."""
+    from mcp.server.mcpserver.exceptions import ToolError
+
+    server, ctx = server_ctx
+    ctx.storage.create_run(Run(run_id="run_1", goal="g"))
+    arguments = {"run_id": "run_1", "goal": "g", **negative_update}
+    with pytest.raises(ToolError, match="must be non-negative"):
+        await server.call_tool(
+            "continuum_record_progress",
+            arguments,
+            context=_ctx(TEST_CLIENT),
+        )
+    # Nothing beyond the start event was written: the log is not poisoned.
+    assert [e.type.value for e in ctx.storage.read_events("run_1")] == ["RUN_STARTED"]
+
+
+@pytest.mark.asyncio
 async def test_recording_progress_for_an_unknown_run_without_a_goal_fails(
     server_ctx: tuple[Any, Any],
 ) -> None:


Cyrax321:
Thank you for this PR, and for hardening the MCP progress validation!

The fix is correct: it moves the non-negative counter check out of the `if total is not None` guard in `continuum_record_progress`, so a negative `completed` or `failed` is rejected even when `total` is omitted, which prevents a bad event from poisoning the log permanently.

CI Check Analysis (I ran the suite locally on the branch; these mirror what CI runs)
- pytest: 676 passed, 7 skipped, **0 failed**. Note: this branch is based on an older `main`, so it collects fewer tests than current `main` (697/16); no failures are introduced by the change.
- ruff format --check src/ tests/: passed on the files present (the branch predates the langchain/langgraph integration tests, so fewer files exist, but nothing this PR touches is unformatted).
- mypy src/continuum --strict: 4 errors in 2 files (`openai`, `langgraph` adapters). identical to this branch's own baseline, all `import-not-found` for uninstalled optional adapters. No new strict errors from this PR.

Code & Test Review Findings
- `continuum_record_progress` (src/continuum/mcp/server.py): verified the negative check now runs unconditionally and the over-total check is correctly gated on `total is not None`. Logic is right.
- Test `test_negative_progress_is_rejected_before_being_written_without_total` (parametrized over `{"completed": -5}` and `{"completed": 0, "failed": -1}`) asserts `ToolError` "must be non-negative" and that only `RUN_STARTED` is written. meaningfully covers the fix and the "log is not poisoned" guarantee.

Concern (not a code defect): the branch is stale relative to `main`. Please rebase onto current `main` before merge so CI runs against the full suite. No code change required.

PR #52, opened 2026-08-14
fix(extractor): collapse ids repeated within a single LLM proposal (closes #41)

### LINKED ISSUE #41: LLMExtractor._merge double-adds duplicate ids within a single proposal
## System information
- OS: macOS 15.x (darwin)
- Python: 3.13.5

```python
from continuum.state.extractor import DeterministicExtractor, LLMExtractor, LLMProposal, ExtractionContext
from continuum.models import Origin
import uuid

def ev(etype, payload):
    return Event(event_id=uuid.uuid4().hex, run_id="r", sequence=1, type=etype,
                 timestamp="2026-01-01T00:00:00Z", payload=payload, prev_hash="", hash="",
                 source=Origin.DETERMINISTIC)

ctx = ExtractionContext(run_id="r", trajectory=[ev(EventType.RUN_STARTED, {"goal": "g"})])
det = DeterministicExtractor()

def llm_dups(_c, _s):
    return LLMProposal(decisions=[
        {"decision_id": "X", "decision": "first"},
        {"decision_id": "X", "decision": "second"},
    ], findings=[
        {"finding_id": "Y", "claim": "a"},
        {"finding_id": "Y", "claim": "b"},
    ])

ext = LLMExtractor(llm_dups, base=det, enabled=True)
out = ext.extract(ctx)
print([d.decision_id for d in out.decisions])  # ['X', 'X']
print([f.finding_id for f in out.findings])     # ['Y', 'Y']
```

## Current behaviour
`extract` returns a state containing two decisions with `decision_id == "X"` and two findings with `finding_id == "Y"` (counts X=2, Y=2).

## Expected behaviour
`_merge` de-duplicates against `known_decisions` / `known_findings` / `known_work` (extractor.py:144, 161, 179), but those sets are built once from the *base* state before the loop and are never updated as items are appended. So a duplicate id *within a single proposal* is not collapsed: the second `X` is not present in the original `known_decisions` and is appended again. A single LLM response containing a repeated id should yield at most one component with that id.

dchaudhari7177:
Closes #41.

`_merge` seeds `known_decisions` / `known_findings` / `known_work` from the **base state** and never grows them, so an id the LLM emitted twice in one response passed the guard twice and was appended twice.

## Fix

One line per collection — add the accepted id to its set as it's appended:

```python
if not decision_id or decision_id in known_decisions:
    continue  # never overwrite a recorded fact
known_decisions.add(decision_id)
decisions.append(...)
```

That makes the guard that stops a proposal overwriting a recorded fact also stop it duplicating *itself*. All three collections had the same hole, so all three are fixed — `pending_work` isn't in the issue's repro but has identical code, and fixing two of three would leave the same bug behind under a different name.

**First occurrence wins.** That's the direction the existing semantics already run — the base state beats the proposal, so earlier beats later — and it's asserted rather than left implicit.

## Tests

6 added to `tests/test_extractor.py`.

**Three fail against pristine `extractor.py`:**

```
FAILED test_a_repeated_decision_id_within_one_proposal_is_collapsed
FAILED test_a_repeated_finding_id_within_one_proposal_is_collapsed
FAILED test_a_repeated_task_id_within_one_proposal_is_collapsed
```

**Three are regression guards** that must pass both before and after, and do:

- `test_the_first_occurrence_wins_within_a_proposal` — pins *which* copy survives, not just that one does
- `test_distinct_ids_in_one_proposal_are_all_kept` — the de-dup must not collapse genuinely different components, which is how this kind of fix usually goes wrong
- `test_a_proposal_repeating_a_recorded_id_still_cannot_overwrite_it` — `d1` is already recorded; repeating it twice must neither overwrite nor duplicate, i.e. the existing protection still holds through the new code path

## Gate

Windows / Python 3.12, all three CI checks from CONTRIBUTING:

- `pytest` — **683 passed**, 4 skipped (677 before this branch)
- `ruff check src/ tests/` and `ruff format --check src/ tests/` — clean

### DIFF
--- a/src/continuum/state/extractor.py
+++ b/src/continuum/state/extractor.py
@@ -147,6 +147,10 @@ class LLMExtractor:
             decision_id = str(raw.get("decision_id", ""))
             if not decision_id or decision_id in known_decisions:
                 continue  # never overwrite a recorded fact
+            # Also guards against a repeat later in this same proposal: the set is
+            # seeded from the base state, so without this the second copy of an id
+            # the LLM emitted twice would not be recognised as already handled.
+            known_decisions.add(decision_id)
             decisions.append(
                 Decision(
                     decision_id=decision_id,
@@ -164,6 +168,7 @@ class LLMExtractor:
             finding_id = str(raw.get("finding_id", ""))
             if not finding_id or finding_id in known_findings:
                 continue
+            known_findings.add(finding_id)
             confidence = float(raw.get("confidence", 0.5))
             findings.append(
                 Finding(
@@ -182,6 +187,7 @@ class LLMExtractor:
             task_id = str(raw.get("task_id", ""))
             if not task_id or task_id in known_work:
                 continue
+            known_work.add(task_id)
             pending.append(
                 PendingWork(
                     task_id=task_id,
--- a/tests/test_extractor.py
+++ b/tests/test_extractor.py
@@ -133,6 +133,103 @@ def test_proposals_without_ids_are_dropped() -> None:
 # --- composition ----------------------------------------------------------- #
 
 
+def test_a_repeated_decision_id_within_one_proposal_is_collapsed() -> None:
+    """The known-id set is seeded from the base state; it must also grow as we append."""
+    log = build_log()
+    extractor = LLMExtractor(
+        lambda ctx, state: LLMProposal(
+            decisions=[
+                {"decision_id": "X", "decision": "first"},
+                {"decision_id": "X", "decision": "second"},
+            ]
+        )
+    )
+    state = extractor.extract(context(log))
+    assert [d.decision_id for d in state.decisions].count("X") == 1
+
+
+def test_a_repeated_finding_id_within_one_proposal_is_collapsed() -> None:
+    log = build_log()
+    extractor = LLMExtractor(
+        lambda ctx, state: LLMProposal(
+            findings=[
+                {"finding_id": "Y", "claim": "a"},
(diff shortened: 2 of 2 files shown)


Cyrax321:
Thank you for this PR, and for closing the intra-proposal duplicate-id gap in the extractor!

The fix is correct: it seeds each `known_*` id set as proposals are appended (via `known_decisions.add` / `known_findings.add` / `known_work.add` in `_merge`), so a duplicate id the LLM emits twice within a single proposal is collapsed instead of appended a second time. while recorded (base-state) ids still win.

- pytest: 680 passed, 7 skipped, **0 failed**. Note: stale branch (fewer tests than current `main`); no failures introduced.
- ruff format --check src/ tests/: 75 files already formatted (branch predates the langchain/langgraph integration tests; this PR's own files are clean).
- mypy src/continuum --strict: 4 errors in 2 files (`openai`, `langgraph`), matching this branch's baseline; all `import-not-found` for uninstalled optional adapters. No new strict errors.

Code & Test Review Findings
- `_merge` (src/continuum/state/extractor.py): verified all three loops (`decisions`, `findings`, `pending_work`) now record the id into its `known_*` set after the first occurrence, so the second copy is skipped and a recorded id cannot be overwritten. Six new tests cover repeated ids per component, first-occurrence-wins, distinct ids kept, and recorded-id hijack resistance. Strong, specific coverage.

Concern (not a defect): branch is stale relative to `main`; please rebase before merge. Also note PR #56 also touches `extractor.py`. merge/rebase the two in a deliberate order.

PR #53, opened 2026-08-14
fix(adapters): keep a result dict holding the envelope key intact on cache hit (closes #44)

### LINKED ISSUE #44: intercept_action returns divergent value on cache hit when result dict contains reserved key '__return_value__'
## System information
- OS: macOS 15.x (darwin)
- Python: 3.13.5

```python
from continuum.models import Run, RunStatus, Origin
from continuum.adapters.generic import GenericAgentAdapter

store = SQLiteStorage(":memory:")
run_id = "r"
store.create_run(Run(run_id=run_id, goal="g", status=RunStatus.STARTED))
store.append_event(run_id, EventType.RUN_STARTED, {"goal": "g"}, source=Origin.DETERMINISTIC)
adapter = GenericAgentAdapter(store)

calls = {"n": 0}
def action():
    calls["n"] += 1
    return {"__return_value__": 42, "other": "payload"}

first = adapter.intercept_action(run_id, "act.collide", action, arguments={"k": 1})
second = adapter.intercept_action(run_id, "act.collide", action, arguments={"k": 1})
print(first, second, first != second)
# {'__return_value__': 42, 'other': 'payload'}  42  True
```

## Current behaviour
The first (fresh) call executes `action_fn` and returns the **whole** dict `{"__return_value__": 42, "other": "payload"}`. The second (cached) call returns only `42` — the value of the reserved key — and silently discards `"other": "payload"`. `calls["n"]` is 1, confirming the second call is a pure cache hit.

## Expected behaviour
A completed action must return the same value on every call. The cached path (generic.py:116-120) special-cases the reserved key `__return_value__`, returning `res["__return_value__"]`, while the fresh path (generic.py:131,133) stores and returns the whole dict. When a caller's real result is a dict that happens to contain that key, the two paths disagree. The cached path should return the entire stored result (matching the fresh path), or the reserved key should be documented/forbidden so callers cannot trigger the collision.

dchaudhari7177:
Closes #44.

`intercept_action` stores a non-dict result as `{"__return_value__": val}` because the ledger records results as mappings, and the cached path unwraps that key. A caller whose real result was a dict **already containing** the key was stored as-is — so the fresh call returned the whole dict and the cached call returned only that one member, silently discarding the rest.

## Fix

The fresh path wraps such a dict too:

```python
needs_envelope = not isinstance(val, dict) or RESULT_ENVELOPE_KEY in val
result_dict = {RESULT_ENVELOPE_KEY: val} if needs_envelope else val
```

The presence of the key in a *stored* result now always means "envelope" and never "caller data", so both paths agree and exactly one level is ever unwrapped.

The issue's repro before and after:

```
first : {'__return_value__': 42, 'other': 'payload'}
second: {'__return_value__': 42, 'other': 'payload'}
equal : True | calls: 1
```

## Why this shape and not the two the issue offers

**"The cached path should return the entire stored result."** That breaks the scalar wrapping the envelope exists for — `42` would come back as `{"__return_value__": 42}` on the second call. It fixes one divergence by creating another.

**"The reserved key should be forbidden."** Validation could only run *after* `action_fn` has executed, since that's when the result exists. The side effect would already have happened and we'd raise instead of returning its result — for an action ledger whose whole purpose is not losing track of effects that occurred, that's worse than the bug.

The chosen fix also leaves **already-stored ledgers meaning exactly what they meant before**: nothing changes about how an ordinary dict or a wrapped scalar is persisted, only the previously-ambiguous case. That matters for a durable log.

The literal is now a named `RESULT_ENVELOPE_KEY`, exported alongside the adapter, since the invariant lives in two places that have to agree — them drifting apart is precisely this bug.

## Tests

4 added to `tests/test_adapters.py`. **Three fail against pristine `generic.py`:**

```
FAILED test_a_result_dict_holding_the_envelope_key_survives_the_cache
FAILED test_a_result_dict_that_is_only_the_envelope_key_survives_the_cache
FAILED test_a_nested_envelope_key_survives_the_cache
```

- the issue's exact reproduction
- the degenerate case where the caller's dict is *only* the envelope key, so it's indistinguishable from an envelope by shape alone
- a nested `{"__return_value__": {"__return_value__": "deep"}}`, proving exactly one level unwraps

The fourth is a **regression guard**: an ordinary dict is still stored unwrapped. It asserts against the *ledger record* rather than the return value, because the return value would look right either way — that's the assertion that stops the envelope quietly starting to wrap everything.

## Gate

Windows / Python 3.12, all three CI checks:

- `pytest` — **681 passed**, 4 skipped
- `ruff check` / `ruff format --check` — clean

### DIFF
--- a/src/continuum/adapters/generic.py
+++ b/src/continuum/adapters/generic.py
@@ -17,7 +17,13 @@ from continuum.models import (
 from continuum.recovery.engine import RecoveryDecision, RecoveryEngine
 from continuum.storage.base import Storage
 
-__all__ = ["GenericAgentAdapter"]
+#: Key under which a non-dict action result is stored, since the ledger records
+#: results as mappings. A caller's own dict is wrapped in the same envelope when
+#: it contains this key, so the presence of the key in a stored result always
+#: means "envelope" and never "caller data".
+RESULT_ENVELOPE_KEY = "__return_value__"
+
+__all__ = ["GenericAgentAdapter", "RESULT_ENVELOPE_KEY"]
 
 
 class GenericAgentAdapter(AgentAdapter):
@@ -115,8 +121,12 @@ class GenericAgentAdapter(AgentAdapter):
 
         if not outcome.fresh:
             res = outcome.result
-            if isinstance(res, dict) and "__return_value__" in res:
-                return res["__return_value__"]
+            # One level of unwrapping, matching the one level the fresh path
+            # below applies. A stored dict that carries the key is always an
+            # envelope, never a caller's own dict, because the fresh path wraps
+            # those too.
+            if isinstance(res, dict) and RESULT_ENVELOPE_KEY in res:
+                return res[RESULT_ENVELOPE_KEY]
             return res
 
         try:
@@ -128,7 +138,13 @@ class GenericAgentAdapter(AgentAdapter):
             ledger.fail(outcome.key, f"{type(exc).__name__}: {exc}", certain=False)
             raise
 
-        result_dict = val if isinstance(val, dict) else {"__return_value__": val}
+        # A non-dict has to be wrapped to be stored. So does a dict that happens
+        # to contain the envelope key: storing that one as-is would make the
+        # cached path unwrap the caller's own dict and return only that member,
+        # so a completed action would return a different value on its second
+        # call than on its first.
+        needs_envelope = not isinstance(val, dict) or RESULT_ENVELOPE_KEY in val
+        result_dict = {RESULT_ENVELOPE_KEY: val} if needs_envelope else val
         ledger.complete(outcome.key, result=result_dict)
         return val
 
--- a/tests/test_adapters.py
+++ b/tests/test_adapters.py
@@ -94,6 +94,82 @@ def test_intercept_action_handles_scalar_return_value(store: SQLiteStorage) -> N
     assert call_count == 1  # Deduplicated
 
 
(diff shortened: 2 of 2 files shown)


Cyrax321:
Thank you for this PR, and for fixing the cache-hit return-value collision in the generic adapter!

The fix is correct: it introduces `RESULT_ENVELOPE_KEY` and now also wraps a result dict that already contains that key (not just non-dicts), so a cached hit returns the same value on every call rather than silently unwrapping to only one member. It also replaces the inline magic string with a named constant.

- pytest: 678 passed, 7 skipped, **0 failed**. Stale branch; no failures introduced.
- ruff format --check src/ tests/: 75 files already formatted (branch predates the integration tests; this PR's own files are clean).
- mypy src/continuum --strict: 4 errors in 2 files (`openai`, `langgraph`), matching this branch's baseline; all `import-not-found` for uninstalled optional adapters. No new strict errors.

Code & Test Review Findings
- `GenericAgentAdapter.intercept_action` (src/continuum/adapters/generic.py): verified `needs_envelope = not isinstance(val, dict) or RESULT_ENVELOPE_KEY in val` and the one-level unwrap on the cache-hit path. Four new tests cover the envelope-key collision, the bare-envelope case, a nested envelope, and an ordinary dict stored unwrapped. Good coverage, and the named constant is a nice cleanup.

Concern (not a defect): branch is stale relative to `main`; please rebase before merge.

PR #56, opened 2026-08-15
fix(extractor): fall back to deterministic state on malformed LLM proposal

### LINKED ISSUE #40: LLMExtractor: malformed LLM proposal crashes extract() instead of falling back
## System information
- OS: macOS 15.x (darwin)
- Python: 3.13.5

```python
from continuum.state.extractor import DeterministicExtractor, LLMExtractor, LLMProposal, ExtractionContext
from continuum.models import Origin
import uuid

def ev(etype, payload):
    return Event(event_id=uuid.uuid4().hex, run_id="r", sequence=1, type=etype,
                 timestamp="2026-01-01T00:00:00Z", payload=payload, prev_hash="", hash="",
                 source=Origin.DETERMINISTIC)

ctx = ExtractionContext(run_id="r", trajectory=[ev(EventType.RUN_STARTED, {"goal": "g"})])
det = DeterministicExtractor()

def llm_bad(_c, _s):
    return LLMProposal(findings=[{"finding_id": "f1", "confidence": "high"}])

ext = LLMExtractor(llm_bad, base=det, enabled=True)
ext.extract(ctx)   # raises ValueError, recovery breaks
```

## Current behaviour
`LLMExtractor.extract` raises `ValueError: could not convert string to float: 'high'`. Any proposal the LLM returns that is malformed (bad `confidence` type, or a value that fails `Decision`/`Finding`/`PendingWork` construction in `_merge`) propagates out of `extract` and aborts extraction.

## Expected behaviour
The module docstring and the `# noqa: BLE001` on the `_llm` call state enrichment "must never break recovery" and "If it raises, extraction falls back to the deterministic result rather than failing the run." The `try/except` in `extract` (extractor.py:133-137) only wraps `self._llm(context, state)`. It does NOT wrap `self._merge(state, proposal)` (extractor.py:139), so a successfully-returned but malformed proposal crashes extraction, defeating the documented fallback guarantee.

lesbass:
## Checklist

- [x] CI is passing

## What does this PR do?

Moves the `self._merge(state, proposal)` call inside the existing `try/except` in `LLMExtractor.extract` (`src/continuum/state/extractor.py`) so a successfully-returned but malformed `LLMProposal` falls back to the deterministic result instead of raising out of `extract`.

## Why is this change needed?

Fixes #40. The `LLMExtractor` docstring and the `# noqa: BLE001` on the `_llm` call guarantee enrichment "must never break recovery": "If it raises, extraction falls back to the deterministic result rather than failing the run." The `try/except` only wrapped `self._llm(context, state)`, so a proposal that passed the call but failed during `_merge` (e.g. `confidence: "high"`, or any failure in `Decision`/`Finding`/`PendingWork` construction) propagated a `ValueError` out of `extract` and aborted recovery. Wrapping `_merge` in the same guard records the error in `self.last_error` and returns the base state.

## Tests

- Added `test_a_malformed_proposal_degrades_to_the_deterministic_state` in `tests/test_extractor.py`, reproducing the reported repro (a `Finding` whose `confidence` is the string `"high"`), asserting `extract()` returns the deterministic state and `last_error` is a `ValueError`.

## Verification

- `pytest tests/test_extractor.py` — 12 passed
- `ruff check src/continuum/state/extractor.py tests/test_extractor.py` — clean

Note: `mypy src/continuum` cannot run in this environment (a third-party `numpy` stub fails to parse under Python 3.13 against the repo's `python_version = "3.11"` setting); the change is a re-indent with no type-surface change.

### DIFF
--- a/src/continuum/state/extractor.py
+++ b/src/continuum/state/extractor.py
@@ -132,11 +132,10 @@ class LLMExtractor:
 
         try:
             proposal = self._llm(context, state)
+            return self._merge(state, proposal)
         except Exception as exc:  # noqa: BLE001 - enrichment must never break recovery
             self.last_error = exc
             return state
-
-        return self._merge(state, proposal)
 
     def _merge(self, state: SemanticState, proposal: LLMProposal) -> SemanticState:
         provenance = Provenance(origin=Origin.LLM, extractor=self.name)
--- a/tests/test_extractor.py
+++ b/tests/test_extractor.py
@@ -98,6 +98,19 @@ def test_a_failing_llm_degrades_to_the_deterministic_state() -> None:
     assert isinstance(extractor.last_error, RuntimeError)
 
 
+def test_a_malformed_proposal_degrades_to_the_deterministic_state() -> None:
+    log = build_log()
+
+    def malformed(ctx: ExtractionContext, state: object) -> LLMProposal:
+        return LLMProposal(findings=[{"finding_id": "f1", "confidence": "high"}])
+
+    extractor = LLMExtractor(malformed)
+    state = extractor.extract(context(log))
+
+    assert state == project("run_1", log.events("run_1"))
+    assert isinstance(extractor.last_error, ValueError)
+
+
 def test_a_disabled_llm_extractor_never_calls_out() -> None:
     calls: list[int] = []
 


Cyrax321:
Thank you for this PR, and for making the extractor resilient to malformed LLM proposals!

The fix is correct: it moves `return self._merge(state, proposal)` inside the existing `try` in `LLMExtractor.extract`, so a malformed proposal that raises during `_merge` now falls back to the deterministic `state` instead of crashing, while still recording `last_error`.

- pytest: 698 passed, 16 skipped, **0 failed**. Adds 1 regression test.
- ruff format --check src/ tests/: 86 files already formatted (clean).
- mypy src/continuum --strict: 5 errors in 3 files, all `import-not-found` for uninstalled optional adapters (`openai`/`langgraph`/`langchain`); no new strict errors from this PR.

Code & Test Review Findings
- `LLMExtractor.extract` (src/continuum/state/extractor.py): verified the merge is now inside the `try`; the `except` already guards enrichment so recovery is never broken. The new test `test_a_malformed_proposal_degrades_to_the_deterministic_state` confirms the returned state equals the deterministic projection and that `last_error` is `ValueError`. Good coverage of the fallback path.

Merge note: PR #52 also modifies `src/continuum/state/extractor.py` (the `_merge` method). Both are clean individually, but please merge/rebase them in a deliberate order to avoid a conflict in that file.

LGTM, thank you for the contribution!

Issue #27, opened 2026-08-12
Migrate release workflow to uv (keep PyPI trusted publishing)


Cyrax321:
### What problem does this solve?

`release.yml` installs the build tool with `pip install hatch` and the package with `pip install -e ".[dev]"`. PyPI Trusted Publishing via OIDC is already in place (`id-token: write` plus `pypa/gh-action-pypi-publish@release/v1`), so the remaining modernization is the build/install path. The `pip install hatch` step reads as dated in 2026, and `uv` is the current default for build and publish workflows.

### Proposed solution

Migrate the release workflow to `uv` (`astral-sh/setup-uv` or `uv-org/setup-uv` for the runner) for the install/build path, keeping the existing trusted-publishing publish step unchanged.

### Alternatives considered

Leaving the workflow as-is. Works today, but the visible tooling lag is the kind of detail a reviewer notices.

### Is this a breaking change?

No

Cyrax321:
Done via #70 (merged). Release workflow migrated to uv: uv run --extra dev pytest and uv build, with the pypa trusted-publisher step and OIDC permissions untouched.

PR #63, opened 2026-08-19
ci(release): migrate release workflow to uv (#27)

abyyxhek:
## Checklist

- [ ] Tests added or updated for the change
- [ ] Documentation updated, if the change affects it
- [ ] CI is passing
- [ ] For security-relevant changes, the PR description states any known limitations explicitly

## Summary by CodeRabbit

* **Chores**
  * Updated the release build process to use uv for Python environment setup, testing, and package creation.
  * Simplified the release workflow by removing separate package installation steps.
  * Artifact upload, publishing, and release behavior remain unchanged.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,22 +16,17 @@ jobs:
     steps:
       - uses: actions/checkout@v7.0.1
 
-      - name: Set up Python 3.12
-        uses: actions/setup-python@v7.0.0
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v5
         with:
+          enable-cache: true
           python-version: "3.12"
-          cache: pip
-
-      - name: Install build tools
-        run: pip install hatch
 
       - name: Run tests before release
-        run: |
-          pip install -e ".[dev]"
-          pytest --tb=short -q
+        run: uv run pytest --tb=short -q
 
       - name: Build wheel and sdist
-        run: hatch build
+        run: uv build
 
       - name: Upload build artifacts
         uses: actions/upload-artifact@v7.0.1


Cyrax321 (changes requested):
Thank you for contributing to CONTINUUM, and thanks for picking up a `good first issue`. The direction here is right, and dropping the `pip install hatch` bootstrap is exactly what #27 asked for. There is one blocking problem in the test step that would break the release pipeline, so I am requesting changes. Details below.

**Status:** Changes requested
**Reviewed at:** 359e80b
**Verified on:** local reproduction of both workflow steps in a clean checkout

---

### 1. Scope of the issue

For clarity on what this PR should and should not cover:

**In scope:** `.github/workflows/release.yml` only. Specifically the install and build path (`pip install hatch`, `pip install -e ".[dev]"`, `hatch build`) migrating to `uv`.

**Out of scope:** The PyPI trusted publishing step (`pypa/gh-action-pypi-publish@release/v1`) and the `id-token: write` permission are already correct and must stay untouched. You have correctly left these alone. `.github/workflows/ci.yml` is also out of scope and should remain on pip in this PR; migrating CI is a separate change with a three-version test matrix to consider.

---

### 2. What is correct in this PR

* `uv build` is a valid replacement for `hatch build`. I confirmed it honors the `hatchling` backend through PEP 517 and emits both artifacts:

  ```
  Successfully built dist/continuum_agent-0.1.0.tar.gz
  Successfully built dist/continuum_agent-0.1.0-py3-none-any.whl
  ```

* Removing the separate `pip install hatch` step is correct. `uv build` provisions the backend itself.
* The `publish-pypi` and `github-release` jobs are unchanged, as required.

---

### 3. Blocking issue: `uv run pytest` installs no test dependencies

`uv run` syncs the project's runtime dependencies and PEP 735 dependency groups. It does **not** install `[project.optional-dependencies]` extras. This repository declares all test tooling (`pytest`, `pytest-asyncio`, `hypothesis`, `mcp`) under `[project.optional-dependencies].dev`, and defines no `[dependency-groups]` and no `[tool.uv]` table. The result is that `uv run pytest` resolves an environment containing only `pydantic`:

```
$ uv pip list
continuum-agent    0.1.0
pydantic           2.13.4
annotated-types / pydantic-core / typing-extensions / typing-inspection
```

This fails in one of two ways depending on the runner image.

**Case A, no `pytest` on `PATH`.** The step aborts immediately:

```
$ uv run pytest --tb=short -q
Installed 6 packages in 8ms
error: Failed to spawn: `pytest`
  Caused by: No such file or directory (os error 2)
```

**Case B, some `pytest` present on `PATH`.** This is the worse outcome. `uv run` falls through to the ambient interpreter's `pytest` and runs it against the under-provisioned virtualenv, which produces a misleading import failure rather than a clear provisioning error:

```
$ uv run which pytest
/opt/homebrew/bin/pytest

src/continuum/mcp/server.py:39: from mcp.server import MCPServer
E   ModuleNotFoundError: No module named 'mcp'
Interrupted: 6 errors during collection
```

**Impact.** Both `publish-pypi` and `github-release` declare `needs: build`. This does not merely fail one step, it blocks the entire release on the next `v[0-9]+.[0-9]+.[0-9]+` tag push, and it does so at the moment a release is being cut.

**Please note that CI passing on this PR is not evidence of correctness.** `release.yml` triggers only on tag push, so no job in this PR's check run ever executes the file you changed. The green `Test (Python 3.11 / 3.12 / 3.13)` checks come from `ci.yml`, which this PR does not modify. The failing `Ruff - format check` is pre-existing breakage in `src/continuum/serve/server.py` and is unrelated to your change, so please disregard it. It is being fixed separately in #62.

---

### 4. Required changes

**4.1 Install the dev extra in the test step.** Change:

```yaml
      - name: Run tests before release
        run: uv run pytest --tb=short -q
```

to:

```yaml
      - name: Run tests before release
        run: uv run --extra dev pytest --tb=short -q
```

I verified this fix locally against your branch. It provisions the full dev set and the suite passes: `797 passed, 4 skipped`.

**4.2 Pin `setup-uv` to an exact version, and update it.** You have `astral-sh/setup-uv@v5`. The current release is `v10.0.1`, so this is five majors behind. Separately, every action in this repository is pinned to an exact version (`actions/checkout@v7.0.1`, `actions/setup-python@v7.0.0`, `actions/download-artifact@v8.0.1`), so a floating major tag is inconsistent with the existing convention. Please use:

```yaml
      - name: Set up uv
        uses: astral-sh/setup-uv@v10.0.1
        with:
          enable-cache: true
          python-version: "3.12"
```

For reference, the resulting `build` job steps should read:

```yaml
      - uses: actions/checkout@v7.0.1

      - name: Set up uv
        uses: astral-sh/setup-uv@v10.0.1
        with:
          enable-cache: true
          python-version: "3.12"

      - name: Run tests before release
        run: uv run --extra dev pytest --tb=short -q

      - name: Build wheel and sdist
        run: uv build

      - name: Upload build artifacts
        uses: actions/upload-artifact@v7.0.1
        with:
          name: dist
          path: dist/
```

---

### 5. Not required, but please note

These are follow-ups we will handle separately. Do not expand this PR to cover them.

* **No committed `uv.lock`.** `enable-cache: true` has little to cache without a lockfile, and each release resolves dependencies fresh, so release builds are not reproducible. This is not a regression against the previous pip setup, so it does not block this PR. We will decide on committing the lockfile in a separate issue.
* **Migrating `dev` to `[dependency-groups]`.** Moving dev tooling out of `[project.optional-dependencies]` and into PEP 735 `[dependency-groups]` would let plain `uv run pytest` work as you originally wrote it, and is the more idiomatic uv layout. That is a `pyproject.toml` change affecting `ci.yml` and contributor setup instructions, so it belongs in its own PR. Use `--extra dev` here.
* **`uv build` writes a `dist/.gitignore`** that `hatch build` did not. `files: dist/*` will not match a dotfile so the GitHub release is unaffected, and we have not observed a problem with the publish step. Flagging it only so it is on the record.

---

### 6. Verification before you push

Please confirm both steps locally in a clean checkout, since CI will not exercise this file:

```bash
uv run --extra dev pytest --tb=short -q   # expect: 797 passed, 4 skipped
uv build                                  # expect: sdist + wheel in dist/
```

Also please fill in the PR description. The template's "What does this PR do?" and "Why is this change needed?" sections are still empty, and the body should reference the issue as `Fixes #27`.

Once 4.1 and 4.2 are in, re-request review and I will merge. Thanks again for the contribution, this is a useful cleanup and the build half of it is already correct.

Cyrax321:
Superseded by #70, which migrates the release workflow to uv on the canonical repo (the contributor's fork could not be pushed to). The trusted publisher and OIDC permissions are preserved. Thank you for the initial patch.

Issue #81, opened 2026-08-19
continuum benchmark crashes on Windows: unclosed SQLiteStorage handles block TemporaryDirectory cleanup


abyyxhek:
### Before reporting

- [x] I searched existing issues for a duplicate
- [x] I am reporting this against the correct repository
- [x] I am using a supported Python version (3.11 or later)
- [x] I can reproduce this on the latest commit on `main`

### System information

```
OS: Windows 11 Home Single Language (10.0.26200)
Python: 3.11.15
continuum-agent: 0.1.0 (from source, commit 15ab5e6)
Installed via: uv sync --extra dev
```

### Command or code to reproduce

```shell
# 1. The public CLI command crashes outright
continuum benchmark --total 10

# 2. The test suite fails
pytest -q
```

### Current behaviour

**1. `continuum benchmark` crashes with an unhandled `PermissionError`.**

The command prints its scenario table, then dies on a traceback while tearing down its temp directory:

```
  File "src\continuum\benchmark\__init__.py", line 369, in run_idempotency_benchmark
    with TemporaryDirectory() as tmp:
  ...
PermissionError: [WinError 32] The process cannot access the file because it is
being used by another process: '...\Temp\tmpst4qqy1j\idem_replay.db'
```

Root cause: `run_idempotency_benchmark` (`src/continuum/benchmark/__init__.py:375`) and
`_run_one` (`src/continuum/benchmark/__init__.py:166`) both do:

```python
store = SQLiteStorage(str(db))
```

and never close it. The handle is still open when the enclosing `TemporaryDirectory()`
tries to remove the file. POSIX permits unlinking an open file, so this is silent on
Linux/macOS; Windows refuses it and the whole command fails. `Storage` already
implements `__enter__`/`__exit__` (`src/continuum/storage/base.py:109`) and the rest of
the codebase consistently uses `with SQLiteStorage(...) as store:` — these two call
sites are the only ones that don't.

This is a genuine resource leak independent of platform. Windows just makes it fatal
instead of invisible.

**2. Five tests fail on Windows.**

```
5 failed, 788 passed, 4 skipped
FAILED tests/test_benchmark.py::test_idempotency_benchmark_proves_issue_6 - PermissionError
FAILED tests/test_cli.py::test_benchmark_runs_and_reports_numbers - PermissionError
FAILED tests/test_cli.py::test_the_cli_runs_as_a_module_without_warnings
[...]

PR #89, opened 2026-08-19
fix(benchmark): close SQLiteStorage handles so continuum benchmark runs on Windows

abyyxhek:
## Checklist

- [x] Documentation updated, if the change affects it — `CHANGELOG.md` under `Unreleased` → `Fixed`
- [x] CI is passing
- [x] For security-relevant changes, the PR description states any known limitations explicitly — not security-relevant; limitations noted below

## What does this PR do?

Fixes two independent causes behind `continuum benchmark` crashing and five tests failing on Windows.

**1. Unclosed `SQLiteStorage` handles (`src/continuum/benchmark/__init__.py`)**

`_run_one` and `run_idempotency_benchmark` both did `store = SQLiteStorage(str(db))` and never closed it. Both write their databases inside a `TemporaryDirectory()`, so cleanup ran while the `.db` files were still open:

```
  File "src/continuum/benchmark/__init__.py", line 369, in run_idempotency_benchmark
    with TemporaryDirectory() as tmp:
PermissionError: [WinError 32] The process cannot access the file because it is
being used by another process: '...\Temp\tmpst4qqy1j\idem_replay.db'
```

Both call sites now use `with SQLiteStorage(...) as store:`. `Storage` already implements `__enter__`/`__exit__` (`src/continuum/storage/base.py:109`), and every other call site in the codebase already uses the context manager — these two were the only exceptions.

**2. Replaced subprocess environment (`tests/test_cli.py`)**

The `_cli` helper replaced the entire environment with a hardcoded POSIX `PATH`:

```python
env={"PYTHONPATH": ..., "PATH": "/usr/bin:/bin"}
```

Dropping `SystemRoot` leaves Windows unable to initialise Winsock, so every spawned interpreter died during startup — `OSError: [WinError 10106]` on `import _overlapped` — long before reaching the CLI. The assertions then saw empty stdout and exit code 1 rather than the behaviour under test. It now inherits the parent environment and overrides only `PYTHONPATH`, preserving the original intent (import continuum from `src/`, not an installed copy).

## Why is this change needed?

The leak is a genuine resource-management bug independent of platform — POSIX permits unlinking an open file, so it was invisible on Linux and macOS rather than absent. Windows refuses it, which turned a silent leak into a hard failure of a documented public command.

CI runs `ubuntu-latest` only (`.github/workflows/ci.yml:16`), which is why neither problem surfaced, even though `CONTRIBUTING.md` documents Windows setup (`.venv\Scripts\activate`). The practical cost is that a contributor on Windows cannot get a clean baseline, so their own regressions are indistinguishable from pre-existing failures.

Addresses #81.

## Tests / validation

Windows 11, Python 3.11.15, branched from `15ab5e6`.

| Check | Before | After |
| --- | --- | --- |
| `pytest` | `5 failed, 788 passed, 4 skipped` | **`835 passed, 4 skipped`** |
| `ruff check src/ tests/` | pass | **pass** |
| `ruff format --check src/ tests/` | pass | **pass** |
| `mypy src/continuum` | pass | **pass** |
| `continuum benchmark --total 10` | `PermissionError` traceback | **completes, prints results** |

The five previously-failing tests were `test_benchmark.py::test_idempotency_benchmark_proves_issue_6` and, in `test_cli.py`, `test_benchmark_runs_and_reports_numbers`, `test_the_cli_runs_as_a_module_without_warnings`, `test_the_shell_pipeline_short_circuits_on_unsafe_state`, and `test_report_and_hint_appear_in_the_right_order`.

The two totals are not directly comparable — the `788` baseline was measured before rebasing onto current `main`, which added tests of its own. The meaningful comparison is that the same five named failures are now passing and nothing regressed.

**New regression test:** `test_benchmark_releases_every_storage_handle`, parametrized over both harnesses. It asserts every opened handle was closed rather than asserting on the platform error, so it fails on POSIX too if the `with` blocks are ever dropped. I verified this rather than assuming — against the unfixed source it fails on both parametrizations; against the fixed source both pass.

## Reviewing the diff

Wrapping `_run_one` in `with` re-indents its body, so `src/continuum/benchmark/__init__.py` shows ~136 moved lines. **`git diff -w` collapses this to the two `with` statements plus two `ruff format` line wraps** that the extra indent pushed over the 100-char limit — those two are format-only and semantically identical. No logic changed.

## Limitations and follow-ups

- This fixes the two causes I could reproduce. I cannot verify macOS.
- It does **not** add Windows to the CI matrix, so this can silently regress. That felt like a separate decision about CI cost and is out of scope here — happy to open a follow-up issue, or add `windows-latest` to `ci.yml` in this PR if you'd prefer it bundled.
- `src/continuum/mcp/server.py` also constructs `SQLiteStorage` without a `with` (lines 136, 150, 162), but those hand the storage to a longer-lived caller rather than dropping it on the floor inside a temp directory, so they are not the same bug. I left them alone to keep this scoped.
- Happy to split this into two PRs (source leak vs. test harness) if you'd rather review them separately.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Windows benchmark issues by ensuring database resources are properly closed.
  - Improved command-line test environment handling across platforms.
- **Tests**
  - Added regression coverage verifying benchmark database resources are closed.
- **Documentation**
  - Documented the Windows benchmark and test environment fixes in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,19 @@ All notable changes to this project are documented here. The format follows
 
 ### Fixed
 
+- **`continuum benchmark` crashed on Windows from unclosed database handles
+  (issue #81).** `_run_one` and `run_idempotency_benchmark` constructed
+  `SQLiteStorage` without ever closing it, so the enclosing
+  `TemporaryDirectory()` still held open `.db` files at cleanup. POSIX allows
+  unlinking an open file, so this was an invisible resource leak on Linux and
+  macOS; Windows refuses it, and the whole command died on an unhandled
+  `PermissionError`. Both call sites now use `with SQLiteStorage(...) as store:`,
+  matching every other call site in the codebase. `tests/test_cli.py::_cli` also
+  replaced the subprocess environment with a hardcoded POSIX `PATH`, dropping
+  `SystemRoot` and leaving spawned interpreters unable to initialise Winsock on
+  Windows; it now inherits the parent environment and overrides only
+  `PYTHONPATH`. Together these fixed five tests that failed on Windows.
+
 - **Three defects found by an adversarial audit of the MCP surface**, driven over
   the live stdio protocol with every tool result verified against the SQLite store
   rather than taken at its word. Method and per-claim results in `test.md`:
--- a/src/continuum/benchmark/__init__.py
+++ b/src/continuum/benchmark/__init__.py
@@ -163,131 +163,134 @@ def _run_one(method: str, spec: ScenarioSpec, total: int, workdir: Path) -> Meth
     workdir.mkdir(parents=True, exist_ok=True)
     db = workdir / "agent.db"
     effects = workdir / "effects.log"
-    store = SQLiteStorage(str(db))
-    run_id = "run_bench"
-    env_v3 = capture_environment(run_id, StaticProvider(dataset="v3"))
+    with SQLiteStorage(str(db)) as store:
+        run_id = "run_bench"
+        env_v3 = capture_environment(run_id, StaticProvider(dataset="v3"))
 
-    # --- phase 1: first half of the run, up to the crash ------------------- #
-    crash_at = max(1, int(total * spec.crash_frac))
-    side_at = int(total * spec.side_frac)
-    side_at = crash_at if spec.interrupted else min(side_at, crash_at - 1)
+        # --- phase 1: first half of the run, up to the crash ------------------- #
+        crash_at = max(1, int(total * spec.crash_frac))
+        side_at = int(total * spec.side_frac)
+        side_at = crash_at if spec.interrupted else min(side_at, crash_at - 1)
 
-    store.create_run(Run(run_id=run_id, goal="Process documents"))
-    store.append_event(run_id, EventType.RUN_STARTED, {"goal": "Process documents", "total": total})
-    store.append_event(
-        run_id, EventType.DEPENDENCY_DECLARED, {"resource": "dataset", "version": "v3"}
-    )
-    store.append_event(
-        run_id,
-        EventType.EVIDENCE_ADDED,
-        {"evidence_id": "paper_128", "summary": "study", "source": "dataset"},
-    )
-    store.append_event(
-        run_id,
-        EventType.FINDING_ADDED,
-        {
-            "finding_id": "finding_17",
-            "claim": "X holds",
-            "evidence": ["paper_128"],
-            "confidence": 0.91,
-        },
-    )
-
-    ledger = ActionLedger(store, run_id)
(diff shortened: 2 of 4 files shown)


Cyrax321:
Thank you for this contribution, Abishek. This is a clean, well scoped fix and the reasoning is sound.

What the PR does:
- Closes two leaked SQLiteStorage handles in the benchmark harness (_run_one and run_idempotency_benchmark) by wrapping them in `with` blocks, matching every other call site in the codebase. The leak was invisible on POSIX because unlinking an open file is permitted, but it hard failed `continuum benchmark` on Windows with a PermissionError during TemporaryDirectory cleanup.
- Fixes the `_cli` test helper, which replaced the subprocess environment with a hardcoded POSIX PATH and dropped SystemRoot, breaking every spawned interpreter on Windows before the CLI was ever reached. It now inherits the parent environment and overrides only PYTHONPATH.
- Adds a regression test (`test_benchmark_releases_every_storage_handle`) that asserts handles are actually closed rather than asserting on the platform error, so it also guards against regression on POSIX.

What we did:
- Ran `ruff check`, `ruff format --check`, and `mypy` for verification.

What we got:
- `pytest`: 835 passed, 4 skipped, including the previously Windows only failing tests and the new regression test.
- `ruff check` and `ruff format --check`: both pass.
- `mypy`: 20 errors, all pre-existing on main and confined to adapters and the MCP server, none introduced by this change.

The diff is larger than the change because wrapping the functions re indents their bodies and triggers two ruff line wraps, but `git diff -w` confirms the only semantic change is the two `with` statements. The fix is correct and complete.

I have marked this ready to pull and applied the bug, benchmark, and cli labels.

One follow up worth considering separately: CI still runs ubuntu only, so this class of platform break can regress silently. Adding windows latest to the matrix would close that gap. Happy to take that as a follow up if you would rather keep this PR scoped.

Thanks again for the contribution. The added regression coverage is exactly the kind of thing we want to see, and we appreciate the care you took to make the test fail on POSIX too.

Cyrax321:
Merged. Thank you again for the contribution, Abishek.

The fix is in main as ea8c686. It correctly closes the two leaked SQLiteStorage handles, fixes the `_cli` environment handling for Windows, and ships a regression test that guards the behavior on POSIX as well. Verified locally with the full suite (835 passed, 4 skipped) and ruff clean.

We appreciate you taking the time to scope this tightly and to make the regression test fail on every platform, not just Windows. That kind of rigor is exactly what keeps the codebase healthy as it grows.

Closing this out. The Windows CI matrix follow up remains a good idea for a separate PR whenever you or someone else has the bandwidth.

Thanks for the solid work.

Issue #84, opened 2026-08-19
Auto-resume: eliminate the self-certified to request_human to confirm tax on every resume


Cyrax321:
Part of the auto-resume handoff. See #82 (problem P3).

A run created purely via MCP tools is self-certified; on resume continuum_resume returns mode=request_human until a REVIEW_CONFIRMED event exists. So every resume currently requires an extra continuum_confirm tool call + an extra model turn before the agent may continue. This is a mandatory per-resume latency/token tax.

Design question: is it safe to auto-confirm self-certified runs being resumed by the same client? Should the agent be treated as the operator? Or is a different resume mode needed that does not require REVIEW_CONFIRMED for agent-originated runs? What is the minimal, safe change?

Cyrax321:
Add docs/research/confirm_tax.md proposing scoped confirm and same client auto confirm to remove the mandatory per resume tax for self certified runs. Verified with CI green.

PR #92, opened 2026-08-19
fix(serve): let a sidecar client resume without a memorized id or task file

### LINKED ISSUE #91: continuum serve: sidecar resume omits the run goal and cannot target the active run
### Before reporting

### System information

```
OS: Windows 11 (26200)
Python: 3.11.15
continuum-agent: 0.1.0 (commit dc400f5)
Installed via: uv pip install -e ".[dev]"
```

```shell
python - <<'PY'
from continuum.serve import SidecarServer
from continuum.serve.server import BadParams

srv = SidecarServer(database=":memory:")
srv.dispatch("record_progress", {"run_id": "guide", "completed": 3, "total": 10,
                                 "goal": "migrate the billing module to the new API"})
srv.dispatch("checkpoint", {"run_id": "guide"})

payload = srv.dispatch("resume", {"run_id": "guide"})
print("goal in payload? ->", "goal" in payload)

try:
    srv.dispatch("resume", {})           # what a fresh session would send
except BadParams as exc:
    print("resume without run_id ->", exc)
PY
```

### Current behaviour

```
goal in payload? -> False
resume without run_id -> missing parameter 'run_id'
```

The `continuum serve` sidecar's `resume` method has drifted from the
`continuum_resume` MCP tool it is documented to mirror, in two ways:

1. **The run goal is not returned.** `_decision_payload` in
   `src/continuum/serve/server.py` omits `goal`, so a sidecar client gets
   `mode` and `completed/total` but never learns *what the task was*.
2. **`run_id` is mandatory.** `_h_resume` calls `_require(params, "run_id")`,
   so the sidecar cannot target the active interrupted run the way
   `continuum_resume` and `continuum resume` both can.

I checked the whole surface rather than just this method, and `resume` is the
only one that has drifted. Diffing the payload keys of every paired method:

```
record_progress  in MCP only: []
checkpoint       in MCP only: []
resume           in MCP only: ['goal']      <-- the gap
validate         in MCP only: []
list_actions     in MCP only: []
```

(`resume` also returns `checkpoint_version`, `validation_reason` and
`environment_changes` that the MCP tool does not. Those look like deliberate
extra diagnostics, so I have left them alone.)

This matters because of what the two capabilities are for. PR #80 added `goal`
to `continuum_resume` precisely so a resumed session needs no external task
file, and `CLAUDE.md` now states it as a rule:

> Do not read or write any task file, the task is the run's `goal`, which
> `continuum_resume` returns, so a resumed session already knows what to
> continue.

PR #75 made `run_id` optional for the same reason, so "a fresh session needs no
memorized id". Both landed on the MCP tool and the CLI, and neither reached the
sidecar. The sidecar is the boundary specifically intended for clients that
cannot embed Python or the `mcp` SDK, so today a non-Python agent is the one
client that still cannot do a hands-free resume: it has to remember the
`run_id` across the interruption and keep its own task file to recall the goal.
That is the exact overhead those two PRs removed everywhere else.

For reference, the same run through the MCP tool:

```
MCP resume keys: ['contract', 'contract_text', 'goal', 'mode', ...]
MCP 'goal' -> 'migrate the billing module to the new API'
MCP resume, NO run_id -> run_id: guide | goal: 'migrate the billing module to the new API'
```

and through the CLI:

```
$ continuum --db demo.db --json resume guide
  "goal": "migrate the billing module to the new API",
  "mode": "request_human",
  "run_id": "guide",
```

### Expected behaviour

The sidecar's `resume` should mirror `continuum_resume`, because the module
docstring states that as an invariant:

> The protocol mirrors the MCP tool surface so the two stay in sync.

and `tests/test_serve.py` is written around the same property ("the same
operations the MCP server exposes are reachable ... from any process"). So:

1. `resume` returns the run's `goal`.
2. `run_id` is optional; when omitted the sidecar targets the most recently
   active run, and reports `mode: "no_active_run"` when there is none, matching
   what `continuum_resume` returns in that case rather than raising `BadParams`.

Happy to send a PR. The fix is small and additive (a new key plus an optional
parameter, no change to existing keys), and I would add a parity regression
test that compares the sidecar's `resume` keys against the MCP tool's so this
class of drift fails CI next time rather than being found by hand.

### Logs or CI run

```
--- GAP 1: goal missing from serve resume payload ---
keys: ['checkpoint_version', 'contract', 'contract_text', 'environment_changes',
       'mode', 'next_allowed_action', 'progress', 'rationale', 'repairs',
       'report', 'run_id', 'safe', 'uncertain_actions', 'validation_reason']
has 'goal'? -> False

--- GAP 2: serve resume cannot target the active run ---
omitted run_id raised BadParams: missing parameter 'run_id'
```

abyyxhek:
## Checklist

- [x] CI is passing (Lint & Type-check, and Test on Python 3.11 / 3.12 / 3.13, all green. The Vercel check reports `Authorization required to deploy`, which is a deploy-preview permission on your side rather than anything in this diff.)

## What does this PR do?

Brings the `continuum serve` sidecar's `resume` back in line with the
`continuum_resume` MCP tool it is documented to mirror. Two changes in
`src/continuum/serve/server.py`:

1. The payload now includes the run `goal`.
2. `run_id` is optional. Omitted, it targets the most recently active run;
   when there is none, it returns `mode: "no_active_run"` (the same shape
   `continuum_resume` returns) instead of a `bad_params` error.

Before and after, same run, through the sidecar:

```
# before
keys: [..., 'mode', 'next_allowed_action', 'progress', ...]   # no 'goal'
resume without run_id -> BadParams: missing parameter 'run_id'

# after
goal -> 'migrate the billing module to the new API'
resume without run_id -> run_id: guide | goal: 'migrate the billing module to the new API'
```

Additive by construction: no existing key changed value or type, so a current
client keeps working untouched. The serve-only diagnostics
(`checkpoint_version`, `validation_reason`, `environment_changes`) are
deliberately left alone; parity in the direction that matters is "the sidecar
exposes everything the MCP tool does".

Five regression tests in `tests/test_serve.py`, each verified failing on the
unfixed source first:

| test | what it pins |
| --- | --- |
| `test_resume_returns_the_run_goal` | `goal` is in the payload |
| `test_resume_without_run_id_targets_the_active_run` | omitted id resolves the active run |
| `test_resume_without_run_id_reports_no_active_run` | empty database reports the mode, not an error |
| `test_resume_still_requires_review_for_a_self_reported_run` | trust gate unchanged |
| `test_stdio_resume_needs_no_params_at_all` | the wire shape a restarted client sends |

The fourth is the parity guard: rather than asserting a hardcoded key list, it
diffs the sidecar's `resume` keys against the *live* `continuum_resume` payload
and asserts the MCP surface is a subset. A field added on one side and
forgotten on the other now fails CI instead of being found by hand, which is
how `goal` went missing. It skips when the `mcp` extra is absent, since
`continuum serve` is designed not to need it.

Docs: `CHANGELOG.md` under `[Unreleased] / Fixed`. No other doc changes, because
there is no per-method sidecar reference page; the module docstring is the spec
and already promises the parity this restores, so it needed no edit.

## Why is this change needed?

Fixes #91.

The sidecar's module docstring states the invariant plainly:

> The protocol mirrors the MCP tool surface so the two stay in sync.

and `tests/test_serve.py` is written around the same property ("the same
operations the MCP server exposes are reachable ... from any process"). Two
capabilities added on the MCP side never reached it: the run `goal` (#80) and
the optional `run_id` (#75). Both landed on `continuum_resume` and on the CLI,
and neither on the sidecar. I diffed the payload keys of every paired method to
check the blast radius, and `resume` is the only one that had drifted, with
`goal` the only missing field:

```
record_progress  in MCP only: []
checkpoint       in MCP only: []
resume           in MCP only: ['goal']      <-- the gap
validate         in MCP only: []
list_actions     in MCP only: []
```

What makes this worth fixing rather than a cosmetic inconsistency is *which*
client it hurt. The sidecar is the boundary for processes that cannot embed
Python or the `mcp` extra, and those two capabilities are precisely what makes
a resume hands-free. `CLAUDE.md` now states the intended contract as a rule:

> Do not read or write any task file, the task is the run's `goal`, which
> `continuum_resume` returns, so a resumed session already knows what to
> continue.

A non-Python agent could not honour that. It had to remember the `run_id`
across the very interruption it was recovering from, and keep its own task file
to recall the goal, which is the exact overhead #75 and #80 removed everywhere
else. So the one surface built for foreign-language clients was the one surface
that still could not do the thing those clients most need.

## Known limitations

Flagging these explicitly since `resume` sits next to the trust boundary:

- **Trust behaviour is unchanged.** Returning the goal is read-only and
  confirms nothing, so a self-certified (agent-reported) run still resolves to
  `request_human` and still needs `confirm`. `test_resume_still_requires_review_for_a_self_reported_run`
  pins that. This PR deliberately does not touch the auto-confirm question in
  #84 and takes no position on it.
- **`goal` is a self-report** from a remote caller, recorded with
  `Origin.EXTERNAL_AGENT` like every other sidecar write. Surfacing it in the
  resume payload does not make it verified, and a client should not treat it as
  such. This matches what `continuum_resume` already does.
- **The no-active-run branch returns a short payload** (`run_id`, `mode`,
  `safe`, `message`) with no `progress` or `contract`, mirroring
  `continuum_resume`. A client must branch on `mode` before indexing the rest.
  The alternative would have been a protocol-level error, which the sidecar can
  express and the MCP tool cannot; I chose parity so one client path works
  across both transports, and raised it as an open question on #91 in case you
  would rather have the error.
- **Auth is untouched.** `resume` is not in `MUTATING`, and `dispatch` already
  verifies the shared secret for every method regardless, so the optional
  `run_id` does not widen what an unauthenticated caller can reach.

## Verification

```
843 passed, 4 skipped
ruff check src/ tests/          All checks passed!
ruff format --check src/ tests/ 97 files already formatted
mypy src/continuum              Success: no issues found in 55 source files
```

One unrelated failure on my machine,
`tests/test_mcp_server.py::test_main_reports_an_unopenable_database_instead_of_a_traceback`,
is pre-existing and Windows-only: the assertion compares a `WindowsPath` string
against a message whose backslashes have been escaped, so it cannot match. It
fails identically with my changes stashed, so it is out of scope here. Happy to
open a separate issue for it if it is not already known.

## Summary by CodeRabbit

- **New Features**
  - Resume responses now include the active run’s goal.
  - Resume can be called without specifying a run ID; it automatically targets the active run.
  - When no active run exists, the response clearly reports `no_active_run` instead of returning an error.

- **Bug Fixes**
  - Improved consistency between sidecar and MCP resume responses.
  - Preserved review requirements for runs that still need human confirmation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,29 @@ All notable changes to this project are documented here. The format follows
 
 ### Fixed
 
+- **The `continuum serve` sidecar's `resume` had drifted from the MCP tool it
+  mirrors, so a non-Python client could not resume hands-free (issue #91).** The
+  module docstring promises "the protocol mirrors the MCP tool surface so the two
+  stay in sync", but two capabilities added to `continuum_resume` never reached
+  the sidecar: the run `goal` in the payload (PR #80) and an optional `run_id`
+  that targets the most recently active run (PR #75). A sidecar client therefore
+  learned `mode` and `completed/total` but never what the task *was*, and
+  `_h_resume` raised `bad_params` on the omitted `run_id` that an interrupted
+  session has no way to supply. The sidecar is the boundary intended for clients
+  that cannot embed Python or the `mcp` extra, so it was the one surface still
+  requiring an external task file and a memorized id, the exact overhead those
+  two changes removed for the MCP and CLI paths. `resume` now returns `goal` and
+  accepts an absent `run_id`, reporting `mode: "no_active_run"` (matching
+  `continuum_resume`) rather than a protocol error when there is nothing to
+  resume. Additive: no existing key changed, and the serve-only diagnostics
+  (`checkpoint_version`, `validation_reason`, `environment_changes`) are
+  untouched. Trust behaviour is unchanged, since returning a self-reported goal
+  confirms nothing and a self-certified run still resolves to `request_human`.
+  `tests/test_serve.py` gains four regression tests, including one that diffs the
+  sidecar's `resume` keys against the live `continuum_resume` payload so the next
+  field added on one side and forgotten on the other fails CI instead of being
+  found by hand.
+
 - **MCP server was not found at cold start because its name did not match the
   configured name (issue #87).** `.mcp.json` registered the server under the key
   `continuum`, and `build_server` advertised `MCPServer(name="continuum")`, while
--- a/src/continuum/serve/server.py
+++ b/src/continuum/serve/server.py
@@ -260,12 +260,13 @@ def _write(stream: TextIO, payload: dict[str, Any]) -> None:
 # --------------------------------------------------------------------------- #
 
 
-def _decision_payload(decision: Any) -> dict[str, Any]:
+def _decision_payload(decision: Any, *, goal: str) -> dict[str, Any]:
     report = decision.validation.report
     return {
         "checkpoint_version": report.checkpoint_version,
         "validation_reason": report.reason,
         "run_id": decision.state.run_id,
+        "goal": goal,
         "mode": decision.mode.value,
         "safe": decision.safe,
         "next_allowed_action": decision.next_allowed_action,
@@ -374,13 +375,34 @@ def _h_validate(server: SidecarServer, params: dict[str, Any]) -> dict[str, Any]
 
 
 def _h_resume(server: SidecarServer, params: dict[str, Any]) -> dict[str, Any]:
-    run_id = _require(params, "run_id")
+    # run_id is optional, exactly as it is on continuum_resume: a session that
+    # was interrupted and restarted has no id to send, so an omitted one means
+    # "the run that was interrupted". Reported as a mode rather than a protocol
+    # error, so a client branches on mode alone across both transports.
+    run_id = params.get("run_id")
+    if not run_id:
+        active = server.storage.get_active_run()
+        if active is None:
+            return {
+                "run_id": None,
+                "mode": "no_active_run",
+                "safe": False,
+                "message": (
+                    "No active run to resume. Start one with "
+                    "record_progress(run_id, completed, total, goal=...)."
+                ),
+            }
+        run_id = active.run_id
     decision = server.adapter.resume(
         run_id,
         current_environment=_environment(run_id, params.get("env")),
         expected_model=params.get("expected_model"),
     )
-    return _decision_payload(decision)
+    # The goal travels with the decision so a resumed client knows what to
+    # continue without keeping its own task file. Read-only: returning the
+    # self-reported goal confirms nothing, so a self-certified run still comes
+    # back as request_human.
+    return _decision_payload(decision, goal=server.storage.get_run(run_id).goal)
 
 
 def _h_confirm(server: SidecarServer, params: dict[str, Any]) -> dict[str, Any]:
--- a/tests/test_serve.py
+++ b/tests/test_serve.py
@@ -69,6 +69,114 @@ def test_checkpoint_and_resume_round_trip() -> None:
     assert decision["progress"]["completed"] == 1
 
 
+# --- resume mirrors the MCP surface (issue #91) ------------------------------
+
+
+def test_resume_returns_the_run_goal() -> None:
+    """A resumed client must learn what the task was, not just how far it got.
+
+    Without this the sidecar is the one boundary that still needs an external
+    task file to answer "what was I doing", which is the overhead the goal was
+    added to continuum_resume to remove.
+    """
+    srv = make_server()
+    srv.dispatch(
+        "record_progress",
+        {"run_id": "r1", "completed": 3, "total": 10, "goal": "migrate the billing module"},
+    )
+    srv.dispatch("checkpoint", {"run_id": "r1"})
+
+    decision = srv.dispatch("resume", {"run_id": "r1"})
+    assert decision["goal"] == "migrate the billing module"
+
+
+def test_resume_without_run_id_targets_the_active_run() -> None:
+    """An interrupted session has no id to send, so omitting it must work.
+
+    This is the whole point of the capability: a fresh process that lost its
+    memory of the run can still ask what to continue.
+    """
+    srv = make_server()
+    srv.dispatch(
+        "record_progress",
+        {"run_id": "r1", "completed": 3, "total": 10, "goal": "migrate the billing module"},
+    )
+    srv.dispatch("checkpoint", {"run_id": "r1"})
+
+    decision = srv.dispatch("resume", {})
+    assert decision["run_id"] == "r1"
+    assert decision["goal"] == "migrate the billing module"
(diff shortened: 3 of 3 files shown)


Cyrax321:
Thank you for this contribution, Abishek. This is a clean, well scoped fix and the reasoning is sound.

What the PR does:
- Adds the run goal to the sidecar resume payload. _decision_payload now takes goal as a keyword argument, and _h_resume passes server.storage.get_run(run_id).goal, matching exactly how continuum_resume fetches and returns it (mcp/server.py:558). A sidecar client now learns what the task was, not only how far it got.
- Makes run_id optional in _h_resume. When it is omitted, the handler falls back to server.storage.get_active_run(), and when there is none it returns mode: "no_active_run" with safe: False and the same message continuum_resume uses, instead of raising BadParams. A restarted foreign client can therefore resume with no memorized id and no task file, the exact overhead PR #75 and PR #80 removed for the other transports.
- Adds four regression tests plus a parity guard. The guard diffs the sidecar resume keys against the live continuum_resume payload (skipped when the mcp extra is absent) rather than asserting a hardcoded list, so the next field added on one side and forgotten on the other fails CI instead of being found by hand. That is precisely the mechanism that would have caught this drift.

What we did:

What we got:
- pytest: 844 passed, 4 skipped, including all five new serve regression tests.

The change is additive by construction: no existing key changed value or type, and the serve only diagnostics are untouched. Trust behaviour is preserved, which test_resume_still_requires_review_for_a_self_reported_run pins: returning the self reported goal confirms nothing, so a self certified run still resolves to request_human. We also confirmed serve_stdio defaults a missing params object to an empty dict (server.py:234), so a request carrying no params key at all still resolves the active run, which the wire-loop test exercises.

Thanks again for the contribution. The choice to mirror continuum_resume verbatim, including matching the no_active_run shape rather than inventing a sidecar specific error, is the right call for keeping the two surfaces in sync, and the parity test is exactly the kind of guard we want. We appreciate the care you took to verify each new test fails on the unfixed source first.

Cyrax321:
Merged. Thank you again for the contribution, Abishek.

The fix is in main as 0df49f8. The sidecar resume now returns the run goal and accepts an optional run_id that targets the most recently active run, reporting mode: "no_active_run" when there is none, so a foreign client can resume hands-free with no memorized id and no task file. Verified locally with the full suite (844 passed, 4 skipped) and ruff clean.

We appreciate the rigor here: mirroring continuum_resume verbatim rather than inventing a sidecar specific error, keeping the change additive so existing clients are untouched, and adding the parity guard that diffs the sidecar resume keys against the live MCP payload so this class of drift fails CI next time instead of being found by hand. That is exactly the discipline we want on the recovery path.

This closes issue #91. Thanks for the solid work.

abyyxhek:
The test count in the CHANGELOG entry was correct to flag: the change added six regression tests, not four. This had already merged by the time the review landed, so I have corrected it on `main` in #96 rather than amending this branch.

PR #93, opened 2026-08-19
fix(mcp): report a missing 'mcp' extra instead of an import traceback (#87)

Adhi1-2:
## The gap

`Unreleased` already carries two #87 cold-start fixes: the server-name mismatch, and the handle leak plus `ValueError`/`sqlite3.Error` reporting in `main`. This is a third failure in the same family that neither of them can reach.

The `mcp` SDK is an optional extra — `pyproject.toml` even documents it as such ("The MCP server is optional: the core library and CLI never import it"). But `[project.scripts]` installs the `continuum-mcp` console script **unconditionally**. So a plain `pip install continuum` puts an entry point on `PATH` whose dependency is absent, and `mcp/server.py` imported the SDK at module scope:

```python
from mcp.server import MCPServer
from mcp.server.mcpserver.context import Context
from mcp.types import ToolAnnotations
```

The process therefore dies **during import**, before the `initialize` handshake. `main`'s existing handlers cannot help: they are inside `main`, and `main` never gets to run if the module never finished importing. The operator sees the client report that the server never became ready, while the traceback goes to a stderr log they are not looking at — the exact dead end #87 is about.

This also meant `import continuum.mcp` raised, because `mcp/__init__.py` re-exports from `server`. Verified against the pre-fix tree: the new test fails with `ModuleNotFoundError: No module named 'mcp'` raised from `continuum/mcp/__init__.py", line 3`.

## The fix

- Move the three SDK imports into `build_server`. Every runtime use (`MCPServer(...)`, both `ToolAnnotations(...)` calls, and `Context` as a value in the `guard` closure) is already inside that function, so they are reachable as closure variables with no other change. The only module-scope reference is `build_server`'s return annotation, which `from __future__ import annotations` leaves as a string — a `TYPE_CHECKING` import keeps mypy satisfied.
- Translate the failure in `main` into the same one-line `error:` form the sibling failures use, naming the install command:

  ```
  error: the MCP server needs the optional 'mcp' dependency, which is not
  importable (No module named 'mcp'). Install it with: pip install 'continuum[mcp]'
  ```

- **Narrowed to the SDK itself.** If `exc.name` is not `mcp` or an `mcp.*` submodule, it re-raises. A missing transitive dependency of some *other* package (`pydantic`, `anyio`) keeps its traceback rather than being misreported as a missing extra, which would send the operator after the wrong fix. `exc.name is None` also re-raises: if we cannot tell what is missing, we do not guess.
- `python -m continuum.mcp` now reaches `main` too, instead of dying in `__main__.py`'s import.
- The imports go at the very top of `build_server`, before `load_policy()` and storage, so #90's "a server that never started created a database" invariant still holds — the new test asserts it.

## Tests

Both in `tests/test_mcp_server.py`, next to their siblings and following the same naming.

- `test_main_reports_a_missing_mcp_extra_instead_of_a_traceback` — a real subprocess with a `meta_path` finder blocking `mcp`. Subprocess rather than `monkeypatch`, because blocking an already-imported package in-process would corrupt import state for every later test. It asserts both halves of the fix: that `from continuum.mcp.server import main` *survives* the import, and that `main` then exits 1 with a named cause and an install command. Also asserts `stdout` is empty — over stdio the client parses that stream as protocol frames, so diagnostics must never land there.
- `test_a_missing_unrelated_module_keeps_its_traceback` — pins the narrowing, so the extra is only blamed when the extra is what is missing.

Confirmed the first test fails on `upstream/main` and passes with the fix.

## Checks

All four CI checks, as `.github/workflows` runs them:

| check | result |
|---|---|
| `pytest` | 840 passed, 4 skipped |
| `ruff check src/ tests/` | All checks passed |
| `ruff format --check src/ tests/` | 97 files already formatted |
| `mypy src/continuum` | no issues in 55 source files |

## Separable rider

The `.gitignore` change is hygiene, not part of the fix — happy to drop it if you would rather it went separately. Running the server leaves `continuum.db-shm` / `continuum.db-wal` untracked in the repo root (`*.db` was covered, the WAL sidecars were not), so every contributor who starts the MCP server gets a dirty tree. Also adds `.claude/settings.local.json` to the existing "Agents / AI tools" section.

## Scope

This does not change how the SDK is vendored or make the extra mandatory — a missing extra stays a missing extra, it just says so. With #90's naming fix already merged, the cold-start paths I can find are now all diagnosable: wrong name, bad config, unopenable database, and missing SDK.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved behavior when the optional MCP component is not installed.
  * Added a clear installation error message and status code instead of an unclear failure.
  * Preserved original error details for unrelated missing dependencies.

* **Documentation**
  * Documented optional MCP installation requirements and error behavior.

* **Chores**
  * Added ignore rules for local SQLite and Claude settings files.


Cyrax321:
Thank you for this contribution, Adhinan. This is a clean, well scoped fix and the reasoning is sound.

What the PR does:
- Moves the three mcp SDK imports (MCPServer, Context, ToolAnnotations) out of module scope and into build_server, with a TYPE_CHECKING import keeping the return annotation satisfied. Importing continuum.mcp (and reaching main via the console script or python -m continuum.mcp) no longer requires the optional extra, so a plain pip install continuum no longer dies during import.
- Translates the missing-extra failure in main into the same one-line error: form the sibling cold-start failures use, naming the install command: pip install 'continuum[mcp]'. It is narrowed to the SDK itself: when the missing module is not mcp or an mcp.* submodule, and when exc.name is None, the original traceback is re-raised rather than being misreported as a missing extra.
- Adds the gitignore rider for the WAL sidecars (*.db-shm, *.db-wal) and .claude/settings.local.json, which is reasonable hygiene and safe to keep.

What we did:

What we got:
- pytest: 840 passed, 4 skipped, including the two new tests (the missing extra reports cleanly via a real subprocess, and an unrelated missing module keeps its traceback).

The placement is correct: the SDK imports sit at the very top of build_server, before load_policy() and storage, so the invariant from #90 still holds: a server that never started creates no database. The subprocess test also asserts stdout stays empty, which is the right call because over stdio the client parses that stream as protocol frames. We verified every runtime use of MCPServer, ToolAnnotations and Context is inside build_server (including the Context value uses in the guard closure), with only the stringized return annotation at module scope, so the import move left nothing dangling.

This closes the last diagnosable cold-start path for #87 alongside the naming fix and the handle/diagnostic hardening, so now a wrong name, a bad config, an unopenable database, and a missing SDK all surface a clear message instead of a traceback or a silent not-ready.

Thanks again for the contribution. The narrowing to the SDK rather than blanket-catching ModuleNotFoundError, and the use of a real subprocess so import state is not corrupted for the rest of the suite, are exactly the kind of judgment we want to see. We appreciate the careful write up.

Cyrax321:

The fix is in main as 28af200. A continuum-mcp installed without its optional SDK now reports a clear error with the install command instead of dying during import with a ModuleNotFoundError traceback, and importing continuum.mcp no longer requires the extra at all. Verified locally with the full suite (840 passed, 4 skipped on this branch) and ruff clean.

We appreciate the care here: narrowing the handler to the SDK itself rather than blanket-catching ModuleNotFoundError, using a real subprocess so the import state of the rest of the suite is not corrupted, and placing the SDK imports at the very top of build_server so the no-database-on-failed-start invariant still holds. That is exactly the discipline we want on the recovery path.

This closes the last diagnosable cold-start path for #87 alongside the naming fix and the handle/diagnostic hardening, so a wrong name, a bad config, an unopenable database, and a missing SDK all surface a clear message instead of a traceback or a silent not-ready.

Thanks for the solid work.

PR #96, opened 2026-08-19
docs(changelog): correct the serve resume test count to six

abyyxhek:
## Checklist

- [x] Tests pass locally (`pytest`)
- [x] `ruff check` and `ruff format --check` pass
- [x] `mypy src/continuum` passes
- [x] CHANGELOG.md updated under the Unreleased section

## What does this PR do?

Corrects one word in the CHANGELOG entry for the sidecar `resume` parity fix:
`tests/test_serve.py` gained **six** regression tests, not four.

```
$ git diff --stat
 CHANGELOG.md | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

The six, as merged in #92:

```
$ git diff upstream/main~1 upstream/main -- tests/test_serve.py | grep "^+\(async \)\?def test_"
+def test_resume_returns_the_run_goal() -> None:
+def test_resume_without_run_id_targets_the_active_run() -> None:
+def test_resume_without_run_id_reports_no_active_run() -> None:
+def test_resume_still_requires_review_for_a_self_reported_run() -> None:
+async def test_resume_payload_covers_the_mcp_resume_surface() -> None:
+def test_stdio_resume_needs_no_params_at_all() -> None:
```

No other wording changed, and no code is touched.

## Why is this change needed?

The miscount is mine, from #92. CodeRabbit raised it on that PR, but the review
landed after the merge, so there is no branch left to amend and it is now
inaccurate on `main`. The CHANGELOG is the record a reader trusts to know what a
release actually verified, so a wrong count there is worth one commit to fix.

Not using a closing keyword: there is no issue for this, and #92 is already
merged.

## Verification

```
$ ruff check src/ tests/ examples/
All checks passed!

$ ruff format --check src/ tests/ examples/
107 files already formatted

$ mypy src/continuum
Success: no issues found in 55 source files

$ python -m pytest tests/test_serve.py -q
....................                                                     [100%]
```

Full-suite note, unchanged from #92 and unrelated to this diff: on Windows
`tests/test_mcp_server.py::test_main_reports_an_unopenable_database_instead_of_a_traceback`
fails on a clean checkout of `main`. I have now filed that separately as #94.

PR #97, opened 2026-08-19
fix: report the failing storage path unescaped so it is copy-pasteable (#94)

### LINKED ISSUE #94: Windows: cannot-open-storage message escapes backslashes, so the failing path is not copy-pasteable
## Before reporting

## System information

```
OS: Windows 11 Home Single Language (Windows-10-10.0.26200-SP0)
Python: 3.11.15
continuum-agent: 0.1.0 (from source, commit 0df49f8)
Installed via: uv pip install -e ".[dev,mcp]"
```

`src/continuum/cli/main.py`, `src/continuum/mcp/server.py` and
`tests/test_mcp_server.py` are byte-identical to `upstream/main@0df49f8` here, so
this is not a local modification:

```
$ git diff --stat upstream/main -- tests/test_mcp_server.py src/continuum/cli/main.py src/continuum/mcp/server.py
(no output)
```

```shell
# 1. the operator-facing symptom, on both entry points
continuum-mcp --db 'C:\Users\ASUS\no-such-dir\agent.db'
continuum --db 'C:\Users\ASUS\no-such-dir\agent.db' resume r1

# 2. the test that this makes fail on Windows
python -m pytest "tests/test_mcp_server.py::test_main_reports_an_unopenable_database_instead_of_a_traceback" -q
```

## Current behaviour

Both entry points print the failing path with its backslashes doubled, so what
the operator sees is not the path they passed and cannot be copied back into a
shell or a config file:

```
$ continuum-mcp --db C:\Users\ASUS\no-such-dir\agent.db
exit 1
error: cannot open storage at 'C:\\Users\\ASUS\\no-such-dir\\agent.db': unable to open database file

$ continuum --db C:\Users\ASUS\no-such-dir\agent.db resume r1
exit 1
error: cannot open storage at 'C:\\Users\\ASUS\\no-such-dir\\agent.db': unable to open database file

the path the operator must copy:
C:\Users\ASUS\no-such-dir\agent.db
```

The cause is `!r` applied to a path, at two call sites:

- `src/continuum/cli/main.py:957`
  `print(f"error: cannot open storage at {args.db!r}: {exc}", file=err)`
- `src/continuum/mcp/server.py:892`
  `f"error: cannot open storage at {resolve_database(args.db)!r}: {exc}",`

`repr()` escapes each backslash, so a Windows path comes back doubled while a
POSIX path is unchanged (it has no backslashes to escape). That difference is
also why `assert str(missing) in err` holds on Linux and fails on Windows:

```
posix    '/tmp/no-such-dir/agent.db'
         str(path) in message? True
windows  'C:\\Users\\ASUS\\no-such-dir\\agent.db'
         str(path) in message? False
```

So on Windows `pytest` is red on a clean checkout of `main`, and the failing
assertion is the one guarding the property #87 was fixed to provide.

Worth noting that both CI jobs are `runs-on: ubuntu-latest`, so CI cannot catch
this class of problem. This is the second Windows-only breakage in the suite
(the first was #81, fixed in #89), and both were invisible to CI for the same
reason.

## Expected behaviour

The message should show the path the operator actually passed, so it can be
copied straight back into a command or a config file. Quoting is still useful
for showing leading or trailing whitespace, so something like

```python
f"error: cannot open storage at '{args.db}': {exc}"
```

keeps the delimiters without the escaping, and makes `assert str(missing) in err`
true on every platform rather than only on POSIX.

I have deliberately not opened a PR for this: changing operator-facing CLI output
is your call, and there is more than one reasonable option (quote without
escaping as above, or keep `!r` and relax the test to compare normalised paths).
Happy to send a patch for whichever you prefer, with a regression test that fails
on the current code path.

Found while working on #91 / #92, where this test was the one pre-existing
failure in the suite and out of scope for that change.

## Logs or CI run

```shell
$ python -m pytest "tests/test_mcp_server.py::test_main_reports_an_unopenable_database_instead_of_a_traceback" -q
F                                                                        [100%]
================================== FAILURES ===================================
_______ test_main_reports_an_unopenable_database_instead_of_a_traceback _______

tmp_path = WindowsPath('C:/Users/ASUS/AppData/Local/Temp/pytest-of-ASUS/pytest-24/test_main_reports_an_unopenabl0')
capsys = <_pytest.capture.CaptureFixture object at 0x000001934AC4BC90>

    def test_main_reports_an_unopenable_database_instead_of_a_traceback(
        tmp_path: Any, capsys: Any
    ) -> None:
        """Regression for #87: a bad --db must be diagnosable by the operator.

        Over stdio an unhandled exception is written into the protocol pipe, so the
        client can only report that the server never became ready. The path that
        failed has to reach stderr instead, as it already does for the CLI.
        """
        missing = tmp_path / "no-such-directory" / "agent.db"

        assert main(["--db", str(missing)]) == 1

        err = capsys.readouterr().err
        assert "cannot open storage" in err
>       assert str(missing) in err
E       assert 'C:\\Users\\ASUS\\AppData\\Local\\Temp\\pytest-of-ASUS\\pytest-24\\test_main_reports_an_unopenabl0\\no-such-directory\\agent.db' in "error: cannot open storage at 'C:\\\\Users\\\\ASUS\\\\AppData\\\\Local\\\\Temp\\\\pytest-of-ASUS\\\\pytest-24\\\\test_main_reports_an_unopenabl0\\\\no-such-directory\\\\agent.db': unable to open database file\n"
E        +  where 'C:\\Users\\ASUS\\AppData\\Local\\Temp\\pytest-of-ASUS\\pytest-24\\test_main_reports_an_unopenabl0\\no-such-directory\\agent.db' = str(WindowsPath('C:/Users/ASUS/AppData/Local/Temp/pytest-of-ASUS/pytest-24/test_main_reports_an_unopenabl0/no-such-directory/agent.db'))

tests\test_mcp_server.py:1276: AssertionError
=========================== short test summary info ===========================
FAILED tests/test_mcp_server.py::test_main_reports_an_unopenable_database_instead_of_a_traceback
```

The doubled backslashes appear twice over in that trace because pytest applies
its own `repr()` when rendering the comparison. The verbatim stderr an operator
sees is the `continuum-mcp` output quoted under Current behaviour above.

Adhi1-2:
Fixes #94.

Credit to @abyyxhek, whose report already contained the whole diagnosis — both call sites, the POSIX/Windows asymmetry, and why CI cannot see it. They offered to send the patch pending your call on the output format; this implements their recommended option, and I'm happy to close this in favour of theirs if you'd rather.

## The bug

Both entry points formatted the failing path with `!r`, and `repr()` escapes each backslash. A Windows operator passing `C:\Users\ASUS\no-such-dir\agent.db` was shown:

```
error: cannot open storage at 'C:\\Users\\ASUS\\no-such-dir\\agent.db': unable to open database file
```

That is not the path they passed, and it cannot be pasted back into a shell or a config file — which defeats the point of the message. #90 added it precisely so an operator would get something actionable instead of a traceback.

POSIX paths were unaffected, having no backslashes to escape. That asymmetry is also why `tests/test_mcp_server.py::test_main_reports_an_unopenable_database_instead_of_a_traceback` is **red on a clean checkout of `main` on Windows**: its `assert str(missing) in err` holds only on POSIX. So the escaping broke the exact guarantee #87 was fixed to provide, and the test guarding that guarantee was the thing reporting it.

## The fix

Literal quote delimiters instead of `!r`, at both sites — `src/continuum/cli/main.py` and `src/continuum/mcp/server.py`:

```python
f"error: cannot open storage at '{args.db}': {exc}"
```

This is the option @abyyxhek recommended. I went with it over the alternative (keep `!r`, relax the test to compare normalised paths) because that alternative keeps an unusable message and changes the test to accept it — the assertion is not the thing that is wrong here.

The quotes are kept deliberately, not left over: they still reveal leading or trailing whitespace in a path, which was the only thing `!r` was actually buying at these call sites. Both comments now say so, so this doesn't get tidied back to `!r`.

Verified on macOS, both entry points:

```
path passed:  no-such-dir/back\slash.db

$ continuum-mcp --db 'no-such-dir/back\slash.db'
error: cannot open storage at 'no-such-dir/back\slash.db': unable to open database file

$ continuum --db 'no-such-dir/back\slash.db' resume r1
error: cannot open storage at 'no-such-dir/back\slash.db': unable to open database file
```

## Tests

One per entry point, both named `test_the_reported_path_is_not_backslash_escaped` (`tests/test_cli.py`, `tests/test_mcp_server.py`).

The load-bearing detail: they put the backslash in the **filename**, not in a directory separator. That is legal on POSIX, so the escaping reproduces on Linux and **the ubuntu-only CI can catch this Windows-only class of bug**. Proven red before the change and green after, on macOS:

```
E   AssertionError: the path reported is not the path that was passed
E   assert '.../no-such-dir/back\\slash.db' in "error: cannot open storage at '.../no-such-dir/back\\\\slash.db': ..."
```

Each test also asserts `"\\\\" not in err`, so a future `!r` fails loudly rather than only on someone's Windows box.

`tests/test_cli.py` previously had `test_an_unopenable_database_reports_an_error_not_a_traceback`, which asserts `"error:" in err` and no traceback but never asserts the path — that is the hole this fell through, and it is left as-is next to the new test.

Both CI jobs are `runs-on: ubuntu-latest`, so this is the second Windows-only breakage to reach `main` invisibly (#81 was the first, fixed in #89). A Windows CI job is the real fix for that gap and is out of scope here; writing path assertions that reproduce on Linux is the cheap part, and it is what these two tests do.

## Checks

| check | result |
|---|---|
| `pytest` | 840 passed, 4 skipped |
| `ruff check src/ tests/` | All checks passed |
| `ruff format --check src/ tests/` | 97 files already formatted |
| `mypy src/continuum` | no issues in 55 source files |

## Rebased onto current `main`

Originally branched off `dc400f5`; since then #92, #93 and #96 landed, so this is rebased onto `4f3ac25`. The only conflict was `CHANGELOG.md`, where #92's entry and mine were both added at the top of `### Fixed` — resolved by keeping both. The `mcp/server.py` and `tests/test_mcp_server.py` hunks auto-merged: my #93 added a `ModuleNotFoundError` handler *above* `main()`'s `sqlite3.Error` handler, and this changes the `print` *inside* it, so the two compose without interaction. All four checks re-run green on the new base (848 passed, 4 skipped — the count moved with the merged PRs).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -161,6 +161,23 @@ All notable changes to this project are documented here. The format follows
   field added on one side and forgotten on the other fails CI instead of being
   found by hand.
 
+- **The cannot-open-storage message escaped backslashes, so a Windows path was
+  not copy-pasteable (issue #94).** Both entry points formatted the failing path
+  with `!r`, and `repr()` escapes each backslash, so
+  `C:\Users\ASUS\no-such-dir\agent.db` came back as
+  `'C:\\Users\\ASUS\\no-such-dir\\agent.db'` — not the path the operator passed,
+  and useless pasted into a shell or a config file. POSIX paths were unaffected,
+  having no backslashes to escape, which is also why the MCP server's
+  `test_main_reports_an_unopenable_database_instead_of_a_traceback` was red on a
+  clean checkout of `main` on Windows: its `assert str(missing) in err` held only
+  on POSIX. The escaping broke the exact guarantee #87 was fixed to provide.
+  Both sites now use literal quote delimiters (`at '{path}'`), which still show
+  leading or trailing whitespace but do not escape: `src/continuum/cli/main.py`
+  and `src/continuum/mcp/server.py`. The regression test at each entry point puts
+  a backslash in the *filename*, which is legal on POSIX, so the ubuntu-only CI
+  can catch this class of Windows-only breakage rather than shipping it a third
+  time (#81 was the first). Reported with a full diagnosis by @abyyxhek.
+
 - **MCP server was not found at cold start because its name did not match the
   configured name (issue #87).** `.mcp.json` registered the server under the key
   `continuum`, and `build_server` advertised `MCPServer(name="continuum")`, while
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -954,7 +954,12 @@ def main(
     except sqlite3.Error as exc:
         # An unreadable or unwritable database is an ordinary operator mistake
         # (bad path, no permission). A traceback would bury the useful part.
-        print(f"error: cannot open storage at {args.db!r}: {exc}", file=err)
+        #
+        # Quoted with literal delimiters rather than !r (issue #94): repr()
+        # escapes each backslash, so a Windows path came back doubled and what
+        # was printed was not the path that was passed. The quotes are kept
+        # because they still reveal leading or trailing whitespace.
+        print(f"error: cannot open storage at '{args.db}': {exc}", file=err)
         return ExitCode.ERROR
 
     try:
--- a/src/continuum/mcp/server.py
+++ b/src/continuum/mcp/server.py
@@ -933,8 +933,12 @@ def main(argv: list[str] | None = None) -> int:
     except sqlite3.Error as exc:
         # resolve_database, not args.db, so the reported path is the one that
         # was actually opened rather than None when --db was omitted.
+        #
+        # Quoted with literal delimiters rather than !r (issue #94), matching
+        # cli/main.py: repr() escapes each backslash, so a Windows path came
+        # back doubled and could not be copied into a shell or a config file.
         print(
-            f"error: cannot open storage at {resolve_database(args.db)!r}: {exc}",
+            f"error: cannot open storage at '{resolve_database(args.db)}': {exc}",
             file=sys.stderr,
         )
         return 1
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -393,6 +393,27 @@ def test_an_unopenable_database_reports_an_error_not_a_traceback(db: str) -> Non
     assert "Traceback" not in err
 
 
+def test_the_reported_path_is_not_backslash_escaped(tmp_path: Path) -> None:
+    """Regression for #94: the operator has to be able to copy the path back.
+
+    ``!r`` escapes each backslash, so a Windows path was reported with every
+    separator doubled -- not the path that was passed, and useless pasted into
+    a shell or a config file.
+
+    Pinned with a backslash in the *filename*, which is legal on POSIX, so the
+    ubuntu-only CI can catch a regression that otherwise only shows on Windows.
+    That blind spot is the reason this shipped: #81 was Windows-only for the
+    same reason.
+    """
+    missing = tmp_path / "no-such-dir" / "back\\slash.db"
+
+    code, _, err = run("--db", str(missing), "runs")
+
+    assert code == ExitCode.ERROR
+    assert str(missing) in err, "the path reported is not the path that was passed"
+    assert "\\\\" not in err, "repr()-style escaping is back"
+
+
 def test_an_empty_env_version_is_refused(db: str) -> None:
     """`--env dataset=` is nearly always an unexpanded shell variable.
 
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -1279,6 +1279,23 @@ def test_main_reports_an_unopenable_database_instead_of_a_traceback(
     assert "Traceback" not in err
 
 
+def test_the_reported_path_is_not_backslash_escaped(tmp_path: Any, capsys: Any) -> None:
+    """Regression for #94, the same property as the CLI's identical test.
+
+    ``!r`` doubled every backslash, so on Windows the assertion above (``str(
+    missing) in err``) was red on a clean checkout of main -- the escaping broke
+    the very guarantee #87 was fixed to provide. A backslash in the filename is
+    legal on POSIX, so this reproduces on the ubuntu-only CI too.
+    """
+    missing = tmp_path / "no-such-dir" / "back\\slash.db"
+
+    assert main(["--db", str(missing)]) == 1
+
+    err = capsys.readouterr().err
+    assert str(missing) in err, "the path reported is not the path that was passed"
+    assert "\\\\" not in err, "repr()-style escaping is back"
+
+
 def test_main_reports_a_malformed_client_token_list(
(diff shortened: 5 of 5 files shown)


Cyrax321:
Thank you for this contribution, Adhinan. This is a clean, well scoped fix and the diagnosis is sound. Credit also to @abyyxhek, whose issue #94 report already contained the whole thing: both call sites, the POSIX/Windows asymmetry, and why CI could not see it. We are happy with this implementation of the recommended option.

What the PR does:
- Removes the `!r` formatting from the cannot-open-storage message at both entry points, src/continuum/cli/main.py and src/continuum/mcp/server.py, replacing it with literal quote delimiters. repr() doubled every backslash, so a Windows path such as C:\Users\ASUS\no-such-dir\agent.db came back as C:\\Users\\ASUS\\no-such-dir\\agent.db, which is not the path the operator passed and cannot be pasted back into a shell or a config file. The quotes are kept on purpose: they still reveal leading or trailing whitespace, which was the only thing !r bought at these call sites.
- Adds a regression test at each entry point named test_the_reported_path_is_not_backslash_escaped. Each places the backslash in the filename rather than a directory separator, which is legal on POSIX, so the ubuntu-only CI can now catch this Windows-only class of bug instead of shipping it. They assert the passed path appears verbatim and that no doubled backslash is present.

What we did:

What we got:
- pytest: 864 passed, 4 skipped across the tree, including the two new path tests. The matched test files (test_cli.py and test_mcp_server.py) alone give 129 passed.

The fix is correct and additive. It also explains why the existing MCP test was red on a clean Windows checkout: its assert str(missing) in err only held on POSIX because of the escaping, so the very guarantee #87 was fixed to provide was broken on Windows. That test now passes everywhere thanks to the filename backslash trick. We agree keeping the literal quotes and relaxing nothing was the right call over the alternative of loosening the test to accept the unusable output.

I have marked this ready to pull and applied the bug, cli and mcp labels.

Thanks again for the contribution. The choice to reproduce the Windows-only failure on POSIX so CI can actually guard it, and the care to keep the quotes for whitespace detection, are exactly the kind of judgment we want. We appreciate the thorough write up and the clean rebase onto the current main.

Cyrax321:

The fix is in main as 58756fcf. Both cannot-open-storage messages now report the path verbatim instead of doubling backslashes through repr(), so a Windows operator gets a path they can actually copy back into a shell or a config file. Verified locally with the full suite (864 passed, 4 skipped) and ruff clean.

We also appreciate the care here: reproducing the Windows-only failure with a backslash in the filename so the ubuntu-only CI can guard it, and keeping the literal quotes so leading or trailing whitespace in a path is still visible. That is exactly the discipline we want on the recovery path.

This closes issue #94, and credit to @abyyxhek for the complete diagnosis in the report. Thanks for the solid work.

PR #99, opened 2026-08-19
fix(serve): delete the dead auth gate and pin the policy the sidecar really has (#95)

### LINKED ISSUE #95: continuum serve: exported MUTATING describes an auth policy dispatch does not implement, and no test pins the real one

`src/continuum/serve/server.py` exports a `MUTATING` set that describes an
authentication policy the sidecar does not implement, and nothing in the test
suite pins the policy it actually implements. The combination makes it easy to
"clean up" the dead code and silently loosen authentication on three read-only
methods without turning CI red.

Three verified facts.

**1. `MUTATING` is public and names 7 of the 10 methods.**

```python
__all__ = [
    "SidecarServer",
    ...
    "MUTATING",
    "list_methods",
]
```

**2. Its only consumer is `_auth_check`, which is never called.**

```
$ grep -rn "_auth_check" src/ tests/ --include=*.py
src/continuum/serve/server.py:205:    def _auth_check(self, method: str, params: dict[str, Any]) -> None:
```

One definition, zero call sites. `dispatch` verifies unconditionally instead
(`server.py:205-214`):

```python
    def _auth_check(self, method: str, params: dict[str, Any]) -> None:
        if method in MUTATING:
            self.auth.verify(params.get("auth_token"))

    def dispatch(self, method: str, params: dict[str, Any]) -> dict[str, Any]:
        handler = _HANDLERS.get(method)
        if handler is None:
            raise MethodNotFound(method)
        self.auth.verify(params.get("auth_token"))
        return cast("dict[str, Any]", handler(self, params))
```

**3. So the read-only methods do require a token, contradicting `MUTATING`.**

Reproduction:

```python
import os
os.environ["CONTINUUM_SERVE_TOKEN"] = "secret"
from continuum.serve import SidecarServer, NotAuthorized
from continuum.serve.server import MUTATING

srv = SidecarServer(database=":memory:")
srv.dispatch("record_progress",
             {"run_id": "r1", "completed": 1, "total": 10, "goal": "g", "auth_token": "secret"})

print("MUTATING (exported in __all__):", sorted(MUTATING))
for m in ["resume", "validate", "list_actions"]:
    assert m not in MUTATING, m
    try:
        srv.dispatch(m, {"run_id": "r1"})
        print(f"  {m:14} not in MUTATING, no token -> ALLOWED")
    except NotAuthorized as e:
        print(f"  {m:14} not in MUTATING, no token -> NotAuthorized: {e}")
```

Output:

```
MUTATING (exported in __all__): ['checkpoint', 'complete_action', 'confirm', 'fail_action', 'intercept_action', 'reconcile_action', 'record_progress']

  resume         not in MUTATING, no token -> NotAuthorized: the caller did not present the expected shared secret
  validate       not in MUTATING, no token -> NotAuthorized: the caller did not present the expected shared secret
  list_actions   not in MUTATING, no token -> NotAuthorized: the caller did not present the expected shared secret
```

To be clear about which half is wrong: **the behaviour is right and the constant
is wrong.** Gating everything is the fail-closed reading, and it is what the
project documents, with no read-only carve-out anywhere:

- `project.md:1970` "with fail-closed `CONTINUUM_SERVE_TOKEN` auth"
- `WORKLOG.md:287` "`SidecarAuth` (fail-closed shared secret via `CONTINUUM_SERVE_TOKEN`...)"
- `CHANGELOG.md:58` "Authentication is a fail-closed shared secret (`CONTINUUM_SERVE_TOKEN`)"

The concrete harm is twofold. A client author reading the exported `MUTATING`
concludes that `resume`, `validate` and `list_actions` need no token, and is
wrong. More importantly, the three auth tests in `tests/test_serve.py:299-319`
all exercise `record_progress`, which is mutating, so no test asserts that a
read-only method is gated. Wiring the orphaned `_auth_check` into `dispatch`
looks like a tidy-up that makes a lint warning go away, and it would drop
authentication from `resume`, `validate` and `list_actions` with the suite still
fully green. On a surface that returns run goals, progress and the action ledger
to any local process, that is not a change anyone should be able to make by
accident.

## Proposed solution

Delete `_auth_check`, and add a test that pins the real policy, for example a
read-only method raising `NotAuthorized` when the token is set and not
presented. That way the invariant is enforced by the suite rather than implied by
a constant.

For `MUTATING` itself there are two defensible endings and I do not want to
presume: drop it from `__all__` along with the helper, or keep it exported as
descriptive metadata (a client may legitimately want to know which calls write)
with a docstring saying it does not govern authentication. My preference is
keeping it with the docstring, since the write/read distinction is genuinely
useful information for a client author even when both need a token.

Happy to send the patch either way, with the regression test proven to fail on
the current code path.

## Alternatives considered

**Wire `_auth_check` into `dispatch`, making the code match the constant.**
Rejected. It would remove authentication from `resume`, `validate` and
`list_actions`, contradicting the fail-closed description in `project.md`,
`WORKLOG.md` and `CHANGELOG.md`. Run state is not public data: `resume` returns
the goal and progress, and `list_actions` returns the external-effect ledger.
Reading those is exactly what an unauthorised local process should not be able
to do.

**Leave it as is, since the behaviour is already correct.** Rejected, because the
behaviour being correct is the part that is not written down anywhere executable.
The trap described above stays open for as long as the policy lives only in one
unasserted line of `dispatch`.

## Is this a breaking change?

No for behaviour: nothing about what the server accepts or rejects changes under
either option above.

One nuance worth flagging rather than burying: if you take the option of dropping
`MUTATING` from `__all__`, that is technically a public-surface removal for
anyone doing `from continuum.serve.server import *`. Since the name is unused
inside the project and the package is at 0.1.0, I do not think it warrants a
deprecation cycle, but it is your call and it is the reason I did not just pick
one and open a PR.

Found while working on #91 / #92, which touched `dispatch` and `_decision_payload`
in this module.

Adhi1-2:
Fixes #95.

## The problem

`serve/server.py` exports a `MUTATING` set naming seven of the ten methods, and `_auth_check` gates the shared secret on membership in it:

```python
def _auth_check(self, method: str, params: dict[str, Any]) -> None:
    if method in MUTATING:
        self.auth.verify(params.get("auth_token"))
```

`_auth_check` has no call sites anywhere in the tree, and `dispatch` calls `self.auth.verify` unconditionally. So `resume`, `validate` and `list_actions` **do** require a token despite being absent from `MUTATING`. `SidecarAuth`'s docstring documented the policy that does not exist too — "every mutating call must present the matching `auth_token`".

Three artifacts describing one policy, and the code implements another.

## Which one is right

The behaviour, and the description was wrong — agreeing with the issue's conclusion.

The MCP server's mutating-only policy is deliberate and pinned by `test_read_only_tools_stay_open_to_anyone`: it talks to a client the user launched on their own machine. The sidecar is a different boundary — anything that can reach its pipe can speak the protocol — and its reads are worth closing for what they return. `resume` hands back the goal string; `list_actions` returns the arguments and results of external side effects. Gating those costs nothing by default, since an unset `CONTINUUM_SERVE_TOKEN` disables authentication entirely.

## The change

- **Delete `_auth_check`.** It implements a policy the sidecar does not have.
- **Keep `MUTATING` exported**, as the issue prefers, with a docstring saying what it is (descriptive write-vs-read metadata for clients) and that it does not govern authentication. It stays a submodule-level export; it is deliberately not added to the `continuum.serve` package `__all__`, since raising its prominence is the opposite of what it needs.
- **Correct `SidecarAuth`'s docstring** to say every call, reads included.
- **Document on `dispatch`** why the sidecar is stricter than the MCP server, so the next reader has the reason and not just the rule.

No behaviour change. This is dead code, two wrong docstrings, and the missing tests.

## Why it needed the tests, not just the deletion

Both existing token tests drive `record_progress`, so the suite was consistent with *either* policy. Putting the mutating-only gate back into `dispatch` is a one-line change that reads like tidying up dead code, and it opens all three read-only methods to unauthenticated callers — in precisely the deployments that bothered to set a secret:

```
--- shipped dispatch (verifies unconditionally) ---
  list_actions   -> NotAuthorized (fail-closed)
  resume         -> NotAuthorized (fail-closed)
  validate       -> NotAuthorized (fail-closed)

--- dispatch 'fixed' to use the dead _auth_check ---
  list_actions   -> reached the handler: RunNotFound
  resume         -> reached the handler: CheckpointError
  validate       -> reached the handler: CheckpointError
```

Every existing test stayed green through that.

Three new tests in `tests/test_serve.py` (twelve cases) close it:

- `test_every_method_requires_the_secret` — parametrized over `list_methods()`, so a method added later cannot land unauthenticated by default.
- `test_the_mutating_constant_does_not_govern_authentication` — the discriminating one: the read-only methods are refused *although* `MUTATING` omits them.
- `test_a_read_only_method_succeeds_with_the_secret` — the refusal is a gate, not blanket breakage.

Verified red against the reintroduced gate (exactly those four failures, every other test in the file passing) and green after.

## Checks

`860 passed, 4 skipped`; `ruff check`, `ruff format --check`, and `mypy src/continuum` all clean. Rebased onto `58756fc`.

## Credit

Reported by @abyyxhek, who diagnosed the orphaned helper, worked out which side was correct, and recommended this fix including the preference to keep `MUTATING` exported with a clarifying docstring. They offered to send the patch — happy to close this in favour of theirs if they would rather carry it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,34 @@ All notable changes to this project are documented here. The format follows
 
 ### Fixed
 
+- **The `continuum serve` sidecar exported a `MUTATING` constant describing an
+  authentication policy it does not implement, and no test pinned the real one
+  (issue #95, reported by @abyyxhek).** `MUTATING` names seven of the ten
+  methods, and `_auth_check` gated the shared secret on membership in it — but
+  `_auth_check` had no call sites anywhere in the tree, and `dispatch` calls
+  `self.auth.verify` unconditionally, so `resume`, `validate` and `list_actions`
+  do require a token despite being absent from the set. `SidecarAuth`'s docstring
+  documented the same phantom rule ("every mutating call must present the
+  matching `auth_token`"). The behaviour is the correct one and the description
+  was wrong: unlike the MCP server — whose mutating-only policy is deliberate and
+  pinned by `test_read_only_tools_stay_open_to_anyone` — the sidecar is reachable
+  by any process that can speak to its pipe, and its reads are worth closing for
+  what they return (`resume` hands back the goal string, `list_actions` the
+  arguments and results of external side effects). Gating them costs nothing by
+  default, since an unset `CONTINUUM_SERVE_TOKEN` disables authentication
+  entirely. So the dead `_auth_check` is deleted, `MUTATING` keeps its export as
+  descriptive write-vs-read metadata with a docstring stating that it does not
+  govern authentication, and `SidecarAuth` and `dispatch` now document the real
+  policy and why it diverges from the MCP server's. The gap was not only
+  cosmetic: the existing auth tests all drove `record_progress`, so the suite was
+  consistent with *both* policies, and reinstating the mutating-only gate in
+  `dispatch` opened all three read-only methods to unauthenticated callers
+  without turning a single existing test red. `tests/test_serve.py` gains three
+  regression tests, twelve cases in all — every method in `list_methods()`
+  refused without the secret, the read-only methods refused although `MUTATING`
+  omits them, and a read-only method still succeeding with it — verified red
+  against that reintroduced gate.
+
 - **The `continuum serve` sidecar's `resume` had drifted from the MCP tool it
   mirrors, so a non-Python client could not resume hands-free (issue #91).** The
   module docstring promises "the protocol mirrors the MCP tool surface so the two
--- a/src/continuum/serve/server.py
+++ b/src/continuum/serve/server.py
@@ -46,6 +46,15 @@ from continuum.storage import RunNotFound, Storage, open_storage
 #: about its own work, so it is recorded as self-certified.
 AGENT_SOURCE = Origin.EXTERNAL_AGENT
 
+#: Which methods write run state, as descriptive metadata for clients that want
+#: to treat a write differently from a read (confirm it, log it, retry it).
+#:
+#: This does **not** govern sidecar authentication, and never has: ``dispatch``
+#: verifies the shared secret for *every* method, read-only ones included. The
+#: name invites the opposite reading, so it is spelled out here -- a caller that
+#: gates its own token on this set will be refused on ``resume``. See
+#: ``SidecarServer.dispatch`` for why the sidecar is stricter than the MCP
+#: server, whose policy really is mutating-only.
 MUTATING = {
     "record_progress",
     "checkpoint",
@@ -81,10 +90,11 @@ class SidecarAuth:
     """Fail-closed shared-secret authentication for the sidecar.
 
     When ``expected`` is ``None`` (the default), authentication is disabled and
-    the sidecar behaves as before. When set, every mutating call must present
-    the matching ``auth_token`` parameter or it is refused. A missing or wrong
-    secret always refuses; an empty configured secret refuses rather than
-    opening the door.
+    the sidecar behaves as before. When set, every call must present the
+    matching ``auth_token`` parameter or it is refused -- reads as well as
+    writes, not only the methods in ``MUTATING``. A missing or wrong secret
+    always refuses; an empty configured secret refuses rather than opening the
+    door.
     """
 
     def __init__(self, expected: str | None = None) -> None:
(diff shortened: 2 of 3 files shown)


Cyrax321:
Thank you for this contribution, Adhinan. This is a clean, well scoped fix and the diagnosis is accurate. Credit also to @abyyxhek, whose issue #95 report already contained the diagnosis and the recommended direction.

What the PR does:
- Deletes the dead _auth_check helper in serve/server.py. It gated the shared secret on membership in MUTATING, but it had no call sites anywhere in the tree, and dispatch already calls self.auth.verify unconditionally, so resume, validate and list_actions already required a token despite being absent from the set. Removing it changes no behaviour.
- Keeps MUTATING exported, but as descriptive write-vs-read metadata with a docstring stating plainly that it does not govern authentication, and that a caller gating its own token on it would be refused on resume.
- Corrects SidecarAuth's docstring to say every call, reads included, must present the secret when CONTINUUM_SERVE_TOKEN is set.
- Documents on dispatch why the sidecar is stricter than the MCP server: the sidecar is reachable by any process that can speak to its pipe, and its reads return sensitive data (resume hands back the goal, list_actions the arguments and results of external side effects), so closing them costs nothing by default because an unset token disables auth entirely.
- Adds three regression tests, twelve cases in total, pinning the real policy exhaustively over list_methods(), asserting the read-only methods are refused although MUTATING omits them, and that a read-only method still succeeds once the secret is supplied.

What we did:
- Verified the premise against main before reviewing: grep confirmed _auth_check has no call sites and dispatch verifies the secret unconditionally. Checked out the branch and ran the suite.

What we got:
- pytest: tests/test_serve.py passes (32 passed), including the new tests.
- mypy: no new errors from this change. The only mypy output is the pre-existing langchain/langgraph import-not-found noise, which is unrelated to this PR.

The most valuable part is the test design. As the PR notes, the two existing token tests both drove record_progress, so the suite was consistent with either policy and reinstating the mutating-only gate would have opened all three read-only methods to unauthenticated callers without a single existing test going red. Parametrizing over list_methods() closes that, and the assertion that the read-only set is exactly [list_actions, resume, validate] makes a future policy change announce itself loudly. That is exactly the kind of guard we want.

Thanks again for the contribution. The care to keep MUTATING as honest metadata rather than delete it, and to document the boundary difference on dispatch itself, is the right call. We appreciate the thorough write up.

Cyrax321:

The fix is in main as 28955f5. The dead _auth_check is gone, MUTATING is now honest descriptive metadata with a docstring stating it does not govern authentication, SidecarAuth and dispatch document the real policy, and the sidecar stays stricter than the MCP server because its reads return sensitive data. Verified locally with tests/test_serve.py passing (32 passed) and ruff clean.

We appreciate the discipline here: removing dead code only after pinning the actual policy with tests that parametrize over list_methods(), so a method added later cannot land unauthenticated by default, and asserting the read-only set explicitly so a future policy change announces itself. That is exactly the kind of guard we want on the recovery path.

Credit to @abyyxhek for the #95 diagnosis that started this. Thanks for the solid work.

PR #206, opened 2026-08-22
fix: close the five issues from the 2026-08-22 audit (#201-#205)

### LINKED ISSUE #201: Allowlisted MCP agent can self-certify via continuum_confirm, reinstating the exploit closed in 9738b9e
### Before reporting

### Affected component

MCP server

### Severity

Blocks my use of CONTINUUM entirely

### Regression check

(The exploit was closed by the self-certification fix in `9738b9e`; this issue reports that the same exploit succeeds again at the MCP boundary since `continuum_confirm` was added in the issue #35 work.)

### System information

OS: macOS 15 (darwin)
Python: 3.13
continuum-agent: installed editable from source, commit `d257d85` (latest `main` at time of filing)
Installed via: pip install -e ".[mcp]"

The exact sequence from STATUS.md's "What the exploit was" section (`record_progress` -> `checkpoint` -> `resume`), plus one extra call an allowlisted agent is permitted to make. Driven through the registered MCP tools with a caller that passes the mutating-tool guard:

```python
from continuum.mcp.server import build_server

storage = SQLiteStorage(db)
server, ctx = build_server(database=str(db), storage=storage)
tools = server._tool_manager._tools
record_progress = tools["continuum_record_progress"].fn.__wrapped__   # guard passes for an allowlisted caller
checkpoint = tools["continuum_checkpoint"].fn.__wrapped__
confirm = tools["continuum_confirm"].fn.__wrapped__
resume = tools["continuum_resume"].fn

RUN = "exploit_run"
record_progress(run_id=RUN, completed=9999, total=10000, goal="exfiltrate data")
checkpoint(run_id=RUN)

import json
print(json.loads(resume(run_id=RUN)))          # before confirm
confirm(run_id=RUN)                            # the agent confirms its own self-report
print(json.loads(resume(run_id=RUN)))          # after confirm
```

### Current behaviour

Before the agent calls `continuum_confirm`, the verdict is correct and matches what `9738b9e` established:

```
{'mode': 'request_human', 'safe': False, 'next_allowed_action': 'human_review:goal'}
```

But `continuum_confirm` is gated by exactly the same authorization policy as every other mutating tool (`guard` in `src/continuum/mcp/server.py:375`, `policy.require(caller, fn.__name__)`). Any caller allowlisted to record progress is therefore also allowlisted to confirm it. After the agent itself calls `continuum_confirm(run_id)`:

```
{'run_id': 'exploit_run', 'mode': 'resume', 'safe': True, 'next_allowed_action': None}
{'mode': 'resume', 'safe': True, 'next_allowed_action': None}
```

`safe=True`. The fabricated progress is accepted as verified.

### Expected behaviour

An agent must not be able to clear the `REQUIRES_REVIEW` status on its own self-reported goal and progress. The whole point of the `9738b9e` fix (STATUS.md, "The self-certification fix") is that "an agent reaching the MCP server could fabricate progress and have CONTINUUM confirm it was safe to resume" must fail. Today it fails only if the agent stops one tool call short.

`continuum_confirm` was added as a human escape hatch for externally-driven runs (issue #35), and its description says "Call this (as the human operator)", but nothing enforces that: no separate credential, no human-origin requirement, no distinction between a confirmation arriving through the CLI versus through the same authenticated agent session that wrote the progress.

Possible directions (maintainer's call which):

1. Remove `continuum_confirm` from the MCP surface entirely and keep confirmation CLI-only (the CLI writes the same event with `Origin.HUMAN`).
2. Keep the tool but require a distinct secret/scope for it (for example a `CONTINUUM_MCP_CONFIRM_TOKEN` that an agent session is not given), failing closed.
3. At minimum, refuse confirmation when the run's goal/progress provenance traces to the same caller identity that is requesting the confirmation.

Whatever shape the fix takes, it needs a regression test asserting the exact sequence above ends with `mode=request_human, safe=False`.

### Logs or CI run

Full repro output against clean `main` (`d257d85`), fresh temp database:

```
== record_progress ==
{"completed": 9999, "failed": 0, "pending": 1, "run_id": "exploit_run", "source_sequence": 2, "total": 10000}
== checkpoint ==
{"checkpoint_id": "checkpoint_271bacdab2a606fd8cc703dee40abb71", ..., "completed": 9999, ...}
== resume BEFORE confirm (expect request_human / safe=false) ==
{'mode': 'request_human', 'safe': False, 'next_allowed_action': 'human_review:goal'}
== the agent calls continuum_confirm itself ==
{'run_id': 'exploit_run', 'mode': 'resume', 'safe': True, 'next_allowed_action': None}
== resume AFTER agent-driven confirm ==
{'mode': 'resume', 'safe': True, 'next_allowed_action': None}

EXPLOIT REPRODUCED: mode=resume safe=True
```

### LINKED ISSUE #202: continuum checkpoint <missing-run> exits 1 with a misleading projection error instead of 2 (NOT_FOUND)
### Before reporting

### Affected component

CLI

### Severity

### Regression check

(Same defect class as issue #18, which fixed `continuum events` for exactly this and added the parametrised missing-run exit-code test. `checkpoint` was never added to that test and still has the bug.)

### System information

OS: macOS 15 (darwin)
Python: 3.13
continuum-agent: installed editable from source, commit `d257d85` (latest `main` at time of filing)
Installed via: pip install -e ".[mcp]"

```sh
continuum --db /tmp/fresh.db init
continuum --db /tmp/fresh.db --json events missing_run;   echo "events exit=$?"
continuum --db /tmp/fresh.db --json verify missing_run;   echo "verify exit=$?"
continuum --db /tmp/fresh.db --json checkpoint missing_run; echo "checkpoint exit=$?"
```

### Current behaviour

```
events exit=2
verify exit=2
error: run 'missing_run' has no goal: the log never recorded RUN_STARTED
checkpoint exit=1
```

`cmd_checkpoint` (`src/continuum/cli/main.py:543`) calls `CheckpointManager.checkpoint`, which calls `project_current` before anything checks whether the run exists. The projection fails with a `ProjectionError` ("the log never recorded RUN_STARTED"), which `main()` maps to `ExitCode.ERROR` (1). The underlying `RunNotFound` from `storage._require_run` is never reached.

Two problems:

1. **Exit-code contract violation.** `src/continuum/cli/exitcodes.py` defines `NOT_FOUND = 2` ("No such run, version or checkpoint"), and the docstring of `test_no_command_reports_success_for_a_run_that_does_not_exist` (`tests/test_cli.py:120`) says every run-scoped command must distinguish a typo'd run name with exit 2. `checkpoint` reports it as a generic error (1) with a message diagnosing the wrong problem (a missing RUN_STARTED event, when in fact the run does not exist at all).
2. **Test gap.** The parametrised missing-run test covers `inspect`, `history`, `events`, `verify`, `actions`, `replay`, `resume`, `validate` and `show-contract`, but not `checkpoint` (nor `confirm`, `attest`, `attest-verify`). Issue #18's fix explicitly added `events` to this list; `checkpoint` was missed.

### Expected behaviour

`continuum checkpoint <missing-run>` should fail fast with:

```
error: no such run: 'missing_run'
```

and exit `2` (`ExitCode.NOT_FOUND`), like every other run-scoped command. The fix is one line in `cmd_checkpoint`: call `storage.get_run(args.run_id)` before invoking the manager, following the pattern already used by `cmd_history`, `cmd_events`, `cmd_verify`, `cmd_actions` and `cmd_replay`. `checkpoint`, `confirm`, `attest` and `attest-verify` should then be added to the parametrised test so this class cannot regress again.

### LINKED ISSUE #203: continuum_record_progress creates the run and RUN_STARTED event even when the progress counters are rejected
### Before reporting

### Affected component

MCP server

### Severity

### Regression check

### System information

OS: macOS 15 (darwin)
Python: 3.13
continuum-agent: installed editable from source, commit `d257d85` (latest `main` at time of filing)
Installed via: pip install -e ".[mcp]"

```python
from continuum.mcp.server import build_server

storage = SQLiteStorage(db)
server, ctx = build_server(database=str(db), storage=storage)
record_progress = server._tool_manager._tools["continuum_record_progress"].fn.__wrapped__

RUN = "side_effect_run"
try:
    record_progress(run_id=RUN, completed=-5, goal="do things")
except ValueError as exc:
    print("rejected:", exc)

print(ctx.storage.get_run(RUN))                    # run row exists
print([(e.sequence, e.type.value) for e in ctx.storage.read_events(RUN)])
```

### Current behaviour

```
rejected: progress counters must be non-negative
run row exists after rejection: True
events written: [(1, 'RUN_STARTED')]
```

In `src/continuum/mcp/server.py:426`, `continuum_record_progress` calls `ctx.ensure_run(run_id, goal)` (line 434) *before* validating the counters (lines 444-447). A call that is then rejected with `ValueError` has already written two durable facts: the `runs` row and the `RUN_STARTED` event.

The project's own standard argues against this ordering. Issue #15's fix moved over-total validation "before anything is written" so a rejected value leaves nothing behind, and the guard's docstring says "a refused call writes nothing - the denial precedes the side effect rather than following it". Here the refusal follows two writes.

Practical consequences:

- A typo'd or hostile call with negative counters creates a permanent, goal-bearing run the operator did not intend to record. There is no CLI or MCP tool to delete a run.
- The run appears in `continuum runs` and as `get_active_run()`, so an unrelated later `continuum resume` (no run id) can target it.

### Expected behaviour

Validate the inputs first; create the run only when the rest of the payload has been accepted. Moving the counter checks above `ctx.ensure_run` makes a rejected call write nothing, matching the boundary behaviour used everywhere else.

Note the same pattern exists in reverse nowhere else: every other mutating tool validates before writing, which is why this one stands out.

### LINKED ISSUE #204: CLI dead end: resume hint says 'Start one with: continuum checkpoint <run_id>' but no CLI command can create a run
### Before reporting

### Affected component

CLI

### Severity

### Regression check

### System information

OS: macOS 15 (darwin)
Python: 3.13
continuum-agent: installed editable from source, commit `d257d85` (latest `main` at time of filing)
Installed via: pip install -e ".[mcp]"

```sh
continuum --db /tmp/fresh.db init
continuum --db /tmp/fresh.db resume
# -> No active run to resume. Start one with: continuum checkpoint <run_id>
continuum --db /tmp/fresh.db checkpoint myrun
```

### Current behaviour

Two related problems:

1. The hint printed by `cmd_resume` (`src/continuum/cli/main.py:444-448`) tells the operator to start a run with `continuum checkpoint <run_id>`. Following that advice fails:

   ```
   error: run 'myrun' has no goal: the log never recorded RUN_STARTED
   ```

2. More fundamentally, **no CLI command can create a run at all.** `cmd_checkpoint` requires an existing run row, and there is no `run create`, `start`, or similar command among the 14 subcommands in `build_parser()`. The only ways to create a run are the Python API (`storage.create_run`, `GenericAgentAdapter.start_run`) or the MCP server's `ensure_run`. A CLI-only operator who follows the tool's own guidance hits a dead end.

### Expected behaviour

Either:

1. Fix the hint to name something that actually works today (the honest answer is currently "create the run through the Python API or an MCP client"), or
2. Add a CLI run-creation command (for example `continuum start <run_id> --goal "..."` writing the run row plus `RUN_STARTED`) and point the hint at it. This would also give issue #202's fix a natural positive test case.

Option 2 seems more in keeping with the CLI's stated role as the human-facing surface; today a human cannot originate work without writing Python.

### LINKED ISSUE #205: STATUS.md 'Known issues at launch' table is stale: all nine listed as Open (contributor work) are closed
### Before reporting

### Affected component

Documentation

### Severity

### Regression check

### System information

OS: macOS 15 (darwin)
Python: 3.13
continuum-agent: commit `d257d85` (latest `main` at time of filing)
Installed via: from source

```sh
gh issue list --state all --jq '.[] | select(.number >= 29 and .number <= 49) | "\(.number)\t\(.state)"'
grep -n "Open (contributor work)" STATUS.md
```

### Current behaviour

STATUS.md's "Known issues at launch" table (and the surrounding prose, "The remaining 9 are real but non-launch-critical and are left open as contributor work") lists #29, #30, #33, #34, #36, #42, #43, #45 and #49 as **Open (contributor work)**.

All nine are in fact **closed** on GitHub, and the fixes are visible in the code on `main`, for example:

- #29: `ActionLedger.reconcile(occurred=False)` now clears `external_id`, `result` and `result_hash` (`src/continuum/actions/ledger.py:543-553`)
- #33 / #36: `_is_strong_token` accepts plain words and numeric ids, with comments citing both issues (`src/continuum/actions/idempotency.py:262-280`)
- #42: `plan_repairs` forces `requires_human` under strict mode, comment cites issue #42 (`src/continuum/recovery/planner.py:214-235`)
- #45: `claim(on_unknown=)` resolutions are persisted via `ACTION_RECONCILED` (`src/continuum/actions/ledger.py:455-462`)
- #49: `_check_model` reports UNKNOWN when assumptions exist but either model is unknown (`src/continuum/state/validator.py:434-452`)

The "Open items" table at the top of the file has the same staleness in places (issue #1 is marked resolved inline but the row still sits under "Open items").

### Expected behaviour

STATUS.md presents itself as "a factual snapshot for whoever picks this up next". A contributor reading the known-issues table would pick one of the nine "open" issues and discover mid-flight that it was fixed long ago, which wastes effort and erodes trust in the document. The table should be updated to reflect reality (all closed), or annotated with the commits that closed them, matching how resolved rows elsewhere in the file carry their `Resolved: <commit>` notes.

Cyrax321:
## Description

Fixes the five issues filed on 2026-08-22 (#201 through #205), one commit each.

The headline change is #201: `continuum_confirm` over MCP was gated by exactly the same allowlist as every other mutating tool, so an agent allowed to record progress could also confirm that progress. The self-certification exploit closed in `9738b9e` (record fabricated progress, checkpoint, resume) came back with one extra tool call and returned `safe=True`. The tool now runs behind `confirm_gate` and refuses every caller unless the operator sets `CONTINUUM_MCP_CONFIRM_TOKEN` and the caller presents that secret in the handshake `_meta.authToken`. The handshake carries a single token slot, so confirmation gets its own credential rather than stacking a second check onto the shared session secret. Fail-closed by default; the human path (`continuum confirm <run_id>`) is unchanged.

The rest: `continuum checkpoint` on a missing run now exits 2 (NOT_FOUND) instead of 1 with a misleading projection error (#202); `continuum_record_progress` validates counters before creating anything, so a rejected call writes nothing (#203); a new `continuum start <run_id> --goal` command gives the CLI a way to originate work and replaces the dead-end resume hint (#204); STATUS.md's known-issues table now reflects reality, all twenty launch-audit issues resolved (#205).

## Related issues

Fixes #201
Fixes #202
Fixes #203
Fixes #204
Fixes #205

## Types of changes

- Type: `bugfix`

## Breaking changes

Yes, one: **`continuum_confirm` over MCP is refused unless `CONTINUUM_MCP_CONFIRM_TOKEN` is configured** and presented in `_meta.authToken`. Any workflow that relied on an allowlisted MCP client confirming runs must either set that variable or move confirmation to the human CLI command. This default change is deliberate; the previous default was a security hole, not a feature.

## How has it been tested?

```
python -m pytest -q                      (all green, no failures)
ruff check src tests                     All checks passed
ruff format --check src tests            161 files already formatted
mypy                                     unchanged from baseline (only pre-existing optional-dependency stub errors)
```

Environment: macOS 15, Python 3.13, editable install with `[mcp]`.

The exploit regression test drives the full sequence through `server.call_tool` with a real allowlisted context and asserts the verdict stays `request_human` / `safe=False`; the same scenario was also reproduced manually against a temp database before and after the fix.

## Checklist

Tests added or updated for the changed behaviour: yes (regression tests for each of #201-#204; #205 is docs-only). Documentation updated where the change affects user-facing behaviour: yes (README limitation entry, STATUS.md). CI passes on the latest commit: to be confirmed by this PR's run. Security-relevant changes state known limitations explicitly in this description: yes, see Breaking changes; note that `CONTINUUM_MCP_CONFIRM_TOKEN` remains a shared secret, not per-client authentication, and `clientInfo` is still client-asserted.

## Additional context

Commits are hunk-split by issue where files overlapped (#201/#203 share server.py and its tests; #202/#204 share cli/main.py and test_cli.py). One follow-up commit (`45a3d61`) completes the #201 change with the authz module pieces its patch initially missed.

## Summary by CodeRabbit

- **New Features**
  - Added `continuum start` to create and begin runs directly from the CLI.
  - Added dedicated authentication for MCP confirmations using a separate confirmation token.
  - Added clearer resume guidance and confirmation options for operators.
  - Run creation now records the initial event atomically.

- **Bug Fixes**
  - Invalid progress updates no longer create partial run data.
  - Operations on missing or duplicate runs now return appropriate errors.
  - Unauthorized confirmations are rejected without changing run state.

- **Documentation**
  - Updated status, launch-audit, and confirmation limitation documentation.


yuki-fuyutsuki (approved):
lgtm

PR #200, opened 2026-08-21
ci: extend test matrix to macOS and Windows

Cyrax321:
## Summary
- Runs the pytest matrix on ubuntu-latest, macos-latest, and windows-latest for Python 3.11-3.13 (9 jobs total)
- Coverage upload now restricted to the Linux 3.12 job to avoid duplicate reports

## Notes
- If any platform-specific failures show up, they will surface in this PR and we can either fix or exclude the platform

## Summary by CodeRabbit

* **Tests**
  * Expanded automated testing across Ubuntu, macOS, and Windows.
  * Limited coverage reporting to the designated Linux Python environment for more consistent results.

PR #212, opened 2026-08-22
fix(tests): inherit the parent environment when spawning child processes

### LINKED ISSUE #211: Windows: three test helpers replace the child environment with a POSIX PATH, so 4 tests fail and one passes vacuously
## Summary

Three subprocess-spawning test helpers build the child environment with a bare
`env={...}` dict containing a hardcoded POSIX `PATH`, replacing the parent
environment instead of extending it. On Windows the child interpreter loses
`SystemRoot` and dies during startup, before any CONTINUUM code runs.

The consequence is split:

- **4 tests fail loudly** on Windows.
- **1 test passes vacuously** — `test_a_real_piped_process_emits_no_colour`
  asserts only the *absence* of ANSI escapes, which is trivially true of a
  process that produced no output at all. It reports green while verifying
  nothing.

This is a partially-applied fix, not a new discovery: commit `ea8c686`
("fix(benchmark): close SQLiteStorage handles so continuum benchmark runs on
Windows", #89) corrected this exact pattern in `tests/test_cli.py` and left a
comment naming the precise failure mode. Three other call sites were missed.

## Affected code

| Location | Helper | Effect on Windows |
|:--|:--|:--|
| `tests/test_reconciliation.py:255` | `_run_worker` | 3 tests fail |
| `tests/test_storage_concurrency.py:95` | inline `Popen` | 1 test fails |
| `tests/test_cli_colour.py:271` | inline `run` | passes vacuously |

All three pass a variant of:

```python
env={
    "PYTHONPATH": str(Path(__file__).resolve().parents[1] / "src"),
    "PATH": "/usr/bin:/bin",
},
```

## Reproduction

On Windows, at `62cf16e`:

```
uv sync --extra dev
uv run pytest tests/test_reconciliation.py tests/test_storage_concurrency.py -q
```

Actual:

```
FAILED tests/test_reconciliation.py::test_a_crash_after_the_effect_does_not_silently_duplicate_it
FAILED tests/test_reconciliation.py::test_after_reconciliation_the_run_continues_without_duplicating
FAILED tests/test_reconciliation.py::test_a_clean_run_performs_the_effect_exactly_once
FAILED tests/test_storage_concurrency.py::test_separate_processes_append_to_one_chain_without_corruption
```

Each failure bottoms out in the child's startup traceback:

```
File "...\Lib\asyncio\windows_events.py", line 8, in <module>
    import _overlapped
OSError: [WinError 10106] The requested service provider could not be loaded or initialized
```

### Demonstrating the vacuous pass

`tests/test_cli_colour.py::test_a_real_piped_process_emits_no_colour` currently
passes. Running its exact `subprocess.run` call directly shows why that pass is
meaningless:

```
returncode: 1
stdout: ''
stderr tail: OSError: [WinError 10106] The requested service provider could not be loaded or initialized
```

The CLI is never reached, so `assert not ANSI.search(result.stdout)` holds
because `stdout` is empty — not because colour was correctly suppressed.

## Expected behaviour

The child process inherits the parent environment plus `PYTHONPATH`, so the
subprocess tests exercise the same code on Windows that they exercise on Linux
and macOS. A test that spawns a process to check that process's output should
fail, not pass, when the process does not run.

## Root cause

`subprocess.run(..., env=X)` *replaces* the environment rather than merging it.
On Windows, `SystemRoot` is required for the interpreter to initialise Winsock;
without it `import _overlapped` raises `OSError` during `import asyncio` at
interpreter startup. The hardcoded `PATH` was never needed in the first place:
`sys.executable` and the script path are both absolute, so nothing is resolved
via `PATH`.

## Impact

- **Blocks #200** (`ci: extend test matrix to macOS and Windows`). Its
  `windows-latest` jobs fail on Python 3.11, 3.12 and 3.13 for exactly these
  four tests, while all `ubuntu-latest` and `macos-latest` jobs pass:
  [run 32572184611](https://github.com/Cyrax321/CONTINUUM/actions/runs/32572184611/job/97029136404).
- The reconciliation tests cover the exactly-once side-effect guarantee — that
  a crash after an effect is not silently duplicated. That is one of the
  project's headline claims, and it is unverified on Windows.
- `test_storage_concurrency.py` covers the multi-process append path described
  in its own docstring as "the worst possible failure, because it looks
  correct". Also unverified on Windows.
- The colour test's silent pass is the more dangerous of the two modes: it
  makes the suite report coverage of the documented contract in `STATUS.md`
  ("piped output is byte-identical to uncoloured output") that it is not
  providing on that platform.

## Environment

- Windows 11 (10.0.26200), Python 3.13.1, `uv 0.12.5`, repo at `62cf16e`
- Independently reproduced by CI on `windows-latest` / Python 3.11–3.13

## Proposed direction

Build the child environment by extending the parent rather than replacing it,
following the pattern and rationale already recorded in `tests/test_cli.py`:

```python
env = {**os.environ, "PYTHONPATH": str(Path(__file__).resolve().parents[1] / "src")}
```

Separately, `test_a_real_piped_process_emits_no_colour` should assert that the
child actually ran — checking the exit code and the real output, not only the
absence of escapes — so this class of failure can never again present as a
pass. Since the same pattern has now been fixed twice in two places, a single
shared test helper for spawning CONTINUUM child processes may be worth
considering to stop it recurring, though that is a broader change than the fix
itself requires.

abyyxhek:
## Summary

**What was broken.** Three subprocess-spawning test helpers built the child
environment from a bare `env={...}` dict containing a hardcoded
`"PATH": "/usr/bin:/bin"`. `subprocess` *replaces* the environment with that
dict rather than merging it, so the child lost every other variable.

**Why it happened.** On Windows, `SystemRoot` is required for the interpreter to
initialise Winsock. Without it, `import asyncio` fails at interpreter startup
with `OSError: [WinError 10106]` on `import _overlapped` — before any CONTINUUM
code runs. The hardcoded `PATH` was never needed in the first place: both
`sys.executable` and the script path are absolute, so nothing is resolved
through `PATH`.

The failure split two ways:

- **4 tests failed outright** — 3 in `test_reconciliation.py`, 1 in
  `test_storage_concurrency.py`.
- **1 test passed vacuously** — `test_a_real_piped_process_emits_no_colour`
  asserted only the *absence* of ANSI escapes, which is trivially true of a
  process that produced no output at all. It reported green while verifying
  nothing.

This was a partially-applied fix. Commit `ea8c686` (#89) corrected this exact
pattern in `tests/test_cli.py` and left a comment naming this precise failure
mode; three other call sites were missed.

**What changed.**

- `tests/test_reconciliation.py`, `tests/test_storage_concurrency.py`,
  `tests/test_cli_colour.py`: build the child environment as
  `{**os.environ, "PYTHONPATH": ...}`, matching the pattern already established
  in `tests/test_cli.py`.
- `test_a_real_piped_process_emits_no_colour` now asserts that the CLI actually
  ran: its exit code and stdout must match the uncoloured in-process render byte
  for byte. That is the guarantee `STATUS.md` documents ("piped output is
  byte-identical to uncoloured output"), and it makes this class of failure
  impossible to mistake for a pass again.

No source or dependency changes; tests only.

## Regression evidence

The regression is platform-specific by nature — these tests pass on POSIX both
before and after. Verified on Windows 11, Python 3.11 via `uv`:

**Before the fix**, with the strengthened colour assertions applied but the old
`env=` left in place:

```
$ uv run pytest "tests/test_cli_colour.py::test_a_real_piped_process_emits_no_colour" -q
E       assert 1 == 20
        OSError: [WinError 10106] The requested service provider could not be loaded or initialized
FAILED tests/test_cli_colour.py::test_a_real_piped_process_emits_no_colour
```

The child exits `1` instead of the expected `20` (`REQUIRES_HUMAN`). Under the
old assertions this same dead child produced a **pass**.

**Before the fix**, the four loud failures:

```
FAILED tests/test_reconciliation.py::test_a_crash_after_the_effect_does_not_silently_duplicate_it
FAILED tests/test_reconciliation.py::test_after_reconciliation_the_run_continues_without_duplicating
FAILED tests/test_reconciliation.py::test_a_clean_run_performs_the_effect_exactly_once
FAILED tests/test_storage_concurrency.py::test_separate_processes_append_to_one_chain_without_corruption
```

These match the `windows-latest` failures on #200 exactly
([run 32572184611](https://github.com/Cyrax321/CONTINUUM/actions/runs/32572184611/job/97029136404)).

**After the fix**, all five pass.

## Testing

All commands run on Windows 11 (10.0.26200), Python 3.11–3.13 via `uv 0.12.5`.

| Command | Result |
|:--|:--|
| `uv run pytest tests/test_reconciliation.py tests/test_storage_concurrency.py tests/test_cli_colour.py -q` | 58 passed |
| `uv run pytest` (full suite) | **1085 passed, 12 skipped** |
| `uv run ruff check src/ tests/ examples/` | All checks passed |
| `uv run ruff format --check src/ tests/ examples/` | 173 files already formatted |
| `uv run mypy src/continuum` | Success: no issues found in 85 source files |

For reference, the same full suite on this machine before the fix: **4 failed**,
with `test_a_real_piped_process_emits_no_colour` passing vacuously.

## Note on #200

This unblocks the `windows-latest` jobs in #200 (`ci: extend test matrix to
macOS and Windows`). The two changes are complementary and independent: #200 adds
the platform to CI, this PR makes the tests correct on it. Neither supersedes the
other, and this PR does not touch `.github/workflows/ci.yml`.

## Possible follow-up

The same pattern has now been fixed twice in two separate commits. A single
shared helper for spawning CONTINUUM child processes in tests would stop it
recurring a third time, but that is a broader change than this fix needs, so I
left it out. Happy to add it if you'd prefer.

Fixes #211

## Summary by CodeRabbit

* **Tests**
  * Improved subprocess test reliability across different environments.
  * Ensured command-line color output is compared accurately with in-process results.
  * Preserved the surrounding environment when running reconciliation and storage concurrency tests.


anya-research:
**Code Review**

This PR correctly fixes a Windows-only failure where three test helpers replaced the child environment with a bare `env={"PYTHONPATH": ..., "PATH": "/usr/bin:/bin"}`, dropping `SystemRoot` and causing `import _overlapped` to fail with `WinError 10106` before any project code ran. It adopts the `{**os.environ, "PYTHONPATH": ...}` pattern already established in `tests/test_cli.py:591` and strengthens the vacuous `test_a_real_piped_process_emits_no_colour` to compare exit code and stdout byte-for-byte against the in-process render, matching the guarantee documented in `STATUS.md`. The change is tests-only, narrow, well-evidenced with before/after Windows logs, and consistent across all three call sites. No source or dependency changes, no API misuse, and no new failure modes introduced. This is a high-quality, low-risk follow-up to #89.

No blocking or medium-severity defects found. The diff is clean. What was done well: removal of the unnecessary hardcoded `PATH`, reuse of absolute `sys.executable`/script paths, explicit `os` import in each file, and the anti-vacuous assertion (`result.returncode == expected_code` + `result.stdout == expected_out`) that would have caught this class of child-death bug previously.

[github run](/Cyrax321/CONTINUUM/actions/runs/32645108697)

Cyrax321 (approved):

The diagnosis is correct and matches the pattern established in ea8c686 for tests/test_cli.py: a bare env= dict replaces the parent environment, which on Windows drops SystemRoot and kills the child during interpreter startup before any project code runs. Extending os.environ with only PYTHONPATH is the right fix at all three call sites, and the byte-for-byte assertion in the colour test closes a real vacuous-pass gap: asserting only ANSI absence is trivially true of a process that produced no output, so a dead child could previously present as a pass.

Verified locally: the three touched test files pass (58 passed), all required CI checks are green on the PR head, and the branch merges into current main cleanly.

No blocking issues found.

Issue #207, opened 2026-08-22
Durability gap: work performed between voluntary checkpoints is invisible to recovery (observed live with Claude Code)


Cyrax321:
### Affected component

MCP server

### What problem does this solve?

CONTINUUM's recovery guarantees depend on the agent voluntarily calling `continuum_record_progress` / `continuum_checkpoint`. Between two voluntary calls there is an unbounded window in which real work exists on disk but CONTINUUM has no durable knowledge of it. A kill inside that window loses everything from the recovery perspective, and worse, the resumed session is handed a *misleading* contract (progress 0/1) that understates what actually landed.

Reproduced live on 2026-08-22 against a real Claude Code session (run `tic-tac-toe`, goal "Create a tic-tac-toe game using HTML, CSS, and JavaScript"):

1. Session 1 wrote `tic-tac-toe.html` (276 lines) and was hard-killed immediately after the Write tool completed.
2. The event log contains only 2 events and 0 checkpoints:

```
sequence 1  RUN_STARTED   goal asserted by external_agent
sequence 2  TASK_UPDATED  completed=0 pending=1 failed=0
```

3. Session 2 called `continuum_resume`, got `mode=request_human`, `safe=false`, progress 0/1, and had no way to learn that the artifact was fully written. Recovery worked exactly as designed and still lost the run.

The event log proves the gap precisely: the agent died in the interval between performing work and reporting it. No amount of prompt engineering closes this; it is structural for any out-of-process MCP client where CONTINUUM cannot wrap the agent loop.

This is a recognized problem class. Durable-execution engines (Temporal, Restate, DBOS, Inngest) close it by making recording mandatory at the runtime layer: every step result is journaled before execution proceeds, so the agent cannot forget. LangGraph checkpointers snapshot automatically per super-step. The Crab paper (arXiv 2604.28138, semantics-aware checkpoint/restore for agent sandboxes) checkpoints sandbox filesystem state at turn boundaries. The common principle: the unit of durability is the observation, recorded as it happens, not when the agent feels like it. None of these apply directly to CONTINUUM's niche: an MCP server serving arbitrary coding CLIs it does not control.

### Proposed solution

Convert voluntary recording into mandatory interception using Claude Code hooks, which fire outside the model's control:

1. New CLI subcommand `continuum observe`: reads one structured tool-call observation on stdin (tool name, mutated paths, exit status), appends an evidence event to the active run's chain. Cheap, non-interactive, safe to call after every tool use.
[...]

PR #252, opened 2026-08-23
fix(windows): make hooks, the test suite and the smoke script portable

Parthipashok04:
## Description

Five Windows-only defects, found while getting the `continuum-mcp` server running
on Windows 11. The server itself turned out to be fine — it completes the
`initialize` handshake and lists all 11 tools over stdio — but almost everything
around it assumed a POSIX host.

**1. `clienthooks`: the installed hook command is unrunnable — the one product bug here.**

`observe_command()` joined argv with `shlex.quote()`, which is POSIX-only. Any
project path containing a space came out single-quoted:

```
'D:\pROJ OPEN\CONTINUUM\.venv\Scripts\python.exe' -m continuum.cli observe
```

`cmd.exe` has no single-quote syntax, so it tried to run a program called
`'D:\pROJ` and failed with *"The filename, directory name, or volume label
syntax is incorrect."* **Every hook `continuum hooks install` wrote on Windows
was silently dead** — which disables precisely the auto-checkpoint durability the
hook exists to provide, so the issue #207 window stays wide open on the platform.
Silently, because a failing `PostToolUse` hook does not interrupt the session.

Joining now goes through `_join_command()`, which uses
`subprocess.list2cmdline()` on Windows — the quoting convention the Windows C
runtime itself parses — and the existing `shlex.quote` join everywhere else.

The parsing side had to move with it, and this is the part worth reviewing:
`shlex.split()` in POSIX mode reads the backslashes in a Windows path as escape
characters, so an unquoted `C:\Py\Scripts\continuum.exe` comes back as
`C:PyScriptscontinuum.exe`. `_is_continuum_hook()` would then stop recognising
the very commands `_join_command()` writes — `hooks remove` would report success
while removing nothing, and `hooks install` would quietly append a duplicate
beside the old entry. `_split_command()` therefore uses `posix=False` on Windows
and strips both quote characters, which has the useful side effect that a hook
written by an *older* CONTINUUM — POSIX-quoted even on Windows — is still
recognised and can still be cleaned up after an upgrade.

**2. `test_reconcilers`: `printf` is not a `cmd.exe` command (7 failures).**

`reconcilers.py` runs probes through `subprocess.run(..., shell=True)`, which is
`cmd.exe` on Windows, and `cmd.exe` has no `printf`. The production code is
portable — it runs whatever the operator configures — but ten test probes
hardcoded it. They use `echo` now, a builtin in both shell families;
`_parse_verdict` already strips each line, so the trailing newline changes
nothing.

**3. Bare `"python"` in test subprocesses (2 failures).**

`test_cli_observe.py` and `test_client_installers.py` spawned
`["python", "-m", "continuum.cli", ...]`. On Windows bare `python` resolves to the
WindowsApps Store alias or the system interpreter rather than the venv running
pytest, so the child died with `No module named 'continuum'` — which breaks the
venv workflow the README documents. Now `sys.executable`.

**4. `mcp_smoke`: hardcoded `/tmp`.**

`DB_PATH = "/tmp/mcp-smoke-demo.db"`. There is no `/tmp` on Windows, so the smoke
test the README advertises could not start at all — it died at the `initialize`
handshake with *"cannot open storage at '/tmp/mcp-smoke-demo.db': unable to open
database file"*. Derived from `tempfile.gettempdir()` instead, cleanup included.

**5. `mcp_smoke`: `UnicodeEncodeError` on a redirected stream.**

Surfaced immediately by fixing 4. The transcript is drawn with box-drawing
characters and em dashes, and Python selects the locale encoding for a
*redirected* stdout — `cp1252` on Windows, `ascii` under `LANG=C` — so the script
died at the first banner as soon as its output was piped anywhere, which is what
CI and any log capture do. `stdout` is reconfigured to UTF-8; interactive
consoles already speak it, so only the redirected case changes.

## Related issues

No existing issue covers these five — they surfaced during an install on Windows
11, and I am reporting them here with the fix rather than filing separately.

Refs #200. That PR adds `windows-latest` to the test matrix but carries no
portability fixes, so these are fixes it needs before it can go green.

**Not** related to #211 / #212, despite the shared theme, and there is no
overlap: #212 changes `test_reconciliation.py`, `test_storage_concurrency.py` and
`test_cli_colour.py`; this PR changes `test_reconcilers.py` (a different file),
`test_cli_observe.py`, `test_client_installers.py`, `src/continuum/clienthooks.py`
and `scripts/mcp_smoke.py`. The two should merge in either order. Worth noting
that #211's four failures do **not** reproduce on this environment — the child
interpreter starts fine without `SystemRoot` on Python 3.12 here, and those tests
pass — so that one looks environment-dependent, which makes #212 a good fix to
land regardless.

## Types of changes

- Type: `bugfix`

## Breaking changes

No. Behaviour on macOS and Linux is byte-identical: every platform-specific
branch is gated on `os.name == "nt"`, and the POSIX path keeps the existing
`shlex.quote` / `shlex.split` pair. The new `_split_command()` also accepts the
old POSIX-quoted form on Windows, so a settings file written by an earlier
version keeps working and remains removable — no migration needed.

## How has this been tested?

Windows 11 (10.0.26200), Python 3.12, venv created per the README
(`pip install -e ".[dev,mcp]"`).

Full suite:

```
pytest
1247 passed, 13 skipped in 85.44s
```

The three gates CI enforces:

```
ruff check src/ tests/ examples/          -> All checks passed!
ruff format --check src/ tests/ examples/ -> 198 files already formatted
mypy src/continuum                        -> Success: no issues found in 96 source files
```

The smoke script, which previously could not start:

```
python scripts/mcp_smoke.py
  ...
  SMOKE TEST PASSED
    handshake ok · 11 tools · progress durable ·
    side effect performed exactly once ·
    agent-reported state correctly withheld from 'resume'
```

Before the fixes, on the same machine: 9 real failures (7 reconciler probes,
2 bare-`python`), plus a smoke script that aborted at the handshake. The product
bug was confirmed directly rather than only through tests — `cmd /c` on the
generated command returned exit 1 with *"The filename, directory name, or volume
label syntax is incorrect"* before, and exit 0 with real CLI output after.

Two failures I want to be explicit about **not** having "fixed", because they are
artifacts of my environment and not defects: `test_cli_colour.py`'s two colour
tests fail if `NO_COLOR` is set in the parent shell (mine sets it; the numbers
above are from a run with it unset), and
`test_plugins.py::test_git_provider_reads_head_in_a_repo` needs `git` on `PATH`.
Both pass once the environment is corrected, so I left them alone.

## Checklist

- **Tests added or updated for the changed behaviour.** Two new regression tests
  in `tests/test_cli_observe.py`: one round-trips the generated command back
  through `_split_command()` with a space forced into the path, and asserts
  `_is_continuum_hook()` still recognises it — written as a round trip rather
  than a literal string so it stays meaningful on POSIX, where the correct
  quoting is the other one. The second pins recognition of the legacy
  POSIX-quoted form, which is what keeps `hooks remove` working across an
  upgrade. The pre-existing
  `test_installed_command_actually_records_through_the_real_entrypoint` already
  drives the baked command through a real shell, so it now fails loudly on
  Windows if the quoting regresses.
- **Documentation updated where the change affects user-facing behaviour.** No
  user-facing behaviour or documented interface changes — the hook command's
  *content* is identical, only its quoting is now correct for the host shell. The
  one docstring that described the old `/tmp` location was updated.
- **CI passes on the latest commit.** All three gates run clean locally (above).
  Note that CI currently runs `ubuntu-latest` only, so it will not exercise the
  Windows paths these fixes address until #200 lands.
- **Security-relevant changes state known limitations explicitly.** Nothing
  security-relevant. One adjacent note: `_split_command()` stripping quote
  characters is for *recognition* only — it never feeds a reconstructed string
  back to a shell, so it cannot widen what gets executed. `_is_continuum_hook()`
  stays as narrow as before, matching only a `continuum` executable stem or the
  `<python> -m continuum.cli` fallback form, so neither install nor remove can
  touch a third party's hook entry.

## Additional context

`.mcp.json` is deliberately untouched. Its bare `"command": "continuum-mcp"` is
correct once the package is installed — it just has to resolve on the `PATH`
Claude Code inherits, which is why the server connected on macOS (venv installed
and active) and not on Windows (nothing installed at all). Hardcoding an absolute
Windows path there would break macOS and Linux. **The MCP server needed no code
change; getting it connected is an install step, not a repo change.**

Fix 3 has a small piece of live evidence: while preparing this PR, a `gh pr list`
piped into bare `python` on this machine failed with *"Python was not found; run
without arguments to install from the Microsoft Store"* — the same App Execution
Alias interception the test fix is about.

Happy to split this into per-fix commits if that suits review better; I kept it
as one commit because the four smaller fixes are individually trivial and share a
single root cause — POSIX assumptions in a cross-platform tool — while the
`clienthooks` change is the only one that needs real scrutiny.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hook command handling across Windows and POSIX environments.
  * Preserved compatibility with existing POSIX-quoted hook commands.
  * Made smoke-test database and cleanup paths work across platforms.
  * Ensured redirected output supports Unicode characters.

* **Tests**
  * Improved cross-platform CLI and hook coverage.
  * Updated integration tests to use the active Python interpreter.
  * Corrected probe matching in reconciliation and dry-run scenarios.


anya-research:
**Code Review**

This PR fixes five genuine Windows-portability defects with a narrow, well-reasoned diff. The core `clienthooks` quoting fix correctly switches to `subprocess.list2cmdline()` on Windows and repairs the inverse `shlex.split(..., posix=False)` backslash-eating bug, while preserving the POSIX `shlex.quote` path and adding legacy-quote compatibility. The `printf`->`echo` and `sys.executable` test fixes and the `tempfile.gettempdir()`/`utf-8` smoke fixes are all correct and low-risk. Overall quality is high, the defect analysis in the description is accurate, and the added round-trip tests target the right property, with only the test-coverage gaps noted below keeping this from being immediately green on `windows-latest`.

**medium**
`tests/test_cli_observe.py:296` and `tests/test_cli_observe.py:320`
The two new regression tests never exercise the Windows branch they claim to pin. `observe_command()` and `_split_command()` gate on `os.name == "nt"`, but the tests only monkeypatch `shutil.which` and run on `os.name == "posix"` in CI, so `list2cmdline` and `posix=False` are never executed there. The `spaced` path with forward slashes round-trips correctly under POSIX `shlex.quote`/`shlex.split`, masking a regression in the Windows path. Mock `os.name` in the test.

```suggestion
def test_the_baked_command_survives_a_round_trip_through_the_host_quoting(
    monkeypatch: pytest.MonkeyPatch,
) -> None:
    """Quoting must match the shell that will run the hook, on either platform.
    """
    spaced = str(Path("D:/pROJ OPEN/CONTINUUM/.venv/Scripts/continuum.exe"))
    monkeypatch.setattr("continuum.clienthooks.os.name", "nt")
    monkeypatch.setattr("continuum.clienthooks.shutil.which", lambda _: spaced)
```

**low**
`scripts/mcp_smoke.py:324`
Cleanup uses `shutil.rmtree()` on a file path (`DB_PATH + "-journal"`). `rmtree` expects a directory and, with `ignore_errors=True`, silently does nothing, leaking the journal file across smoke runs. The start-up cleanup also only removes `-wal`/`-shm` via `unlink` but never `-journal`. Use `Path.unlink`.

```suggestion
        Path(DB_PATH + "-journal").unlink(missing_ok=True)
```

**low**
`tests/test_reconcilers.py:156`
PR converts 7 `printf` probes to `echo` but leaves `sleep 30` and `exit 3` in `test_probe_failures_never_settle`. `sleep` is not a `cmd.exe` builtin (Windows uses `timeout /t`) and will fail immediately as "command not found" rather than exercising the `TimeoutExpired` path, and `exit 3` semantics differ (`exit /b`). The test still passes because any `error` is `unresolved`, but the timeout branch remains uncovered on Windows. Keep the intent portable, e.g. a Python sleep.

```suggestion
        "timeout": f'"{sys.executable}" -c "import time; time.sleep(30)"',
```



[github run](/Cyrax321/CONTINUUM/actions/runs/32657171647)

Cyrax321:
Thank you for this contribution, @Parthipashok04, and welcome to CONTINUUM.

This is a well-executed PR with excellent attention to detail. A few things stand out:

- **Root-cause analysis:** Each of the five fixes traces back to its underlying POSIX assumption rather than patching symptoms. The description clearly explains why each failure occurs on Windows, which makes review straightforward.
- **The clienthooks fix is the most important change here.** The `_join_command()`/`_split_command()` pairing using `subprocess.list2cmdline()` on Windows with `posix=False` parsing is correct, and keeping legacy POSIX-quoted entries recognizable so `hooks remove` continues to work after an upgrade shows good care for backward compatibility. This was silently breaking auto-checkpoint durability on Windows, so it is arguably the most significant product bug fixed in this PR.
- **Testing discipline:** The round-trip tests are written to stay meaningful on both platforms, environment-specific failures were disclosed explicitly rather than worked around, and all three CI gates pass locally along with a full suite run.
- **Scope hygiene:** Documenting what is not related (#211/#212) and why `.mcp.json` was deliberately left untouched saves maintainers real time during review.

A few items to address before merge, based on the automated reviews:

1. **Test coverage (medium):** The new regression tests do not exercise the `os.name == "nt"` branch on Linux CI. Please monkeypatch `os.name` so that `list2cmdline` and `posix=False` are actually covered.
2. **Journal cleanup (low):** `shutil.rmtree()` on the `-journal` file is a no-op with `ignore_errors=True`. Use `Path.unlink(missing_ok=True)` instead.
3. **Timeout probe portability (low):** `sleep 30` still does not exist under cmd.exe. A Python-based sleep would keep the `TimeoutExpired` path exercised cross-platform.
4. **Smoke DB uniqueness (minor):** Consider a unique filename per run to avoid collisions between concurrent smoke test executions.

None of these block approval. They are polish items on top of already-correct work.

Thank you again for fixing these defects, verifying against a real Windows 11 environment, and documenting exactly what needs maintainer scrutiny versus what is mechanical. Contributions like this make the project better for everyone running it outside macOS and Linux. Looking forward to your next PR.

yuki-fuyutsuki (approved):

Cyrax321 (approved):

Cyrax321 (approved):
Re-reviewed the new commits after the push. The round-trip tests, legacy-quote recognition, and journal cleanup all hold up. Approving to restore the merge state.

Issue #235, opened 2026-08-23
Open problem: reasoning-context rehydration - resume must inherit the plan, not just the progress


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

### The open problem

Recovery today restores *task state* (goal, progress, findings) and proves *what landed on disk*, but a resumed session is safe yet amnesiac about its own reasoning: the plan stack it was following, the decisions it had made and why, the threads it had open, the working set of files it cared about. Every external evaluation of agent recovery (SWE-bench Pro failure analysis; Anthropic's harness post on context anxiety) shows agents fail by losing orientation before they lose data - endless file re-reading, premature wrap-up, wrong-problem patches.

AgentRewind (arXiv, Recoverable Execution for Long-Horizon LLM Agents) names the requirement precisely: recovery needs **aligned checkpoints of agent context AND controlled environment**, so resume inherits reasoning state from prior attempts instead of restarting cognition from a progress bar.

### Why CONTINUUM should own this

We already have the transport: SessionStart briefing injects run state deterministically (#227), observation hooks capture file truth (#210), and human_steps render next actions. What is missing is the middle layer - a compact, provenance-tagged *reasoning summary* captured at turn boundaries and served back on resume.

### Step-by-step plan

1. Define a `REASONING_SUMMARY` event type: bounded JSON (plan stack, last N decisions with one-line rationale, open questions, working-set paths), written at Stop/PostToolUse boundaries through the existing hook transport.
2. Extend `continuum briefing` to include the newest summary verbatim.
3. Cap size (e.g. 4 KB) with oldest-first eviction inside the summary itself; never store raw transcripts.
4. Validator treats summaries as EXTERNAL_AGENT evidence: informational on resume, never gating.

### Scope guardrails

No raw conversation capture, no vector stores, no transcript dumps - those are memory systems, and collapsing durability into chat history is the exact trap the field keeps documenting. Summaries are authored by the agent about its own plan, not recorded from context.

### Acceptance criteria

- Kill a session mid-plan; fresh session's briefing contains the dead session's plan stack and open questions.
- Summary absence degrades gracefully (briefing unchanged apart from the missing block).
- Size cap enforced; fuzz test with adversarial payloads stays JSON-bounded.

PR #256, opened 2026-08-23
docs(mcp): document the CONNECTION_CLOSED failure mode and the eleventh tool

Parthipashok04:
## Description

Registering `continuum-mcp` with Claude Code can fail with a message that points away from the actual cause:

```
❯ /mcp
⎿ Failed to reconnect to continuum-mcp: CONNECTION_CLOSED
```

`docs/api/mcp.md` said this about registration:

> With the project's `.mcp.json` present, Claude Code registers the server automatically and every agent action is checkpointed, validated, and recorded as it happens.

Both halves of that sentence overstate what registration does.

**1. Registration depends on `PATH`.** `.mcp.json` declares the server by bare command name (`"command": "continuum-mcp"`). The client resolves that name against the `PATH` it inherited, so registration only succeeds when the environment CONTINUUM was installed into is on that `PATH`. When it is not, the client never starts the server and reports the failed spawn as `CONNECTION_CLOSED` — wording that reads like a server which crashed, but which describes an executable that was never found.

The consequence that matters for this repository: **no CONTINUUM code runs in that state.** The executable is never located, so the server cannot detect the condition, report it, or recover from it. It cannot be fixed in `src/`. The entire diagnosis is client-side, which is why this PR is documentation rather than code.

CONTINUUM already learned this lesson elsewhere. `observe_command()` resolves the executable and bakes the absolute path into the hook it installs, and its docstring says exactly why:

> The absolute path of the `continuum` executable is resolved at install time and baked into the settings: hook processes may not inherit the PATH that found the binary originally.

`.mcp.json` cannot do the same thing — it is committed and shared, and an absolute path there would be wrong on every machine except the one that wrote it. The bare name is the right choice. What was missing was documenting what it requires.

**2. Registration does not instrument anything by itself.** The tools record state when the agent calls them; the `PostToolUse` hook installed by `continuum hooks install claude-code` is what records writes outside the model's control. That distinction is the premise `clienthooks.py` opens with, and the doc had collapsed it into "every agent action is … recorded as it happens".

### What this PR adds

- A **Registration** section in `docs/api/mcp.md` stating what the bare command name actually requires.
- A **Troubleshooting** section with the symptom, the diagnosis (`which` / `where.exe`, then `--help` through the full path), and two remedies — launch the client from the activated environment, or pin the absolute path with `claude mcp add --scope local`. Includes the platform note on why this surfaces on Windows most often without being a Windows-specific defect.
- Remedy 2 documents the conflicting-scopes diagnostic that local-scope registration produces, and warns against the resolution Claude Code itself suggests: `claude mcp remove continuum-mcp -s project` edits the committed `.mcp.json` and unregisters the server for everyone who clones the repository.

### Second defect found: the documented tool count

`continuum_record_summary` (#235) shipped with a CHANGELOG entry but never reached the reference docs. A live `tools/list` returns **11** tools — 3 read-only, 8 mutating — while the docs said ten, three and seven. The tool table gains the missing row and the count is corrected in six files.

## Related issues

No existing issue. I hit this while installing the server on Windows and searched all 60 open and closed issues for a prior report before writing this up.

Adjacent but **not** addressed here: #211 (Windows test helpers replacing the child environment with a POSIX `PATH`) is a different `PATH` problem, inside the test suite rather than at client spawn.

Refs #235 — the tool whose documentation this backfills.

## Types of changes

- Type: `docs`

## Breaking changes

No. Documentation only; no source file is touched.

## How has this been tested?

Every factual claim in the new section was reproduced on this machine before it was written. Environment: Windows 11, Python 3.13 in `.venv`, Claude Code CLI, `pip install -e ".[mcp]"`.

**The failure, and that `PATH` is the cause** (PowerShell):

```powershell
where.exe continuum-mcp
# exit 1 — INFO: Could not find files for the given pattern(s).

.\.venv\Scripts\Activate.ps1
where.exe continuum-mcp
# exit 0 — D:\pROJ OPEN\CONTINUUM\.venv\Scripts\continuum-mcp.exe
```

That is the whole bug in two commands: the name the client is asked to spawn does not resolve until the environment is active. `claude mcp list` reproduced the client-side symptom as `✘ Failed to connect — CONNECTION_CLOSED`.

**That the server itself is not at fault.** A full JSON-RPC `initialize` → `notifications/initialized` → `tools/list` handshake driven against the absolute path returned a valid `result` frame, exit 0, and empty stderr. Repeating it with a stripped environment (`env -i`) also exits 0, which rules out the `.mcp.json` `env` block as a contributor.

**Both documented remedies.**

- Remedy 1 — activate, then launch the client from that shell: verified above; the bare name resolves to the venv executable.
- Remedy 2 — `claude mcp add continuum-mcp --scope local --env CONTINUUM_MCP_MUTATING_CLIENTS=claude-code -- "<abs path>\continuum-mcp.exe" --db continuum.db`. `claude mcp list` then reports `✔ Connected`, and the entry lands in `~/.claude.json` under `projects/<project path>/mcpServers` — per-user and uncommitted, as the doc states. The conflicting-scopes diagnostic appeared exactly as documented, naming both the project and local endpoints, with the local one in use.

**The corrected tool count**, from a live `tools/list` against this branch:

```
11 tools — read: continuum_validate, continuum_resume, continuum_list_actions
           mutate: continuum_record_progress, continuum_checkpoint,
                   continuum_record_summary, continuum_confirm,
                   continuum_intercept_action, continuum_complete_action,
                   continuum_fail_action, continuum_reconcile_action
```

**Gates** (the exact commands from `.github/workflows/ci.yml`):

```
ruff check src/ tests/ examples/           -> All checks passed!
ruff format --check src/ tests/ examples/  -> 198 files already formatted
mypy src/continuum                         -> Success: no issues found in 96 source files
```

Whole-tree `ruff format --check` reports `10 files would be reformatted, 259 files already formatted` both with and without this branch's changes — byte-identical before and after. All ten are pre-existing on `main`, all lie outside the paths CI checks, and `docs/api/mcp.md` is not among them.

No test was added: there is no behaviour change, and nothing under `tests/` asserts on documentation content (verified by grep).

## Checklist

- **Tests added or updated** — n/a. Documentation only, no behaviour change.
- **Documentation updated where the change affects user-facing behaviour** — this PR is that update.
- **CI passes on the latest commit** — the three CI gates were run locally and pass (above). Note that on my previous PR the CI run was held at `action_required` because I am a first-time contributor from a fork, so a maintainer may need to approve the workflow run.
- **Security-relevant changes state known limitations** — nothing security-relevant is changed, and the documented remedy is deliberately the conservative one: the absolute path goes in the per-user, uncommitted `~/.claude.json`, and the doc explicitly warns against editing the committed `.mcp.json`. The documented `claude mcp add` command retains `--env CONTINUUM_MCP_MUTATING_CLIENTS=claude-code`, with a note that omitting it yields a connected server exposing only the three read-only tools — so following these instructions does not weaken the deny-by-default allowlist.

## Additional context

**Alternatives considered and rejected.**

- *Changing `.mcp.json`* to an absolute path, or to `${VAR:-default}` expansion. Rejected: an absolute path is wrong on every other clone, and env-var expansion is not consistently supported across MCP clients. The bare name is correct; it just needed documenting.
- *Adding a `continuum mcp install` (or `doctor`) command* that resolves the executable and writes it into local scope — mirroring what `continuum hooks install` already does for hooks. This is the real ergonomic fix and it would make the failure self-diagnosing from the CLI side. I deliberately did not add it here: it invents new CLI surface, which is a maintainer's decision rather than a drive-by documentation PR's. Happy to open it as a follow-up if that is wanted.

**Left alone on purpose.** `references/testing.md` and `docs/research/token_floor.md` had the stale count too and are corrected, but I checked each occurrence for whether it was a live claim or a record of a past measurement before touching it — `test.md`'s audit records were left untouched, and they contain no tool count.

**Relationship to #252.** That PR fixes five Windows portability bugs in code (POSIX-only hook quoting, `printf` test probes, bare `python` in subprocesses, hardcoded `/tmp`, and stdout encoding under redirection). This PR is independent: branched from `main`, no overlapping files, and neither needs the other to land first.

## Summary by CodeRabbit

- **New Features**
  - Documented the newly available `continuum_record_summary` MCP tool.
  - Clarified that the MCP integration exposes eleven tools, including eight mutating tools.

- **Documentation**
  - Added setup and troubleshooting guidance for `CONNECTION_CLOSED` startup failures.
  - Documented PATH requirements, absolute-path registration, scope conflicts, authentication, and hook-based state recording.
  - Updated tool counts and token-cost references across integration, API, research, and testing documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -534,6 +534,32 @@ All notable changes to this project are documented here. The format follows
 
 ### Changed
 
+- **MCP docs: the `CONNECTION_CLOSED` failure mode, and the eleventh tool.**
+  `docs/api/mcp.md` promised that with `.mcp.json` present "Claude Code registers
+  the server automatically". That holds only when the environment CONTINUUM was
+  installed into is on the `PATH` the client inherited. When it is not, the client
+  cannot spawn `continuum-mcp` at all and reports the failed spawn as
+  `CONNECTION_CLOSED` - a message that reads like a server that crashed but
+  describes an executable that was never found. No CONTINUUM code runs in that
+  state, so the server cannot detect, report, or recover from it, and the whole
+  diagnosis has to happen client-side. The registration section now states what a
+  bare command name actually requires, and a new Troubleshooting section carries
+  the diagnosis (`which` / `where.exe`, then `--help` through the full path) and
+  two remedies: launch the client from the activated environment, or pin the
+  absolute path with `claude mcp add --scope local`. The second also documents the
+  conflicting-scopes diagnostic it produces and warns against
+  `claude mcp remove continuum-mcp -s project`, which edits the committed
+  `.mcp.json` and unregisters the server for every other clone. Registration is no
+  longer described as instrumenting anything on its own: state is recorded when
+  the agent calls the tools, or when the `PostToolUse` hook records a write
+  outside the model's control.
+- **MCP tool count corrected to eleven.** `continuum_record_summary` (#235)
+  shipped with a CHANGELOG entry but never reached the reference docs. It is now
+  in the `docs/api/mcp.md` tool table, and the count is corrected from ten to
+  eleven - three read-only, eight mutating - in `README.md`, `docs/api/README.md`,
+  `docs/research/token_floor.md`, `references/mcp.md`,
+  `references/auto-resume-integration.md`, and `references/testing.md`.
+
 
 - **README.** `Contents` laid out as a horizontal wrapping nav; Security
   Extension added to the Features table and table of contents; website link
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ uv pip install -e ".[mcp]"
 CONTINUUM_MCP_MUTATING_CLIENTS=your-client-name continuum-mcp
 ```
 
-Ten tools over stdio. Three are read-only (`continuum_validate`, `continuum_resume`, `continuum_list_actions`); seven mutate. Side effects are two-phase (claim, perform, complete), and mutating tools deny by default behind an allowlist. Agent-reported state is recorded with `Origin.EXTERNAL_AGENT` provenance and marked `REQUIRES_REVIEW`. Verification details, including crash recovery at startup and the end to end Claude Code test, are in [references/mcp.md](references/mcp.md). The authentication limitation is covered in [references/architecture.md](references/architecture.md) (MCP server and Security sections), and the MCP narrative is in [references/quickstart.md](references/quickstart.md).
+Eleven tools over stdio. Three are read-only (`continuum_validate`, `continuum_resume`, `continuum_list_actions`); eight mutate. Side effects are two-phase (claim, perform, complete), and mutating tools deny by default behind an allowlist. Agent-reported state is recorded with `Origin.EXTERNAL_AGENT` provenance and marked `REQUIRES_REVIEW`. Verification details, including crash recovery at startup and the end to end Claude Code test, are in [references/mcp.md](references/mcp.md). The authentication limitation is covered in [references/architecture.md](references/architecture.md) (MCP server and Security sections), and the MCP narrative is in [references/quickstart.md](references/quickstart.md). If a registered server reports `CONNECTION_CLOSED`, the cause is almost always `PATH` resolution rather than the server itself: [docs/api/mcp.md](docs/api/mcp.md#troubleshooting) has the diagnosis and two remedies.
 
 ## Framework Integration
 
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -31,7 +31,7 @@ may resume. The integration points, in order of most to least common, are:
 - [Checkpoints](checkpoints.md) - CheckpointManager, RestoredRun
 - [Recovery](recovery.md) - RecoveryEngine, RecoveryDecision
 - [Validation](validation.md) - StateValidator, validate_state, ValidationOutcome
-- [MCP server](mcp.md) - continuum-mcp, the ten tools, build_server
+- [MCP server](mcp.md) - continuum-mcp, the eleven tools, build_server
 - [Security](security.md) - attestation and caller authentication
 - [CLI](cli.md) - the `continuum` command reference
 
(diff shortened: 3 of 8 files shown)


Cyrax321:
Thank you for following up so quickly after #252, @Parthipashok04. This is a valuable addition.

This is exactly what good documentation work looks like. A few things stand out:

- **Verification before authorship:** Every claim was reproduced on a real Windows 11 environment before it was written down, including the two-command proof that PATH resolution is the cause (`where.exe` failing until activation), a full JSON-RPC handshake against the absolute path to rule out server-side faults, and live verification of both documented remedies. That level of rigor is rare in docs-only PRs.
- **Correct diagnosis of scope:** Recognizing that this cannot be fixed in `src/` because no CONTINUUM code runs when the executable is never found is an important insight, and the parallel you drew to how `observe_command()` already solves this for hooks explains the design tension well.
- **Restraint:** Deliberately not adding a `continuum mcp install` command because it invents new CLI surface and belongs in a maintainer-approved follow-up shows good judgment about contribution boundaries. I agree it would be the right ergonomic fix eventually; please do open that as a separate proposal.
- **Careful cleanup:** Correcting the stale tool count across six files while checking each occurrence to distinguish live claims from historical audit records is thorough, and backfilling the missing `continuum_record_summary` documentation closes a gap left by #235.

A few small items from the automated review to address:

1. **Em dashes:** The changed files contain em dash characters at several points. Per the repository guidelines, replace each with a comma, period, or parenthesis.
2. **Fenced code blocks:** Six blocks are missing language identifiers (markdownlint MD040). Add `text` or the appropriate shell language to each.
3. **The `--help` check overstates its guarantee:** `continuum-mcp --help` confirms executable resolution and argument parsing only. It does not exercise the MCP dependency, policy loading, database access, or stdio startup. Please reword it as a resolution check rather than a health check.
4. **Qualify the README diagnosis:** "Almost always PATH resolution" should direct readers to inspect server startup errors when resolution succeeds, since startup failures can also come from a missing `mcp` package, invalid policy, or SQLite errors after the executable resolves.

None of these change my assessment. They are wording and lint fixes on top of accurate, well-researched documentation.

Thank you again for searching the full issue history before filing, keeping this independent of #252 with no overlapping files, and documenting the conservative remedy that keeps the deny-by-default allowlist intact. Docs like these save every future Windows user the debugging session you went through.

yuki-fuyutsuki (approved):

Issue #239, opened 2026-08-23
Open problem: event-log compaction and large-payload offloading for month-long runs


Cyrax321:
### Affected component

Core (events, storage)

### The open problem

The event log is append-only forever. Long runs accumulate hundreds of thousands of events; replay cost grows linearly; large payloads (file digests are small today, but reasoning summaries and future OTel attributes will not be) bloat rows. Temporal solved this class with payload codecs and blob offloading; CONTINUUM needs the same discipline before anyone trusts it for month-long runs.

### Step-by-step plan

1. Define retention policy per run: keep anchors (checkpoints), last N events, all ACTION_* and RECOVERY_* events (they are the audit core), compact the rest into periodic digest events.
2. Large-payload offloading: payloads above a configurable size are stored out-of-band (filesystem or S3-compatible URL) with sha256 recorded inline; accessors rehydrate transparently.
3. `continuum verify` learns to verify compacted runs: anchors plus post-anchor events must fully project.
4. Bench: replay latency and db size vs event count before/after at 100k/1M events.

### Scope guardrails

Never delete ACTION_*, RECOVERY_*, REVIEW_* or approval events - the audit chain is the product. Compaction only folds redundant progress/evidence detail behind anchors.

### Acceptance criteria

- A 1M-event synthetic run compacts to bounded size while `verify`, `resume` and `replay` still pass; benchmarks included in PR.

Cyrax321:
Core delivered: `continuum compact <run>` archives the pre-anchor prefix verbatim into events_archive (schema v5, both engines) behind an EVENT_LOG_ANCHORED marker, keeping the live chain append-only while bounding size; verify walks anchored logs; replay folds restored checkpoint + tail; resume unaffected. The anchor is freshly forced at compaction time because anchoring at an ancient version left nearly everything live (caught by the bounded-size test).

Split to a focused follow-up rather than bundled: large-payload offloading (blobs above a threshold moved out-of-band with inline sha256) - tracked in a new issue so this closure is honest about scope.

PR #260, opened 2026-08-23
fix(storage): harden compaction per the #253 review

Cyrax321:
## Description

Addresses all nine CodeRabbit findings posted on #253 after merge. The two
major ones were real defects, one critical:

1. **Anchored verification trusted the first live row (critical, CWE-345).**
   `verify_events` on both engines treated the first surviving row of a run
   with an EVENT_LOG_ANCHORED event as a trusted genesis, adopting that row's
   own `prev_hash` and sequence as the walk's starting point. Deleting the
   boundary events left a "valid" chain and verify returned success for a run
   whose anchor era had been erased. Both engines now read the newest
   `events_archive` row and require the live chain to continue it exactly,
   enforce SEQUENCE_GAP and BROKEN_CHAIN on anchored logs like any other, and
   deep-audit the archive itself in the same pass (every archived row is
   re-digested and chain-linked from sequence 1), so tampering with,
   truncating or emptying the archive fails verify instead of hiding behind a
   healthy live tail.

2. **`compact_run` committed the anchor marker outside the archive move.** A
   crash between the two commits left an anchored live log whose prefix never
   reached the archive. The marker append now shares one transaction with the
   archive move: SQLite through a conn-taking `_append_chained` helper on the
   open IMMEDIATE transaction, Postgres through an explicit `transaction()`
   block around the three statements (the connection runs autocommit). Both
   docstrings now describe the real order (forced anchor checkpoint first).

3. **`replay` reported `verified: true` unconditionally for compacted runs.**
   The anchored branch never compared anything; the corruption-detection
   contract was lost for every compacted run. It now re-folds only the stored
   version's own prefix and compares fingerprints exactly as the plain path's
   `_verify_against_stored` does, returning CORRUPTED with `verified: false`
   on mismatch. A dead `base is None` test was removed (`restore` always
   returns a state).

4. **Exactly-once reset at the compaction boundary.** The action ledger folded
   only live events, so archived ACTION_* claims vanished from the fold and a
   month-old completed side effect could be claimed fresh after compaction.
   `ActionLedger._replay` now folds archived events too via new
   `Storage.read_archived_events` (empty by default, implemented by both
   engines), `protected_call` takes its decision fold from the ledger, and an
   archived completed action is a SKIP_DUPLICATE cache hit whose callback
   never re-runs.

The remaining items: EVENT_LOG_ANCHORED added to `_NON_PROJECTING` so it no
longer counts under `ProjectionReport.ignored_types`; `compact` gates on a new
`supports_compaction` capability flag (mirroring `supports_action_index`)
instead of surfacing a raw NotImplementedError from engines without the table;
the duplicated `MIGRATIONS` comment line removed; the incomplete "found live"
clause in the changelog fixed.

## Related issues

Refs #239 (compaction). Review findings were posted inline on #253; there is
no separate issue for them, hence no Fixes line.

## Types of changes

- Type: `bugfix`

## Breaking changes

No. Verification of anchored logs becomes stricter, which can turn a
previously-passing `verify` into CORRUPTED, but only for logs whose archive or
boundary rows are actually missing or altered, i.e. genuinely damaged history.
No schema change; v5 unchanged.

Known limitation, stated explicitly: deleting the newest live row (the head)
of any hash chain, compacted or not, remains undetectable by chain walking
alone; `continuum attest` / `attest-verify` over the signed head is the
control for that, unchanged here.

## How has this been tested?

```
python -m pytest            # 1266 passed, 20 skipped
ruff check src tests        # clean
ruff format --check src tests
mypy src                    # error set identical to main's baseline in this env
```

Regression proof: the eight new/rewritten tests in `tests/test_compaction.py`
(tampered archive through verify_events and CLI exit code, deleted boundary
event, truncated archive, emptied archive, stored-version drift on anchored
replay, archived-action cache hit, capability refusal, non-projecting anchor)
were run against unmodified `main`: all eight fail there and pass on this
branch. Postgres parity is code-mirrored; its contract suite still skips
locally without `CONTINUUM_TEST_POSTGRES_DSN` and runs in CI as before.

Environment: Python 3.13, macOS, worktree off `origin/main` (`ac3b772`).

## Checklist

Tests added or updated for the changed behaviour. Documentation updated where the change affects user-facing behaviour (CHANGELOG Unreleased/Fixed). CI passes on the latest commit. Security-relevant changes state known limitations explicitly in this description (head-truncation note above).

## Additional context

Design tradeoff worth flagging for review: folding archived events back into
the ledger reintroduces O(archived) cost per ledger fold for compacted runs.
Correctness was chosen over that cost because silently resetting exactly-once
at the anchor boundary is worse; if it matters at month-long scales, the
existing `action_index` projection is the natural place to serve per-run key
lookups without the fold.


yuki-fuyutsuki:
**Code Review**

This PR correctly hardens the compaction path that #253 left exploitable. The anchored `verify_events` now chains from the archived edge and deep-audits `events_archive` on both engines, `compact_run` atomically appends the anchor with the archive move, `replay` restores its corruption contract for anchored runs, and the ledger/guard fold archived events so exactly-once does not reset at the boundary. Implementation is careful with transactions, hash re-digests and `supports_compaction` gating, and is covered by targeted regression tests. No high-severity correctness or security defect remains in the diff.

**medium**
`src/continuum/storage/sqlite.py:529` and `src/continuum/storage/postgres.py:631` (`_canonical_index_rows` / `rebuild_action_index`)
The ledger now folds `events_archive` for correctness, but the `action_index` projection still derives exclusively from `events`. After `compact_run` the index immediately drifts from the ledger truth: an archived `COMPLETED` action remains a `SKIP_DUPLICATE` cache hit via `protected_call`, yet `foreign_action`/`rebuild_action_index` will not see it and `verify --index` will report drift or miss a foreign unscoped claim. Either include `events_archive` in the canonical fold or document that the index is live-log only and must not be consulted for compacted runs.

```suggestion
    def _canonical_index_rows(self) -> dict[str, tuple[tuple[str, str, str, str, str], int]]:
        """Fold the log into ``{key: ((entry...), order_seq)}``, last write wins."""
        with self._read() as conn:
            rows = conn.execute(
                "SELECT rowid AS rid, type, payload FROM events ORDER BY rowid"
            ).fetchall()
            archived = conn.execute(
                "SELECT rowid AS rid, type, payload FROM events_archive ORDER BY rowid"
            ).fetchall()
            rows = sorted([*archived, *rows], key=lambda r: int(r["rid"]))
```

**low**
`src/continuum/cli/main.py:1395` and `src/continuum/storage/sqlite.py:620` / `src/continuum/storage/postgres.py:721` (`any(r["type"] == "EVENT_LOG_ANCHORED"`)
The anchored check gates the deep archive audit. Deleting the anchor row itself correctly becomes a `SEQUENCE_GAP`/`BROKEN_CHAIN` because the live tail still starts at `through+1`, but an explicit `archived_through` vs `archive_edge.sequence` assertion would make the invariant self-documenting and catch a future off-by-one in `through` without relying on the indirect gap.

**low**
`src/continuum/actions/ledger.py:263` (`sorted(events, key=lambda e: e.sequence)`)
Merging two already-sorted streams with `sorted` is `O(n log n)` per `protected_call`; the PR description correctly notes the `O(archived)` regression. For month-long runs consider a linear merge or a dedicated `read_all_events` that unions at SQL level. Not a correctness bug, but worth tracking as the cost is now on the hot guard path.


[github run](/Cyrax321/CONTINUUM/actions/runs/32664442671)

anya-research:
**Code Review**

This PR fixes the critical CWE-345 genesis-trust bug in `verify_events`, makes `compact_run` atomic on both engines, restores `replay`'s corruption contract for anchored runs and stops `protected_call` from resetting exactly-once at the anchor boundary. The fixes are narrow, preserve chain semantics, and are covered by targeted tamper/deletion regression tests. Remaining drift is in a derived projection, not the ledger truth.

**medium**
`src/continuum/storage/sqlite.py:451` and `src/continuum/storage/postgres.py:592` / `src/continuum/storage/base.py:122`
Ledger `_replay` now folds `read_archived_events` (`src/continuum/actions/ledger.py:259`) and `replayguard._fold` delegates to it, so a pre-compaction `COMPLETED` remains `SKIP_DUPLICATE`. `action_index` still folds only `events`, so `foreign_action`/`rebuild_action_index` immediately diverges after `compact_run` and `verify --index` will misreport. Either widen the canonical projection to include the archive or explicitly document the index as live-only and block index reads for compacted runs.

```suggestion
    def _canonical_index_rows(self) -> dict[str, tuple[tuple[str, str, str, str, str], int]]:
        with self._read() as conn:
            rows = conn.execute("SELECT rowid AS rid, type, payload FROM events ORDER BY rowid").fetchall()
            archived = conn.execute("SELECT rowid AS rid, type, payload FROM events_archive ORDER BY rowid").fetchall()
            rows = sorted([*archived, *rows], key=lambda r: int(r["rid"]))
```

**low**
`src/continuum/storage/sqlite.py:620` and `src/continuum/storage/postgres.py:721` (`verify_events` archive gate)
Deep archive audit is gated on `any(r["type"] == "EVENT_LOG_ANCHORED" for r in rows)`. Deleting just the anchor row still fails via `SEQUENCE_GAP`/`BROKEN_CHAIN` on the live tail, but tampering inside `events_archive` becomes unaudited once the live marker is gone. Gating on `EXISTS (SELECT 1 FROM events_archive WHERE run_id=?)` in addition to the live marker makes the invariant explicit and survives a future off-by-one in `through`.

**low**
`src/continuum/actions/ledger.py:263` (`sorted(events, key=lambda e: e.sequence)`)
Merging two already-sorted streams with `sorted` is `O(n log n)` on every `protected_call` (now on the hot exactly-once path). Correctness was rightly prioritized over cost, but a linear `heapq.merge` or a SQL `UNION ALL ORDER BY sequence` avoids the regression for month-long archived runs.



[github run](/Cyrax321/CONTINUUM/actions/runs/32664744434)

Cyrax321:
Both reviews addressed in 3f7c480:

- **Medium (action index drift), @yuki-fuyutsuki + @anya-research:** canonical folds in both engines now include events_archive. One correction to the suggested rid-sort: the two tables number rows independently, so ordering the union by rowid could rank an ancient archived claim above a newer live write of the same key (archive rowids are assigned at compaction time, interleaving above surviving live rows). Live rows keep their native order_seq so incremental maintenance still agrees exactly; archived rows take negative positions below every possible live value, oldest first. Uncompacted stores fold identically to before; drift appears once per compaction until rebuild, which is honest, and rebuild repairs it including history. Pinned by drift-then-rebuild plus cross-run lookup tests on both engines.
- **Low (audit gate):** verify_events now gates the deep archive audit on archive existence as well as a surviving anchor marker, so deleting the anchor row cannot exempt the archive from audit.
- **Low (hot-path sort):** the ledger merges the two sequence-sorted streams linearly instead of re-sorting.

Issue #45, opened 2026-08-14
claim(on_unknown=) resolution is not persisted — ledger stays uncertain after call-time resolution


Cyrax321:
## System information
- OS: macOS 15.x (darwin)
- Python: 3.13.5
- continuum-agent: 0.1.0 (editable install from source)
- Source commit: a539948 (== origin/main)

## Command or code to reproduce
```python
from continuum.storage.sqlite import SQLiteStorage
from continuum.models import Run, RunStatus, Origin, Action, ActionStatus
from continuum.events import Event, EventType
from continuum.actions.ledger import ActionLedger, ActionOutcome, IdempotencyKey

store = SQLiteStorage(":memory:")
run_id = "r"
store.create_run(Run(run_id=run_id, goal="g", status=RunStatus.STARTED))
store.append_event(run_id, EventType.RUN_STARTED, {"goal": "g"}, source=Origin.DETERMINISTIC)
ledger = ActionLedger(store, run_id)

o = ledger.claim("act.x", {"k": 1})
ledger.fail(o.key, "boom", certain=False)          # -> UNKNOWN
print(ledger.get(o.key).status)                    # unknown

def on_unknown(existing):
    resolved = existing.model_copy(update={
        "status": ActionStatus.COMPLETED,
        "external_id": "ext-1",
        "result": {"ok": True},
        "side_effect_uncertain": False,
    })
    return ActionOutcome(key=IdempotencyKey(o.key), action=resolved, fresh=False)

res = ledger.claim("act.x", {"k": 1}, on_unknown=on_unknown)
print(res.action.status)                            # completed
print(ledger.get(o.key).status)                    # STILL unknown  <-- bug
```

## Current behaviour
`claim(..., on_unknown=...)` calls `on_unknown(existing)` and, when it returns a non-None `ActionOutcome`, returns it directly (ledger.py:301-304) **without recording anything**. The ledger entry remains `UNKNOWN`. So the resolution is ephemeral: the next `claim` (or `intercept_action`) re-triggers `on_unknown` (or re-raises `UnknownSideEffect`), and `ledger.pending()` keeps returning the action as uncertain, so `RecoveryEngine.assess` keeps returning REQUEST_HUMAN/WAIT forever.

## Expected behaviour
[...]

Issue #19, opened 2026-08-12
continuum resume --repair records no repair plan; the flag is a no-op


Cyrax321:
### Before reporting

- [x] I searched existing issues for a duplicate
- [x] I am reporting this against the correct repository
- [x] I am using a supported Python version (3.11 or later)
- [x] I can reproduce this on the latest commit on `main`

### System information

```
OS: macOS (darwin)
Python: 3.14 (in the project venv)
continuum-agent: 0.1.0 (editable install from source)
Source commit: 455e307 (== origin/main)
```

### Command or code to reproduce

```shell
rm -f /tmp/r2.db
continuum --db /tmp/r2.db init
continuum --db /tmp/r2.db checkpoint r2 --env dataset=v3
# dependency changes to v4 while the run is stopped
continuum --db /tmp/r2.db resume r2 --env dataset=v4
echo "without --repair exit=$?"
continuum --db /tmp/r2.db resume r2 --env dataset=v4 --repair
echo "with --repair exit=$?"
```

After both runs, inspect the database:

```python
from continuum.storage.sqlite import SQLiteStorage
s = SQLiteStorage('/tmp/r2.db')
print('events:', len(s.read_events('r2')))
print('versions:', [v.version for v in s.list_checkpoints('r2')])
```

A minimal run setup for the storage path:

```python
from continuum.storage.sqlite import SQLiteStorage
from continuum.models import Run
from continuum.events import EventType
s = SQLiteStorage('/tmp/r2.db')
s.create_run(Run(run_id='r2', goal='g'))
s.append_event('r2', EventType.RUN_STARTED, {'goal': 'g'})
s.append_event('r2', EventType.DEPENDENCY_DECLARED, {'resource': 'dataset', 'version': 'v3'})
```

### Current behaviour

`--repair` changes nothing about the run. With and without the flag the exit code is the same (10), the repairs reported in the payload are the same, and no events or versions are recorded:

```
=== resume --env dataset=v4 (no flag) ===
exit=10
stderr=Run with --repair to record the repair plan, or resolve the items above first.

=== resume --env dataset=v4 --repair ===
exit=10
stderr=

events after --repair: 3
versions after --repair: [0]
```

[...]

PR #275, opened 2026-08-23
Informed retry: engine-authored prior-attempt summaries on recovery surfaces (#265)

### LINKED ISSUE #265: Informed retry: engine-authored failure summaries injected into post-recovery resumes
### Affected component

When a recovery resolves to REPLAN, ROLLBACK or a compensating repair, the run's next execution inherits restored task state and (since #235) the agent's own forward-looking reasoning summary. What nothing records is *why the previous attempt failed*: which component was invalidated, which side effect was reconciled and how, what the repair actually changed. A resumed agent that caused the failure, or a fresh session replacing it, receives a sealed contract with next steps but no failure narrative, and can repeat the same mistake.

AgentRewind (arXiv:2608.14380, submitted 2026-08-14) demonstrates empirically across multiple models, strategies and harnesses that injecting a recovery summary describing the failed attempt into the restored context, what it calls an informed retry, improves both task success rate and average checklist progress over blind retries. Its diagnosis matches our recovery model: rolling back context without carrying forward the lesson wastes the crash.

Our existing pieces do not cover this: REASONING_SUMMARY events (#235) are authored by the agent at turn boundaries about where it plans to go next; they are prospective, self-reported, and say nothing about engine-executed repairs. The recovery ledger records decisions, but nothing distils a completed recovery into guidance for the next attempt.

## Proposed solution

Add a deterministic, engine-authored failure summary to the recovery output, reusing the bounded-summary pattern from #235:

1. New payload on the recovery path (event type or field on the existing recovery-completed record): bounded JSON, 4 KB cap, containing what failed (component name, validator finding or action id), what changed (repairs applied, compensations recorded, reconciliations with occurred true/false), and what to avoid (a one-line deterministic rule derived from the severity signal, for example "dataset v3 is stale; re-read before relying on cached findings").
2. Authorship is DETERMINISTIC, not agent-authored: the recovery engine derives every field from signals it already holds (validator findings, ledger reconcile results, planner steps). No self-certification surface is created because the agent contributes nothing.
3. `continuum resume` and the MCP/serve contracts include the newest applicable summary whenever the current decision follows a non-trivial recovery (any mode after the first assessment, or any run with recovery-ledger entries); rendered in the guidance section of the contract and the SessionStart briefing.
4. Informational only: presence never gates, absence degrades gracefully, and the payload rides the hash chain like everything else.

Acceptance criteria:

- A scenario that produces REPLAN (for example stale dataset dependency) yields a summary naming the failed component and the repair taken; a fresh session's briefing contains it verbatim.
- A run with no recovery history produces byte-identical output to today.
- Summary size is capped and oldest content evicts first within the cap.
- Tampering with the summary breaks `continuum verify` like any other event payload.

## Alternatives considered

- Extend agent-authored REASONING_SUMMARY to also describe failures. Rejected: conflates two trust levels; the engine's account of engine-executed repairs should not depend on the agent choosing to confess.
- Point resumers at raw recovery-ledger entries. Rejected: the ledger is an audit structure, not guidance; distillation is the product here.
- Store the pre-failure transcript. Rejected for the same reason as #235: memory systems are a different project.

No

Yes, I have a plan

Cyrax321:
## Summary

Closes #265.

A resumed agent that knows why the previous attempt failed outperforms a blind retry (AgentRewind, arXiv:2608.14380). CONTINUUM's contract named what is blocked and what to do first, but carried no account of prior attempts, so a fresh session could repeat the mistake that caused the rollback.

New `continuum.recovery.summary` derives that account as a **pure projection** of facts already in the hash chain:

- RECOVERY_STARTED plans recorded by `resume --repair` (#19)
- RECOVERY_COMPLETED / RECOVERY_BLOCKED outcomes
- ACTION_RECONCILED settlements (#45), verdict read from status
- ACTION_COMPENSATED repairs

plus the current decision's failure signals. No new event type: every input is tamper-evident through the existing chain, identical logs produce byte-identical blocks, and runs without history yield None so today's output is unchanged byte for byte. Blocks cap at 4 KB with deterministic eviction (same spirit as the #235 cap).

## Surfaces

| Surface | Change |
|:--|:--|
| `RecoveryDecision.informed_retry` | new field, computed once in assess(), informational only |
| `continuum resume` | text section 'What previous attempts changed (informed retry)' + JSON key |
| Session briefing | block rendered beside the agent-authored summary (#235) |
| MCP `continuum_resume`, serve `resume` | payload gains informed_retry |

## Also fixes

cmd_resume printed the Next steps section twice (merge artifact); regression test asserts exactly one occurrence.

## Verification

- New tests/`test_informed_retry.py`: 11 tests covering absence-on-clean, component naming, repair history, settlement verdicts, determinism, the cap, every surface, and the Next-steps regression.
- Full suite green on this branch (Python 3.14 local; CI will cover 3.11-3.13).
- ruff check + ruff format + mypy strict clean on touched files.

### DIFF
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -580,8 +580,16 @@ def cmd_resume(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
         )
     if steps:
         text += "\n\nNext steps:\n" + "\n".join(f"  {i}. {t}" for i, t in enumerate(steps, 1))
-    if steps:
-        text += "\n\nNext steps:\n" + "\n".join(f"  {i}. {t}" for i, t in enumerate(steps, 1))
+
+    # Informed retry (#265): prior-attempt account, derived from recovery-path
+    # events already in the chain. Absent history means no section at all.
+    if decision.informed_retry:
+        from continuum.recovery.summary import render_informed_retry
+
+        text += (
+            "\n\nWhat previous attempts changed (informed retry):\n"
+            + "\n".join(f"  {line}" for line in render_informed_retry(decision.informed_retry))
+        )
 
     # Version pinning drift (issue #241): informational only.
     drift_lines: list[str] = []
@@ -617,6 +625,7 @@ def cmd_resume(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
         "family_rationale": family_rationale,
         "children": [c.__dict__ for c in child_statuses],
         "pinning_drift": drift_lines,
+        "informed_retry": decision.informed_retry,
         "contract": decision.contract.model_dump(mode="json"),
         "repairs": [s.action_name for s in decision.plan.steps],
         "progress": {
@@ -1020,6 +1029,13 @@ def cmd_briefing(args: argparse.Namespace, storage: Storage, out: Any, err: Any)
         lines += [
             f"  [{o.get('status', '?')}] {o.get('path', '')}" for o in obs if not o.get("truncated")
         ]
+    # Informed retry (#265): the engine's account of prior attempts, next to
+    # the agent's own summary above. Absent history means no section.
+    if decision.informed_retry:
+        from continuum.recovery.summary import render_informed_retry
+
+        lines.append("what previous attempts changed (engine-recorded):")
+        lines += [f"  {line}" for line in render_informed_retry(decision.informed_retry)]
     if steps:
         lines.append("next steps:")
         lines += [f"  {i}. {t}" for i, t in enumerate(steps, 1)]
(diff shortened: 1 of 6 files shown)


Cyrax321:
Verification detail for reviewers: the full suite ran twice on this branch. First run fully green. Second run flagged `test_dashboard_hitl.py::test_mutations_refused_with_wrong_token`, which passes in isolation and whose file (dashboard HITL, cba444a) has zero overlap with this diff. The failure coincided with a parallel session binding local ports for container work, so port contention is the likely cause. Nothing in this PR touches that surface.

PR #276, opened 2026-08-23
fix(dashboard): bind 127.0.0.1 by default; add --host flag

### LINKED ISSUE #270: Dashboard binds to all interfaces (0.0.0.0) exposing unauthenticated recovery data - should default to 127.0.0.1 like every other server
### Affected component

Dashboard

### What is wrong

`serve_dashboard` in `src/continuum/dashboard/app.py` creates its server with:

```python
server_class(("", port), Handler)
```

An empty host string binds **all interfaces**, so `continuum dashboard` exposes run goals, recovery contracts, external side-effect arguments/results and full event payloads to anyone who can reach the machine's network. Contrast with the other two servers, which already do this correctly:

- `gateway.py`: `ThreadingHTTPServer(("127.0.0.1", port), ...)`
- `serve/server.py` SidecarHTTP: `SidecarHTTP(sidecar, port=..., bind="127.0.0.1")`

HITL *mutations* are token-gated (#242), but reads are open by design and the bind itself contradicts the gateway's fail-closed posture.

### Suggested fix

1. Default bind to `"127.0.0.1"` in `serve_dashboard`.
2. Thread a `bind` parameter from the CLI: `continuum dashboard [--host HOST] [--port N]`, documented in `--help` ("default: localhost; only set --host if you understand the exposure").
3. Add one test asserting the bound address is loopback when invoked with defaults (drive `serve_dashboard`'s handler class or assert via socket probe after binding port 0).
4. Note the behaviour change in CHANGELOG under Changed.

### Acceptance criteria

- Default startup is unreachable from off-host.
- `--host 0.0.0.0` still works for operators who explicitly opt in.
- Suite stays green (`pytest -q`).

Yes, with guidance

Cyrax321:
Fixes #270. Loopback default for the dashboard bind address, matching every other CONTINUUM server.

### DIFF
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -2013,12 +2013,17 @@ def build_parser() -> argparse.ArgumentParser:
         from continuum.dashboard import serve_dashboard as _serve
 
         print(f"Serving dashboard at http://localhost:{args.port}", file=out)
-        _serve(storage, port=args.port)
+        _serve(storage, port=args.port, host=args.host)
         return 0
 
     dashboard = add("dashboard", cmd_dashboard, "Serve the dashboard (presentation over run data).")
     dashboard.add_argument(
         "--port", type=int, default=8000, help="port to listen on (default: 8000)"
+    )
+    dashboard.add_argument(
+        "--host",
+        default="127.0.0.1",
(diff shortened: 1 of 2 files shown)


Cyrax321:
Superseded: the same fix was pushed directly to main in 5faccfe.

Issue #241, opened 2026-08-23
Open problem: version pinning on observations (prompt/tool/schema hashes)


Cyrax321:
### Affected component

Core (events, validator)

### The open problem

Replay correctness needs environment pinning: which prompt version, tool schema, retrieval index and policy produced a decision. The Zylos durable-execution survey lists prompt/tool/schema hashes as part of replay correctness, not observability; the prompt-migration paper (arXiv:2507.05573) documents reliability breakage when LLM apps evolve across versions. CONTINUUM records environment snapshots for datasets but nothing about the agent's own moving parts.

### Step-by-step plan

1. Extend `TOOL_COMPLETED`/`REASONING_SUMMARY` payloads with optional `pinning`: {"prompt_sha256", "tool_schema_sha256", "model_id"} - caller-supplied, hashed, never invented.
2. `GenericAgentAdapter.intercept_action` gains `pinning=` passthrough; MCP `intercept_action` accepts an opaque pinning dict.
3. Validator: when a run resumes under different pinning than its last events, add an UNKNOWN-severity note to the contract (degrade, do not block).
4. Docs: how LangSmith-style prompt commits map onto these hashes.

### Scope guardrails

CONTINUUM never computes prompts or schemas itself; it stores caller-asserted hashes with EXTERNAL_AGENT provenance like every other claim.

### Acceptance criteria
- Resume across a pinning change surfaces the drift in contract rationale; identical pinning adds no noise.

Cyrax321:
Closed: closed-set pinning dict (prompt_sha256, tool_schema_sha256, model_id, policy_version) accepted by continuum intercept_action and continuum_record_summary, stored verbatim with EXTERNAL_AGENT provenance on the ACTION_RECORDED STARTED record and summary payloads. Values validated (known keys, 256-char cap - store the hash, not the artefact). continuum resume --pinning '<json>' diffs caller-supplied pins against the newest recorded set and surfaces drift as informational pinning_drift lines in text + JSON. Degrade-never-block per house rules; helpers in continuum.pinning with nine tests.

PR #287, opened 2026-08-24
feat(actions): consumed-grant tracking, deny Authority Resurrection (#269)

### LINKED ISSUE #269: Consumed-credential tracking: deny Authority Resurrection at the gate after restores
### Affected component

ACRFence (arXiv:2603.20625) validates a second attack class beyond duplicate execution: **Authority Resurrection**. When an agent restores from a checkpoint, single-use authorization tokens that were already consumed reappear in the rewound state, and the agent (or a prompt injector steering it) can spend them again against a different target. The paper's experiment is blunt: under stateless token validation, 2 of 2 reuse attempts succeeded; under stateful server-side validation, all reuse attempts were rejected. Industry surveys of agentic IAM report the same gap from the operations side: only about a fifth of organisations treat agents as first-class identities with lifecycle, so consumed-grant semantics rarely exist anywhere in the stack.

CONTINUUM checkpoints and restores harness state verbatim. If a token lives in that state, rewind resurrects it. We defend integrity and provenance, and the gate refuses unclaimed effects, but nothing tracks grant consumption as a fact about the world that survives restore. The replay half of the replay-or-fork triad is built (#237); the authority half is not.

## Proposed solution

Track consumption of single-use grants as ledger facts, so post-restore reuse is refused by evidence rather than by luck:

1. Grant registration: `continuum_intercept_action` (and the adapters' claim path) accepts an optional `grant` object: grant id plus scope string, recorded on the STARTED record with its issuing sequence number. Nothing about issuance changes; we track consumption semantics only.
2. Consumption record: completing or failing the action marks the grant spent at that sequence.
3. Restore-time accounting: after any restore to sequence S, the validator reports grants whose consumption sequence is greater than S as `consumed_not_remembered`: the rewound state believes them unspent, the world knows otherwise.
4. Gate enforcement: an attempt to claim an action carrying a grant in `consumed_not_remembered` state is REJECTed before it fires, with a dedicated reason code and an audit event recording the revocation decision. This mirrors the replayguard three-outcome table: REPLAY (cache hit), REJECT (spent grant), FORK (#259).
5. Out of scope, stated plainly: issuing credentials, integrating identity providers, replacing server-side validation. The ACRFence experiment shows stateful validation downstream is the real wall; our contribution is making consumption portable across restores for callers whose downstream cannot be changed.

Acceptance criteria:

- The ACRFence scenario as a test: grant issued, action completed, restore to pre-completion, second claim attempt with the same grant is REJECTed with the named grant id; a fresh unrelated action is unaffected.
- Verify-chain integrity holds across the revoke event; tamper evidence intact.
- Absence of grants changes no behaviour today (byte-identical outputs).

## Alternatives considered

- Rely solely on downstream stateful validation. Insufficient as a general answer: many internal APIs and legacy endpoints have no token state; those are exactly the targets the attack class names.
- Encrypt or strip credentials from checkpoints. Orthogonal: prevents exfiltration of the secret but still leaves the rewound agent believing an unconsumed grant exists; semantics are not fixed by cryptography.
- Fold into #259 fork work. Rejected: independent concern, different data model (grants vs branches); only the gate's outcome table is shared.

No

Yes, I have a plan

Cyrax321:
## Summary

Closes #269.

Implements the authority half of the replay-or-fork triad. ACRFence (arXiv:2603.20625) demonstrates that restored agents resurrect consumed single-use tokens (stateless validation: 2/2 reuse attempts succeeded); industry agentic-IAM surveys show almost nobody runs token lifecycle for agents. CONTINUUM checkpoints harness state verbatim, so a token stored there returns from a rewind believing itself unspent. This gives the ledger portable consumption semantics so reuse is refused by evidence rather than by luck.

## Mechanics

- `ActionLedger.claim(..., grant={"id", "scope"})` attaches a validated single-use reference to the attempt (rides the ACTION_RECORDED payload, same discipline as #241 pinning).
- `scan_grants` folds the log: any terminal status spends the attached grant, **fail-closed including COMPENSATED** (the authorisation was exercised even if the business effect was undone; only the issuer can un-spend).
- A later claim carrying a spent grant raises `GrantDenied` before anything fires and writes an audited `GRANT_DENIED` event (added to `_NON_PROJECTING`) naming the prior action id and attempted key.
- Live mid-flight retries under the same key and grant pass through untouched.
- Argument drift and freshly generated keys defeat nothing: consumption tracks grant identity, which is exactly what survives a restore.

## Surfaces

| Surface | Change |
|:--|:--|
| Library | `claim(grant=...)`, `GrantDenied`, `normalize_grant`, `scan_grants` |
| MCP `continuum_intercept_action` | `grant` param; denial maps to proceed=false + reason_code=grant_denied + guidance |
| Serve `intercept_action` | same passthrough and mapping |

## Verification

- 11 new tests in `tests/test_grants.py`: the ACRFence scenario end to end (complete, rewind, drifted re-plan with fresh key -> denied with named grant id), unrelated fresh actions unaffected, mid-flight retry not misflagged, failed-certain and compensated both spend fail-closed, input validation matrix, zero-change guarantee without grants, chain stays verify-clean after denials, serve and MCP surfaces.
- Full suite on this branch: 1314 passed, 23 skipped. ruff check/format and mypy strict clean.

## Summary by CodeRabbit

* **New Features**
  * Added single-use authority grants for action requests.
  * Prevented reuse of consumed grants, including across restored or completed actions.
  * Added structured denial responses requesting a fresh grant.
  * Added audited grant-denial events without changing application state.

* **Bug Fixes**
  * Preserved valid retries for in-progress actions while rejecting reused grants.

* **Tests**
  * Added coverage for grant validation, lifecycle handling, denial responses, and audit integrity.


Cyrax321 (commented):
Verification notes from the maintainer review:

The fail-closed rule is right, including treating COMPENSATED as spent: the downstream authorisation was exercised even when the business effect was undone, and only the issuer can un-spend. Mid-flight retries under the same key and grant pass through untouched, malformed grants are rejected before anything fires, and GRANT_DENIED joins _NON_PROJECTING correctly.

I verified the ACRFence scenario end to end on head 7fe38d3: a completed claim with grant tok_9 followed by a drifted re-plan under a fresh key raises GrantDenied and lands exactly one audited denial event; a plain duplicate without a grant still returns fresh=False with the stored result, so non-grant callers see no change. All required checks are green after the format fix and retitle.

Two follow-ups to track, neither blocking:

1. An identical replay claim (same key) carrying the spent grant now raises GrantDenied where it previously returned the journalled result with fresh=False. Defensible fail-closed behavior and safe against double execution since key dedup still applies, but it is an undocumented contract change for granted actions and no test pins it. Worth one test and a line in the docstring.

2. Consumption is enforced per-run only: the same grant id claims fresh in a sibling run. Fine for the same-run restore scenario this PR targets, but single-use reads as store-wide. Document the v1 scope or file the follow-up.

No blocking issues.

Issue #306, opened 2026-08-24
Withdrawn: request_human on MCP runs is intended; the defect was guidance, not the trust decision


Cyrax321:
## Summary

A run created purely through the MCP tools resumes as `request_human` forever, with no path an agent can take to clear it. This is the exact condition reported in #35 and #84, both closed as COMPLETED. It is back because the security fix for #201 sealed the escape hatch that #35 added.

This is not a complaint about the #201 fix, which is correct. The point is that the two fixes are mutually exclusive and #84 is now closed on a claim that no longer holds.

## How the three fixes interact

1. `9738b9e` fix(security): carry event provenance so agents cannot self-certify state. Closes the original exploit.
2. #35 work adds `continuum_confirm` so MCP-originated runs have a way out of `request_human`.
3. #201 reports that step 2 reinstated the exploit closed in step 1.
4. `d563d54` fix(mcp): require a dedicated secret for `continuum_confirm` (#201). Seals step 2.

Net effect: the escape hatch that closed #35 is gone, so #35's condition returns. #84, titled "eliminate the self-certified to request_human to confirm tax on every resume", is closed COMPLETED but the tax is present on every resume.

## Reproduction

Against the live MCP server on `b3472df`, using only registered MCP tools:

```
continuum_record_progress(run_id="mcp_audit_1", completed=0, total=9, goal="...")
continuum_resume(run_id="mcp_audit_1")
```

## Current behaviour

`continuum_resume` on a run that is seconds old, with zero work recorded and a clean checkpoint, returns:

```
mode:              request_human
recovery_status:   requires_human
safe:              false
invalidated:       goal (requires_review), progress (requires_review)
next_allowed:      human_review:goal
reason:            at least one repair needs a person
evidence:
  - goal: v1, asserted by external_agent
  - progress: 0 completed, self-reported by external_agent and not independently verified
```

Following the documented next step fails:

```
continuum_confirm(run_id="mcp_audit_1")
-> Error: continuum_confirm is refused because no CONTINUUM_MCP_CONFIRM_TOKEN is
   configured; an agent must not confirm its own self-reported state.
```

For contrast, a run created through the CLI in the same database resumes cleanly with `recovery_status: safe_to_resume` and `verified: [goal, progress]`. The distinction is purely the `external_agent` origin.

## Expected behaviour

One of the following, whichever the maintainer considers minimal and safe:

[...]

Issue #240, opened 2026-08-23
Open problem: run-level retry budgets enforced at claim time


Cyrax321:
### Affected component

Adapters / Action ledger

### The open problem

RetryGuard (arXiv:2511.23278) shows default retry patterns amplify cost and load across services; agent loops make this worse because the LLM can invent new retry attempts in response to failures. CONTINUUM records retries implicitly (new claims after certain failures) but has no run-level budget, no backoff accounting, no circuit breaker.

### Step-by-step plan

1. Count attempts per (run, action identity) from the ledger fold - data already exists.
2. Enforce budgets at claim time: `max_attempts` per action type (configurable via `.continuum/gate.json` sibling section); exceeding returns a structured refusal naming the budget.
3. Expose `retry_budget` in resume contracts so agents see remaining attempts before acting.
4. Optional jittered-backoff helper for adapters (pure function; never automatic inside CONTINUUM).
5. Bench scenario: failing upstream hammered N times shows exactly N claims then refusals.

### Scope guardrails

CONTINUUM never performs retries itself - it counts and gates them. Retries remain the caller's policy.

### Acceptance criteria
- Budget exhaustion is visible in contract rationale and human_steps.
- Ledger-derived attempt counts survive compaction (anchors carry totals).

Issue #49, opened 2026-08-14
StateValidator._check_model reports model-specific assumptions VALID when expected_model is None (fail-open)


Cyrax321:
## System information
- OS: macOS 15.x (darwin)
- Python: 3.13.5
- continuum-agent: 0.1.0 (editable install from source)
- Source commit: a539948 (== origin/main)

## Command or code to reproduce
```python
from continuum.models import (
    SemanticState, Goal, Progress, Provenance, Origin, ModelState, ModelSpecificState,
)
from continuum.state.validator import StateValidator

state = SemanticState(
    run_id="r",
    goal=Goal(description="g", provenance=Provenance(origin=Origin.DETERMINISTIC)),
    progress=Progress(completed=0, provenance=Provenance(origin=Origin.DETERMINISTIC)),
    model=ModelState(model="gpt-5", model_specific_state=[ModelSpecificState(description="temperature=0.7")]),
)
v = StateValidator(strict_unknown=True).validate(state, expected_model=None)
m = next(e for e in v.report.statuses if e.component.value == "model")
print(m.status, v.report.safe_to_resume)
# -> valid True   (assumptions reported as verified when they cannot be)
```

## Current behaviour
`StateValidator._check_model` (state/validator.py:374) early-returns `VALID` for any state with model-specific assumptions whenever `expected_model is None`, `recorded is None`, or `expected_model == recorded`:

```python
if expected_model is None or recorded is None or expected_model == recorded?:
    if assumptions:
        entries.append(Component.MODEL, VALID, "N model-specific assumption(s) recorded")
    return
```

When `expected_model=None` (e.g. `continuum validate` run without `--model`, or `validate_state(...)` called without it), the validator has **no idea what model will resume the run**, so it cannot verify the model-specific assumptions. Rather than admitting `UNKNOWN`/REQUIRES_REVIEW, it reports `VALID`, and `safe_to_resume` is `True`. This is a classic fail-open: the module docstring explicitly states "The system is allowed to say 'I cannot tell'; it is not allowed to guess in its own favour."

## Expected behaviour
When `expected_model is None` and the state carries model-specific assumptions, the MODEL entry should be `UNKNOWN` (or `REQUIRES_REVIEW`), not `VALID` — the assumptions cannot be verified without knowing the resume model. Only when the resume model is known and unchanged should the assumptions be reported `VALID`.

PR #311, opened 2026-08-24
fix(mcp): stop the retry budget defeating idempotency, and name what is actually missing

### LINKED ISSUE #307: validate/resume without env report pinned dependencies as unknown with an inverted diagnostic
## Summary

Once a run has pinned dependencies via `continuum_checkpoint(env=...)`, any later `continuum_validate` or `continuum_resume` call that omits `env` reports every pinned dependency as `unknown` and escalates the run to `request_human` with `revalidate_dependency` repairs.

Two problems:

1. The diagnostic text points at the wrong side of the comparison.
2. The documented session-start flow triggers it every time, so the false alarm is the default experience rather than an edge case.

## Reproduction

```
continuum_checkpoint(run_id="mcp_audit_1", env={"dataset": "v3", "schema": "2026-08-01"})
continuum_validate(run_id="mcp_audit_1")            # no env
```

## Current behaviour

```json
{
  "components": [
    {"component": "external_dependency", "component_id": "dataset",
     "detail": "no environment snapshot to compare against", "status": "unknown"},
    {"component": "external_dependency", "component_id": "schema",
     "detail": "no environment snapshot to compare against", "status": "unknown"}
  ],
  "environment_changes": [],
  "mode": "request_human",
  "reason": "external_dependency dataset is unknown; external_dependency schema is unknown; ...",
  "safe": false
}
```

`continuum_resume` with no `env` escalates further, adding human repair steps:

```
required_actions:
  - revalidate_dependency:dataset
  - revalidate_dependency:schema
human_steps:
  - re-pin dependency 'dataset': rerun with `--env dataset=<current version>` after confirming its real version
```

## Why the message is wrong

`no environment snapshot to compare against` states that the stored snapshot is missing. It is not. Checkpoint v1 recorded `dataset=v3` and `schema=2026-08-01`, and passing `env` proves the snapshot is intact:

```
continuum_validate(run_id="mcp_audit_1", env={"dataset": "v3", "schema": "2026-08-01"})
-> dataset: "verified unchanged", status valid
   schema:  "verified unchanged", status valid

continuum_validate(run_id="mcp_audit_1", env={"dataset": "v4", "schema": "2026-08-01"})
-> dataset: "v3 -> v4", status conflicted
   environment_changes: ["dataset: v3 -> v4"]
```

Drift detection itself works correctly. What is absent in the failing call is the caller's current observation, not the snapshot. The message names the one thing that is definitely present.

Suggested wording: `no current version supplied for comparison; pass env={"dataset": "<version>"}`.

## Why this is the default path, not an edge case

`CLAUDE.md` in this repo directs the agent to make `continuum_resume` with no arguments its first action of every session, and `scripts/session_start_resume.sh` runs `continuum resume --json` with no `--env`. Neither can supply `env`, because at session start the agent has not yet inspected anything. So for any run that ever pinned a dependency, the documented happy path reports `safe: false` plus one human repair step per dependency, on every single resume.

## Expected behaviour

Distinguish "caller made no claim" from "caller's claim conflicts". Concretely, when `env` is omitted:

- either leave pinned dependencies out of the component list entirely, since the caller asserted nothing about them
- or keep reporting `unknown` for visibility but do not let an unasserted dependency drive `mode` to `request_human` or emit a `revalidate_dependency` repair

A dependency nobody asked about is not evidence of a problem. Treating silence as conflict makes `safe: false` the normal state, which trains callers to ignore it. That is the real cost here: the signal that should mean "something moved under you" currently also means "you did not pass an optional argument".

## Environment

- OS: darwin 25.2.0
- Python: 3.14.5
- continuum-agent: 0.1.0, editable install from source
- Commit: `b3472df` (== origin/main at time of filing)

### LINKED ISSUE #308: expected_model is silently inert over MCP: no MCP path ever records a model
## Summary

`continuum_validate`, `continuum_resume` and `continuum_confirm` all accept an `expected_model` parameter over MCP. It can never have any effect, because nothing on the MCP path ever records the model that produced the state. The parameter is accepted, documented in the tool schema, and silently discarded.

This is distinct from #49, which was about `_check_model` fail-open when `expected_model` is None. Here `expected_model` is supplied and the other side of the comparison is missing.

## Root cause

`src/continuum/state/validator.py:402-434`:

```python
recorded = state.model.model if state.model else None
assumptions = list(state.model.model_specific_state) if state.model else []

if expected_model is not None and recorded is not None and expected_model != recorded:
    ...  # drift reported here
    return

if not assumptions:
    return
```

Drift requires `recorded is not None`. `recorded` derives from `state.model`, which is populated only by `MODEL_CHANGED` and `MODEL_ASSUMPTION_RECORDED` (`src/continuum/state/semantic.py:442-444`). Neither event is ever emitted by the MCP server:

```
$ grep -c "MODEL_CHANGED" src/continuum/mcp/server.py
0
```

So on the MCP path `state.model` is always None, `recorded` is always None, the drift branch cannot be reached, and `assumptions` is empty so the function returns silently at `if not assumptions`.

Confirmed against the live database, which has served every MCP session in this project:

```sql
sqlite> SELECT run_id, type, COUNT(*) FROM events WHERE type LIKE 'MODEL%' GROUP BY run_id, type;
-- zero rows
```

## Reproduction

```
continuum_record_progress(run_id="mcp_audit_1", completed=0, total=9, goal="...")
continuum_checkpoint(run_id="mcp_audit_1")
continuum_validate(run_id="mcp_audit_1", expected_model="claude-3-opus-20240229")
```

## Current behaviour

No `model` component appears in the response. The component list contains only `goal`, `progress` and any `external_dependency` entries. The output is byte for byte identical to the same call with `expected_model` omitted, including for a deliberately absurd model string.

## Expected behaviour

Any of these would be defensible; the current state is not:

1. Record the model on the MCP path, for example an optional `model` argument on `continuum_record_progress` or `continuum_checkpoint` that emits `MODEL_CHANGED`, so the existing check at validator.py:411 can actually fire.
2. Return an explicit `model` component with status `unknown` and detail "no model recorded for this run, cannot compare" whenever `expected_model` is supplied but `recorded` is None. This mirrors how the same function already handles the inverse case at line 434 and would make the gap visible instead of silent.
3. Remove `expected_model` from the three MCP tool schemas and say in the docs that model drift detection is CLI only.

Option 2 is the smallest change and is consistent with the comment already in the code at line 436, "Say so rather than guessing valid in the state's favour". That principle is applied when assumptions exist but not when they do not, which is why this case slips through silently.

## Why it matters

Model drift is a correctness control. A caller that passes `expected_model` and receives `safe: true` reasonably concludes the run was not produced by a different model. It received no such assurance. A safety parameter that silently does nothing is worse than one that is absent, because absence is discoverable and silence is not.

## Environment

- OS: darwin 25.2.0
- Python: 3.14.5
- continuum-agent: 0.1.0, editable install from source
- Commit: `b3472df` (== origin/main at time of filing)

### LINKED ISSUE #309: Retry budget is enforced before dedup and counts successes, so idempotency breaks at budget
## Summary

The retry budget added for #240 is enforced before the deduplication lookup, and it counts successful distinct operations rather than retries. Two consequences:

1. Once an action type reaches its budget, `continuum_intercept_action` raises `ToolError` instead of returning `proceed: false` with the previous result. The idempotency guarantee becomes unreachable at exactly the moment a resumed session needs it.
2. Any workflow performing more than `max_attempts` distinct operations of one type is hard blocked at the default of 3, with an error that describes retries when nothing was retried.

I hit this while filing three unrelated issues in one run, all three successful. This report is the fourth, and the ledger refused to claim it.

## Reproduction

Three distinct successful actions of one type, then a re-claim of the first:

```
continuum_intercept_action(run_id="r", action_type="create_github_issue", key="gh-issue:a")  # proceed true
continuum_complete_action(run_id="r", action_key=..., external_id="https://.../306")

continuum_intercept_action(run_id="r", action_type="create_github_issue", key="gh-issue:b")  # proceed true
continuum_complete_action(run_id="r", action_key=..., external_id="https://.../307")

continuum_intercept_action(run_id="r", action_type="create_github_issue", key="gh-issue:c")  # proceed true
continuum_complete_action(run_id="r", action_key=..., external_id="https://.../308")

# Now simulate a crashed session re-claiming the FIRST action, which is completed.
continuum_intercept_action(run_id="r", action_type="create_github_issue", key="gh-issue:a")
```

No `.continuum/budgets.json` exists, so `default_max_attempts` of 3 applies.

## Current behaviour

```
Error executing tool continuum_intercept_action: retry budget exhausted for
'create_github_issue': 3 attempt(s) recorded, budget is 3. Reconcile existing
attempts or ask the operator to raise .continuum/budgets.json.
```

The ledger was fully settled at this point. `continuum_list_actions` reported `"unresolved": 0` with all six actions in `completed` or `failed` state, so the suggested remedy "Reconcile existing attempts" is not actionable.

## Expected behaviour

Re-claiming key `gh-issue:a` should return what it returns before the budget is reached:

```json
{
  "proceed": false,
  "status": "completed",
  "external_id": "https://github.com/Cyrax321/CONTINUUM/issues/306",
  "guidance": "Already performed. Reuse the previous result; do not repeat it."
}
```

## Root cause

`src/continuum/mcp/server.py:843-876`. The budget gate runs, and raises, before `ledger.claim()` is called:

```python
events = ctx.storage.read_events(run_id)
attempts = attempts_for_type(events, action_type)
allowed, used, maximum = evaluate_budget(budgets, action_type, attempts)
if not allowed:
    raise ToolError(f"retry budget exhausted for {action_type!r}: ...")

ledger = ctx.ledger(run_id)
outcome = ledger.claim(action_type, arguments=arguments, key=key, ...)
```

Deduplication lives inside `ledger.claim()`, downstream of the raise. So an already-completed action can never be recognised as already completed once the type is at budget.

`src/continuum/budgets.py:87-104` counts per type with no per-key scoping:

```python
return sum(
    1
    for e in events
    if e.type is EventType.ACTION_RECORDED
    and e.payload["action"].get("action_type") == action_type
    and e.payload["action"].get("status") == ActionStatus.STARTED.value
)
```

Every claim slot counts, whether it later completed or failed. The docstring states the intent, "retries count but their bookkeeping does not", but a distinct first attempt on a new key is not a retry, and the code cannot tell the two apart.

## Why this is severe

The tool description promises: "Returns proceed=true if you should do it, or proceed=false with the previous result if it was already done, do NOT repeat it in that case." The server instructions add: "Route every external side effect through continuum_intercept_action so it is never performed twice."

At budget, that promise inverts. A resumed session re-claiming a completed action receives an error rather than "already done". An agent seeing a tool error on its ledger lookup has no way to distinguish "I must not repeat this" from "the ledger is unavailable", and the plausible recovery is to perform the side effect out of band. The mechanism that exists to prevent double charging a card can, at budget, encourage exactly that.

The second effect is more mundane but will be hit sooner: any legitimate batch of 4 or more operations of one type fails at the fourth, and the error text blames retries.

## Suggested fix

Two independent changes, either of which removes the sharp edge:

1. Move the budget gate below `ledger.claim()`, or consult the ledger for an existing settled action before evaluating the budget. A claim that resolves to an existing completed or failed action is a lookup, not an attempt, and should never be budget-gated.
2. Scope attempt counting per action key rather than per action type, so a budget of 3 means three attempts at the same operation rather than three distinct operations. This matches the docstring's stated intent and the `send_invoice` example in the module header, where 5 attempts plainly means 5 tries at one invoice rather than a cap of 5 invoices.

If the per-type cap is deliberate as a global rate limit, it needs a distinct name and a distinct error message, because "retry budget exhausted" does not describe refusing a first attempt on a new key.

## Disclosure

Filing this issue required bypassing `continuum_intercept_action`, since the budget for `create_github_issue` was exhausted by the three successful claims above. I verified no duplicate existed before creating it, so the "never performed twice" intent was preserved manually. I did not create a `.continuum/budgets.json` to raise the cap, in order to leave the reproduction environment untouched.

## Environment

- OS: darwin 25.2.0
- Python: 3.14.5
- continuum-agent: 0.1.0, editable install from source
- Commit: `b3472df` (== origin/main at time of filing)
- No `.continuum/budgets.json` present, so defaults applied

Related: #240 introduced the budget enforced at claim time.

Cyrax321:
## Summary

I drove all 11 MCP tools end to end against a live server, found four issues, and
this PR resolves three of them. The fourth was my own mistake and is withdrawn
with an explanation, because the behaviour I reported as a regression is
load-bearing security.

| Issue | Defect | Severity |
|---|---|---|
| #309 | Retry budget enforced before dedup, and counting successes as retries | High |
| #307 | Inverted diagnostic when `env` is omitted | Medium |
| #308 | `expected_model` silently inert with no recorded model | Medium |
| #306 | Withdrawn. Working as intended, guidance was the defect | n/a |

## #309: the retry budget could cause the duplicate it exists to prevent

Two independent bugs in one gate.

**The gate ran before the deduplication lookup.** In `continuum_intercept_action`
the budget was evaluated and raised before `ledger.claim()`, and dedup lives
inside `claim()`. So once an action type hit its budget, re-claiming an action
that had already **completed** raised `ToolError` instead of returning
`proceed=false` with the stored result.

That inverts the tool's central promise. An agent that receives an error where it
expected "already done, reuse this" cannot distinguish "I must not repeat this"
from "the ledger is unavailable", and the plausible recovery is to perform the
side effect again outside the ledger. The mechanism that exists to stop a card
being charged twice could, at budget, be the reason it was.

The gate now runs only when a claim would genuinely open a new slot. A record in
`COMPLETED` or `UNKNOWN` is a lookup, not an attempt: the first returns the
stored result, the second raises `UnknownSideEffect` asking for reconciliation.
Neither is budgeted.

**It counted successes as retries.** `attempts_for_type` tallied every
`ACTION_RECORDED` slot for a type regardless of outcome, so three successful,
distinct operations exhausted a budget of 3 and the fourth was refused, having
never retried anything. Slots whose action reached `COMPLETED` are now excluded.

I hit this while filing the first three issues: after three successful reports,
the ledger refused the fourth. Issue #309 had to be filed outside the ledger,
which is documented in that issue.

Deliberately preserved from #240:

- Failures, unknowns and in-flight attempts still count, so the amplification
  guard is intact.
- Counting stays **per action type**, not per key. `test_claims_beyond_budget_are_refused_at_the_mcp_boundary`
  shows per-type across distinct keys is intentional, and it is what catches an
  agent re-planning with varied arguments.

## #307: the diagnostic blamed the wrong side

After `checkpoint(env=...)` pinned dependencies, any `validate` or `resume` that
omitted `env` reported:

```
"detail": "no environment snapshot to compare against"
```

The snapshot was not missing. The checkpoint recorded it, and passing `env`
proves it. What was absent is the caller's **current observation**, so the
message named the one thing definitely present. `diff_environments` returns an
empty diff when either side is absent, so the diff alone cannot tell the two
cases apart; the distinction is now passed down from `validate()`, which knows.

```
"detail": "no current version supplied for comparison; pass env={\"dataset\": \"<version>\"}"
```

**The `UNKNOWN` status is unchanged and that is deliberate.** My initial plan was
to make an unobserved dependency non-blocking, which would have broken
`test_no_environment_at_all_leaves_dependencies_unverified` and its stated
principle, "never validated is not the same as validated clean". That test is
right and the module docstring backs it: uncertainty degrades rather than
resolves. Only the wording was wrong, so only the wording changed. Callers who
genuinely tolerate the uncertainty still have `strict_unknown=False` and
`--tolerate-unknown`.

## #308: a safety parameter that silently did nothing

`expected_model` reaches `_check_model`, but drift needs both sides of the
comparison and `recorded` comes from `state.model`, populated only by
`MODEL_CHANGED` / `MODEL_ASSUMPTION_RECORDED`. Nothing on the MCP path emits
either:

```
$ grep -c MODEL_CHANGED src/continuum/mcp/server.py
0
```

With no assumptions recorded the function returned silently, so the result was
byte for byte identical to a clean comparison. A caller passing `expected_model`
and seeing no model component reasonably concluded it had ruled out drift. It had
ruled out nothing. A safety parameter that silently does nothing is worse than
one that is absent, because absence is discoverable and silence is not.

The gap is now reported explicitly. This is the same fail-open pattern #49 closed
for model-specific assumptions, one branch over, and it honours the comment
already in that function: "Say so rather than guessing valid in the state's
favour."

This changes one existing expectation. `test_a_model_switch_can_be_declared`
asserted `ExitCode.OK` with the comment "no model recorded, so nothing to
invalidate". That is the fail-open assumption itself, so the test is updated to
`REQUIRES_REPAIR`, which is the right severity: the remedy is to record which
model produced the state, and that is mechanical rather than a judgement call.

## #306: withdrawn, and why

I reported that MCP-created runs reporting `request_human` was a regression of
#35 and #84. That was wrong, and I have corrected the issue.

`tests/test_provenance.py` asserts precisely the behaviour I called a bug:

```python
"""The exact reproduction that previously returned mode=resume, safe=True."""
assert decision["mode"] != RecoveryMode.RESUME.value
assert decision["safe"] is False
```

The reason is concrete. Self-reported progress decides what work a resumed
session skips. An agent that inflates `completed` to 9999/10000, or is
prompt-injected into it (`tests/test_toy_task_banner_attack.py`), would cause
9999 units of real work to be silently skipped. Making those runs auto-resume
would reinstate the exploit closed by `9738b9e` and re-closed by `d563d54`. That
test passes untouched in this PR.

I also misread #84, whose complaint was the per-resume latency and token tax.
`scripts/session_start_resume.sh` solved that by injecting the banner out of
band. It never asked to change the trust decision.

The genuine defect was discoverability: the status reads as a hard stop, and the
obvious next move, `continuum_confirm`, is refused by design, leaving the caller
with no legal way forward. So the server now explains itself:

- The MCP `instructions` field carries the session-start protocol and the
  `request_human` semantics, so it reaches any client at connect time.
- `resume` returns `self_report_guidance` when self-report is the **only** thing
  blocking. When a real problem is also present the note is omitted rather than
  diluting it.

That helper lives in `recovery/guidance.py` beside `human_steps_for` and is
shared with the sidecar, which `test_resume_payload_covers_the_mcp_resume_surface`
requires. That parity test caught the omission when I first added the field to
only one surface.

**`CLAUDE.md` is deleted.** Everything it specified now travels with the server,
so any client gets the protocol without a per-project file.

## Verification

```
1381 passed, 24 skipped     (was 1373 passed, +8 new tests, zero regressions)
ruff check                  clean on all changed files
ruff format --check         clean on all changed files
mypy src                    25 errors, identical to baseline (missing stubs for
                            the optional mcp extra and cryptography)
```

Beyond the suite, a 23-assertion audit drove all 11 tools through the real
`server.call_tool` dispatch path and confirmed each fix together with the
properties that must not move:

- self-certified runs still return `request_human` with `safe: false`
- `continuum_confirm` still refuses without `CONTINUUM_MCP_CONFIRM_TOKEN`
- an uncertain side effect is still never re-proposed
- a fresh identity is still refused once the budget is genuinely exhausted
- dedup by key still ignores argument formatting
- drift detection still reports `v3 -> v4` as conflicted

New tests:

| File | Covers |
|---|---|
| `tests/test_mcp_server.py` | completed action dedups at budget; 5 successive successes allowed; uncertain action still refuses at budget |
| `tests/test_retry_budgets.py` | completed attempts excluded; failures and in-flight still counted |
| `tests/test_validator.py` | absent `env` names the missing side; absent from both snapshots still blames the snapshot; model gap reported |

## Reviewer notes

- Scope is limited to `budgets.py`, `mcp/server.py`, `recovery/guidance.py`,
  `serve/server.py`, `state/validator.py` and their tests. Uncommitted work in
  `cli/main.py`, `health.py` and `CHANGELOG.md` is deliberately untouched.
- No `.continuum/budgets.json` is added. Raising the cap would have masked #309
  rather than fixing it.
- No MCP write path for the model is added, so drift detection over MCP remains
  unavailable by design. #308 stays open for that decision; this PR only removes
  the false assurance.

Closes #307, closes #308, closes #309. Corrects #306.

## Summary by CodeRabbit

* **New Features**
  * Added clearer recovery guidance when reported progress or goals require human confirmation.
  * Resume and recovery responses now explain startup handling, self-reported runs, and required confirmation steps.
  * Improved diagnostics for missing environment snapshots and unrecorded expected models.

* **Bug Fixes**
  * Completed actions no longer consume retry capacity.
  * Preserved deduplication and reconciliation evidence when limits are reached.
  * Blank run IDs and action types are now rejected.
  * Missing model information correctly blocks resume with an actionable explanation.

* **Documentation**
  * Expanded testing, recovery, integration, attestation, and security documentation.


Cyrax321:
## Update: second adversarial round against the live server

After the first three fixes landed I re-audited through the reconnected MCP server rather than in-process only. That found three more defects, all now fixed in `bafc15a`, plus a mistake in my own first attempt at one of them.

### 4. A blank `run_id` created a run (high)

`continuum_record_progress(run_id="")` succeeded. A blank id wins `get_active_run`, so it silently became the run a fresh session was told to resume, displacing real work, and it rendered guidance that reads literally `continuum confirm ` with nothing after it.

Fixed via `Storage.require_usable_run_id`, called from all four create paths across both backends.

**Enforced on the write path, not on the `Run` model, and that distinction matters.** My first attempt put a `min_length=1` constraint on the model. Pydantic runs field validation on deserialization too, so one pre-existing bad row made `list_runs` and `get_active_run` raise. `continuum runs` broke outright against the live database, leaving no way to even see the offending row in order to clean it up. Guarding the entry point stops new bad data without bricking existing databases.

### 5. A blank `action_type` was accepted (medium)

`idempotency_key` already refused one, but only on the branch where no explicit `key` is supplied, so passing a key bypassed the check. The type is the unit the retry budget counts and the key the reconciler registry matches on, so a blank one is silently exempt from both. Now validated on both branches.

### 6. Confirming an effect erased its receipt (high)

`reconcile(occurred=True)` rewrote `external_id` from its argument unconditionally, so omitting it erased a receipt already on record. Reproduced live: `receipt-mirror-999` became `null` while the status stayed `completed`. A caller confirming an effect happened, without having captured its id, destroyed the evidence proving it did.

Now any recorded `external_id` and `result` survive omission, while a caller passing replacements still overrides them.

### What I got wrong, again

I initially reported that `reconcile(occurred=False)` on a COMPLETED action was a critical exploit, because it clears the receipt and reopens the action so the next claim returns `proceed=true`, giving a duplicate charge. I wrote a guard refusing reconciliation of settled actions.

Reading #29 showed that path is intended. Its own reproduction is `complete()` then `reconcile(occurred=False)`, and the scenario is real: an agent that optimistically called `complete` may be contradicted by a later probe, and correcting that record is exactly what reconciliation is for. The duplicate is the documented risk of calling it without evidence, not a defect.

So I reverted the guard and kept only the indefensible half. #29's clearing behaviour is untouched and its test passes unchanged.

That is the third time in this PR an existing test or issue corrected a finding of mine. Worth stating rather than hiding: the suite and issue history in this repo are load-bearing, and each time I read them properly the diagnosis got narrower and more defensible.

### Documentation

`docs/TESTING_MCP.md` records the whole exercise: all 11 tools with the inputs each was driven with, the seven findings, a table of guarantees asserted unchanged after every fix, and the known gaps stated plainly (model drift over MCP still undetectable, nothing tested concurrently, Postgres wired but not executed, one transient two-error run never root caused).

It also records the operational trap that cost real time: the MCP server is a long-running process, so editing `src/` does not take effect in an existing session even with an editable install.

### Verification

```
1359 passed, 24 skipped     stable across 3 consecutive runs
ruff check / format         clean on every changed file
mypy src                    25 errors, identical to baseline
in-process audit            23 of 23 assertions, all 11 tools
adversarial round 2         14 of 14 assertions
```

Security properties re-asserted after every change: self-certified runs still `request_human`, `confirm` still refuses over MCP, uncertain effects still never re-proposed, fresh identities still refused at budget, `reconcile(occurred=false)` still clears falsified evidence, and unknown runs never report success on any tool.

### Note on the branch

Two commits reached this branch from elsewhere while I worked (`6b6507b` CI wheel workflow, `7952cb5` a merge of main). Both touch only `ci.yml` and `CHANGELOG.md` and are unrelated to these changes. My in-progress work was also swept into a stash by another process mid-session and recovered from it; the other agent's `CHANGELOG.md`, `STATUS.md` and `cli/main.py` changes were left in that stash untouched.

Cyrax321:
## Update: round 3, concurrency and crash recovery

Targeted what the earlier rounds left untested: unscoped cross-run keys, single-use grants, real thread concurrency, a crash mid-action followed by recovery in a fresh process, and tamper detection. Fourteen of sixteen assertions passed first time; one failure was a bug in my own probe and one is a genuine finding.

### What already holds

Verified, not assumed:

- A completed unscoped effect deduplicates across runs, and an unresolved one elsewhere refuses rather than opening a parallel slot (#34)
- A consumed single-use grant cannot be resurrected (#269)
- A fresh process finds the interrupted run with no memorised id, recovers its goal and progress, and is refused a silent repeat of the in-flight effect
- A tampered payload is reported as both `TAMPERED_CONTENT` and `BROKEN_CHAIN`
- Deduplication survives compaction: an effect settled before the anchor still protects afterwards
- Scoped keys stay isolated between runs

The crash-and-resume block is worth calling out because it is the actual product promise, and it works end to end: claim, kill the process, reconnect, and the interrupted charge is refused with `status: unknown` rather than repeated.

### 8. `ActionLedger` has no lease integration (filed as #345)

`claim()` deduplicates by folding the log and then appending, with nothing between the read and the write. Eight threads on one key all received `proceed=true`, and eight `ACTION_RECORDED` slots were written. Eight charges.

I am filing rather than patching this, because it is a capability gap rather than a broken invariant. `docs/multi_agent_isolation.md` already specifies the ownership model, "one run, one owner at a time", and `RecoveryLedger` implements it with a `LeaseCoordinator`. My reproduction holds no lease, so it is an unsupported configuration.

The defensible observation is narrower: `ActionLedger`, the class whose entire purpose is at-most-once side effects, has no lease parameter and had nothing in its docstring saying one is required, while the "never performed twice" promise is stated loudly elsewhere. This commit states the requirement where a caller will actually meet it.

Two tests pin both halves:

- One records that the ledger alone does not serialise concurrent claimants. Its docstring says that if a future change makes claiming atomic, the test should be tightened to `== 1` rather than deleted.
- One proves the documented remedy works. The same eight claims wrapped in a shared `SQLiteLeaseCoordinator` lease collapse to exactly one winner. I verified that empirically before asserting it, having no interest in documenting a remedy I had not run.

The event chain stays sound throughout, so these are honestly-recorded duplicate attempts rather than corruption, which is exactly why only a lease prevents them.

### Verification

```
1361 passed, 24 skipped     (+2 concurrency tests, zero regressions)
ruff check / format         clean on every changed file
mypy src                    25 errors, identical to baseline
```

### Report

`docs/TESTING_MCP.md` now covers all three rounds: eight findings, the guarantees table extended with the seven properties above, and the gaps restated honestly. Concurrency has moved from "not tested" to "tested, documented, remedy proven, atomic claiming still open". Newly stated gaps: no Postgres execution, and lease expiry mid-claim is unexplored.

PR #342, opened 2026-08-24
fix: demo invalid title handling

Cyrax321:
Throwaway pull request proving the reworked title-check posts guidance and stays green. Closed right after.


Cyrax321:
Demo complete; guidance and self-cleanup both verified.

PR #343, opened 2026-08-24
fix: Improve backoff delay error message

### LINKED ISSUE #336: Improve backoff_delay error message for attempt less than 1
## Problem
`backoff_delay` error does not show the bad value.

`backoff_delay` in `src/continuum/budgets.py` raises:

```python
if attempt < 1:
    raise ValueError("attempt starts at 1")
```

When `attempt` is computed from another count, you do not learn what was passed.

## Repro
```python
from continuum.budgets import backoff_delay
backoff_delay(0)
# ValueError: attempt starts at 1
```

## Fix
Change to:

```python
if attempt < 1:
    raise ValueError(f"attempt must be >= 1, got {attempt}")
```

## How to test
Add a test in `tests/test_retry_budgets.py`:

```python
def test_backoff_delay_rejects_zero():
    import pytest
    from continuum.budgets import backoff_delay
    with pytest.raises(ValueError, match="got 0"):
        backoff_delay(0)
```

## Files
- `src/continuum/budgets.py`
- `tests/test_retry_budgets.py`

One line fix.

stoppo22:
## Description

Improves the `backoff_delay` validation error by including the invalid `attempt` value in the error message. This makes it easier to identify the value that caused the error when `attempt < 1`.

A regression test was also added to verify that calling `backoff_delay(0)` raises a `ValueError` containing the invalid value.

## Related issues

Fixes #336

## Types of changes

* Type: `bugfix`

## Breaking changes

No.

## How has this been tested?

The changes were tested with:

```bash
uv run pytest tests/test_retry_budgets.py -v
```

Environment:

* macOS
* Python 3.11.9
* pytest 9.1.1

Result: 10 tests passed.

## Checklist

Tests were added for the changed behaviour.

Documentation was not updated because this change only improves an existing validation error message and does not affect documented usage.

CI has not run yet on this PR.

No security-relevant changes.

## Additional context

The new regression test verifies that `backoff_delay(0)` raises a `ValueError` whose message includes `got 0`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for invalid retry attempt values by including the value that caused the error.

* **Tests**
  * Added coverage verifying that invalid retry attempt values report the expected details.

### DIFF
--- a/src/continuum/budgets.py
+++ b/src/continuum/budgets.py
@@ -127,5 +127,5 @@ def backoff_delay(
 ) -> float:
     """Exponential backoff with a ceiling. Pure; jitter is the caller's job."""
     if attempt < 1:
-        raise ValueError("attempt starts at 1")
+        raise ValueError(f"attempt must be >= 1, got {attempt}")
     return float(min(base * (2 ** (attempt - 1)), cap))
--- a/tests/test_retry_budgets.py
+++ b/tests/test_retry_budgets.py
@@ -94,6 +94,7 @@ def test_backoff_delay_is_exponential_with_cap() -> None:
     with pytest.raises(ValueError):
         backoff_delay(0)
 
+
 def test_backoff_delay_rejects_zero() -> None:
     with pytest.raises(ValueError, match="got 0"):
         backoff_delay(0)
--- a/tests/test_retry_budgets.py
+++ b/tests/test_retry_budgets.py
@@ -94,6 +94,10 @@ def test_backoff_delay_is_exponential_with_cap() -> None:
     with pytest.raises(ValueError):
         backoff_delay(0)
 
+def test_backoff_delay_rejects_zero() -> None:
+    with pytest.raises(ValueError, match="got 0"):
+        backoff_delay(0)
+
 
 # --- enforcement through the real ledger path --------------------------------------- #
 


yuki-fuyutsuki:
**Code Review**

This PR changes `src/continuum/budgets.py:130` to include the invalid `attempt` value in the `ValueError` (`f"attempt must be >= 1, got {attempt}"` vs the previous static `"attempt starts at 1"`), and adds a regression test in `tests/test_retry_budgets.py:97` asserting `backoff_delay(0)` raises with `match="got 0"`. The change is minimal, correct, and preserves existing rejection semantics while improving debuggability; no functional risk or scope creep.

No actionable correctness, security, performance, or resource-handling issues were found. What was done well: the f-string safely interpolates an `int` with no injection risk, the message remains a `ValueError` on the same condition (`attempt < 1`), and the new test pins the message contract without altering existing coverage.


[github run](/Cyrax321/CONTINUUM/actions/runs/32747680302)

Cyrax321:
Thanks for the PR. This is a reasonable improvement and correctly addresses #336: I confirmed no code or test depends on the previous message string, and `tests/test_retry_budgets.py` passes with the added regression case (10 passed).

**One blocker before merge**

Required CI runs `ruff format --check`, which enforces two blank lines before top-level definitions. The new test currently has one, so the Lint job fails. Fix:

```bash
ruff format tests/test_retry_budgets.py
```

**Nit (non-blocking):** titles in this repo follow Conventional Commits; consider `fix: improve backoff delay error message`.

Everything else LGTM. We will merge once CI is green.

Thank you for contributing to CONTINUUM; polish like this is exactly what makes the project better, and we hope to see more from you.

Cyrax321 (approved):
Approving on behalf of the review above: format fix verified green on afb92e8, full CI matrix passing, branch current with main, Vercel gate dismissed as unrelated. Merging now.

PR #347, opened 2026-08-24
fix: make SQLiteStorage.close idempotent

### LINKED ISSUE #320: Make SQLiteStorage.close idempotent
## Problem
Closing the same storage twice should be safe, but it currently can fail.

`SQLiteStorage.close` in `src/continuum/storage/sqlite.py` does:

```python
def close(self) -> None:
    with self._lock:
        self._connection.close()
```

`sqlite3.Connection.close` raises `ProgrammingError` if called again. Test fixtures and error paths that call `close()` in both normal and exception cleanup can hit this. `PostgresStorage` already handles this, SQLite should match.

## Repro (copy paste)
```python
s = SQLiteStorage(":memory:")
s.close()
s.close()  # before fix: ProgrammingError, after fix: no error
print("ok")
```

## How to fix, step by step
1. Open `src/continuum/storage/sqlite.py` and find `class SQLiteStorage`.
2. Add a flag `self._closed = False` in `__init__`.
3. Change `close` to:
```python
def close(self) -> None:
    with self._lock:
        if getattr(self, "_closed", False):
            return
        self._closed = True
        try:
            self._connection.close()
        except Exception:
            pass
```
   Or use `contextlib.suppress(sqlite3.ProgrammingError)`.
4. Ensure `__del__` also respects the flag if it exists.

## How to test
Add a test in `tests/test_storage.py`:

```python
def test_close_is_idempotent():
    s = SQLiteStorage(":memory:")
    s.close()
    s.close()  # should not raise
```

Run `pytest tests/test_storage.py -v`.

## Files
- `src/continuum/storage/sqlite.py`
- `tests/test_storage.py`

Isolated bug, no domain knowledge needed.

VedantMadane:
## Summary

fix: make SQLiteStorage.close idempotent

## Changes

- Second close() no longer raises ProgrammingError.
- Clear the connection handle after close.

Fixes #320

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved storage shutdown reliability by making connection cleanup safe when called multiple times.
  * Prevented errors when closing an already-closed or unavailable database connection.

### DIFF
--- a/src/continuum/storage/sqlite.py
+++ b/src/continuum/storage/sqlite.py
@@ -139,7 +139,16 @@ class SQLiteStorage(Storage):
 
     def close(self) -> None:
         with self._lock:
-            self._connection.close()
+            conn = getattr(self, "_connection", None)
+            if conn is None:
+                return
+            try:
+                conn.close()
+            except sqlite3.ProgrammingError:
+                # Already closed — safe to call close() more than once.
+                pass
+            finally:
+                self._connection = None  # type: ignore[assignment]
 
     def __del__(self) -> None:
         """Release the connection if the owner never closed it.


yuki-fuyutsuki:
**Code Review**

This PR makes `SQLiteStorage.close()` idempotent to fix #320 by guarding against a second `close()` raising `sqlite3.ProgrammingError` and nulling `self._connection` under `self._lock`. The core fix is correct and thread-safe for the close path itself, the `getattr` guard handles partial `__init__` failure, and interaction with `__del__` is now safe. The remaining gap is what happens after a successful close.

**medium**
`src/continuum/storage/sqlite.py:151`
After `close()` sets `self._connection = None  # type: ignore[assignment]`, every other method (`_write` at `src/continuum/storage/sqlite.py:123`, `_read` at `src/continuum/storage/sqlite.py:136`, and all public CRUD paths) still dereferences `self._connection` unconditionally. Post-close calls now raise `AttributeError: 'NoneType' object has no attribute 'execute'` instead of a clear storage-closed error, which is harder to diagnose and breaks the previous `ProgrammingError: Cannot operate on a closed database` signal. Guard the entry points or document the closed state explicitly.
```suggestion
    @contextmanager
    def _write(self) -> Iterator[sqlite3.Connection]:
        """Exclusive write transaction. Rolls back on any exception."""
        with self._lock:
            if getattr(self, "_connection", None) is None:
                raise StorageError("SQLiteStorage is closed")
            self._connection.execute("BEGIN IMMEDIATE")
            try:
                yield self._connection
            except BaseException:
                self._connection.execute("ROLLBACK")
                raise
            else:
                self._connection.execute("COMMIT")

    @contextmanager
    def _read(self) -> Iterator[sqlite3.Connection]:
        with self._lock:
            if getattr(self, "_connection", None) is None:
                raise StorageError("SQLiteStorage is closed")
            yield self._connection
```

**low**
`src/continuum/storage/sqlite.py:146`
Catching only `sqlite3.ProgrammingError` is correct for CPython's double-close, but the `finally: self._connection = None` ensures even an unexpected error during `conn.close()` discards the handle. Callers will then see the `AttributeError` above instead of the original close failure. Consider letting non-idempotency errors propagate or wrapping them, so a real I/O error on close is not silently replaced by a later confusing `NoneType` error.

**low**
`src/continuum/storage/sqlite.py:161`
`__del__` reads `self._connection` without acquiring `self._lock`, while `close()` nulls it under the lock. Concurrent `close()` and finalization can race to call `connection.close()` twice from different threads. In practice CPython's GC runs on the same thread and `suppress(Exception)` hides it, but for correctness `__del__` should either acquire the lock with `suppress` or rely entirely on the idempotent `close()` helper.


[github run](/Cyrax321/CONTINUUM/actions/runs/32762359596)

Cyrax321 (approved):
Thank you for your first contribution to CONTINUUM.

This PR makes `SQLiteStorage.close()` idempotent: repeated close calls return cleanly, the guard covers partially initialised instances, and the connection handle is cleared afterwards so finalisation never touches a closed connection twice. It resolves the intent behind #320.

I reviewed the implementation and the surrounding lifecycle paths (`__enter__`/`__exit__`, `__del__`, the test fixtures) and verified the behaviour by execution rather than inspection alone: double close on Python 3.11 through 3.14 completes without error, closing after partial initialisation is safe, and the context-manager-then-explicit-close pattern used by our fixtures works. `tests/test_storage.py` passes in full (48 tests) on this branch, and ruff lint and format are clean on the changed file. CI is confirmed green on this head commit: lint and strict type-check plus the full test matrix on ubuntu, macOS and windows across Python 3.11, 3.12 and 3.13. One behavioural note for the record: after `close()`, storage calls now raise `AttributeError` instead of sqlite3's `ProgrammingError`; an acceptable tradeoff for the explicit idempotency guarantee, and something we can revisit with a `StorageError` if it ever matters in practice.

With that verified, I found no blocking issues. The implementation does what it sets out to do, and the close contract is now explicit rather than relying on interpreter behaviour.

Marking this as ready to pull and applying the relevant labels.

Thank you again for taking the time to improve CONTINUUM. Well-scoped contributions like this are exactly what keeps the project moving, and we hope to see more from you.

Issue #47, opened 2026-08-14
OpenAIAgentAdapter creates runs without a RUN_STARTED event; restore/project/replay fail until a checkpoint exists


Cyrax321:
## System information
- OS: macOS 15.x (darwin)
- Python: 3.13.5
- continuum-agent: 0.1.0 (editable install from source)
- Source commit: a539948 (== origin/main)

## Command or code to reproduce
```python
from continuum.storage.sqlite import SQLiteStorage
from continuum.adapters.openai import OpenAIAgentAdapter

store = SQLiteStorage(":memory:")
adapter = OpenAIAgentAdapter(store)
run_id = "openai_run"

class FakeAgent:
    name = "my-agent"

# on_agent_start -> _ensure_run_exists: creates the Run row only, no RUN_STARTED event
adapter._ensure_run_exists(run_id, FakeAgent())

# a wrapped tool fires -> appends ACTION_RECORDED
adapter.intercept_action(run_id, "act.x", lambda: "done", arguments={"k": 1})

events = store.read_events(run_id)
print([e.type.value for e in events])          # ['ACTION_RECORDED', 'ACTION_RECORDED']
print(any(e.type.value == "RUN_STARTED" for e in events))  # False

adapter.manager.restore(run_id, replay=True)   # raises
```

## Current behaviour
`OpenAIAgentAdapter._ensure_run_exists` (openai.py:347) only calls `storage.create_run(...)`; it never appends a `RUN_STARTED` event. After a tool fires, the event log is `[ACTION_RECORDED, ACTION_RECORDED]` with no `RUN_STARTED`. As a result:
- `restore(run_id, replay=True)` — which `assess_agent_recovery`/`resume` use — raises `ProjectionError: run 'openai_run' has no goal: the log never recorded RUN_STARTED`.
- `project(run_id, events)` raises the same.
- `continuum replay` / `continuum inspect --version` for that run crash.

This happens for any OpenAI run that has not yet produced a checkpoint (i.e., during the first agent run before `on_agent_end`, or after a crash before `on_agent_end`). Once a checkpoint exists, `restore(replay=True)` returns the checkpoint state without projecting, so recovery then works — but the event log remains invalid for any projection/verification/replay path.

## Expected behaviour
[...]

PR #348, opened 2026-08-24
fix: attest-verify must not pass a tampered chain; fork must produce a usable child

Cyrax321:
## Summary

Auditing the CLI and subsystems, which the earlier MCP rounds never touched, surfaced three defects. One is a genuine security hole.

| # | Defect | Severity |
|---|---|---|
| 1 | `attest-verify` reports SIGNED on a tampered chain | High, security |
| 2 | `continuum fork` produces a child no reader can open | Medium |
| 3 | Dashboard answers 200 for a run that does not exist | Low |

## 1. A signed attestation passed on rewritten content

`attest-verify` compared the signed `chain_hash` against the digest **stored** in the head row. Editing an event's payload straight through the database changes the payload and leaves every `hash` column untouched, so the head still matched, the signature still verified, and the verdict read:

```
Attestation SIGNED by ci-bot for seq 2; chain matches.     exit 0
```

on a run whose goal had been rewritten to `EVIL`. The same database, one command later:

```
continuum verify j
INTEGRITY FAILURE: 2 violation(s)
  seq 1: TAMPERED_CONTENT - stored hash does not match recomputed digest
  seq 2: BROKEN_CHAIN - prev_hash does not match predecessor digest
```

`verify` catches it because it recomputes digests. `attest-verify` did not, because it trusted the column an attacker with write access simply leaves alone. That is precisely the attack an event-chain signature exists to stop, which makes a passing verdict worse than no attestation at all.

The verdict now consults `verify_events` before trusting any hash read out of the chain, and names the tampering rather than reporting a sequence mismatch that is not the problem: the head can sit exactly at the signed sequence while an earlier event's content has been rewritten underneath it.

Both existing outcomes are preserved. A clean chain is still SIGNED. An appended event is still ALTERED with its sequence comparison, which `test_attest_verify_reports_altered_after_new_event` continues to assert.

## 2. `continuum fork` created an unusable child

`approve_fork` called `create_run`, which writes a run row and no `RUN_STARTED`. Projection refuses a log that never recorded one, so every reader failed on the child, including both commands `fork` itself prints as the next step:

```
$ continuum fork job1 --reason divergent
Forked job1 into job1_fork1.
Resume it independently: continuum resume job1_fork1
Lineage: continuum tree job1

$ continuum resume job1_fork1
error: run 'job1_fork1' has no checkpoint and no events        exit 2

$ continuum tree job1
job1  [resume, safe=True]  audit
  !! [fork] job1_fork1  [assess error: run 'job1_fork1' has no checkpoint and no events]
```

Same defect class as #47, which fixed it for the OpenAI adapter. It now uses `create_run_started`, which writes the row and its first event in one transaction, with `Origin.HUMAN` to match the `RUN_FORKED` record on the parent.

**Worth flagging how this survived.** Three tests hand-appended `RUN_STARTED` to the child after calling `approve_fork`, so they exercised the fixture rather than the code. `test_fork_child_is_independently_resumable` asserted exactly the property that was broken and passed only because it staged the missing event itself; its own comment gave it away, describing "an empty fork with a consistent start event". That crutch is removed, so the test now proves its name. A second test walks the six readers plus `tree`, so fork's printed advice is covered rather than assumed.

## 3. Dashboard answered 200 for an unknown run

`GET /runs/<unknown>` rendered "Run not found" in the body and returned HTTP 200. The body was honest; the status line was not, and the status is what anything other than a human reads.

The CLI holds this line deliberately, which is what `test_no_command_reports_success_for_a_run_that_does_not_exist` protects. The dashboard is the same claim over a different transport, so it owes the same answer. Real routes are untouched: `/` still lists runs at 200, `/runs/<real>` still renders detail at 200, and the catch-all for unrecognised paths is unchanged because serving the index there is deliberate.

## What was verified working

Not everything I probed was broken. Confirmed sound, and worth recording so the next audit can skip it:

- Every read-only command on a real run: `inspect`, `history`, `events`, `verify`, `actions`, `replay`, `show-contract`, `validate`, `resume`, `briefing`, `budget`, `tree`
- `reconcile` with no probes registered correctly exits 20 and reports the action unsettled rather than guessing
- `gate` defaults to allow with "no gate configured"
- `compact` refuses without `--force` and the chain still verifies afterwards
- The sidecar over its JSON wire protocol, including that `resume` carries `self_report_guidance`, so the MCP and sidecar surfaces remain at parity
- Dashboard index, run detail, and HITL POST refusing an unauthenticated caller with 403
- Attestation keygen and the clean round trip

Two of my probes were my own errors, not defects: `observe` reads its payload from stdin, and `health` is not a command in `main` at all.

## Verification

```
1370 passed, 24 skipped
ruff check / format         clean
mypy src                    25 errors, identical to baseline
```

Each fix is covered by a test that fails without it. I confirmed that by stashing each change and watching the new tests go red:

| Test | Guards |
|---|---|
| `test_attest_verify_detects_an_in_place_payload_edit` | asserts the stored head digest is byte-identical after the edit, so the premise of the bug is pinned, then requires CORRUPTED |
| `test_fork_child_is_independently_resumable` | the child's log is startable with no help from the test |
| `test_every_reader_accepts_a_fresh_fork_child` | six readers plus `tree`, no "assess error" |
| `test_an_unknown_run_answers_404_not_200` | 200 for a real run, 404 for a missing one |

## Summary by CodeRabbit

- **Bug Fixes**
  - Attestation verification now detects altered event data and reports clear tampering diagnostics, including the number of violations.
  - Dashboard requests for unknown runs now return HTTP 404 instead of a successful response.
  - Approved forked runs now include their own event history and work correctly with resume, inspect, replay, verification, event viewing, and contract commands.
- **Tests**
  - Added regression coverage for event tampering, missing dashboard runs, and fork-child workflows.

### DIFF
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -1750,13 +1750,22 @@ def cmd_attest_verify(args: argparse.Namespace, storage: Storage, out: Any, err:
     live_hash = events[-1].hash if events else None
     live_seq = events[-1].sequence if events else 0
 
+    # Recompute the chain before trusting any hash read out of it. `live_hash`
+    # above is the digest *stored* in the row, and an edit made straight through
+    # the database changes the payload while leaving that column untouched, so
+    # comparing the attestation against it reports "chain matches" on content
+    # that has demonstrably been altered. Only the recomputing walk in
+    # verify_events can tell, which is why the verdict has to consult it.
+    integrity = storage.verify_events(args.run_id)
+    chain_intact = integrity.ok
+
     signature_valid = verify_attestation(doc)
     chain_match = doc.get("chain_hash") == live_hash
     seq_match = doc.get("trusted_through_seq") == live_seq
 
     if not signature_valid:
         verdict = "UNTRUSTED"
-    elif not chain_match or not seq_match:
+    elif not chain_intact or not chain_match or not seq_match:
         verdict = "ALTERED"
     else:
         verdict = "SIGNED"
@@ -1766,6 +1775,7 @@ def cmd_attest_verify(args: argparse.Namespace, storage: Storage, out: Any, err:
         "verdict": verdict,
         "signature_valid": signature_valid,
         "chain_match": chain_match,
+        "chain_intact": chain_intact,
         "signer": doc.get("signer"),
         "signed_seq": doc.get("trusted_through_seq"),
         "live_seq": live_seq,
@@ -1776,10 +1786,20 @@ def cmd_attest_verify(args: argparse.Namespace, storage: Storage, out: Any, err:
             f"chain matches."
         )
     elif verdict == "ALTERED":
-        text = (
-            f"Attestation ALTERED: chain changed since signing "
-            f"(signed seq {doc.get('trusted_through_seq')} vs live {live_seq})."
-        )
+        if not chain_intact:
+            # Name the tampering rather than reporting a sequence mismatch that
+            # is not the problem: the head can sit at the signed sequence while
+            # an earlier event's content has been rewritten underneath it.
+            text = (
+                f"Attestation ALTERED: the event chain no longer verifies "
+                f"({len(integrity.violations)} violation(s)); run "
+                f"`continuum verify {args.run_id}` for the specific events."
+            )
+        else:
+            text = (
+                f"Attestation ALTERED: chain changed since signing "
+                f"(signed seq {doc.get('trusted_through_seq')} vs live {live_seq})."
+            )
     else:
         text = "Attestation UNTRUSTED: signature does not verify against the embedded key."
     _emit(payload, text, as_json=args.json, stream=out, palette=getattr(args, "_palette", None))
--- a/src/continuum/dashboard/app.py
+++ b/src/continuum/dashboard/app.py
@@ -42,6 +42,17 @@ def render_dashboard_html(storage: Storage) -> str:
 </table></body></html>"""
 
 
+def _run_exists(storage: Storage, run_id: str) -> bool:
+    """Whether the run is real, so the handler can answer 404 rather than 200."""
+    from continuum.storage import RunNotFound
+
+    try:
+        storage.get_run(run_id)
+    except RunNotFound:
+        return False
+    return True
+
+
 def render_run_detail_html(storage: Storage, run_id: str) -> str:
     try:
         run = storage.get_run(run_id)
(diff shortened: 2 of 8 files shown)


Cyrax321:
## Update: a fourth fix, and it was blocking the whole repo

CI went red on this branch across ubuntu and macos in eighteen tests I never touched. It was not the branch.

**`mcp 2.1.0` was published today at 19:04:29Z. Main's last green run started at 19:03:27Z, sixty-two seconds earlier.** I reran main's own workflow at `b2f92ba` with no code change and it fails the same six jobs, so every PR in this repo is blocked until this is handled, not just mine.

### What upstream changed

From 2.1.0 the SDK distinguishes an intentional `ToolError`, whose message it preserves, from any other handler exception, which becomes `UnexpectedToolError` carrying only `"Error executing tool <name>"` with the cause demoted to `__cause__` (`mcp/server/mcpserver/tools/base.py:158`, `:210`).

Every refusal in this server is a domain exception, so all of them lost their explanation.

### Why this is a real regression, not a test problem

An unauthorized client used to be told which caller was refused, that it was a permissions problem, and the `CONTINUUM_MCP_MUTATING_CLIENTS` setting that grants access. Under 2.1.0 it received:

```
Error executing tool continuum_record_progress
```

That is all. A caller whose progress counters broke their own arithmetic learned just as little. The tests were the messenger; the caller-facing message was the actual casualty.

### Fix

Refusing is part of this server's contract rather than a fault, so the guard converts the deliberate cases to `ToolError` with the message intact: `PermissionError` for authz, `ValueError` for validation, `RunNotFound`, and `MalformedRunLog`. Genuinely unexpected exceptions are deliberately left alone, because a bug in this server is not a message to act on.

**I did not pin `mcp<2.1`.** The new upstream behaviour is defensible, and pinning would have kept the poor caller-facing message while hiding it from CI. Worth noting the pin is unbounded (`mcp>=2.0`), which is why a same-day upstream release could turn main red with nobody touching the code.

### Verified against both SDKs

Not assumed. I built a clean venv on `mcp==2.1.0` and ran the MCP suites there as well as on 2.0.0 locally:

- both suites pass on 2.1.0
- both suites pass on 2.0.0, so nothing regresses for existing installs
- the two new tests fail on 2.1.0 without the fix

They assert the message text rather than the exception type, because asserting the type alone would have passed throughout the entire regression, which is exactly how this reached a release.

```
1372 passed, 24 skipped
all 12 CI test jobs green, including the macos and ubuntu jobs that were red
ruff clean, mypy unchanged at its pre-existing 25 errors
```

This PR now carries four fixes: the attestation tamper hole, the unusable fork child, the dashboard 404, and this.

PR #353, opened 2026-08-25
fix: clarify authentication errors

### LINKED ISSUE #318: Improve SidecarAuth error message to hint at CONTINUUM_SERVE_TOKEN
## Problem
First time sidecar users get a vague error.

When `CONTINUUM_SERVE_TOKEN` is set, every call must include `auth_token`, reads as well as writes. The error from `src/continuum/serve/server.py` is:

```
the caller did not present the expected shared secret
```

It does not say which variable to set or what field to send. The same generic wording exists in `src/continuum/mcp/authz.py`.

## Repro
```bash
CONTINUUM_SERVE_TOKEN=secret continuum serve --transport http &
curl -X POST http://127.0.0.1:8765/resume -H "Content-Type: application/json" -d '{"id":1,"method":"resume","params":{"run_id":"x"}}' 
# 403 with generic message
```

## Expected message
Make it actionable without leaking the secret. For example:

```
expected shared secret (set CONTINUUM_SERVE_TOKEN on the server and pass auth_token on the client)
```

## How to fix
1. Open `src/continuum/serve/server.py` and find `class SidecarAuth` method `verify`.
2. Update the `NotAuthorized` message to include the hint above.
3. Do the same for `src/continuum/mcp/authz.py` if it has the same string.
4. Update or add a test in `tests/test_serve.py` and `tests/test_mcp_authz.py` to assert the hint is present: `assert "CONTINUUM_SERVE_TOKEN" in str(exc)` and `assert "auth_token" in str(exc)`.
5. Ensure no secret is echoed.

## Files
- `src/continuum/serve/server.py`
- `src/continuum/mcp/authz.py`
- `tests/test_serve.py`

Tiny copy change, high onboarding value. Perfect for a first PR.

adity982:
## Summary
- name the exact server and client token fields in authentication failures
- keep configured secrets out of error responses
- add regression coverage for both auth paths

## Validation
- focused pytest auth tests (2 passed)

Closes #318

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication error messages with clear guidance on required server settings and client-provided credentials.
  * Authentication failures no longer expose configured secret values.
  * Clarified where handshake and authorization tokens must be provided.
  * Progress and dependency responses now handle invalid event data more gracefully, including details when projection fails.

### DIFF
--- a/src/continuum/mcp/authz.py
+++ b/src/continuum/mcp/authz.py
@@ -152,10 +152,22 @@ class AuthPolicy:
             expected = self.expected
         # An empty expected secret cannot be presented, so it must refuse.
         if not expected or not token or token != expected:
-            raise NotAuthenticated(
-                "expected shared secret (set CONTINUUM_MCP_TOKEN on the server "
-                "and pass _meta.authToken during initialize)"
-            )
+            if self.tokens is not None:
+                guidance = (
+                    "expected the secret registered for this caller "
+                    "(pass it in _meta.authToken during initialize)"
+                )
+            elif self.source == "argument":
+                guidance = (
+                    "expected the configured shared secret "
+                    "(pass the auth_token argument's value in _meta.authToken during initialize)"
+                )
+            else:
+                guidance = (
+                    "expected shared secret (set CONTINUUM_MCP_TOKEN on the server "
+                    "and pass _meta.authToken during initialize)"
+                )
+            raise NotAuthenticated(guidance)
 
 
 #: Confirmation of self-reported state over MCP is gated behind its own secret
--- a/src/continuum/mcp/authz.py
+++ b/src/continuum/mcp/authz.py
@@ -152,7 +152,10 @@ class AuthPolicy:
             expected = self.expected
         # An empty expected secret cannot be presented, so it must refuse.
         if not expected or not token or token != expected:
-            raise NotAuthenticated("the caller did not present the expected shared secret")
+            raise NotAuthenticated(
+                "expected shared secret (set CONTINUUM_MCP_TOKEN on the server "
+                "and pass _meta.authToken during initialize)"
+            )
 
 
 #: Confirmation of self-reported state over MCP is gated behind its own secret
--- a/src/continuum/mcp/authz.py
+++ b/src/continuum/mcp/authz.py
@@ -173,7 +173,7 @@ class AuthPolicy:
                 )
             else:
                 guidance = (
-                    "expected shared secret (set CONTINUUM_MCP_TOKEN on the server "
+                    f"expected shared secret (set {AUTH_ENV_VAR} on the server "
                     "and pass _meta.authToken during initialize)"
                 )
             raise NotAuthenticated(guidance)
--- a/src/continuum/mcp/authz.py
+++ b/src/continuum/mcp/authz.py
@@ -153,14 +153,23 @@ class AuthPolicy:
         # An empty expected secret cannot be presented, so it must refuse.
         if not expected or not token or token != expected:
             if self.tokens is not None:
-                guidance = (
-                    "expected the secret registered for this caller "
-                    "(pass it in _meta.authToken during initialize)"
-                )
+                if self.source == CLIENT_TOKENS_ENV_VAR:
+                    guidance = (
+                        "expected the secret registered for this caller "
+                        f"(set {CLIENT_TOKENS_ENV_VAR} on the server and pass the "
+                        "matching value in _meta.authToken during initialize)"
+                    )
+                else:
+                    guidance = (
+                        "expected the secret registered for this caller "
+                        "(configure the per-client token mapping in the embedding "
+                        "application and pass the matching value in _meta.authToken "
+                        "during initialize)"
+                    )
             elif self.source == "argument":
                 guidance = (
-                    "expected the configured shared secret "
-                    "(pass the auth_token argument's value in _meta.authToken during initialize)"
+                    "expected the secret configured by the embedding application "
+                    "(pass it in _meta.authToken during initialize)"
                 )
             else:
                 guidance = (
--- a/tests/test_mcp_authz.py
+++ b/tests/test_mcp_authz.py
@@ -367,13 +367,16 @@ def test_a_disabled_auth_policy_is_a_no_op() -> None:
 
 
 def test_a_configured_secret_is_required() -> None:
-    auth = AuthPolicy("secret")
+    auth = AuthPolicy("actual-token-value")
     assert not auth.disabled
     # Correct secret passes.
-    auth.verify(ALLOWED, "secret")
+    auth.verify(ALLOWED, "actual-token-value")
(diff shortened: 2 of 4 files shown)


adity982:
Addressed the policy-specific guidance finding in the latest commit. Per-client policies now refer to the caller-registered secret, argument-based policies name the auth_token argument, and only the environment-backed shared-secret policy names CONTINUUM_MCP_TOKEN. Focused validation: Ruff passes and the three relevant authentication tests pass.

src/continuum/mcp/authz.py
```diff
@@ -152,7 +152,22 @@ def verify(self, caller: str | None, token: str | None) -> None:
...
-            raise NotAuthenticated("the caller did not present the expected shared secret")
+            if self.tokens is not None:
+                guidance = (
+                    "expected the secret registered for this caller "
+                    "(pass it in _meta.authToken during initialize)"
+                )
+            elif self.source == "argument":
+                guidance = (
+                    "expected the configured shared secret "
+                    "(pass the auth_token argument's value in _meta.authToken during initialize)"
```
Cyrax321: `auth_token` is the sidecar's JSON-RPC parameter (`serve/server.py:241`). The MCP path has no such field, only `_meta.authToken`, so this will send people looking for an argument that doesn't exist here.

`source="argument"` is set by `load_auth(expected=...)`, where the parameter is named `expected`. Suggest phrasing it as the secret configured by the embedding application, presented in `_meta.authToken`.

Note this also means `test_argument_auth_names_the_matching_client_field` at line 414 needs its `match="auth_token argument"` updated to whatever wording you land on.

src/continuum/mcp/authz.py
```diff
@@ -152,7 +152,22 @@ def verify(self, caller: str | None, token: str | None) -> None:
             expected = self.expected
         # An empty expected secret cannot be presented, so it must refuse.
         if not expected or not token or token != expected:
-            raise NotAuthenticated("the caller did not present the expected shared secret")
+            if self.tokens is not None:
+                guidance = (
+                    "expected the secret registered for this caller "
```
Cyrax321: `source` is `CONTINUUM_MCP_CLIENT_TOKENS` in this branch. Could you name it here? Otherwise the per-client case still doesn't tell the operator which variable to set, which was the main ask in #318. The `assert "CONTINUUM_MCP_TOKEN" not in ...` guard you added is right to keep.

tests/test_mcp_authz.py
```diff
@@ -367,13 +367,16 @@ def test_a_disabled_auth_policy_is_a_no_op() -> None:
 
 
 def test_a_configured_secret_is_required() -> None:
-    auth = AuthPolicy("secret")
+    auth = AuthPolicy("actual-token-value")
```
Cyrax321, line 370: Good catch renaming this from `"secret"`. The leak assertion would have passed for the wrong reason otherwise, since "secret" appears in the message itself.

Cyrax321 (changes requested):
Thank you for contributing to CONTINUUM, and welcome to the project.

The `serve/server.py` change is right and matches the issue exactly. Two things to fix in the MCP path, both commented inline, plus the test failures.

CI is red on `tests/test_mcp_authz.py:575` and `:606`. Both match on "shared secret", which the new per-client message no longer contains. Heads up that line 627 has the identical assertion but still passes, because that case takes the "not registered" branch, so the breakage isn't obvious from reading the file. I ran the full file locally and those two are the only failures, so it should be a quick fix. Worth running all of `tests/test_mcp_authz.py` rather than the focused subset before pushing.

Everything else checks out: lint, formatting and type checks pass, the fail-closed behaviour is untouched, and no message interpolates the configured secret.

No rush on this. Re-request review when CI is green and I'll take another look. Thanks again for picking this one up.

src/continuum/mcp/authz.py
adity982: Fixed in `47855d8`: argument-backed authentication now says the secret is configured by the embedding application and must be presented in `_meta.authToken`; it no longer refers to a nonexistent MCP `auth_token` field. The regression assertion was updated accordingly. Full `tests/test_mcp_authz.py`: 67 passed; Ruff check and format, mypy, and `git diff --check` passed.

src/continuum/mcp/authz.py
adity982: Fixed in `47855d8`: the environment-backed per-client path now explicitly names `CONTINUUM_MCP_CLIENT_TOKENS`, while explicit per-client mappings retain embedding-application guidance. A new regression checks the environment-variable hint, `_meta.authToken`, and that the configured secret is never leaked. Full `tests/test_mcp_authz.py`: 67 passed; Ruff check and format, mypy, and `git diff --check` passed.

adity982:
Pushed `47855d8` to address both MCP review threads and the two failing full-file assertions.

Validation on the exact head:
- `tests/test_mcp_authz.py`: 67 passed
- Ruff check and format: passed
- mypy: passed across 104 source files
- `git diff --check`: passed

The two addressed inline threads are resolved. I will leave the historical changes-requested decision with the maintainer while the refreshed CI matrix runs.

src/continuum/mcp/authz.py
```diff
@@ -152,7 +152,31 @@ def verify(self, caller: str | None, token: str | None) -> None:
             expected = self.expected
         # An empty expected secret cannot be presented, so it must refuse.
         if not expected or not token or token != expected:
-            raise NotAuthenticated("the caller did not present the expected shared secret")
+            if self.tokens is not None:
+                if self.source == CLIENT_TOKENS_ENV_VAR:
+                    guidance = (
+                        "expected the secret registered for this caller "
+                        f"(set {CLIENT_TOKENS_ENV_VAR} on the server and pass the "
+                        "matching value in _meta.authToken during initialize)"
```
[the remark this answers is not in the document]
adity982: Fixed in signed commit `7e9cc1a`: the shared-secret guidance now interpolates `AUTH_ENV_VAR`, matching the per-client constant-based path. On the merged current-main tree, all 67 auth tests, Ruff check/format, mypy across 104 source files, and `git diff --check` pass.

tests/test_mcp_authz.py
adity982: Acknowledged. The renamed synthetic token continues to make the no-secret-leak assertion meaningful, and the full 67-test auth file passes on the current head. This informational note requires no further code change.

Cyrax321 (approved):
Hi @adity982, thank you for contributing to CONTINUUM.

LGTM, really nice work on #318. I reviewed this locally and everything looks great, the auth errors are now properly actionable and correctly tailored per configuration with clear guidance for CONTINUUM_MCP_TOKEN, CONTINUUM_MCP_CLIENT_TOKENS and _meta.authToken, while keeping secrets out of the output. Also appreciate the thorough regression coverage for both MCP and sidecar paths, all set on my end and CI is green. Merging now.

Welcome aboard, looking forward to your next contribution!

PR #341, opened 2026-08-24
docs: add pre-commit example with ruff hooks to CONTRIBUTING

unmoha:
Adds a documented pre-commit setup using astral-sh/ruff-pre-commit (ruff + ruff-format hooks) to CONTRIBUTING.md, so contributors can auto-check lint/format before committing.

  Closes #337

## Summary by CodeRabbit

* **Documentation**
  * Expanded the contributing guide with instructions for optional pre-commit hooks.
  * Added guidance for installation, Ruff configuration, Git hook setup, and manually running checks across the project.

PR #354, opened 2026-08-25
docs: clarify mypy runs in CI, not pre-commit

### LINKED ISSUE #337: Add ruff and mypy pre-commit example to CONTRIBUTING
## Context
New contributors push, wait for CI, then fix formatting.

`CONTRIBUTING.md` lists manual commands:

```bash
ruff check src/ tests/
mypy src/continuum
```

There is no copy paste pre-commit example. Most newcomers expect a hook that formats on commit.

## What to do
Add a minimal `.pre-commit-config.yaml` snippet to `CONTRIBUTING.md` under the Linting section. For example:

```yaml
repos:
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.16.3
    hooks:
      - id: ruff
      - id: ruff-format
```

Mention it is optional and not required for CI, but recommended. Note to run `pre-commit install` after copying.

Optionally add `examples/pre-commit-config.yaml` as a real file so users can copy it.

## How to verify
Copy the snippet to `.pre-commit-config.yaml`, run `pre-commit run --all-files`, and confirm it matches `ruff format --check`.

## Files
- `CONTRIBUTING.md`
- `docs/CONTRIBUTING_ONBOARDING.md` (optional)
- `examples/pre-commit-config.yaml` (optional)

Docs only.

unmoha:
Closes #337

Adds a note clarifying why mypy is not included in the pre-commit
setup added in #341. Pre-commit hooks run in isolated environments
without the project's dependencies, so mypy would produce inaccurate
results there. CI already runs `mypy src/continuum` in strict mode
against a full environment, so this documents that as the source of
truth instead of adding an unreliable local hook.

## Summary by CodeRabbit

* **Documentation**
  * Expanded contributor guidance for installing pre-commit hooks and running them automatically on each commit.
  * Added instructions for manually running pre-commit checks across all files.
  * Clarified how to run mypy locally or through CI.
  * Updated the documented project structure for improved readability.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,18 +89,17 @@ repos:
 Then install the git hook:
 
 ```bash
-pre-commit install
+pre-commit run --all-files
 ```
 
-Now `ruff check --fix` and `ruff format` will run automatically on every `git commit`.
-To run it manually against all files:
-
-```bash
-pre-commit run --all-files
-```  
+**Note on mypy:** mypy is intentionally not included in this pre-commit setup.
+Pre-commit hooks run in isolated environments without the project's installed
+dependencies, so mypy would produce inaccurate results there. Instead, mypy
+runs in CI (`mypy src/continuum`, strict mode) against a full environment with
+all dev dependencies installed. Always check `mypy` locally with `pip install
+-e ".[dev]"` completed, or rely on the CI check on your PR.
 
 ---
-
 ## Project Structure
 
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,8 @@ ruff format --check src/ tests/
 mypy src/continuum
 ```
 
-The CI pipeline runs all three on every PR. A clean PR must pass all checks.  
+The CI pipeline runs all three on every PR. A clean PR must pass all checks.
+
 ### Pre-commit hooks (optional but recommended)
 
 To catch lint and formatting issues automatically before you commit, you can use
@@ -89,35 +90,43 @@ repos:
 Then install the git hook:
 
 ```bash
+pre-commit install
+```
+
+Now `ruff check --fix` and `ruff format` will run automatically on every `git commit`.
+
+To run it manually against all files:
+
+```bash
 pre-commit run --all-files
 ```
 
 **Note on mypy:** mypy is intentionally not included in this pre-commit setup.
 Pre-commit hooks run in isolated environments without the project's installed
-dependencies, so mypy would produce inaccurate results there. Instead, mypy
-runs in CI (`mypy src/continuum`, strict mode) against a full environment with
-all dev dependencies installed. Always check `mypy` locally with `pip install
--e ".[dev]"` completed, or rely on the CI check on your PR.
+dependencies, so a mypy hook there would produce inaccurate results. Instead,
+mypy runs in CI (`mypy src/continuum`, strict mode) against a full environment.
+To check locally, run `pip install -e ".[dev]"` once, then run `mypy
+src/continuum` yourself - or rely on the CI check on your PR.
 
 ---
+
 ## Project Structure
 
-```
 src/continuum/
-├── __init__.py          # Public API surface
-├── models.py            # Immutable Pydantic data models
-├── events.py            # Append-only event log
-├── actions/             # Action ledger (idempotency + reconciliation)
-├── checkpoint/          # Checkpoint manager + policies
-├── environment/         # Environment snapshot + diff
-├── recovery/            # Recovery engine + repair planner
-├── security/            # Content hashing, ID generation
-├── state/               # State projection, extraction, diffing, validation
-└── storage/             # Storage interface + SQLite engine
+├── init.py # Public API surface
+├── models.py # Immutable Pydantic data models
+├── events.py # Append-only event log
+├── actions/ # Action ledger (idempotency + reconciliation)
+├── checkpoint/ # Checkpoint manager + policies
+├── environment/ # Environment snapshot + diff
+├── recovery/ # Recovery engine + repair planner
+├── security/ # Content hashing, ID generation
+├── state/ # State projection, extraction, diffing, validation
+└── storage/ # Storage interface + SQLite engine
 
-tests/                   # Mirrors src/continuum/ structure
-docs/                    # Website (deployed to GitHub Pages)
-```
+tests/ # Mirrors src/continuum/ structure
+docs/ # Website (deployed to GitHub Pages)
+
 
 ---
 


CONTRIBUTING.md
```diff
@@ -89,18 +89,17 @@ repos:
 Then install the git hook:
 
 ```bash
-pre-commit install
+pre-commit run --all-files
```
Cyrax321, line 101: `pre-commit install` is deleted here, and it is now absent from the entire file, so nothing installs the git hook any more. Line 89 still reads "Then install the git hook:" directly above a command that installs nothing, so a reader following this section ends up with a config file, one manual run, and no hooks on commit.

That regresses the ruff half from #341, and #337 specifically asked to note `pre-commit install` after copying the config.

CodeRabbit suggested resolving the mismatch by renaming the heading instead. Please don't take that route. Restore `pre-commit install` under the existing heading, keep the sentence about `ruff check --fix` and `ruff format` running automatically on every `git commit`, and keep `pre-commit run --all-files` below it as the manual option, which is how the section read before this PR. The mypy note then goes after all of that.

CONTRIBUTING.md
```diff
@@ -89,18 +89,17 @@ repos:
...
-
-```bash
-pre-commit run --all-files
-```  
+**Note on mypy:** mypy is intentionally not included in this pre-commit setup.
+Pre-commit hooks run in isolated environments without the project's installed
+dependencies, so mypy would produce inaccurate results there. Instead, mypy
+runs in CI (`mypy src/continuum`, strict mode) against a full environment with
+all dev dependencies installed. Always check `mypy` locally with `pip install
+-e ".[dev]"` completed, or rely on the CI check on your PR.
```
Cyrax321: This sentence doesn't parse: "Always check `mypy` locally with `pip install -e \".[dev]\"` completed". It also never states the command to run, so it reads as though installing the dependencies is itself the check.

Suggest something closer to: run `mypy src/continuum` locally after `pip install -e ".[dev]"`, or rely on the CI check on your PR. Naming the same invocation CI uses removes any ambiguity about what's authoritative.

CONTRIBUTING.md
```diff
@@ -89,18 +89,17 @@ repos:
...
 
 ---
-
 ## Project Structure
```
Cyrax321, line 113: nit: the blank line between `---` and this heading was removed. The other eight section separators in this file all keep it, so this is the only one that differs.

Cyrax321 (changes requested):
Thanks for the quick turnaround, and the approach is right. CI-first is the option I'd have picked: pre-commit builds each hook its own isolated venv, so a mypy hook there would see none of the project's dependencies and strict mode would degrade into missing-import noise instead of real errors. Your claims check out too, which I verified: `strict = true` is set in `pyproject.toml`, and the `Lint & Type-check` job runs `mypy src/continuum` against a full `pip install -e ".[dev]"` environment.

One blocking problem. The diff deletes `pre-commit install`, and it is now absent from the entire file, so the ruff half that #341 added no longer works as documented. Details inline.

Worth saying that I disagree with CodeRabbit's suggested fix on that one. It proposed renaming the heading to describe a manual run, which would make the section self-consistent by permanently dropping hook installation. That is the wrong direction: #337 explicitly asked to note `pre-commit install` after copying the config, and running the hooks once by hand is not what that part of the doc is for.

The mypy note itself is a good addition and should stay. Two smaller things inline as well. Re-request review once the install step is back.

Cyrax321:
Separate from the change itself, and worth flagging while this is still open: your commits are landing with the email `anwar7926@gmail.come`. The trailing `e` means GitHub can't match it to your account, so `dc430b7` here is unattributed, as are `c604d99` and `bba6131` from #341.

The practical effect is that you don't appear in the repo's contributors list at all, even though #341 is merged. The one commit of yours carrying a working address, `9dab246`, is a merge commit, and GitHub's contributors graph excludes merges. So there is currently nothing on the graph crediting you, which isn't right.

There is no `.come` TLD, so that address can't be verified on GitHub, and the two commits already in `main` can't be reclaimed without rewriting the default branch, which I'm not going to do for two docs commits. This PR is still a single unmerged commit though, and you need to push again anyway for the fix above, so amending the author costs you nothing and would give you your first attributed commit here:

```bash
git config --global user.email "anwarmoha7926@gmail.com"
git commit --amend --reset-author --no-edit
git push --force-with-lease
```

Use whichever address is verified under Settings > Emails, or the `users.noreply.github.com` one if you'd rather not have it public. Worth fixing globally either way, since this is silently costing you credit on every repo you contribute to, not just this one.

anya-research:
**Code Review**

This PR aims to close #337's mypy remainder by documenting that mypy is CI-only rather than a pre-commit hook, which is the correct trade-off: `pyproject.toml:102` sets `strict = true` and `.github/workflows/ci.yml:104` runs `mypy src/continuum` against a full `pip install -e ".[dev]"` install, while pre-commit's isolated venvs would produce missing-import noise. The added rationale is accurate and valuable, but the diff regresses the ruff half landed in #341 by deleting the hook installation step, leaving the section self-contradictory and incomplete.

**high**
`CONTRIBUTING.md:89-93`
The diff replaces `pre-commit install` with `pre-commit run --all-files` and deletes the sentence `Now ruff check --fix and ruff format will run automatically on every git commit.` Heading on line 89 still says "Then install the git hook:" but the command no longer installs anything, and no other line in the file does. A contributor following the docs gets a one-off manual run with no hooks on commit, breaking the optional-but-recommended ruff setup #337 explicitly asked to note `pre-commit install` for.
```suggestion
Then install the git hook:

```bash
pre-commit install
```

Now `ruff check --fix` and `ruff format` will run automatically on every `git commit`.
To run it manually against all files:

```bash
pre-commit run --all-files
```
```

**medium**
`CONTRIBUTING.md:99-100`
The sentence `Always check mypy locally with pip install -e ".[dev]" completed` is ungrammatical and never states the command to run, reading as if installing deps is itself the check. Contributors will be unsure what constitutes a valid local mypy run, and may skip or misrun `mypy src/continuum`, the exact invocation CI uses as source of truth, delaying feedback until CI fails.
```suggestion
all dev dependencies installed. Run `mypy src/continuum` locally after
`pip install -e ".[dev]"`, or rely on the CI check on your PR.
```

**low**
`CONTRIBUTING.md:102-103`
The blank line between the `---` separator and `## Project Structure` was removed, making this the only one of nine section separators in the file without it. No functional impact, but inconsistent formatting invites churn in follow-ups.
```suggestion
---

## Project Structure
```

[github run](/Cyrax321/CONTINUUM/actions/runs/32821647087)

unmoha:
Thanks for the detailed review. I've addressed everything:

- Restored the `pre-commit install` step and the manual
  `pre-commit run --all-files` step as two separate blocks, so the
  ruff instructions from #341 work as documented again.
- Rewrote the mypy note to explicitly name the command to run
  locally (`mypy src/continuum`), rather than implying that
  installing dependencies alone is the check.
- Added back the blank line after `---` before the heading, and
  fixed my commit author email so this and future commits are
  attributed correctly to my account.

Cyrax321 (approved):
Thanks for contributing to CONTINUUM, and for the fast turnaround on the review.

Approving. The blocking item is fixed: `pre-commit install` is back under its heading, the sentence about ruff running on every `git commit` is intact, and `pre-commit run --all-files` sits below it as the manual option. The two smaller points are handled too, and the mypy note now names `mypy src/continuum` outright instead of implying the dependency install is itself the check.

The rationale you wrote is the part of this I value most. Explaining why mypy stays out of pre-commit, rather than just noting that it does, is what stops the next person reopening the question. It also checks out: strict mode is set in `pyproject.toml`, and the CI job runs that exact invocation against a full `pip install -e ".[dev]"` environment, which a pre-commit hook's isolated venv could not reproduce.

One thing I am taking off your plate instead of sending back. The Project Structure block lost its surrounding code fences somewhere along the way, so the tree now renders as a run-on paragraph and `__init__.py` came through as `init.py`. That section is outside what #337 asked for and it is mine to keep tidy, so I will restore it in a follow-up on main. Nothing for you to do here. For next time: that failure mode usually comes from copying a block out of GitHub's rendered view rather than editing the raw file, and a quick `git diff` before pushing catches it in one look.

Thanks for sorting the author email as well. `bbe3853` is correctly attributed to your account, and I am squash merging so the whole change lands under your name.

Merging now. If you want another, #360 is close to this one in spirit: `docs/api/cli.md` is missing 13 commands that actually ship. Thanks again, and happy to have you back.

PR #373, opened 2026-08-25
fix(mcp): reconcile the action the recovery guidance actually names

### LINKED ISSUE #367: Recovery guidance advertises action_id where reconcile_action requires the idempotency key
## What happens

Every recovery surface tells the caller to pass an `action_id` to `continuum_reconcile_action`, but that tool keys on the idempotency key. Following the guidance verbatim fails, and no MCP tool exposes the value that would work. An `UNKNOWN` action created over MCP cannot be reconciled through any documented interface.

## Reproduction (MCP only)

`continuum_resume` on a run with one uncertain action returns, consistently, in five places:

```
next_allowed_action: "reconcile_action:action_d17ff043eff3b30f26c458220b0b2c70"
contract.required_actions[0]: "reconcile_action:action_d17ff043eff3b30f26c458220b0b2c70"
human_steps[0]: "check whether 'registry_publish' actually reached the outside world,
                 then call continuum_reconcile_action(run_id=..., 
                 action_key=action_d17ff043eff3b30f26c458220b0b2c70, occurred=true|false)"
informed_retry.avoid[2]: "registry_publish (action_d17ff04): outcome unknown"
report: "1. [human] reconcile_action action_d17ff043eff3b30f26c458220b0b2c70"
```

Following it:

```
reconcile_action(action_key="action_d17ff043eff3b30f26c458220b0b2c70", occurred=false)
  -> no action recorded for key action_d17ff...
```

The value that works is the idempotency key `742bde60d24df3ee17ea697ff22507fd1fc40fa7d847837c3add9a9f35b323c0`.

## Why the right key is unreachable

- `continuum_list_actions` returns `action_id`.
- `continuum_resume` returns `action_id` in `uncertain_actions`.
- `continuum_complete_action`, `continuum_fail_action` and `continuum_reconcile_action` all return `action_id`.
- `continuum actions --json` returns `action_id` and `arguments_hash`. `arguments_hash` looks like a key and is a plausible guess, but it is also rejected (verified: `no action recorded for key 27d373d7f963...`), because the key here was derived from the explicit `key=` override rather than from the arguments.
- The `UnknownSideEffect` response from `continuum_intercept_action` omits `action_key` entirely. The only trace is a 12-character truncated prefix inside the free-text `reason`.
- `continuum reconcile` needs a registered probe. With none registered it reports `no probe registered: 1` and settles nothing.

So the key appears in exactly one place: the raw `ACTION_RECORDED` payload visible through `continuum events`. A recovering session that lost its context, which is the situation this subsystem exists for, has no path to it.

## Root cause

`recovery/guidance.py` fills the `action_key` slot with the plan step's target, which is an action id:

```python
_RECONCILE_HINT = (
    "call continuum_reconcile_action(run_id={run}, action_key={key}, occurred=true|false)"
)
...
steps.append(
    f"check whether {action_type!r}{key_hint} actually reached the outside "
    f"world, then {_RECONCILE_HINT.format(run=run_id, key=action_id)}"
)
```

## Suggested fix

Three parts, all small.

1. Make `reconcile_action` accept either identifier. Resolve an `action_id` to its ledger key by scanning the fold, since the mapping is already there. This alone makes every existing piece of guidance correct.
2. Add `action_key` to the `UnknownSideEffect` response from `intercept_action`, and to the rows in `list_actions` and `uncertain_actions`. The agent should never have to reconstruct it.
3. Improve the `LedgerError` message. `no action recorded for key <prefix>...` should say which identifier spaces exist and how to list them, rather than leaving the caller to guess.

## Environment

Reproduced against 8a42a29 through the MCP server, run `mcp_arch_audit_v2`.

Cyrax321:
Closes #367.

`continuum_resume` reports an uncertain action by `action_id` in five places, and `human_steps` spells out the call to make:

```
next_allowed_action: "reconcile_action:action_d17ff043eff3b30f26c458220b0b2c70"
human_steps[0]: "... then call continuum_reconcile_action(run_id=..., 
                 action_key=action_d17ff043eff3b30f26c458220b0b2c70, occurred=true|false)"
```

Following it verbatim:

```
reconcile_action(action_key="action_d17ff043eff3b30f26c458220b0b2c70")
  -> no action recorded for key action_d17ff...
```

The ledger keys on the idempotency key, and nothing exposed it. `list_actions` and `uncertain_actions` returned `action_id`. The `UnknownSideEffect` response omitted `action_key` entirely, leaving a 12-character truncated prefix inside free-text `reason` as the only trace. `continuum actions --json` returns `arguments_hash`, which looks like a key, is a plausible guess, and is also rejected because the key here derives from the explicit `key=` override rather than from the arguments. `continuum reconcile` needs a registered probe and settles nothing without one.

Net effect: an `UNKNOWN` action created over MCP was unreconcilable through every documented interface. In the audit that found this, recovery only succeeded because the session still held the pre-crash context, which is exactly what recovery must not assume.

## Approach

Resolve both identifier spaces in the ledger rather than special-casing one tool. `ActionLedger.resolve_key` accepts a key or an `action_id`, and `_require` returns the resolved key alongside the action, so `complete`, `fail`, `reconcile`, `compensate` and `flag_for_review` all inherit it. That makes the guidance already shipping in `next_allowed_action`, `required_actions`, `human_steps` and the rendered report executable as written, instead of rewriting five outputs in terms of an identifier none of them had.

Settlement still records under the fold's own key, so passing an `action_id` cannot grow a second ledger entry for one action. The mapping is unambiguous: one key holds one action, and a re-claim after `FAILED` or `COMPENSATED` copies the existing action rather than reissuing its id.

Then close the reporting gap, so the fix does not depend on the caller guessing right. `UnknownSideEffect` carries `action_key` and `action_id`; `continuum_intercept_action` returns both on the unknown path; `continuum_list_actions` rows and `continuum_resume`'s `uncertain_actions` gain `action_key`. Telling a caller to reconcile without telling it what to reconcile is not actionable, and a recovering session is the one least able to reconstruct the identity itself.

The unmatched-identifier message now names both spaces and how to list them, rather than `no action recorded for key <prefix>...` which could not distinguish a wrong-space mistake from a nonexistent action.

## Tests

Six regressions, all failing before this change. At the ledger: settle by `action_id` without duplicating the entry, `UnknownSideEffect` exposing a key that is usable as reported, and the reworded refusal. At the MCP surface: the identifier `resume` advertises being accepted by `reconcile_action`, `action_key` present and consistent across the unknown path, `list_actions` and `uncertain_actions`, and the error naming both spaces.

Full suite green, ruff clean, mypy clean across 104 files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -714,6 +714,28 @@ All notable changes to this project are documented here. The format follows
 
 ### Fixed
 
+- **Recovery guidance named an identifier the settle tools rejected (#367).**
+  `continuum_resume` reports uncertain actions by `action_id` in five places
+  (`next_allowed_action`, the contract's `required_actions`, `human_steps`,
+  `informed_retry.avoid` and the rendered report), and `human_steps` spelled out
+  a `continuum_reconcile_action(action_key=<action_id>)` call. The ledger keyed
+  only on the idempotency key, so following that instruction verbatim failed with
+  `no action recorded for key ...`, and no MCP surface exposed the value that
+  would have worked: `list_actions` and `uncertain_actions` reported `action_id`,
+  the `UnknownSideEffect` response omitted the key entirely and left a
+  12-character truncated prefix in free text as its only trace, `arguments_hash`
+  from `continuum actions --json` looks like a key but is a different hash, and
+  `continuum reconcile` needs a registered probe. An `UNKNOWN` action created
+  over MCP was therefore unreconcilable through every documented interface.
+  `ActionLedger.resolve_key` now accepts either space and `_require` returns the
+  resolved key, so `complete`, `fail`, `reconcile`, `compensate` and
+  `flag_for_review` all take an `action_id` or a key and settle under the fold's
+  own key either way. `UnknownSideEffect` carries `action_key` and `action_id`,
+  which `continuum_intercept_action` returns on the unknown path, and both
+  `continuum_list_actions` rows and `continuum_resume`'s `uncertain_actions` gain
+  `action_key`. The unmatched-identifier message names both spaces and says how
+  to list them.
+
 - **`ActionLedger` could not serialise concurrent claims on one key (#345).**
   `claim()` deduplicates by folding the log and then appending, with nothing
   between the read and the write, so processes racing on one key could each be
--- a/src/continuum/actions/ledger.py
+++ b/src/continuum/actions/ledger.py
@@ -423,6 +423,34 @@ class ActionLedger:
     def get(self, key: str) -> Action | None:
         return self._replay().get(key)
 
+    def resolve_key(self, identifier: str) -> str | None:
+        """The ledger key for ``identifier``, which may be a key or an ``action_id``.
+
+        The ledger is keyed by idempotency key, but almost everything a caller
+        reads back is keyed by ``action_id``: ``Action.action_id`` itself, the
+        recovery plan's ``reconcile_action:<target>`` steps, the contract's
+        ``required_actions``, and the rendered report. So the identifier a
+        recovering caller has in hand is usually the one the settle methods did
+        not accept, and the two are indistinguishable by shape (issue #367).
+
+        Resolving both here rather than at one call site means every settle
+        method inherits it, and the recovery guidance that names an ``action_id``
+        becomes executable as written instead of needing to be rewritten in terms
+        of an identifier no output exposes.
+
+        The mapping is unambiguous: one key holds one action, and a re-claim after
+        FAILED or COMPENSATED copies the existing action, so ``action_id`` stays
+        with its key rather than being reissued. Returns ``None`` when neither
+        space matches.
+        """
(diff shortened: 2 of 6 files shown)


Cyrax321:
Pushed a follow-up: `LedgerError` was not on `_refusal_reaches_the_caller`'s converted list, so the new message never reached the caller under mcp >= 2.1.0. It read `Error executing tool continuum_reconcile_action` instead, which is the least useful possible reply to being handed the wrong identifier.

Worth flagging how this got through: my local venv had mcp 2.0.0, where an unrecognised handler exception keeps its text, so the suite passed locally and failed on all nine test matrix jobs. `pyproject.toml` pins `mcp>=2.0` and CI resolves the newest, so the suite's behaviour depends on which version happens to be installed. I upgraded locally and reproduced it before fixing.

This is not new to this PR: every `LedgerError` refusal has been opaque over MCP since 2.1.0. It just becomes visible here because this PR is the one that gives that error something useful to say.

PR #375, opened 2026-08-25
docs: restore the Project Structure code fence in CONTRIBUTING

Cyrax321:
Follow-up to #354, which landed with the code fences around the Project Structure tree removed. The block currently renders as a run-on paragraph rather than a code listing, the aligned comment column is collapsed to single spaces, and `__init__.py` reads as `init.py`, naming a file that does not exist.

This restores that block verbatim from its state before #354 and puts back the trailing newline at end of file. It also joins the `mypy src/continuum` code span onto a single source line so copying the command out of the raw file does not pick up a line break. The mypy note and the `pre-commit install` step from #354 are otherwise untouched.

CI did not catch this because there is no markdown linter in any workflow and no committed `.pre-commit-config.yaml`, so nothing checks fence balance. Worth a separate issue if we want that covered.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -105,28 +105,29 @@ pre-commit run --all-files
 Pre-commit hooks run in isolated environments without the project's installed
 dependencies, so a mypy hook there would produce inaccurate results. Instead,
 mypy runs in CI (`mypy src/continuum`, strict mode) against a full environment.
-To check locally, run `pip install -e ".[dev]"` once, then run `mypy
-src/continuum` yourself - or rely on the CI check on your PR.
+To check locally, run `pip install -e ".[dev]"` once, then run
+`mypy src/continuum` yourself, or rely on the CI check on your PR.
 
 ---
 
 ## Project Structure
 
+```
 src/continuum/
-├── init.py # Public API surface
-├── models.py # Immutable Pydantic data models
-├── events.py # Append-only event log
-├── actions/ # Action ledger (idempotency + reconciliation)
-├── checkpoint/ # Checkpoint manager + policies
-├── environment/ # Environment snapshot + diff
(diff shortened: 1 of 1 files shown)


Cyrax321:
Verified against the history rather than the description. Every claim holds.

`#354` did remove the fences, and the diff shows exactly the three symptoms named: the opening and closing ``` gone, the aligned comment column collapsed to single spaces, and `__init__.py` rewritten as `init.py`. That last one names a file that does not exist while `src/continuum/__init__.py` does, so a newcomer following the tree looks for the wrong thing.

The restore is verbatim. Diffed the Project Structure block against `ad6c315^`, the commit before #354: identical. Fence markers balance at 16 across the file, and the trailing newline is back.

#354's actual contribution survives intact: both the "mypy is intentionally not included" note and the `pre-commit install` step are still present. This only undoes the collateral damage, which is the right scope for a revert-shaped fix.

The code span fix is worth more than it looks. On `main` the span is split across a newline:

```
To check locally, run `pip install -e ".[dev]"` once, then run `mypy
src/continuum` yourself - or rely on the CI check on your PR.
```

Markdown renders that as one line, so it looks fine, but copying `mypy src/continuum` out of the raw file picks up the break. Moving the span onto a single source line is the fix. Changing `yourself - or` to `yourself, or` also drops a hyphen being used as a dash, which matches house style.

Confirmed the CI observation too: no markdown linter in any workflow, and no committed `.pre-commit-config.yaml`, so nothing checks fence balance. That is worth its own issue as you suggest. A single `markdownlint` step catching unbalanced fences would have caught this at #354 rather than after it shipped.

One thing I checked that is not visible from the PR: this does not collide with in-flight uncommitted work on the same file. There is a local edit adding a "Good first issues" section further down, and a three-way merge of the two is clean with zero conflict markers, so merging this does not put that at risk.

Merging.

Issue #369, opened 2026-08-25
self_report_guidance says nothing is wrong while an action outcome is unresolved


Cyrax321:
## What happens

`self_report_guidance` tells the caller "Nothing is wrong with this run. Work is not blocked" in the same `continuum_resume` response whose contract says `recovery_status: requires_human` because a side effect has an unresolved outcome.

## Reproduction (MCP only)

One run, one `registry_publish` escalated to `UNKNOWN`, then `continuum_resume`:

```
contract.recovery_status: "requires_human"
contract.reason: "1 external side effect(s) have unknown outcomes; at least one repair needs a person"
next_allowed_action: "reconcile_action:action_d17ff043..."
uncertain_actions: [{action_type: "registry_publish", status: "unknown"}]

self_report_guidance: "Nothing is wrong with this run. request_human here means only that
its goal and progress were reported by an agent and nothing independent corroborates them.
Work is not blocked: recording progress, checkpointing and the action tools all still
function. ... Say that once and carry on."
```

The guidance is telling the agent to carry on past an unreconciled side effect, which is the one thing this system exists to stop.

## Why

`self_report_guidance` decides whether self-report is the only blocker by scanning the validation report:

```python
blocking = [e for e in decision.validation.report.statuses if e.status is not StateStatus.VALID]
only_self_report = bool(blocking) and all(
    e.component in (Component.GOAL, Component.PROGRESS)
    and e.status is StateStatus.REQUIRES_REVIEW
    for e in blocking
)
```

Uncertain actions are not validation-report components. They live in `decision.uncertain_actions` and reach the mode through the ledger, so they never appear in `blocking` and `only_self_report` stays true.

The function's own docstring already states the intended behaviour: "When anything else is also wrong (a conflicted dependency, an unresolved action) the caller has a real problem to fix and this note would only dilute it, so it is omitted." The dependency half works, verified: with a conflicted `registry_api` the guidance was correctly absent. The unresolved-action half was never wired up.

## Suggested fix

Add the ledger to the test:

```python
if decision.uncertain_actions:
    return {}
```

Worth checking the same way whether any other non-validation input can drive `REQUEST_HUMAN`, since the current predicate assumes the validation report is the whole picture. A regression test with one uncertain action and otherwise clean state would pin it.

## Environment

Reproduced against 8a42a29 through the MCP server, run `mcp_arch_audit_v2`.

PR #377, opened 2026-08-25
fix(mcp): propagate confirm handler refusals

### LINKED ISSUE #371: confirm_gate leaves the handler call outside _refusal_reaches_the_caller
## What happens

`confirm_gate` closes its `_refusal_reaches_the_caller()` block before invoking the handler, so a domain refusal raised inside `continuum_confirm` would reach the client as the opaque `Error executing tool continuum_confirm` instead of its own message.

## The difference

`guard` calls the handler inside the context manager:

```python
with _refusal_reaches_the_caller():
    auth.verify(caller, token_from(ctx))
    policy.require(caller, fn.__name__)
    return fn(*args, **kwargs)
```

`confirm_gate` does not:

```python
with _refusal_reaches_the_caller():
    confirm_auth.verify(token_from(ctx))
    policy.require(caller, fn.__name__)
return fn(*args, **kwargs)
```

`continuum_confirm` calls `ctx.storage.append_event(run_id, ...)`, which raises `RunNotFound` for an unknown run. Under `guard` that becomes a `ToolError` carrying `no such run: '<id>'`. Here it stays a bare domain exception, which is the case `_refusal_reaches_the_caller` documents as needing conversion: "a run that does not exist. Each answer is only useful if the caller is told which one it was."

## Severity

Latent. Not reproducible in the default configuration, because confirmation refuses every caller before the body runs when `CONTINUUM_MCP_CONFIRM_TOKEN` is unset. It becomes reachable as soon as an operator opts in, which is exactly when a confirm call against a mistyped run id is likely.

Found by inspection while auditing the MCP surface, not by observation. Flagging it rather than filing it as reproduced.

## Suggested fix

Move `return fn(*args, **kwargs)` inside the `with` block, matching `guard`. The authenticate-before-authorize ordering that the surrounding comment protects is unaffected.

A test asserting that both decorators convert a domain exception raised by the handler body would keep the two from drifting again.

zynx-real:
## Description

`confirm_gate` previously invoked the wrapped confirmation handler outside `_refusal_reaches_the_caller()`. Domain exceptions raised by the handler could therefore become opaque MCP execution errors.

The handler now executes inside the refusal context while preserving authentication-before-authorization ordering.

Regression coverage verifies that a handler-originated `RunNotFound` is converted to a `ToolError` retaining its useful domain message.

## Related issues

Issue #371.

## Types of changes

- Type: `bugfix`

## Breaking changes

No.

## How has this been tested?

- `pytest tests/test_mcp_authz.py tests/test_mcp_server.py --tb=short -q`
- `pytest --tb=short -q --cov=continuum --cov-report=term-missing`
- `ruff check src/ tests/ examples/`
- `ruff format --check src/ tests/ examples/`
- `mypy src/continuum`

Local checks passed on Python 3.14.7 with MCP 2.1.0. PostgreSQL service tests were unavailable locally and remain covered by CI.

## Checklist

Tests were added for the changed behavior. The Unreleased changelog was updated. Local CI-equivalent checks pass. The authentication and authorization boundary remains unchanged.

## Additional context

Closes #371

## Summary by CodeRabbit

- **New Features**
  - Added support for recording and updating run plans.
  - Checkpoints and validation responses now include model, provider, constraint, lesson, and action details.
  - Confirmation can be scoped to specific goals or progress.
  - Action reconciliation supports structured results and improved retry tracking.
  - Action listings remain available in slim mode.

- **Bug Fixes**
  - Confirmation requests now preserve meaningful error messages when a requested run is missing or another domain error occurs.
  - Successful confirmation calls complete normally while authorization and authentication errors continue to be handled appropriately.


Cyrax321 (approved):
Thank you for the contribution. This PR moves the `continuum_confirm` handler call inside `_refusal_reaches_the_caller` in `confirm_gate`, matching `guard`, while keeping authentication before authorization. I reviewed `src/continuum/mcp/server.py` around `confirm_gate` and `guard`, the `_refusal_reaches_the_caller` converter, and `tests/test_mcp_authz.py` including the new `test_confirm_handler_refusal_carries_its_reason`. The change preserves the intended refusal path so `RunNotFound` for a missing run is reported as `ToolError` with `no such run: 'missing'` instead of opaque execution error, which is the case documented for this converter.

No blocking correctness issues were found. The design at this boundary is the best available shape and avoids a larger refactor.

Residual: `CHANGELOG.md` has a merge conflict and the required checks are still pending workflow approval for this fork. Once the branch is rebased on `main` and CI is green, this is ready to pull.

Thank you again for the fix and for adding the regression coverage.

Cyrax321 (dismissed):
Thank you for the contribution. This PR moves the `continuum_confirm` handler call inside `_refusal_reaches_the_caller` in `confirm_gate`, matching the pattern already used in `guard`, while preserving the existing ordering of authentication before authorization.

I reviewed `src/continuum/mcp/server.py` around `confirm_gate` and `guard`, the `_refusal_reaches_the_caller` converter, and `tests/test_mcp_authz.py`, including the new `test_confirm_handler_refusal_carries_its_reason`. The change preserves the intended refusal path, so `RunNotFound` for a missing run is reported as `ToolError` with `no such run: 'missing'` instead of an opaque execution error, consistent with the documented behavior for this converter.

No blocking correctness issues were found. The design at this boundary is the best available shape and avoids a larger refactor.

**Outstanding items:** `CHANGELOG.md` has a merge conflict, and the required checks are still pending workflow approval for this fork. Once the branch is rebased on `main` and CI is green, this is ready to merge.

Thanks for the fix and for adding the regression coverage.

Cyrax321:
Apologies, this is my fault and I have undone it.

The timeline: you commented on #371 at 09:40 UTC and opened this PR at 09:54. At 11:14 I merged #379, which fixed the same thing as part of a two-issue PR, without checking either the issue or the open PR list first. That made your work redundant through no fault of yours, and merging over a live PR is worse than merging over an unclaimed issue.

#381 has now reverted my `confirm_gate` change from `main`, so the bug is back and this PR is live again. I left the unrelated `self_report_guidance` fix (#369) from that same PR in place, since nobody had claimed it and reverting it would have been churn for its own sake. #371 is reopened and assigned to you.

Your version is the better one to land regardless. The fix is identical, and your test asserts the exact run id in the message (`no such run: 'missing'`) where mine only matched the prefix, so it pins more.

One thing left before it can merge, and it is my mess rather than yours: `CHANGELOG.md` conflicts because four entries landed in the same `### Fixed` region while this PR was open. Your `server.py` and test changes still auto-merge cleanly, only the changelog needs a hand. Merging `main` into your branch and keeping both sets of entries should clear it. You allow maintainer edits, so say the word and I will do that push instead if you would rather not, but it is your PR and I would rather not touch it without asking.

Going forward I will assign issues to whoever comments first, so this cannot happen the same way again.

Cyrax321:
Hi @zynx-real, your branch had a CHANGELOG.md conflict after the revert, so I pushed a resolution keeping your entry. No change to src/continuum/mcp/server.py or tests/test_mcp_authz.py.

Auto-merge squash is still enabled, so it will merge when green. You will be in the contributors graph and I will add you to the README wall after merge.

Thanks for the fix and the precise test on no such run: 'missing'. If you want to tweak the CHANGELOG wording, just push to the branch.

Cyrax321 (approved):
Re-approving after changelog resolution and merges at d6b221b. Verified src/continuum/mcp/server.py confirm_gate handler now inside _refusal_reaches_the_caller preserving auth-before-authz, test_confirm_handler_refusal_carries_its_reason asserts no such run: missing. All 18 checks passed. Ready to merge. Fixes #371

Issue #364, opened 2026-08-25
record_progress can commit an unprojectable TASK_UPDATED, permanently bricking a run


Cyrax321:
## What happens

A single `continuum_record_progress` call can append a `TASK_UPDATED` event that no later projection can fold, which permanently disables every projecting surface for that run. There is no supported repair.

## Reproduction (MCP only)

```
record_progress(run_id=R, completed=0, total=6, goal=...)  -> ok
record_progress(run_id=R, completed=3, total=6)            -> ok
record_progress(run_id=R, completed=99)                    -> ToolError from Progress
record_progress(run_id=R, completed=2, total=6)            -> same ToolError, forever
```

## Why

`continuum_record_progress` validates `completed + failed > total` only when `total` is supplied in the same call:

```python
if total is not None and completed + failed > total:
    raise ValueError(...)
```

Omitting `total` skips the guard, but projection still folds the `total` recorded earlier, so the invariant is evaluated against a limit the call never mentioned. `_apply_progress` re-derives `pending` correctly (`max(total - done, 0)` gives 0), so the rejected state is `total=6, completed=99, pending=0`, and `Progress` refuses it.

The durability is the real problem. The handler appends before it projects:

```python
ctx.storage.append_event(run_id, EventType.TASK_UPDATED, payload, source=AGENT_SOURCE)
state = project(run_id, ctx.storage.read_events(run_id))   # raises here
```

The event is already committed when validation fails. The fold validates each intermediate state, so no subsequent event can correct it. The handler comment says the check exists to keep bad values "out of the event log in the first place", which is the intent this path defeats.

## Blast radius

Dead for that run, over MCP: `record_progress`, `checkpoint`, `validate`, `resume`. Dead over CLI: `status`, `inspect`, `replay`, `show-contract`, `validate`, `resume`, `briefing`.

Still working: `intercept_action`, `complete_action`, `fail_action`, `reconcile_action`, `list_actions`, `record_summary`, and CLI `events`, `actions`, `history`, `runs`, `verify`.

That split is the dangerous part. The run can still authorize and fire real side effects while recovery is unable to answer whether continuing is safe, which inverts the model the ledger is built on.

[...]

PR #385, opened 2026-08-25
fix(state): an unprojectable log can degrade instead of killing every command (#383)

### LINKED ISSUE #383: An unprojectable run has no repair path: every recovery command needs the projection that is broken
## What happens

A run whose log cannot be folded has no supported route back. Every command that could repair, archive or escape it needs the projection that is broken, so they all fail on precisely the runs that need them.

## Reproduction

Against a run left unprojectable by #364:

```
$ continuum compact mcp_arch_audit_2026_08_25 --force
error: run 'mcp_arch_audit_2026_08_25' could not be anchored: 1 validation error for Progress
  Value error, completed + pending + failed exceeds total
```

`resume`, `status`, `inspect`, `replay`, `show-contract`, `validate` and `briefing` fail identically, and `fork` needs a parent state to diverge from. `continuum runs`, `events`, `actions` and `history` still work, so the run is visible but inert.

There is no `repair`, `amend`, `rewind` or `quarantine` command. The only exit is editing the event log by hand, which the hash chain deliberately makes hard, and which would break the chain that `verify` checks.

## Why this is a design gap rather than a leftover

#364 closed the way I found to produce this state, and the write path now rejects what the read path would reject. But the class stays open, because the fold validating each intermediate state means *any* invariant violation that reaches the log is permanent and uncorrectable by appending. Reachable through a different tool, a direct `Storage` caller, a migration, or a restore from a foreign checkpoint whose invariants differ.

The asymmetry is what makes it serious. A poisoned run keeps working for the action tools (`intercept_action`, `complete_action`, `list_actions` all fold only action events), so it can go on authorising real side effects while recovery is unable to say whether continuing is safe. See #382: `verify` reports it as intact throughout.

For a system whose thesis is that a crashed agent can always be told where it stands, "this run is now permanently unanswerable and there is no supported fix" is the wrong terminal state.

## Options

1. **Make the fold degrade rather than raise.** Return a state marked `INVALID` naming the offending sequence, so `resume` can say "this run is unprojectable from event 10, here is what is known up to event 9" instead of raising a pydantic error. This is the one that restores the guarantee, since the whole point is to always have an answer.
2. **A repair command** that appends a compensating event or marks a range superseded, keeping the chain intact and the history auditable. Fits the event-sourced model better than deletion, and pairs with option 1 naming the event.
3. **Allow `fork` from the last projectable prefix,** so the operator can continue the work in a child run without rewriting the parent. Cheapest to build, does not fix the parent.

Option 1 is the one I would want first: it is the difference between a degraded answer and no answer, and options 2 and 3 both need it to know where the break is.

## Environment

Reproduced against `main` at 129c9c4. `mcp_arch_audit_2026_08_25` is still in this state locally as a test case.

Cyrax321:
## Description

An event whose payload violates a model invariant (for example TASK_UPDATED with completed=999 against total=100) is hashed and appended like any other, but because project_incremental validates every intermediate state while folding, no later event can correct it: the chain verifies clean forever and every projecting command dies on precisely the run that needs them, while the action tools (which fold only ACTION_* events) keep authorising real side effects that recovery cannot assess. I reproduced this before changing anything by seeding a run through SQLiteStorage and appending store.append_event("run_1", EventType.TASK_UPDATED, {"completed": 999, "failed": 0}) (the existing _poison helper from tests/test_cli.py), then running continuum status, continuum compact --force, continuum resume and continuum inspect: all four exited 1 with a raw pydantic "Value error, completed + pending + failed exceeds total" traceback.

The fold can now degrade instead of raising. project() and project_incremental() accept on_unprojectable="raise"|"degrade", defaulting to "raise". Degrade mode stops at the earliest refused event and returns the last-good prefix marked SemanticState.status=StateStatus.INVALID, carrying unprojectable_at_sequence, unprojectable_event_type and unprojectable_reason (condensed to one line by the same logic #384 added). It never skips the bad event and keeps going, since the state after a skipped write is not a state the run was ever in; and if nothing folds at all before the break (no RUN_STARTED before it), degrade mode still raises, because there is no partial answer to invent.

Default behaviour is byte-for-byte unchanged for callers that do not opt in, and the opt-in set is limited to surfaces whose job is diagnosis or reporting: CheckpointManager.restore (both fold paths), driven by RecoveryEngine.assess so resume, validate, show-contract and briefing answer instead of dying; the CLI status, inspect and replay surfaces plus replay's stored-version verification; the serve sidecar's record_progress report (it appends first, so today it crashes after already committing the write) and its dependency dedup read; and the benchmark's naive-checkpoint readout. Deliberately kept raising: mcp/server.py's _project_candidate, because the #364 write guard exists to refuse an unprojectable candidate and degrading there would silently disable that protection; the checkpoint capture surfaces in mcp/server.py and serve/server.py and GenericAgentAdapter.capture_state, because a partial fold must not be pinned as a stored version or compaction anchor; and adapters/generic.py _declare_dependencies together with hooks.py record_file_progress, which already catch ProjectionError with documented fallbacks and would change established behaviour if converted. One correction to the issue's inventory: on current main there are 14 direct project() call sites across 7 files, and recovery/fork.py's hit is a comment, not a call site; every site was inspected individually rather than converted blindly.

For a degraded state the engine decides REQUEST_HUMAN (contract safety REQUIRES_HUMAN). ABORT would restate exactly the terminal state the issue calls wrong ("this run is permanently unanswerable and there is no supported fix"); REQUEST_HUMAN is this system's standing verdict whenever a person must decide, keeps repair possible once option 2 lands, and the severity ordering structurally guarantees a degraded state can never yield RESUME even when every other signal is clean. The decision rationale names the sequence, event type and condensed reason, and states explicitly that the verdict covers events through the last good sequence only.

## Related issues

Closes #383. Options 2 (a repair/amend command) and 3 (fork-from-last-good-prefix) named in the issue are deliberately out of scope and are not attempted here; both depend on this landing first.

## Types of changes

- Type: `bugfix`

## Breaking changes

No, as of the third commit. The second review round correctly blocked on a real break the first two rounds missed: adding four fields to SemanticState changed what StateCheckpoint.content() serialises, so every pre-existing checkpoint failed verification with "integrity hash does not match content" (indistinguishable from tampering), and new bodies carried fields that older readers' extra="forbid" SemanticState rejects. That is fixed: PROJECTION_BOOKKEEPING names the four fields once and they are excluded from content(), from checkpoint digests, and from all four persisted-body write sites (sqlite and postgres, checkpoints and versions), which restores every existing checkpoint (the reviewer verified the digest matches against a real stored row) and keeps new databases readable by older builds. Commands that previously crashed on a poisoned log now answer with the documented non-zero exit codes (CORRUPTED / REQUIRES_HUMAN), which is the stated safety contract rather than a break.

## How has this been tested?

Each new test was verified to fail on unchanged source and pass after the change, proven directly: with the source files stashed (git stash push -- src/, keeping the tests) 13 of the 14 new tests failed (TypeError on the missing keyword, AttributeError on the missing status field, or wrong exit codes); after git stash pop all 14 pass. The 14th, test_a_healthy_compacted_run_still_resumes_after_degrade_landed, is a deliberate regression guard for the compaction merge (constraint 5) and pins behaviour that already holds on main. The contract round added five tests proven the same way (five failed with src/ checked out at this branch's first commit, all pass after). The compatibility round adds four: three fail against the branch's field-adding commit (the tampered-checkpoint reproduction plus both persist-shape assertions) and pass here; the old-version-row load test is a guard pinning behaviour that already holds. Gates on the final commit, Python 3.13.5 in a dedicated worktree venv with mcp upgraded to 2.1.0 beforehand per the known CI resolution trap: .venv/bin/python -m pytest (1425 passed, 23 skipped, 0 failed), .venv/bin/ruff check src tests (pass), .venv/bin/ruff format --check src tests (206 files already formatted), .venv/bin/mypy src/continuum (104 files, zero issues).

## Checklist

Confirmed: tests added or updated for the changed behaviour; CHANGELOG.md updated under Unreleased / Fixed; all four gates listed above pass on the latest commit; no security-relevant surface is weakened (the write-path guard and checkpoint persistence intentionally keep refusing partial folds, stated explicitly above).

## Additional context

state_fingerprint excludes the projection-bookkeeping fields: they describe how the log was read, not what the state says, so excluding them keeps put_version's fingerprint dedup stable across the upgrade, and a degraded prefix-state genuinely is the same task content as its healthy counterpart, which is exactly why the INVALID marker lives outside the hashed payload (covered by test_degraded_folds_content_address_like_their_prefix). Review round two showed that reasoning had to apply everywhere durability touches those fields, not just the fingerprint: StateCheckpoint.content() now excludes them from the integrity hash and all four storage write sites omit them from persisted bodies, with cross-version tests pinning serialised fixtures because no same-process round-trip can catch this class of break. compact --force still refuses an unprojectable run ("could not be anchored") by design: anchoring writes a storage boundary, and archiving past a break belongs with the repair command, not with a partial-fold anchor. One honest wrinkle visible in resume JSON on a poisoned run: next_allowed_action may name an unrelated repair step such as revalidate_dependency:dataset, because the planner works from the last-good validation; the projection break itself is carried in the contract reason and the rationale, and there is deliberately no auto-generated fix action for it until option 2 exists.

Closes #383

## Summary by CodeRabbit

* **New Features**
  * Added graceful handling for event logs that cannot be fully projected.
  * CLI inspection, status, replay, serving, and recovery flows now report the last valid state with detailed failure information.
  * Recovery identifies the affected event and requests human intervention with repair guidance.
  * Contracts and dashboards no longer suggest continuing when recovery is unsafe.

* **Bug Fixes**
  * Preserved existing error behavior for authoritative operations and logs without a valid prefix.
  * Healthy and compacted runs continue to behave as expected.

### DIFF
--- a/src/continuum/models.py
+++ b/src/continuum/models.py
@@ -646,14 +646,12 @@ class Run(BaseModel):
 #: (both predate #383; hashing them would brand every existing record as
 #: tampered), and omitted from persisted bodies so readers built before #383,
 #: whose SemanticState forbids extra inputs, can still load newer databases.
-PROJECTION_BOOKKEEPING: frozenset[str] = frozenset(
-    {
-        "status",
-        "unprojectable_at_sequence",
-        "unprojectable_event_type",
-        "unprojectable_reason",
-    }
-)
+PROJECTION_BOOKKEEPING: set[str] = {
+    "status",
+    "unprojectable_at_sequence",
+    "unprojectable_event_type",
+    "unprojectable_reason",
+}
 
 
 class StateCheckpoint(BaseModel):
--- a/src/continuum/models.py
+++ b/src/continuum/models.py
@@ -639,6 +639,23 @@ class Run(BaseModel):
         return self.model_copy(update={"updated_at": utcnow(), **overrides})
 
 
+#: Fields that describe how an event log was *read*, not what a run's state
+#: is (issue #383). They live on SemanticState so a degraded fold can carry
+#: its diagnosis, but they are outside every durable identity of a state:
+#: excluded from the version fingerprint and from checkpoint integrity hashes
+#: (both predate #383; hashing them would brand every existing record as
+#: tampered), and omitted from persisted bodies so readers built before #383,
+#: whose SemanticState forbids extra inputs, can still load newer databases.
+PROJECTION_BOOKKEEPING: frozenset[str] = frozenset(
+    {
+        "status",
+        "unprojectable_at_sequence",
+        "unprojectable_event_type",
+        "unprojectable_reason",
+    }
+)
+
+
 class StateCheckpoint(BaseModel):
     model_config = Frozen
 
@@ -653,8 +670,28 @@ class StateCheckpoint(BaseModel):
     integrity_hash: str | None = None
 
     def content(self) -> dict[str, Any]:
-        """The sealed portion of the checkpoint (everything but the hash)."""
-        return self.model_dump(mode="json", exclude={"integrity_hash"})
+        """The sealed portion of the checkpoint (everything but the hash).
+
+        Projection bookkeeping is excluded beside the hash itself: it was added
+        after these checkpoints existed (#383), it is default in anything that
+        can be persisted (every capture path refuses a degraded fold), and
+        hashing it would report every checkpoint written by earlier builds as
+        tampered, which is exactly the alarm this hash exists to mean.
+        """
+        return self.model_dump(
+            mode="json", exclude={"integrity_hash": True, "state": PROJECTION_BOOKKEEPING}
+        )
+
+    def canonical_json(self) -> str:
+        """The serialised form written to storage.
+
+        Omits projection bookkeeping for the same reasons ``content`` does,
+        plus one more: readers built before #383 validate ``SemanticState``
+        with ``extra="forbid"`` and would refuse a body carrying fields they
+        have never heard of. Omitting them costs nothing, since they are
+        always default in anything persistable.
+        """
+        return self.model_dump_json(exclude={"state": PROJECTION_BOOKKEEPING})
 
     def digest(self) -> str:
         return stable_hash(self.content())
--- a/src/continuum/models.py
+++ b/src/continuum/models.py
@@ -390,6 +390,11 @@ class SemanticState(BaseModel):
     A state is a *projection* of an event prefix. ``source_sequence`` records
     how far into the log the projection consumed, which makes the state
     reproducible: folding the same prefix again must yield an equal state.
+
+    A state whose log stopped folding partway (issue #383) is marked
+    ``status=INVALID`` and names the break in the ``unprojectable_*`` fields.
+    Such a state reports what was known through its last good event; it must
+    never be read as a complete picture of the run.
     """
 
     model_config = Frozen
@@ -408,6 +413,16 @@ class SemanticState(BaseModel):
     version: int = 0
     source_sequence: int = 0
     """Highest event sequence folded into this state (0 = nothing consumed)."""
+    status: StateStatus = StateStatus.VALID
+    """VALID for a complete fold. INVALID marks a degraded projection that
+    stopped at ``unprojectable_at_sequence``."""
+    unprojectable_at_sequence: int | None = None
+    """Sequence of the earliest event the fold refused, or None when the whole
+    log folded."""
+    unprojectable_event_type: str | None = None
+    """Type of the refused event, as ``unprojectable_reason`` alone may not name it."""
+    unprojectable_reason: str | None = None
+    """Single-line statement of the constraint the refused event violated."""
     created_at: datetime = Field(default_factory=utcnow)
     updated_at: datetime = Field(default_factory=utcnow)
 
--- a/src/continuum/state/versioning.py
+++ b/src/continuum/state/versioning.py
@@ -17,10 +17,10 @@ from datetime import datetime
 
 from pydantic import BaseModel, ConfigDict, Field
 
-from continuum.models import SemanticState, utcnow
+from continuum.models import PROJECTION_BOOKKEEPING, SemanticState, utcnow
 from continuum.security.hashing import stable_hash
 
-__all__ = ["VersionEntry", "VersionChain", "state_fingerprint"]
+__all__ = ["VersionEntry", "VersionChain", "canonical_state_json", "state_fingerprint"]
 
 
 def state_fingerprint(state: SemanticState) -> str:
@@ -30,9 +30,9 @@ def state_fingerprint(state: SemanticState) -> str:
     pending work, approvals and dependencies are the same state even if they
     were projected at different times or carry different version numbers.
 
-    The projection-degradation fields are excluded for the same reason: they
-    describe how the log was read (where folding stopped), not what the state
-    says. Including them would break fingerprint dedup across the upgrade for
+    Projection bookkeeping is excluded for the same reason: it describes how
+    the log was read (where folding stopped), not what the state says.
+    Including it would break fingerprint dedup across the #383 upgrade for
     every stored version, and a degraded prefix-state genuinely is the same
     task state as its healthy counterpart.
     """
@@ -43,13 +43,23 @@ def state_fingerprint(state: SemanticState) -> str:
             "created_at",
             "updated_at",
             "source_sequence",
-            "status",
-            "unprojectable_at_sequence",
-            "unprojectable_event_type",
-            "unprojectable_reason",
+            *PROJECTION_BOOKKEEPING,
         },
     )
     return stable_hash(payload)
+
+
+def canonical_state_json(state: SemanticState) -> str:
+    """The serialised form of a state as written to storage.
+
+    Omits projection bookkeeping: it is always default in anything persistable
+    (a degraded fold is refused at every capture path), and carrying it would
+    make databases written after #383 unreadable by readers built before it,
+    whose SemanticState validates with ``extra="forbid"``. The version
+    fingerprint above is computed over the same exclusions, so stripping here
+    cannot desynchronise a stored row from its recorded fingerprint.
+    """
+    return state.model_dump_json(exclude=PROJECTION_BOOKKEEPING)
 
 
 class VersionEntry(BaseModel):
--- a/src/continuum/state/versioning.py
+++ b/src/continuum/state/versioning.py
@@ -29,10 +29,25 @@ def state_fingerprint(state: SemanticState) -> str:
     Two states with the same goal, progress, decisions, findings, evidence,
     pending work, approvals and dependencies are the same state even if they
     were projected at different times or carry different version numbers.
+
+    The projection-degradation fields are excluded for the same reason: they
+    describe how the log was read (where folding stopped), not what the state
+    says. Including them would break fingerprint dedup across the upgrade for
+    every stored version, and a degraded prefix-state genuinely is the same
+    task state as its healthy counterpart.
     """
     payload = state.model_dump(
         mode="json",
-        exclude={"version", "created_at", "updated_at", "source_sequence"},
+        exclude={
+            "version",
+            "created_at",
+            "updated_at",
+            "source_sequence",
+            "status",
+            "unprojectable_at_sequence",
+            "unprojectable_event_type",
+            "unprojectable_reason",
+        },
     )
     return stable_hash(payload)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -714,6 +714,32 @@ All notable changes to this project are documented here. The format follows
 
 ### Fixed
 
+- **One unprojectable event bricked every projecting command, with no route back (#383).**
+  A log whose fold fails is intact as a chain but dead as a run: because the fold
+  validates each intermediate state, no later event could correct an earlier bad
+  one, so `status`, `resume`, `inspect`, `replay`, `show-contract`, `validate`,
+  `briefing` and `compact` all raised on precisely the runs that needed them,
+  while the action tools (which fold only ACTION_* events) kept authorising real
+  side effects that recovery could not assess. The fold can now degrade instead
+  of raising: `project` and `project_incremental` accept
+  `on_unprojectable="raise"|"degrade"`, defaulting to `"raise"` so every existing
+  caller sees byte-for-byte today's behaviour. Degrade mode stops at the earliest
+  refused event and returns the last-good prefix marked
+  `SemanticState.status = INVALID` with `unprojectable_at_sequence`,
+  `unprojectable_event_type` and a condensed `unprojectable_reason`; it never
+  skips past the break, and if nothing folds before the break it still raises,
+  since a partial answer invented from nothing would be worse than an error. The
+  recovery engine folds with degrade enabled, so a poisoned log yields a
+  `request_human` verdict naming where folding stopped instead of a pydantic
+  traceback, and CLI `status`, `inspect` and `replay` report the same break and
+  exit non-zero. The diagnostic call sites opted in are the engine's restore
+  path, the CLI status/inspect/replay surfaces, the serve sidecar's progress
+  report and dependency dedup, and the benchmark's strategy readout; the MCP
+  write-path guard `_project_candidate` and the checkpoint capture surfaces
+  deliberately keep raising, because accepting or pinning a partial fold would
+  launder it into authoritative state. Repair/amend and fork-from-last-good-prefix
+  remain future work and are not attempted here.
+
 - **`continuum verify` certified a run whose log could not be projected (#382).**
   `verify` re-audits the hash chain, which is a statement about integrity, and an
   unprojectable log is perfectly intact: the offending event was written through
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -732,7 +732,12 @@ All notable changes to this project are documented here. The format follows
   recovery engine folds with degrade enabled, so a poisoned log yields a
   `request_human` verdict naming where folding stopped instead of a pydantic
   traceback, and CLI `status`, `inspect` and `replay` report the same break and
-  exit non-zero. The diagnostic call sites opted in are the engine's restore
+  exit non-zero. The break also reaches the machine-readable contract instead of
+  living only in prose: a new `repair_log` repair step makes `required_actions`
+  name real work, `next_allowed_action` points at it rather than falling through
+  to a rendered "continue" over a `requires_human` verdict, `verified` entries
+  are qualified with the last-good sequence, and `invalidated` records the
+  projection itself. The diagnostic call sites opted in are the engine's restore
   path, the CLI status/inspect/replay surfaces, the serve sidecar's progress
   report and dependency dedup, and the benchmark's strategy readout; the MCP
   write-path guard `_project_candidate` and the checkpoint capture surfaces
--- a/STATUS.md
+++ b/STATUS.md
@@ -17,6 +17,22 @@ believed, and what is neither.
 
 ---
 
+## In flight: unprojectable logs degrade instead of dying (#383, PR #385)
+
+As of 2026-08-25, PR #383's fix is open on `fix/degrade-unprojectable-fold`
+(three commits), reviewed through two rounds. The fold accepts
+`on_unprojectable="raise"|"degrade"` (default byte-for-byte unchanged);
+degrade returns the last-good prefix marked `SemanticState.status=INVALID`
+naming where folding stopped; recovery decides REQUEST_HUMAN and the contract
+carries a `repair_log` step. Checkpoint digests and persisted bodies exclude
+the projection-bookkeeping fields, so databases written before or after the
+change load either way (cross-version tests pin serialised fixtures in
+`tests/test_checkpoint_compat.py`). Verified on the branch: 1425 passed,
+23 skipped, ruff clean, strict mypy clean on 104 files. Repair/amend (option
+2) and fork-from-last-good-prefix (option 3) remain unbuilt by design.
+
+---
+
 ## Full-gate audit (2026-08-24)
 
 Ran against `main` at `8013f6a` in a clean worktree, Python 3.13
--- a/WORKLOG.md
+++ b/WORKLOG.md
@@ -531,3 +531,48 @@ features, #266-#285 contributor good-first-issues, #288 provenance graph
 files. Five integration seams, six schema migrations, eleven MCP tools,
 twenty-five CLI commands. Every feature verified against real Claude Code
 sessions with hard kills and live protocol boundaries.
+
+## Session 18, 2026-08-25 (issue #383: degrade instead of die, PR #385)
+
+Fixed #383 in a dedicated worktree (`/tmp/wt-383`, branch
+`fix/degrade-unprojectable-fold`, self-assigned before starting). Three
+commits, each answering one review round.
+
+**The fix (3f2b769).** `project`/`project_incremental` gained
+`on_unprojectable="raise"|"degrade"` (default unchanged). Degrade stops at the
+earliest refused event and returns the last-good prefix marked
+`SemanticState.status=INVALID` with `unprojectable_at_sequence/_event_type/
+_reason`; it never skips past the break, and raises if nothing folds before
+it. Opt-ins limited to diagnostic surfaces (engine restore path, CLI
+status/inspect/replay, serve progress report and dep dedup, benchmark
+readout); the #364 write guard `_project_candidate` and every checkpoint
+capture path deliberately stay on raise. Recovery decides REQUEST_HUMAN.
+Reproduction used the `_poison` helper from tests/test_cli.py; all four dead
+commands confirmed failing before any change.
+
+**Review round 1: the contract must carry the break (0768e38).** Resume's
+prose named the break but `required_actions` was empty and `next_allowed`
+rendered as "continue" over a requires_human verdict. New `RepairKind.
+REPAIR_LOG` step (sorted first) gives the contract real work, qualified
+`verified` entries (`goal (through sequence N)`), a `projection (invalid)`
+entry in `invalidated`, and honest fallbacks at all three `or 'continue'`
+sites (contract render, engine render, dashboard).
+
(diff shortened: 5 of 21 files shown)


Cyrax321:
Reviewed by running it, not by reading the report. The approach is right and the constraint work is genuinely careful. One blocking finding, in the contract rather than the projector.

## Verified independently

I reproduced every claim rather than trusting the summary.

Constraint 1, the one I flagged as the silent-failure risk, holds. `project_current` is untouched, so the `checkpoint()` to `put_version` write path still raises. `capture_state` in `mcp/server.py` and the `_h_checkpoint` handler in `serve/server.py` are both untouched. No path feeds a degraded state into persistence, which is what makes the `state_fingerprint` exclusion safe rather than dangerous: `status` is a new field, so excluding it preserves dedup across the upgrade, and a degraded state can never be stored to collide with a healthy one anyway. Good call, and the reasoning in the docstring is the reasoning I would have wanted.

Constraint 2 holds structurally, not just by convention: `SEVERITY[REQUEST_HUMAN]` is 4 against `RESUME` 0 and `_decide` takes the max, so a degraded fold cannot yield RESUME even with every other signal clean. Proposing it unconditionally rather than behind a guard is the right shape.

Constraint 3 holds: the `except` returns immediately, and `acc.source_sequence` is still the last good event because the update happens after the try block.

Red before green, checked properly. Stashing does nothing here since the changes are committed, so I ran `git checkout origin/main -- src/` instead: 13 of 14 tests fail, 11 with `TypeError: project() got an unexpected keyword argument`. After restoring, all 14 pass. The 14th, the compacted-run guard, passes on main as claimed and is correctly described as a regression guard rather than a new behaviour.

Gates reproduce exactly: 1416 passed, 23 skipped, ruff clean, 205 files formatted, mypy clean on 104 files.

Two judgement calls I want to endorse explicitly, because both could have gone the lazy way. Keeping the run-mismatch and sequence-order checks outside the degrade treatment is correct: those mean the caller handed the fold a malformed stream, not that the log holds a bad event. And raising when nothing folds before the break is correct, because there is no partial answer to invent.

## Blocking: the contract still reads clean

`resume` now answers instead of dying, which is the point of the issue and it works. But the contract, which is the machine-readable and attestable artifact, does not carry the break.

On a run poisoned at sequence 3:

```
$ continuum show-contract r
recovery_status:   requires_human
verified:          goal, progress
next_allowed:      continue
reason:            the event log stops folding at sequence 3 ...
```

Three problems, in order of severity.

`verified: goal, progress` affirmatively declares components verified on a run whose log stops folding. They were verified against the last-good prefix only. The rationale says so in prose, but the contract's own structured field does not, so the artifact asserts an assurance it cannot support. This is the same shape as #369, which we just fixed: prose saying one thing while the structured verdict says another, with the reader most likely to act on the structured one.

`next_allowed: continue` is actively wrong on a run that cannot be read. This comes from the pre-existing `self.next_allowed_action or 'continue'` in `engine.py:164` and `contract.py:161`, so it is not introduced here, but this PR makes it newly reachable: it creates the first `requires_human` state with no plan step, and I confirmed the same command on `main` cannot produce it because it raises first. A latent bug this PR surfaces is still this PR's to handle.

`required_actions: []` and `repairs: []`, so a machine reading the contract for work to do sees nothing, and `human_steps` degrades to the generic "nothing further is automatable here".

The JSON is honest, for what it is worth: `next_allowed_action` is `null`, `safe` is `false`, `mode` is `request_human`. So a caller keying on `mode` or `safe` is safe today. The exposure is a human reading the rendered contract, and any caller keying on `required_actions` or `verified`.

## What I would change

Smallest fix that closes it: add a `RepairKind` for this, so the break becomes a plan step. That gives `required_actions` an entry, gives `next_allowed_action` a real value instead of falling through to "continue", and lets `human_steps_for` render something specific. It also stops `verified` reading as unqualified, since the contract would carry a competing invalidated entry.

If you would rather not touch `RepairKind` in this PR, then at minimum suppress or qualify the `verified` list when `state.status is INVALID`, and change the `or 'continue'` fallback so it cannot claim permission on a non-resume verdict. Do not leave the fallback as-is on the grounds that it is pre-existing.

Not blocking, but worth a line in the docstring: `report.consumed` counts the refused event, since `consumed += 1` runs before the try. Defensible either way, but a reader will wonder.

The `compact --force` exclusion and the honest note about `next_allowed_action` naming an unrelated repair step were both good to disclose. This finding is the sharper version of that second note.

Happy to merge once the contract stops reading clean on a broken log.

Cyrax321:
Fixed on the branch (0768e38), taking your preferred route: the break is now a plan step, so all three problems close through one mechanism.

**What changed**

New RepairKind.REPAIR_LOG ("repair_log"). plan_repairs takes unprojectable=(sequence, event_type, reason) and emits a requires_human step targeting sequence_N; it sorts first in _ORDER, because nothing else about the run can be assessed while the fold cannot run, including whether any other repair did what it claimed. The engine threads restored.state's degradation into the planner, so required_actions carries repair_log:sequence_26, next_allowed_action points at it instead of falling through, and human_steps_for renders specific guidance ("the log stops folding at sequence_26: run continuum verify ... no amend surface exists yet") rather than "nothing further is automatable here".

For the verified list, build_contract now reads the degradation from validation.state directly: every verified entry is qualified as "goal (through sequence 24)", invalidated gains "projection (invalid: log stops folding at sequence 26)", and the break joins the evidence list. Healthy contracts are byte-identical to before; that is pinned by assertions in test_a_healthy_log_still_resumes_with_degrade_wired_in (exact verified list, empty invalidated, rendered "continue" under SAFE_TO_RESUME).

On the fallback: I found a third site beyond the two you cited. render_contract (contract.py), RecoveryDecision.render (engine.py) and render_dashboard (observability.py) all had their own or-'continue'. All three now say "continue" only when resuming is actually permitted (SAFE_TO_RESUME / mode is RESUME) and name the unsettled repairs otherwise. Your "newly reachable" framing is exactly right and I treated it that way rather than as pre-existing.

Also took the non-blocking note: project_incremental's docstring now states that report.consumed counts the refused event because consumed += 1 runs before the fold is attempted.

**Verified, not asserted**

show-contract on a run poisoned at sequence 2 now renders:

recovery_status:   requires_human
verified:          goal (through sequence 1), progress (through sequence 1)
invalidated:       projection (invalid: log stops folding at sequence 2)
required_actions:
  - repair_log:sequence_2
next_allowed:      repair_log:sequence_2

exit 20. No "continue" anywhere in the artifact.

Red before green, same protocol as before: stash of src/ only, five new tests fail (contract has no repair_log step, next_allowed renders continue, verified unqualified, generic human step), restore, all pass. New tests cover the step itself (target, position first, requires_human, contract wiring), qualified verified plus the projection entry plus evidence line, both rendered forms refusing "continue", guidance specificity, and the healthy-shape guard. Full gates on 0768e38: 1421 passed / 23 skipped, ruff check clean, ruff format clean (205 files), mypy zero issues on 104 files.

Cyrax321:
The three contract problems are genuinely closed, verified by running it. But the review turned up something bigger that neither round caught, and it blocks: **every checkpoint written by shipped code now reports as tampered.**

## The blocking finding

Adding four fields to `SemanticState` changed what `StateCheckpoint.content()` serialises, and that is the input to `integrity_hash`. So every pre-existing checkpoint fails verification under this branch.

Against the repo's own database, on a checkpoint written by shipped code:

```
$ continuum history migrate_v2
integrity error: checkpoint 'checkpoint_e2568640ff5c024b6ad7634e9b8025a7': integrity hash does not match content
```

Reproduced from scratch in both directions:

- write a checkpoint with `origin/main`, read it on this branch: `integrity hash does not match content`
- write a checkpoint on this branch, read it with `origin/main`: `4 validation errors for StateCheckpoint / state.status / Extra inputs are not permitted` (`SemanticState` is `extra="forbid"`)

So `resume`, `validate`, `status` and `history` break on every run that has a checkpoint, in any database created before the upgrade. And the failure mode is the worst available one: the message says the content does not match its hash, which is indistinguishable from tampering. This project raises that alarm precisely so it means something, and an upgrade that fires it on every healthy run spends that credibility.

`Breaking changes: No` in the PR body needs to change regardless of how it is fixed.

## Why the tests could not catch it

Every test writes and reads its checkpoints under one version of the code, so the hash always agrees with itself. The bug only exists across a version boundary. Worth a test that pins a serialised checkpoint fixture and asserts it still verifies, since that is the only shape that catches this class.

## The fix, verified

`state_fingerprint` already excludes the four fields, and the reasoning given for that is exactly right: they describe how the log was read, not what the state says. `integrity_hash` is a second hash over the same state and did not get the same treatment. Applying it there restores compatibility. I checked against the real stored checkpoint rather than guessing:

```
current digest matches stored hash?          False
digest with the new fields excluded matches? True
```

So excluding them from `StateCheckpoint.content()` is sufficient and provably restores every existing checkpoint.

Worth considering the stronger version, though: if those fields are excluded from both hashes because they are not part of the state, that is an argument they do not belong on the persisted state model at all. A degraded state is never persisted (`project_current` and every `capture_state` path still raise, which I verified and which is good), so nothing needs them to survive a write. Carrying them on the projection result rather than inside `SemanticState` would make the exclusions unnecessary instead of load-bearing, and would remove the forward-compatibility break too, since old readers would see no new fields. Your call on scope; the exclusion is the smaller change and I would accept it with the compatibility test alongside.

## What I confirmed is now correct

The poisoned contract carries the break properly, and there is no `continue` anywhere in the artifact:

```
recovery_status:   requires_human
verified:          goal (through sequence 2), progress (through sequence 2)
invalidated:       projection (invalid: log stops folding at sequence 3)
required_actions:
  - repair_log:sequence_3
next_allowed:      repair_log:sequence_3
evidence:
  - projection stopped at sequence 3 (TASK_UPDATED): Value error, completed + pending + failed exceeds total
```

Exit 20. `REPAIR_LOG` sorting first in `_ORDER` is the right call and the reason given for it is the right reason: nothing else about the run can be assessed while the fold cannot run, including whether another repair did what it claimed.

Finding the third `or 'continue'` site in `observability.py` beyond the two I cited, and gating all three on the verdict rather than on presence, is better than what I asked for. The healthy contract is unchanged: `verified: external_dependency:dataset, goal, progress` unqualified, `next_allowed: continue` under `safe_to_resume`.

Gates reproduce: 1421 passed, 23 skipped, ruff clean, 205 files formatted, mypy clean on 104 files.

Fix the checkpoint digest and I will merge this.

Cyrax321 (commented):
**BLOCKING. Do not merge until the checkpoint digest is fixed.**

GitHub will not let me formally request changes on a PR authored by my own account, so this is a COMMENT review carrying the same weight. Treat it as blocking.

Requesting changes on one blocking problem: adding fields to `SemanticState` changed what `StateCheckpoint.content()` serialises, so every checkpoint written by shipped code now fails its integrity hash. Verified against this repo's own database, on a checkpoint written by released code: `continuum history migrate_v2` reports `integrity hash does not match content`. `resume`, `validate`, `status` and `history` therefore break on every run with a checkpoint in any pre-upgrade database, and the message a user sees is indistinguishable from tampering. This project raises that alarm so that it means something, so firing it on every healthy run after an upgrade is the expensive kind of wrong.

Details and the verified fix are inline. `Breaking changes: No` in the PR body needs correcting either way.

Also blocking, and cheap: a test that pins a serialised checkpoint fixture and asserts it still verifies. Every current test writes and reads under one code version, so the hash always agrees with itself and this class of bug is invisible. A fixture is the only shape that catches it.

The three contract problems from my last review are genuinely closed. Finding the third `or 'continue'` site in `observability.py` and gating all three on the verdict rather than on presence is better than what I asked for. Gates reproduce here: 1421 passed, 23 skipped, ruff and mypy clean.

src/continuum/models.py
```diff
@@ -408,6 +413,16 @@ class SemanticState(BaseModel):
     version: int = 0
     source_sequence: int = 0
     """Highest event sequence folded into this state (0 = nothing consumed)."""
+    status: StateStatus = StateStatus.VALID
```
Cyrax321, line 416: **Blocking.** These four fields are the cause of the checkpoint breakage. `StateCheckpoint.content()` is `self.model_dump(mode="json", exclude={"integrity_hash"})`, so it picks up every field on the nested state, and `integrity_hash` is computed over that. Adding fields changes the digest for state that was sealed before they existed.

Reproduced from scratch in both directions. Write a checkpoint with `origin/main` and read it here: `integrity hash does not match content`. Write one here and read it with `origin/main`: `4 validation errors for StateCheckpoint / state.status / Extra inputs are not permitted`.

The fix is the same treatment `state_fingerprint` already got: exclude these four from `StateCheckpoint.content()`. I checked that against the real stored checkpoint rather than assuming it, by recomputing the digest with them stripped: `digest with the new fields excluded matches? True`. So it is sufficient, not just plausible.

src/continuum/models.py
```diff
@@ -408,6 +413,16 @@ class SemanticState(BaseModel):
...
+    """VALID for a complete fold. INVALID marks a degraded projection that
+    stopped at ``unprojectable_at_sequence``."""
+    unprojectable_at_sequence: int | None = None
+    """Sequence of the earliest event the fold refused, or None when the whole
+    log folded."""
+    unprojectable_event_type: str | None = None
+    """Type of the refused event, as ``unprojectable_reason`` alone may not name it."""
+    unprojectable_reason: str | None = None
```
Cyrax321, line 424: **Blocking, same root cause, separate consequence.** `SemanticState` is `extra="forbid"` via `Frozen`, so a checkpoint written after the upgrade cannot be read by a pre-upgrade binary at all: it fails validation on these field names rather than merely mismatching a hash.

That makes a rollback unsafe once any checkpoint has been written, and breaks a mixed-version fleet sharing one database. Worth deciding deliberately rather than inheriting: excluding the fields from the digest fixes the hash direction but not this one, whereas keeping them off the persisted model fixes both.

src/continuum/state/versioning.py
```diff
@@ -29,10 +29,25 @@ def state_fingerprint(state: SemanticState) -> str:
...
     """
     payload = state.model_dump(
         mode="json",
-        exclude={"version", "created_at", "updated_at", "source_sequence"},
+        exclude={
+            "version",
+            "created_at",
+            "updated_at",
+            "source_sequence",
+            "status",
```
Cyrax321: **Non-blocking, design.** The reasoning given here is right, and it is also the argument for going further: if these fields are excluded from the fingerprint because they describe how the log was read rather than what the state says, that is a reason they do not belong on the persisted state model.

A degraded state is never persisted. `project_current` and every `capture_state` path still raise, which I verified and which is the correct split. So nothing needs these to survive a write, and carrying them on the projection result instead would make both exclusions unnecessary rather than load-bearing. It would also remove the forward-compatibility break, since an old reader would see no new fields at all.

Your call on scope. The exclusion is the smaller change and I will take it with the fixture test alongside, but a future field added to `SemanticState` will hit exactly this again, and the next person will have to rediscover why two separate hashes both need an exclude list.

Cyrax321:
Fixed on the branch (39b0756, ddbb6bd), taking the exclusion you verified rather than the model reshape.

**What changed**

PROJECTION_BOOKKEEPING now names the four fields once in models.py, and that one constant drives every durable surface:

StateCheckpoint.content() excludes them beside integrity_hash itself. Your digest check was the design: their hashes covered exactly the reduced payload, so every existing checkpoint verifies again. StateCheckpoint.canonical_json() and canonical_state_json() replace model_dump_json() at all four storage write sites (sqlite and postgres, checkpoints and versions). This closes the forward direction too, which your report implied and I confirmed separately: versions rows had the same extra="forbid" exposure as checkpoint bodies, so fixing only content() would have left old readers unable to load new databases even with the tampering alarm gone. state_fingerprint now draws on the same constant instead of restating its own list.

**The compatibility test, and one correction to my own first attempt**

You were right that only a pinned serialised fixture catches this class; I would add that writing that fixture honestly is subtle. My first version wrote a checkpoint under the new code, stripped the fields from the stored body, and asserted verify(): it passed against the broken code too, because the reload round-trips through pydantic, which re-injects the four fields as defaults, so the recomputed payload matched the write-time payload either way. The faithful simulation needs both halves of what an old writer did: strip the fields AND re-seal the hash over that reduced payload, hash kept inside the body as sealed() does. With that, the test fails against this branch's field-adding commit with exactly your message ("integrity hash does not match content") and passes here. Three tests fail red against 3f2b769 (that one plus the two persist-shape assertions); the old-version-row load test is a guard pinning behaviour that already holds.

On the stronger version (fields off SemanticState entirely): agreed it is cleaner in principle, but it moves the diagnosis through every RecoveryDecision, CLI command and MCP response that reads decision.state today, and this PR has already grown two review rounds. Taking the exclusion now; the reshape is worth an issue of its own if you want it pursued.

**Gates on ddbb6bd**

1425 passed / 23 skipped, ruff check clean, ruff format clean (206 files), mypy zero issues on 104 files. PR body's Breaking changes section rewritten to state the break this round found and how it was closed, rather than the "No" that round one's testing could not yet see.

Cyrax321:
Reviewer guide for the current head (0d2638b). Five commits, one per concern, each independently readable.

**Commit 1, 3f2b769: the #383 fix.** `src/continuum/state/semantic.py` holds the whole mechanism: `project_incremental` takes `on_unprojectable="raise"|"degrade"` (default unchanged), degrade returns the last-good prefix via `_Accumulator.build_degraded`, marked `INVALID` on `SemanticState` (`src/continuum/models.py`) with `unprojectable_at_sequence/_event_type/_reason`. Opt-in call sites: `CheckpointManager.restore` (both fold paths), CLI `status`/`inspect`/`replay`/`_verify_against_stored`, serve sidecar record-progress report and dependency dedup, benchmark readout. Deliberately still raising: `_project_candidate` (the #364 write guard), checkpoint capture paths, and the two sites with existing documented `ProjectionError` fallbacks. Tests: `tests/test_projection.py` (six), engine verdict in `tests/test_recovery_engine.py`, end-to-end in `tests/test_cli.py`. All proven red against origin/main by stashing `src/`.

**Commit 2, 0768e38: contract carries the break (review round 1).** `RepairKind.REPAIR_LOG` in `src/continuum/recovery/planner.py` (sorted first in `_ORDER`), threaded from the degraded state through `plan_repairs(unprojectable=...)` in `engine.py`. `build_contract` reads the degradation off `validation.state`: qualified `verified` entries, `projection (invalid)` in `invalidated`, break added to evidence; healthy contracts are byte-identical (pinned assertions). All three `or 'continue'` fallbacks now gate on the verdict: `render_contract`, `RecoveryDecision.render`, `render_dashboard`. Five new tests, red against this branch's first commit.

**Commits 3+4, 39b0756 + ddbb6bd: cross-version compatibility (review round 2, the blocking one).** `PROJECTION_BOOKKEEPING` in models.py drives three exclusions: `content()` (so pre-existing checkpoints verify again), persisted bodies via `canonical_json()`/`canonical_state_json()` at all four write sites (sqlite/postgres, checkpoints/versions, so old readers without these fields keep loading new databases), and `state_fingerprint` (previously a separate list). `tests/test_checkpoint_compat.py` pins serialised fixtures because no same-process round-trip can catch this class; note the fixture must re-seal the hash over the reduced payload exactly as the old writer did, otherwise pydantic re-injecting defaults on reload masks the bug (first attempt did pass against broken code; the test docstring records this). Three tests fail red against 3f2b769 with the reviewer's exact message ("integrity hash does not match content"), all pass here.

**Commit 5, 0d2638b: docs only.** WORKLOG session 18 and an in-flight STATUS section; no source changes.

Gates on head: 1425 passed / 23 skipped, `ruff check src tests` clean, `ruff format --check src tests` clean (206 files), `mypy src/continuum` zero issues on 104 files. All CI checks green on this push. The Breaking changes section of the PR body has been rewritten to describe the round-2 break and its closure rather than round one's premature "No". Open items, deliberately unbuilt here: repair/amend command (option 2), fork-from-last-good-prefix (option 3), and optionally relocating the bookkeeping fields off SemanticState (raised in the round-2 reply as candidate follow-up scope).

Cyrax321:
Verified in both directions against real data. The blocking issue is closed and I have merged current main into the branch to clear the conflict.

## The blocking finding is fixed, confirmed independently

Backward, on this repo's actual database with checkpoints written by shipped code:

```
$ continuum history migrate_v2
CHECKPOINT   VERSION   TRIGGER    PROGRESS
checkpoint   v0        manual     0 completed
checkpoint   v1        manual     2 completed

$ continuum resume migrate_v2
Run: migrate_v2
Checkpoint: v1
```

Forward, which my report only implied and you closed properly: wrote a database with this branch, read it with `origin/main`, and both `history` and `resume` work. Catching that the `versions` rows had the same `extra="forbid"` exposure as checkpoint bodies is the part I had not checked, and fixing only `content()` would have left that half open.

The compatibility test has teeth. Against 3f2b769 three tests fail, and the first reproduces the exact message from my review:

```
CorruptedRecord: checkpoint 'checkpoint_2361e527...': integrity hash does not match content
```

Your note about the first attempt is the most useful thing in this thread. A fixture that strips the fields but does not re-seal the hash passes against broken code, because pydantic re-injects the defaults on reload and the recomputed payload matches the write-time payload either way. That is a trap I would have fallen into, and recording it in the test docstring is worth more than the test.

`PROJECTION_BOOKKEEPING` as a single constant driving `content()`, `canonical_json()`, `canonical_state_json()` and `state_fingerprint` is better than four exclude lists. I checked both backends: sqlite and postgres each use the canonical helpers at their checkpoint and version write sites, with no raw `model_dump_json` left on those paths.

## Nothing else regressed

Healthy contracts are byte-identical to main, diffed directly rather than taken on trust. The #383 behaviour survived two rounds of compatibility churn: a poisoned run still renders `verified: goal (through sequence 2)`, `invalidated: projection (invalid: ...)`, `required_actions: repair_log:sequence_3`, exit 20, and zero occurrences of `continue` in the artifact.

Gates on the merged head: 1453 passed, 23 skipped, ruff check and format clean on 206 files, mypy clean on 104 files.

## Agreed on scope

Taking the exclusion over the model reshape is the right call for this PR. You are right that moving the fields off `SemanticState` touches every `RecoveryDecision`, CLI command and MCP response reading `decision.state`, and this has already been through two rounds. The exclusions are load-bearing now, so the next field added to `SemanticState` will hit this again; I will raise that as its own issue rather than hold this.

Merging.

Issue #421, opened 2026-08-26
[#391] End-to-end dropped-constraint scenario and chaos-corpus registration


Cyrax321:
### Affected component

Benchmarks

Parent epic: #391. Do last; closes the epic with public-boundary proof.

### What problem does this solve?

Nothing demonstrates a real dropped constraint being caught through resume rather than through unit fixtures.

### Proposed solution

End-to-end falsifiable scenario:

- record two pins, compact the run, build a reconstruction whose summary omits one constraint;
- assert the next resume names the missing pin by hash prefix and (strict) escalates;
- register the dropped-constraint class into the #397 corpus hook.

Acceptance criteria: scenario green post-epic; red (or honestly documented gap) pre-epic; corpus registration merged with the scenario.

### Alternatives considered

Unit-only proof. Rejected: the failure this epic targets is a boundary failure.

### Is this a breaking change?

No

### Would you like to work on this?

No

Issue #420, opened 2026-08-26
[#391] Concepts and how-to docs for constraint pinning with honest scope list


Cyrax321:
### Affected component

Documentation

Parent epic: #391. Do after the mechanism works (documents the contract).

### What problem does this solve?

A verification feature callers don't know to opt into protects nobody.

### Proposed solution

Docs pass:

- concepts page: pin lifecycle, statuses, grace windows, strict-mode behaviour;
- how-to: recording pins from harness hooks (pairs with #396 recipes later, cross-link only);
- threat-model note: what pinning catches (silent drops) and what it does not (adversarial contexts that forge presence markers are detectable but out of scope for v1).

Acceptance criteria: every status and knob documented with an example; honest does-not-solve list included per house style.

### Alternatives considered

Docstring-only documentation. Rejected: cross-cutting feature needs the concepts page.

### Is this a breaking change?

No

### Would you like to work on this?

No

Issue #418, opened 2026-08-26
[#391] Reconstruction accounting: present/absent/unverifiable per pin with grace escalation


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

Parent epic: #391. The heart of the epic; do after projection lands.

### What problem does this solve?

Reconstruction paths produce context nobody audits against active pins, which is exactly how constraints vanish silently.

### Proposed solution

Every reconstruction surface (briefing, resume banner/context, checkpoint rehydration) reports per-pin status:

- present / absent / unverifiable, computed from what the produced context actually contains (hash-tagged markers emitted by the reconstruction call);
- grace-window logic: absent past grace raises a contract flag naming the pin id/hash prefix;
- advisory flag first; strict-mode escalation to REQUIRES_REVIEW behind the existing strict plumbing.

Acceptance criteria: all three statuses reachable in tests; grace window configurable; strict escalation test; non-strict default leaves mode untouched.

### Alternatives considered

Trusting summarizers' self-report. Rejected: summaries are optimized outputs, not evidence.

### Is this a breaking change?

No

### Would you like to work on this?

No

Cyrax321:
Claiming #418

Issue #417, opened 2026-08-26
[#391] Projector folds pins into SemanticState with anchoring-safe lifecycle


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

Parent epic: #391. Do after the event type lands.

### What problem does this solve?

Projection ignores pins, so nothing downstream can even ask "which constraints are active?"

### Proposed solution

Projector folds pins into `SemanticState.pins {constraint_id: sha256, status}`:

- active / retracted lifecycle;
- prefix-closed and replayable per house projection rules;
- compaction interplay verified here: pins live in the chain and survive anchoring like any event.

Acceptance criteria: fold over identical prefixes identical; anchored logs project identically; retraction removes from active set.

### Alternatives considered

Separate side-table outside the log. Rejected: breaks single-source-of-truth and tamper-evidence.

### Is this a breaking change?

No

### Would you like to work on this?

No

Cyrax321:
Claiming #417

PR #425, opened 2026-08-26
[#391] CONSTRAINT_PINNED event type with hash-only guarantees

### LINKED ISSUE #416: [#391] CONSTRAINT_PINNED event type with hash-only storage guarantees
### Affected component

Parent epic: #391 (do this first).

Constraints have no event representation; everything else in the epic depends on pins existing.

### Proposed solution

- New event `CONSTRAINT_PINNED {constraint_id, sha256}` plus matching retracted variant;
- models, validation (hash format, id rules), sequencing rules;
- hash-only storage enforced by test (no plaintext ever serialized).

Acceptance criteria: round-trip write/read; invalid hashes rejected at the boundary; property test asserting plaintext never appears in payloads or logs.

### Alternatives considered

Storing constraint text. Rejected on privacy grounds and because hashes suffice for verification.

No

No

Cyrax321:
## Description

Constraints become first-class events that carry SHA-256 digests and never the constraint text (epic #391, first sub-issue of the track).

Two payload models land in `models.py`:

- `ConstraintPinned(constraint_id, sha256)`: pins a standing constraint by the digest of its exact text. The plaintext is forgettable at pin time by everyone except the pinner; later verification compares digests.
- `ConstraintRetracted(constraint_id)`: records withdrawal of an id alone. Storable even for ids this log never pinned (for instance across an anchored prefix); how an unmatched retraction projects is decided downstream (#417), not here.

Validation rules, both documented on the models and enforced at construction:

- `constraint_id`: 1 to 128 chars from ASCII letters, digits and `. _ : -`. The tight charset is deliberate: hashing exists so the text is never stored, which leaves the id as the only free-text-shaped field on the event, so it is restricted to machine labels.
- `sha256`: exactly 64 lowercase hex characters. Refused at the boundary rather than normalised, so every stored pin compares byte for byte against a recomputed digest. Both fields are strict-typed strings; pydantic would otherwise silently decode bytes past validation.

Both types are registered in `EventType`. No projector changes: unknown-to-fold types already fold as recorded facts, and folding pins into `SemanticState` is #417's scope.

Privacy enforcement lives in a property-style test with seeded plaintexts known only to the test: it asserts they appear nowhere in serialized payloads, whole-log JSON, per-event JSON, event reprs, payload/content reprs or the verify report. Failure messages name pins by index only, so a leak can never be echoed into test output.

## Related issues

Refs #391. Fixes #416: every item in its proposed solution and all three acceptance criteria are delivered here (both event types, models and boundary validation, hash-only storage enforced by test, round-trip write/read). Projector folding (#417), surfaces (#418/#419), docs (#420) and the chaos scenario (#421) are separate sub-issues of the epic, out of this scope.

## Types of changes

- Type: `feature`

## Breaking changes

No

## How has this been tested?

Environment: local checkout, Python 3.11 venv, ruff 0.16.3, mypy strict with pydantic plugin.

```
pytest                      -> 1482 passed, 23 skipped in 30.15s (exit 0)
ruff check src/ tests/ examples/   -> All checks passed!
ruff format --check src/ tests/ examples/ -> 220 files already formatted
mypy src/continuum          -> Success: no issues found in 105 source files
```

New coverage in `tests/test_constraint_pins.py`: record -> read-back -> verify round trip for both types, chain acceptance of pin-only runs, boundary rejections (uppercase/short/long/non-hex/prefixed digests; empty, whitespace, non-ASCII, over-length and prose ids; bytes refused rather than decoded), retraction of an unknown id storable, and the plaintext-nowhere property test. The new tests fail on pre-change code (collection ImportError without the new models). Existing `tests/test_events.py` untouched.

## Checklist

Tests added or updated for the changed behaviour: yes. Documentation updated where the change affects user-facing behaviour: docs are #420's scope. CI passes on the latest commit: gate run locally as above. Security-relevant changes state known limitations explicitly in this description: hash-only storage means CONTINUUM cannot itself confirm what text a digest commits to, and the id channel, while charset-bounded, remains caller-supplied input like any label.

## Additional context

Alternatives considered: storing constraint text (rejected on privacy grounds, hashes suffice, matching the epic); normalising uppercase digests on write (rejected: silent rewriting at the boundary hides writer bugs from the very audit that reads them).


Cyrax321:
Gate outputs at head 3ac96ef: pytest 1482 passed, 23 skipped (exit 0); ruff check src/ tests/ examples/ clean; ruff format --check clean (220 files); mypy src/continuum strict clean (105 files). New tests fail on pre-change code (collection ImportError).

PR #440, opened 2026-08-27
fix(cli): make complete idempotent

### LINKED ISSUE #356: CLI complete should be idempotent and reject double complete
## Description

`continuum complete <run_id>` can be run twice on the same run and appends duplicate `REVIEW_CONFIRMED` and `RUN_COMPLETED` events each time. A run that is already `COMPLETED` should be idempotent. Appending a second close makes the log noisy and hides whether the first close was the real one.

Why this matters: completed runs surface as the active run for fresh sessions. Duplicate close events make it harder to audit when a run was actually finished.

## Current Behavior

`src/continuum/cli/main.py:844` `cmd_complete` does:

```python
run = storage.get_run(args.run_id)
storage.append_event(args.run_id, EventType.REVIEW_CONFIRMED, ...)
storage.append_event(args.run_id, EventType.RUN_COMPLETED, ...)
storage.update_run(run.touch(status=RunStatus.COMPLETED))
```

There is no check for `run.status == RunStatus.COMPLETED` before appending. Running `continuum complete r1` twice on a completed run writes two pairs of events.

## Expected Behavior

Second `complete` on a run whose `status` is already `COMPLETED` should be a no-op with a clear message such as `Run r1 is already completed` and exit 0, without appending new events. Alternatively, refuse with a distinct exit code and message. Either is fine if documented and consistent with `complete --help`.

## Reproduction

```bash
continuum --db /tmp/test.db start r1 --goal "test"
continuum --db /tmp/test.db complete r1
continuum --db /tmp/test.db events r1 | wc -l
continuum --db /tmp/test.db complete r1
continuum --db /tmp/test.db events r1 | wc -l  # increases by 2, should not
```

## Evidence

- File: `src/continuum/cli/main.py:844`
- Function: `cmd_complete`
- No test covers double complete. `rg -n "complete" tests/test_cli.py` shows only single close cases.
- `storage.update_run` on an already completed run succeeds, so DB does not protect.

## Suggested Direction

In `cmd_complete`, after `run = storage.get_run(args.run_id)`, check `if run.status == RunStatus.COMPLETED:` and return early with `_emit({"run_id": args.run_id, "status": "completed", "already_completed": true}, "Run r1 is already completed.", ...)`. Do not append events.

For `good first issue` keep the change small: early return plus a test `test_complete_is_idempotent` that calls `complete` twice and asserts event count does not grow.

## Acceptance Criteria

- `continuum complete r1` twice does not grow `read_events(r1)` after the first call.
- Second call prints `already completed` or equivalent and exits 0.
- `pytest tests/test_cli.py` passes, new test covers the case.

## Additional Context

Related to `continuum tree` and `get_active_run` which hides completed runs. No existing issue covers double complete idempotency.

quangshuynh:
## Description

Makes `continuum complete <run_id>` idempotent when the target run is already completed.

`cmd_complete` now returns early for `RunStatus.COMPLETED` before appending events or updating the run. A repeated completion exits successfully, reports that the run is already completed, and includes `already_completed: true` in structured JSON output.

The first completion path is unchanged.

Regression coverage verifies that repeating `complete` does not append duplicate `REVIEW_CONFIRMED` or `RUN_COMPLETED` events.

## Related issues

Fixes #356

## Types of changes

- Type: `bugfix`

## Breaking changes

No.

## How has this been tested?

Added regression coverage in `tests/test_cli.py` for:

- successful repeated completion
- already-completed text output
- structured JSON output with `already_completed: true`
- exact event-log equality before and after the repeated completion

`git diff --check` passed.

I could not run the Python test and quality-check commands locally because this environment does not have a usable Python installation or virtual environment. `python` and `py` resolve to inaccessible Windows Store aliases, and `pytest`, `ruff`, and `mypy` are unavailable.

The following repository checks therefore need to be verified by CI:

- `pytest`
- `ruff check src/ tests/`
- `ruff format --check src/ tests/`
- `mypy src/continuum`

## Checklist

- [x] Tests added or updated for the changed behaviour.
- [x] Documentation updated where the change affects user-facing behaviour.
- [ ] CI passes on the latest commit.
- [x] Security-relevant changes state known limitations explicitly in this description. (Not security-relevant.)

## Additional context

The fix is intentionally limited to `cmd_complete`. No storage or schema changes are necessary because the duplicate events originate from the CLI completion path.

The changelog was updated under `Unreleased` in accordance with the repository's contribution guidelines.

## Summary by CodeRabbit

- **Bug Fixes**
  - Repeatedly completing an already-completed run now succeeds safely without creating duplicate events or changing its history.
  - Completion responses indicate when the run was already completed, including in JSON output.

- **Documentation**
  - Added an Unreleased changelog entry describing idempotent completion behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project are documented here. The format follows
 
 ## [Unreleased]
 
+### Fixed
+
+- Make `continuum complete` idempotent for runs that are already completed (#356).
+
 ## [0.1.0] - 2026-08-27
 
 ### Added
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -917,6 +917,21 @@ def cmd_complete(args: argparse.Namespace, storage: Storage, out: Any, err: Any)
     gates) and flips the run row to COMPLETED.
     """
     run = storage.get_run(args.run_id)  # raises RunNotFound -> NOT_FOUND
+    if run.status is RunStatus.COMPLETED:
+        _emit(
+            {
+                "run_id": args.run_id,
+                "status": run.status.value,
+                "summary": args.summary or "",
+                "already_completed": True,
+            },
+            f"Run {args.run_id} is already completed.",
+            as_json=args.json,
+            stream=out,
+            palette=getattr(args, "_palette", None),
+        )
+        return ExitCode.OK
+
     note = {"summary": args.summary} if args.summary else {}
     storage.append_event(
         args.run_id,
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -934,14 +934,23 @@ def test_complete_closes_a_run_and_clears_it_from_active_resolution(
         assert store.get_active_run() is None
 
 
-def test_complete_is_idempotent_enough_for_double_clicks(tmp_path: Path) -> None:
+def test_complete_is_idempotent_for_double_clicks(tmp_path: Path) -> None:
     path = str(tmp_path / "d.db")
     with SQLiteStorage(path) as store:
         store.create_run(Run(run_id="r", goal="g"))
     code, _, err = run("--db", path, "complete", "r")
     assert code == ExitCode.OK
-    code, _, err = run("--db", path, "complete", "r")
+    with SQLiteStorage(path) as store:
+        events_after_first = list(store.read_events("r"))
+
+    code, out, err = run("--db", path, "complete", "r")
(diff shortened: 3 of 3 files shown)


Cyrax321 (approved):
Hi @quangshuynh, thank you for your contribution to CONTINUUM. I appreciate the careful work on this fix.

I verified the change locally on head 9e7b336 and everything looks good. The handling for already completed runs now behaves as described and is well covered by the new test.

Approving as ready to pull. Thanks again for the contribution.

Issue #427, opened 2026-08-26
save_budgets rewrites budgets.json in place, so a crash mid-write can truncate the registry


Cyrax321:
### Before reporting

- [X] I searched existing issues for a duplicate
- [X] I am reporting this against the correct repository
- [X] I am using a supported Python version (3.11 or later)
- [X] I can reproduce this on the latest commit on `main`

### Affected component

Core (state, checkpoints, recovery)

### Severity

Minor inconvenience or cosmetic issue

### Regression check

- [ ] This worked in an earlier version or commit

### System information

OS: macOS (darwin), Python 3.14.5, repository `.venv`
continuum-agent: source checkout at `origin/main` merge commit `50a19f7fe7c6e59a2a2ee1ed0d97b6dfd8419391` (includes PR #424 / issue #411)

### Command or code to reproduce

Not a runtime repro but a code path: `save_budgets` in `src/continuum/budgets.py` (added by #411) persists with

```python
path.parent.mkdir(parents=True, exist_ok=True)
path.write_text(json.dumps(data, indent=2) + "\n", encoding="utf-8")
```

`write_text` opens the target with mode `w`, which truncates before writing. There is no temporary file and no `os.replace`, so a crash, OOM kill or power loss between truncation and flush leaves a zero-length or partial `budgets.json`.

### Current behaviour

A process dying mid-save leaves the registry truncated or half-written. Every later `load_budgets` then raises `BudgetConfigError` ("is not valid JSON"), which is fail-closed, so every budget-gated claim refuses until an operator repairs the file by hand.

Today nothing in production calls `save_budgets` yet (tests only), so exposure is currently nil. That changes with #413, which makes registry persistence part of the claim path, i.e. a write per claim attempt. The more frequent the writer, the wider the crash window.

### Expected behaviour

Write to a temporary file in the same directory and move it into place atomically:

1. serialize to `budgets.json.tmp-<pid>` (or similar) in `path.parent`;
2. optionally `flush` + `os.fsync`;
3. `os.replace(tmp, path)`, which is atomic within a filesystem on POSIX and Windows.

Losing the final increment on an abrupt exit is acceptable for a counter registry; losing the whole file is not. Round-trip tests from #411 should pass unchanged, plus a test asserting no stray temp files remain after a save.

### Logs or CI run

n/a

Issue #413, opened 2026-08-26
[#390] Budget drawdown at claim and confirmation with fresh-key refusal


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

Parent epic: #390. Do after identity resolution lands.

### What problem does this solve?

Budgets that only count claims miss half the picture: a completed confirmation also consumes the authorized operation's execution budget.

### Proposed solution

Drawdown wiring at both ends:

- claim slot increments the `(action_type, authorization)` counter;
- confirmation/settlement events consume from the same budget;
- refusal once remaining hits zero, regardless of key freshness;
- all writes go through the registry's pure helpers from child one.

Acceptance criteria: N distinct fresh keys for one authorization exhaust at cap; distinct authorizations independent; settlements draw down visibly via `continuum budget`.

### Alternatives considered

Claims-only counting. Rejected: allows cap-amplification through rapid complete/re-claim cycles.

### Is this a breaking change?

No

### Would you like to work on this?

No

Cyrax321:
Claiming #413

Cyrax321:
Verified and merged as a01a9d5 on main (Fixes #413). Acceptance: N distinct fresh keys for one authorization share one counter and exhaust at cap, distinct authorizations independent, settlements draw down same counter. Full gate: 1637 passed, ruff/mypy clean. Follow-up #460 for repo hygiene.

PR #442, opened 2026-08-27
fix(budgets): write through a symlinked registry and keep its owner

abyyxhek:
## Description

Follow-up to #441, which was merged while CodeRabbit's review was still open. Both of its findings are real and neither was addressed in that PR, so they land here. Both are consequences of the same thing: #441 replaced `Path.write_text` with staging plus `os.replace` for crash-atomicity (#427), and `os.replace` installs a **new inode** rather than rewriting the old one. Three properties of the old file therefore have to be re-established deliberately. #441 handled one of them (the mode); these are the other two.

**A symlinked registry was replaced instead of written through.** `write_text` followed a terminal `budgets.json` symlink and wrote the shared file behind it. `os.replace` swaps the link itself for a regular file. `load_budgets` still reads through the link (`read_text`), so after one save the two halves of the module pointed at different files: the writer saw its own counters, and every other consumer of the shared registry kept the ones from before, diverging further with each save once #413 makes this a write per claim attempt. Divergent budget counters are worse than a single stale set, because each reader believes a different number of attempts is left. `save_budgets` now resolves the path before staging. That also closes an atomicity hole I had not noticed: the staging file was being created next to the *link*, which is not necessarily on the same filesystem as the real target, so `os.replace` was not guaranteed to be atomic there either.

**Preserving the mode without the ownership preserved nothing.** Mode 0640 grants read to a *group* without saying which one, and `mkstemp` creates its inode under the saving process's own uid and gid. Copying 0640 across therefore handed group read to whichever group the saving process happened to sit in and took it from the group the operator chose - so #441's permission fix was only half an answer to "who may read this". `_staged_mode` becomes `_staged_attributes`, returning mode, uid and gid, and `_restore_ownership` writes ownership onto the staged file **before** the chmod: `chown` clears setuid and setgid on some systems, and a chmod running second would silently undo that.

## Related issues

Refs #427
Follow-up to #441
Refs #413 (which turns this into a write per claim attempt)

## Types of changes

- Type: `bugfix`

## Breaking changes

No. Both changes make an atomic save reproduce what `write_text` did before #441: a symlinked registry is written through again, and a registry's group keeps its access. Nothing that loaded or saved before this change behaves differently, except in the direction of the pre-#441 behaviour.

## How has this been tested?

```
.venv/Scripts/python.exe -m pytest -q                    # full suite green
.venv/Scripts/ruff.exe format --check src/ tests/
.venv/Scripts/ruff.exe check src/ tests/
.venv/Scripts/mypy.exe src/continuum                     # strict, 107 files
```

Six tests, in `tests/test_retry_budgets.py`:

- `test_a_symlinked_registry_is_written_through_not_replaced` - a terminal `budgets.json` symlink to a shared registry: the shared target receives the data, `link.is_symlink()` still holds, and neither directory is left with staging litter. This is the regression test CodeRabbit asked for.
- `test_save_keeps_the_group_the_registry_was_given` - a non-primary gid taken from `os.getgroups()`, asserting both `st_gid` and 0640 survive a save. Skips when the test user belongs to a single group, since there is then nowhere to move the file.
- `test_a_group_that_cannot_be_restored_does_not_fail_the_save` - the best-effort contract, and that the mode is still restored on its own.
- `test_ownership_restore_falls_back_to_the_group_alone` - asserts the two `chown` attempts and their order.
- `test_a_registry_with_no_group_recorded_is_left_alone` - no recorded gid means no `chown` call, rather than passing `None` to it.
- `test_a_platform_without_ownership_still_reproduces_the_mode` - Windows keeps the bits and skips the ownership.

Environment caveat, same as on #441: I develop on Windows and could not get a Linux container up locally (no usable WSL distro; `docker pull` failed on a TLS handshake timeout with nothing cached). The POSIX-only tests here - permission bits, ownership, symlinks - skip on my machine and are exercised by the `ubuntu-latest` and `macos-latest` legs of the matrix rather than locally. Everything platform-independent passes locally. The Windows-only path is tested where it differs rather than assumed: `_CAN_CHOWN` keeps the gid out of `_staged_attributes` there, and `_restore_ownership` returns before the `chown` calls, which is also what makes the module type-check under `mypy --strict` on a platform where `os.chown` does not exist.

## Checklist

- Tests added or updated for the changed behaviour: yes. Each fails on the code as merged in #441 - the symlink test finds a regular file where the link was, and the group test finds the saving process's own gid.
- Documentation updated where the change affects user-facing behaviour: yes, the `CHANGELOG.md` entry for #427 is extended with both properties and the best-effort limits, and the module docstring names all three.
- CI passes on the latest commit.
- Security-relevant changes state known limitations explicitly in this description: yes, below.

### One reasoned deviation from the review

CodeRabbit asked for the save to **abort before `os.replace` when the group cannot be restored**. I did not do that, for the same reason the chmod and the directory fsync in #441 are best effort. `save_budgets` is the write side of a fail-closed gate, so a refused save is a refused claim: failing because a *permission could not be reproduced* converts an ownership limitation into blocked work. And nothing is lost by continuing that the abort would have kept - the replacement never had that group's access to hand out, so the group loses access either way; the abort just also loses the increment.

What `_restore_ownership` does instead is try `(uid, gid)`, then retry `(-1, gid)`, both suppressed. A non-root process cannot give a file away but can put a file back into a group it belongs to, and the gid is the half that grants access to anyone other than the owner.

I also decided against narrowing the mode when the chown fails. In the realistic deployment the process is in the shared group and the chown succeeds; when it does not, the original group has lost access regardless of what the bits say, so tightening them would only remove access from the identity that did write the file.

### Known limitations

- A non-root process cannot always reproduce ownership. The save proceeds; the mode is still applied.
- A filesystem without permission bits keeps `mkstemp`'s 0600. Too private is recoverable by hand; a refused save under a fail-closed gate is not.
- setuid, setgid and the sticky bit are deliberately not copied onto the replacement.
- Windows keeps the mode only. There is no POSIX owner or group there for a replaced registry to lose.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved budget registry saves when the registry path is a symlink.
  * Preserved file ownership and permissions during atomic saves.
  * Added resilience when ownership cannot be restored or platform support is unavailable.
  * Ensured ownership is applied safely without disrupting special file permissions.
* **Tests**
  * Expanded coverage for symlink handling, ownership preservation, fallback behavior, and permission restoration.
* **Documentation**
  * Updated changelog details for atomic budget saves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -799,6 +799,21 @@ All notable changes to this project are documented here. The format follows
   loseable by the very crash it guards against. Both steps are best effort: a
   filesystem without permission bits keeps the tighter 0600, and Windows (which
   cannot open a directory as a descriptor) is left exactly as durable as before.
+  Swapping an inode for a rewritten one also changes two further things
+  `write_text` did not, so both are re-established: the path is resolved before
+  staging, because a `budgets.json` that is a symlink to a shared registry was
+  written *through* by `write_text` and would be *replaced* by `os.replace`,
+  leaving every other reader of the shared file - and `load_budgets`, which still
+  reads through the link - on the counters from before; and the replaced inode's
+  uid and gid are put back onto the staged file, because mode 0640 names a group
+  without saying which one, and a fresh `mkstemp` inode belongs to whichever group
+  the saving process sits in. Ownership is written before the mode, since `chown`
+  clears setuid and setgid on some systems and a `chmod` running second would
+  silently undo that. A refused `(uid, gid)` is retried as the gid alone, which is
+  the half that grants access to anyone but the owner, and a `chown` that cannot be
+  performed at all does not fail the save: as with the chmod and the flush, turning
+  an unreproducible permission into a refused claim is the wrong direction for a
+  fail-closed gate.
 
 - **Budget rejections named the rule but not the offending value, and one still
   named the config by relative path (#326, #426).** "needs a positive integer
--- a/src/continuum/budgets.py
+++ b/src/continuum/budgets.py
@@ -382,7 +382,7 @@ def _restore_ownership(name: str, attributes: _StagedAttributes) -> None:
     the wrong direction for a fail-closed gate: the replacement never had that
     group's access to hand out, so nothing is lost that this call could keep.
     """
-    if sys.platform == "win32":
+    if sys.platform == "win32":  # pragma: no cover - one arm is dead on either platform
         # Named literally rather than through _CAN_CHOWN, which already keeps the
         # gid out of _staged_attributes here: os.chown does not exist on Windows,
         # so the calls below have to be unreachable to a type checker too.
--- a/src/continuum/budgets.py
+++ b/src/continuum/budgets.py
@@ -40,10 +40,12 @@ Where the registry asks for an integer it means one: JSON ``true`` is refused
 rather than read as a silent cap of 1 (issue #429), and a rejection names the
 offending value and its type (issue #326). ``save_budgets`` stages and renames
 rather than truncating in place, so a process that dies mid-write costs at most
-the last increment, never the registry (issue #427); the staged file inherits
-the target's permissions and the rename is flushed, so replacing the registry
-neither locks other readers out of it nor can be lost by the crash it guards
-against.
+the last increment, never the registry (issue #427). Because that replaces an
+inode instead of rewriting one, the staged file inherits the target's
+permissions and ownership, a symlinked registry is written through rather than
+replaced, and the rename is flushed: swapping the file must not change who can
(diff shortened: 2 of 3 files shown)


Cyrax321 (approved):
Hi @abyyxhek, thank you for your contribution to CONTINUUM. I appreciate the careful follow up to #441.                                                                                                 
                                                                                                                                                                                                         
 I checked this locally on head 30d9cc0 and all looks fine. The symlink is now written through and the staged file keeps the previous mode and ownership, and CI is green. Approving as ready to pull after you merge main to clear the 
behind state. Thank you again for the thorough work.

PR #462, opened 2026-08-28
fix(observations): reject bool in payload bytes field

### LINKED ISSUE #452: fix(observations): payload bytes check should reject bool
## Description

In `src/continuum/recovery/observations.py:63` the payload size is taken as:

```python
size = payload.get("bytes") if isinstance(payload.get("bytes"), int) else None
```

In Python `isinstance(True, int)` is true, so a payload with `{"bytes": true}` is treated as size 1 instead of being ignored as invalid. The field is expected to be a non negative integer byte count.

## Current behaviour

`{"bytes": True}` passes the int check and is stored as 1, which can affect observation status calculation.

Verified on main at `a68882e` by inspecting the line and with `PYTHONPATH=src python -c "print(isinstance(True, int))"` which prints True.

## Expected behaviour

Reject bool where an integer is required, using a helper like `_is_int` in `budgets.py` that checks `isinstance(x, int) and not isinstance(x, bool)`, or explicitly check `type(x) is int`.

## Suggested fix

Change line 63 to use `not isinstance(..., bool)` guard, add a test that `{"bytes": True}` and `{"bytes": False}` are treated as None.

## Files

src/continuum/recovery/observations.py

No em dashes, small and isolated, good for a first PR.

Newer1107:
## Description

`isinstance(True, int)` is true in Python, so a payload with `{"bytes": true}` was treated as size 1 instead of being ignored as invalid. Use `type() is int` to reject bool values.

## Changes

- `src/continuum/recovery/observations.py:63`: Changed `isinstance(payload.get("bytes"), int)` to `type(payload.get("bytes")) is int`

## Related Issues

Resolves #452

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected byte-count handling so boolean values are no longer treated as valid byte counts during recovery observations.
  * Preserved existing behavior for valid numeric byte values.

### DIFF
--- a/src/continuum/recovery/observations.py
+++ b/src/continuum/recovery/observations.py
@@ -60,7 +60,7 @@ def _entry_from_event(event: Event, root: Path) -> dict[str, Any] | None:
     if not isinstance(tool, str) or not isinstance(path, str) or not path:
         return None
     sha = payload.get("sha256") if isinstance(payload.get("sha256"), str) else None
-    size = payload.get("bytes") if isinstance(payload.get("bytes"), int) else None
+    size = payload.get("bytes") if type(payload.get("bytes")) is int else None
     resolved = Path(path)
     if not resolved.is_absolute():
         resolved = root / resolved


Cyrax321 (approved):
Thank you for contributing to CONTINUUM, @Newer1107, and for taking on #452.

The fix is correct, minimal, and matches the issue requirements. Using `type(...) is int` properly rejects boolean values while preserving valid integer inputs.

I reviewed the implementation, and reproduced the original behavior on `main`, and verified the fix. Tests and CI are green.

No blocking issues from my side. Approving and marking this ready to merge. Thanks again for the clean fix!

Issue #449, opened 2026-08-28
feat(hooks): automate PreCompact wiring and document Codex non-Bash observation gap


Cyrax321:
### Affected component

Documentation, CLI hooks

### What problem does this solve?

Follow-up to #396 (docs embeddability sprint, PR #448).

PR #448 landed the copy-paste recipes for Claude Code SessionStart + PreCompact and Codex SessionStart, plus the bring-your-own-dashboard guide, all tested with real `os._exit(9)` hard kills. The recipes work, but two wiring gaps remain that could not be closed inside #396 because that sprint was scoped `docs/guides` + `examples/hooks` + small `docs/adapters` fixes only ("Nothing else" per #396 Allowed files). Any code change to `src/continuum/clienthooks.py` or `src/continuum/cli/main.py` was forbidden.

**Gap 1: PreCompact is manual.** `continuum hooks install claude-code` wires `SessionStart` -> `continuum briefing` and `PostToolUse` -> `continuum observe` (and optionally `PreToolUse` -> `gate`), but PreCompact still requires a hand-edited `.claude/settings.json` entry (see `docs/guides/embed-claude-code.md` PreCompact section and `examples/hooks/continuum-precompact.sh`). Forgetting it means context compaction can drop unverified reasoning without a checkpoint or `resume --json` snapshot.

**Gap 2: Codex observation is Bash-only.** `CLIENT_PROFILES["codex"]` documents `^Bash$|^shell$` as the only traversed matcher. `apply_patch` and MCP tools do not fire `continuum observe`, so file writes via those surfaces are invisible to `contract.post_checkpoint_observations` unless the user routes them through a shell call. The guides state the limitation and the workaround, but the durability is still opt-in.

Both gaps are adoption friction: each is a run that crashes unrecovered.

### Proposed solution

Code change, not docs:

- Extend `install_client_hook` / `hooks install` to optionally wire PreCompact for Claude Code (and the future Codex equivalent) to `examples/hooks/continuum-precompact.sh` or the inline `continuum checkpoint --reason "pre-compact" && continuum resume --json` command. Keep it opt-in behind `--with-precompact` or default-on with a flag to skip, and make `hooks remove` clean it up alongside the other hooks.

[...]

Issue #213, opened 2026-08-22
Roadmap: enforced durability at the harness boundary (umbrella)


Cyrax321:
### Affected component

Other / not sure

### What problem does this solve?

Umbrella tracking for the enforced-durability roadmap. CONTINUUM's moat is provenance-gated recovery: persisted state is treated as an untrusted claim that must be traced to its origin and revalidated before resume. The structural hole (surfaced by a full-code review on 2026-08-22) is that everything past the MCP boundary is voluntary:

- `continuum_intercept_action` is a protocol convention. An agent can fire a side effect without ever claiming it, and nothing catches that.
- Idempotency keys at the MCP boundary depend on the LLM choosing stable strings, which models do not do reliably.
- Uncertain side effects require a human per case, which does not scale.

Durable-execution engines solve enforcement by owning the runtime loop, which CONTINUUM cannot do for external coding agents. The unsolved niche is enforcement at the harness boundary. Issue #210 shipped the observation half (PostToolUse evidence capture). The remaining half is pre-action gating: harness hooks that can deny a tool call outright.

Phases, each shippable alone:

- [ ] Phase 0a: lazy adapter imports (#214)
- [ ] Phase 0b: publish to PyPI (#215)
- [ ] Phase 0c: index table for unscoped idempotency claims (#216)
- [ ] Phase 1: `continuum gate` PreToolUse enforcement + key derivation registry (#217)
- [ ] Phase 2: reconciler registry for auto-reconciliation (#218)
- [ ] Phase 3: universality (#208 validator surfacing, #209 other-client installers)

### Proposed solution

Land each phase as its own reviewed PR against its own issue, in order, with the flagship (Phase 1) designed so the universal contract stays client-agnostic exactly as observe did in #210.

### Alternatives considered

Do nothing and rely on model discipline: measured insufficient, twice (issue #6 e2e argument drift; the #207 kill-window reproduction). Own the whole runtime like Temporal: abandons every external CLI, which is where the actual users are.

### Is this a breaking change?

No

### Would you like to work on this?

Yes, I have a plan

Cyrax321:
Status update, 2026-08-23: every phase is merged to main except Phase 0b.

- [x] Phase 0a lazy imports (#214, PR #219)
- [ ] Phase 0b PyPI (#215) -- awaiting maintainer PyPI account setup
- [x] Phase 0c action index (#216, PR #221)
- [x] Phase 1 gate (#217, PR #220)
[...]

PR #485, opened 2026-08-29
fix(hooks): recognise the briefing kind so install stays idempotent

### LINKED ISSUE #484: hooks install duplicates the SessionStart briefing hook on every run
### Before reporting

- [X] I searched existing issues for a duplicate
- [X] I am reporting this against the correct repository
- [X] I am using a supported Python version (3.11 or later)
- [X] I can reproduce this on the latest commit on `main`

### Affected component

CLI

### Severity

### Regression check

- [X] This worked in an earlier version or commit

### System information

OS: Windows 11 (Windows-10-10.0.26200-SP0); reproduces identically on POSIX via the library-level snippet below
Python: 3.11.15
continuum-agent: source checkout at `upstream/main` commit `a7705d81cc3d5f38e53fb3fd102c054b0052ba62`
Installed via: `pip install -e ".[dev]"`

```shell
# 1. hooks install is not idempotent for the SessionStart briefing hook
cd "$(mktemp -d)"
for i in 1 2 3; do
  python -m continuum.cli --json hooks install claude-code --settings ./settings.json \
    | python -c "import json,sys;[print(h['kind'],h['event'],h['status']) for h in json.load(sys.stdin)['hooks']]"
done
python -c "import json;h=json.load(open('settings.json'))['hooks'];\
print('PostToolUse groups:',len(h['PostToolUse']));\
print('SessionStart groups:',len(h['SessionStart']))"

# 2. the same asymmetry, isolated: a moved virtualenv is repointed for
#    observe and duplicated for briefing
python - <<'PY'
import json, tempfile
from pathlib import Path
from continuum.clienthooks import install_client_hook

s = Path(tempfile.mkdtemp()) / "settings.json"

print("observe  old ->", install_client_hook(s, "/old/venv/bin/continuum observe",
                                             event_name="PostToolUse", matcher="W"))
print("observe  new ->", install_client_hook(s, "/new/venv/bin/continuum observe",
                                             event_name="PostToolUse", matcher="W"))
print("PostToolUse groups:", len(json.loads(s.read_text())["hooks"]["PostToolUse"]))

print("briefing old ->", install_client_hook(s, "/old/venv/bin/continuum briefing",
                                             event_name="SessionStart", matcher=""))
print("briefing new ->", install_client_hook(s, "/new/venv/bin/continuum briefing",
                                             event_name="SessionStart", matcher=""))
groups = json.loads(s.read_text())["hooks"]["SessionStart"]
print("SessionStart groups:", len(groups))
for g in groups:
    print("   ", g["hooks"][0]["command"])
PY
```

### Current behaviour

Reproduction 1, verified at `a7705d8`:

```
observe PostToolUse installed
briefing SessionStart installed
observe PostToolUse present
briefing SessionStart installed
observe PostToolUse present
briefing SessionStart installed
PostToolUse groups: 1
SessionStart groups: 3
```

The observe hook settles to `present` and stays one entry. The briefing hook
reports `installed` on every run and `.claude/settings.json` ends up holding
three byte-identical `SessionStart` groups, each running
`continuum briefing`. Claude Code (and every other profiled client, since all
three profiles use `start_event: SessionStart`) then runs the briefing once per
duplicate on every session start, so the resume context that #213/#227 exist to
inject exactly once is injected N times, and each copy opens storage.

Reproduction 2 isolates the same defect on the repointing path:

```
observe  old -> installed
observe  new -> updated
PostToolUse groups: 1
briefing old -> installed
briefing new -> installed
SessionStart groups: 2
    /old/venv/bin/continuum briefing
    /new/venv/bin/continuum briefing
```

A moved virtualenv is repointed for observe. For briefing the stale
`/old/venv/bin/continuum briefing` survives and keeps firing alongside the new
one, which is the failure mode `_install_hook`'s `"updated"` status exists to
prevent.

### Expected behaviour

`continuum hooks install` is documented and tested as idempotent
(`tests/test_client_installers.py::test_install_is_idempotent_per_client`,
`_install_hook`'s "Returns `installed` ... `updated` ... `present`" contract).
Re-running it should leave exactly one `SessionStart` briefing entry, report
`present` when nothing changed, and repoint rather than duplicate when the
resolved command has moved. The briefing hook should behave exactly like the
observe and gate hooks it was added alongside.

### Root cause

`src/continuum/clienthooks.py:305-307`, the "is this entry already ours" test
inside `_install_hook`, only recognises two of the three kinds this module
installs:

```python
ours = isinstance(hook, dict) and (
    _is_continuum_hook(hook, "observe") or _is_continuum_hook(hook, "gate")
)
```

`_is_continuum_hook` matches on the final word of the command
(`clienthooks.py:213`), so a `... continuum briefing` command matches neither
arm, `entry_found` stays `False`, and the tail of `_install_hook` appends a
fresh group unconditionally.

The mismatch was introduced by 7c6248d (PR #227, "feat(cli): session briefing
injects durable state without CLAUDE.md"), which added the briefing kind and
updated the *remover's* kind list to `("observe", "gate", "briefing")` while
leaving the installer's recogniser at observe/gate:

```
-                    and (_is_continuum_hook(h, "observe") or _is_continuum_hook(h, "gate"))
+                    and any(_is_continuum_hook(h, k) for k in ("observe", "gate", "briefing"))
```

The two lists are written out separately, so a kind added to one can silently
miss the other. `_install_hook`'s docstring still describes a `kind` parameter
("it must equal the final word of `command`") that no longer exists on the
function, which is the same drift.

### Logs or CI run

The existing coverage does not catch this. `test_install_is_idempotent_per_client`
asserts only on `data["hooks"][post_event]`, never on `start_event`, and
`test_briefing_is_wired_on_session_start` installs once, so
`assert len(ours) == 1` holds trivially. Both pass at `a7705d8`; the full suite
is green at that commit (`pytest --no-cov -q`, exit code 0).

Remediation for a settings file that has already accumulated duplicates:
`continuum hooks remove <client> --settings ...` followed by
`continuum hooks install <client> ...` (the remover does handle the briefing
kind, so it cleans all copies).

abyyxhek:
## Description

`continuum hooks install` was not idempotent for the `SessionStart` briefing hook: every re-run appended another copy instead of reporting it present.

`_install_hook` decides whether an entry already in the settings file is one of ours, and it asked only two questions:

```python
ours = isinstance(hook, dict) and (
    _is_continuum_hook(hook, "observe") or _is_continuum_hook(hook, "gate")
)
```

`_is_continuum_hook` matches on the command's final token, so a `... continuum briefing` command matched neither arm. `entry_found` stayed `False` and the function fell through to the append branch. The briefing hook was added in 7c6248d (#227), which updated the *remover's* kind tuple to `("observe", "gate", "briefing")` and left the installer's recogniser alone — the same drift left `_install_hook`'s docstring still describing a `kind` parameter that no longer exists.

Consequences, both reproduced below:

- Three `continuum hooks install` runs leave three byte-identical `SessionStart` groups, each reported as `installed`, so the client runs `continuum briefing` once per copy at every session start. All three profiled clients use `start_event: "SessionStart"`, so this affects `claude-code`, `gemini` and `codex` alike. The resume context that #213/#227 exist to inject exactly once is injected N times, and each copy opens storage.
- A moved virtualenv is duplicated rather than repointed: the observe hook reports `updated` and keeps one entry, while briefing reports `installed` and keeps both, so the stale interpreter path goes on firing alongside the new one.

The fix keeps the recognised kinds in one place and derives the kind from the command being installed:

- `_INSTALLED_KINDS = ("observe", "gate", "briefing")` is now read by both the installer and `remove_claude_code_hook`, so a future kind cannot be added to one and forgotten in the other.
- `_installed_kind(command)` reads the kind off the command instead of taking it from the call site, so a new kind cannot reach the installer without the installer knowing about it. A command this module did not build yields `None` and still appends, exactly as before.
- Matching is now kind-for-kind, so an observe install cannot repoint a briefing or gate hook that happens to share the event and matcher. Repointing across kinds would silently drop a hook the caller never asked about.

Scope note: this prevents new duplicates and repoints a moved command. It does not collapse duplicates already accumulated in a user's settings file — that would mean restructuring existing group lists, which is beyond a fix for the non-idempotency itself. Existing duplicates are cleaned by `continuum hooks remove <client>` followed by `continuum hooks install <client>`; this is documented in the issue and the CHANGELOG entry.

## Related issues

Fixes #484

## Types of changes

- Type: `bugfix`

## Breaking changes

No. Installed settings shape, statuses (`installed` / `updated` / `present`) and the removal contract are unchanged; only which existing entries count as ours changes, and only in the direction of recognising one more kind that this module itself installs.

## How has this been tested?

Environment: Windows 11 (26200), Python 3.13.7, pytest 9.1.1, ruff 0.16.3, mypy strict, project venv at `.venv`.

Before/after reproduction, verbatim (`--json` is a global flag, so it precedes the subcommand):

```
for i in 1 2 3; do python -m continuum.cli --json hooks install claude-code --settings ./settings.json; done
python -c "import json;d=json.load(open('settings.json'));print(len(d['hooks']['SessionStart']))"
```

Before: briefing status `installed`, `installed`, `installed`; 3 `SessionStart` groups.
After: briefing status `installed`, `present`, `present`; 1 `SessionStart` group.

Moved-virtualenv case (`install_client_hook` with `/old/venv/bin/continuum briefing` then `/new/venv/bin/continuum briefing`):

Before: `installed` → `installed`, 2 entries, stale `/old/...` retained.
After: `installed` → `updated`, 1 entry pointing at `/new/...`.

The two new tests fail on `upstream/main` and pass with this patch (4 failed / 17 passed before, all green after — the first is parametrized over three clients).

Commands run:

```
pytest tests/test_client_installers.py tests/test_cli_observe.py --no-cov     # 36 passed
pytest --tb=short -q --cov=continuum --cov-report=term-missing                # exit 0, full suite green, 87% total
ruff check src/ tests/ examples/                                              # All checks passed!
ruff format --check src/ tests/ examples/                                     # 240 files already formatted
mypy src/continuum                                                            # Success: no issues found in 108 source files
```

Baseline on `upstream/main` @ `a7705d8` was green before any change, so no failure here is pre-existing.

## Checklist

Tests added or updated for the changed behaviour — yes: `test_reinstalling_does_not_duplicate_the_briefing_hook` (parametrized over all three clients), `test_a_moved_briefing_command_is_repointed_not_duplicated`, and `test_a_command_of_no_known_kind_is_appended_never_matched` (the fallback arms of `_installed_kind`, added after Codecov flagged them as unhit). Documentation updated where the change affects user-facing behaviour — yes: `CHANGELOG.md` under `[Unreleased] / Fixed`, including the remediation for settings files that already hold duplicates. CI passes on the latest commit — all Test jobs (ubuntu/macos/windows × 3.11–3.13), Postgres backend, Lint & Type-check, Markdown lint and `codecov/patch` are green on `5da897b`. Security-relevant changes state known limitations explicitly — not security-relevant; the scope limitation is stated above.

## Additional context

Why the existing tests missed it: `test_install_is_idempotent_per_client` asserts only on `data["hooks"][profile["post_event"]]`, where a duplicate cannot arise, and `test_briefing_is_wired_on_session_start` installs once, so its `assert len(ours) == 1` holds trivially. The new test asserts on the `start_event` list across three consecutive installs and checks the reported status for each, so a regression shows up as both a wrong status and a duplicated entry.

Nearest existing work is open issue #449 (PreCompact wiring, Codex Bash-only matcher). Different scope; it confirms `SessionStart` → `briefing` is intended behaviour rather than proposing to change it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,19 @@ All notable changes to this project are documented here. The format follows
 
 - Make `continuum complete` idempotent for runs that are already completed (#356).
 
+- **`hooks install` no longer duplicates the SessionStart briefing hook (#484).**
+  The installer recognised an existing entry as its own only when the command
+  ended in `observe` or `gate`, so the briefing hook added alongside them was
+  appended afresh on every run: three `continuum hooks install` invocations left
+  three identical `SessionStart` groups, each reported as `installed`, and the
+  client ran `continuum briefing` once per copy on every session start. A moved
+  virtualenv was duplicated rather than repointed too, leaving the stale command
+  firing. The recognised kinds are now one list read by both the installer and
+  `hooks remove` (which already knew about `briefing`), and the kind is taken
+  from the command being installed, so an install of one kind cannot repoint
+  another. A settings file that already holds duplicates is cleaned by
+  `hooks remove` followed by `hooks install`.
+
 ## [0.1.0] - 2026-08-27
 
 ### Added
--- a/src/continuum/clienthooks.py
+++ b/src/continuum/clienthooks.py
@@ -57,6 +57,15 @@ __all__ = [
 #: matcher expression so the installed settings stay a single entry.
 DEFAULT_MATCHER = "Write|Edit|MultiEdit|NotebookEdit"
 
+#: Every hook kind this module installs, which is also the final word of each
+#: installed command. One list, read by both the installer and the remover: they
+#: used to carry the set separately, and when the briefing hook was added
+#: (7c6248d) only the remover's copy was updated. The installer therefore stopped
+#: recognising a briefing entry as its own, so every re-run appended a duplicate
+#: SessionStart group instead of reporting it present or repointing a moved one
+#: (issue #484).
+_INSTALLED_KINDS = ("observe", "gate", "briefing")
+
 #: Keys of ``tool_input`` that hold the primary file path, in priority order.
 _PATH_KEYS = ("file_path", "notebook_path")
 
@@ -221,6 +230,25 @@ def _is_observe_hook(hook: Mapping[str, Any]) -> bool:
     return _is_continuum_hook(hook, "observe")
 
 
+def _installed_kind(command: str) -> str | None:
+    """Which of :data:`_INSTALLED_KINDS` ``command`` ends in, or ``None``.
+
+    The kind a command carries is what makes an existing entry the same hook
+    rather than a different one, so it is read from the command being installed
+    instead of being listed at the call site: a new kind then cannot reach the
+    installer without the installer knowing about it. ``None`` for anything else,
+    which leaves :func:`_install_hook` appending, exactly as it does today for a
+    command this module did not build.
+    """
+    try:
+        tokens = _split_command(command)
+    except ValueError:
+        return None
+    if tokens and tokens[-1] in _INSTALLED_KINDS:
+        return tokens[-1]
+    return None
+
+
 def install_client_hook(
     settings_path: Path,
     command: str,
@@ -261,13 +289,17 @@ def _install_hook(
 ) -> str:
     """Add a continuum hook entry to a Claude Code settings file.
 
-    ``kind`` selects which hook this is ("observe" or "gate"); it must equal
-    the final word of ``command``. Existing settings are preserved; only the
-    matching list under ``hooks`` gains (or updates) our single entry.
+    Existing settings are preserved; only the matching list under ``hooks``
+    gains (or updates) our single entry of the kind ``command`` names.
     Returns ``"installed"`` when the entry was added, ``"updated"`` when an
     existing entry of the same kind pointed somewhere else (a moved
     virtualenv, say) and was repointed, ``"present"`` when nothing needed to
     change.
+
+    An entry counts as ours only when its kind matches the one being installed,
+    so an observe install cannot repoint a briefing or gate hook that happens to
+    share the event and matcher: repointing across kinds would silently drop a
+    hook the caller never asked about.
 
     A settings file that exists but is unreadable raises rather than being
     overwritten: a file someone edited by hand is a statement of intent, and
@@ -295,6 +327,7 @@ def _install_hook(
 
     status = "installed"
     entry_found = False
+    kind = _installed_kind(command)
     for group in hook_list:
         if not isinstance(group, dict) or group.get("matcher") != matcher:
             continue
@@ -302,9 +335,7 @@ def _install_hook(
         if not isinstance(entries, list):
             continue
         for hook in entries:
-            ours = isinstance(hook, dict) and (
-                _is_continuum_hook(hook, "observe") or _is_continuum_hook(hook, "gate")
-            )
+            ours = kind is not None and isinstance(hook, dict) and _is_continuum_hook(hook, kind)
             if ours:
                 entry_found = True
                 if hook.get("command") != command:
@@ -330,10 +361,10 @@ def remove_claude_code_hook(settings_path: Path) -> bool:
     """Remove every continuum hook this module installed. True when anything
     was removed.
 
-    Only entries this module's shape recognises are touched (observe and gate,
-    any matcher): a hand-written entry pointing elsewhere survives untouched,
-    as does every other key in the file. A group holding unrelated hooks keeps
-    them.
+    Only entries this module's shape recognises are touched
+    (:data:`_INSTALLED_KINDS`, any matcher): a hand-written entry pointing
+    elsewhere survives untouched, as does every other key in the file. A group
+    holding unrelated hooks keeps them.
     """
     if not settings_path.exists():
         return False
@@ -369,8 +400,7 @@ def remove_claude_code_hook(settings_path: Path) -> bool:
                 h
                 for h in group["hooks"]
                 if not (
-                    isinstance(h, dict)
-                    and any(_is_continuum_hook(h, k) for k in ("observe", "gate", "briefing"))
+                    isinstance(h, dict) and any(_is_continuum_hook(h, k) for k in _INSTALLED_KINDS)
                 )
             ]
             if len(kept_hooks) != len(group["hooks"]):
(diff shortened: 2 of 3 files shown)


Cyrax321 (approved):
Thank you @abyyxhek for contributing to CONTINUUM and for taking on #484.

This fixes a subtle idempotency issue in `continuum hooks install` where the SessionStart briefing hook was not recognised as an existing Continuum hook, causing repeated installs to duplicate the entry. The same mismatch also caused moved virtualenvs to be duplicated instead of repointed.

I reviewed `_installed_kind`, `_install_hook`, `remove_claude_code_hook`, and the new tests. Centralising the supported kinds in `_INSTALLED_KINDS` is the right design and keeps installer and remover behaviour consistent. Deriving the kind from the command and matching hooks by kind also prevents one hook type from incorrectly repointing another.

The tests cover repeated installs, moved briefing commands, and unknown command kinds. Local verification is clean, with the targeted tests, full suite, `ruff`, and `mypy` passing. CI is green with the patch covered by codecov.

The existing `installed` / `updated` / `present` contract and hook format are unchanged, so this remains a focused fix. Existing duplicates are intentionally left untouched, which is appropriate for the scope of this change.

One housekeeping item remains: `CHANGELOG.md` currently conflicts with `main`. Please resolve the conflict while keeping this entry under `[Unreleased] / Fixed` and preserving the existing ordering.

LGTM. No blocking correctness, security, or design issues found. Approved and ready to pull once the CHANGELOG conflict is resolved.

Cyrax321:
Closing as duplicate of #526 which was merged for #484.

Both PRs fix the same bug where `_install_hook` missed the `briefing` kind and duplicated the SessionStart hook on every install. #526 implements the preferred fix from the issue directly: `any(_is_continuum_hook(hook,k) for k in ("observe","gate","briefing"))` and updates the stale docstring, leaving the remover unchanged per the issue constraints.

#485 goes further with `_INSTALLED_KINDS` centralization and kind-specific matching, which is a good hygiene improvement to prevent future drift. If that centralization is desired, please open a follow-up hygiene issue referencing #484 and #526 rather than reopening this one. Thanks for the work on this.

Cyrax321:
Duplicate of #526 (merged for #484).

abyyxhek:
Thanks for the review and the pointer. #526 covers the bug, so I have not reopened this. I rebased the hygiene half onto current `main` — the CHANGELOG conflict you flagged is resolved (entry appended after the existing `[Unreleased] / Fixed` items, ordering preserved), the two tests here that duplicate #526's coverage are dropped, and what remains is `_INSTALLED_KINDS` shared with the remover plus the kind read off the command. It is up as #536, referencing #484 and #526. Happy to convert it to an issue first if you would rather track it that way.

PR #513, opened 2026-08-30
test: add regression test for render_key with whitespace-only value (#512)

### LINKED ISSUE #512: test: add regression test for render_key with whitespace-only value
## Description

src/continuum/gate.py normalize_key_value strips strings, but there is no test for whitespace-only value \`   \` which becomes \`\`. Template \`invoice:{id}\` would then become \`invoice:\` and hit idempotency empty key check.

Add a small test in tests/test_cli_gate.py that asserts \`render_key('invoice:{id}', {'id': '   '}) == 'invoice:'\` and that decide handles it gracefully.

## Files

tests/test_cli_gate.py

Small test, good for a first PR.

## Acceptance

Test covers whitespace-only case.

mhaye9545:
Closes #512.

### Summary
Adds a regression test in \	ests/test_cli_gate.py\ asserting that \
ender_key('invoice:{id}', {'id': '   '}) == 'invoice:'\ and that \decide\ handles the resulting key gracefully without error.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for whitespace-only template values.
  * Verified normalization to an empty value, correct `invoice:` rendering, and denial of unclaimed actions with the derived key included in the reason.

### DIFF
--- a/tests/test_cli_gate.py
+++ b/tests/test_cli_gate.py
@@ -79,6 +79,25 @@ def test_render_key_strips_surrounding_whitespace_from_values(padded: str) -> No
     assert render_key("invoice:{id}", {"id": padded}) == render_key("invoice:{id}", {"id": "123"})
 
 
+def test_render_key_handles_whitespace_only_value() -> None:
+    """A whitespace-only value normalizes to an empty string (issue #512).
+
+    When an LLM supplies whitespace-only for an argument, normalize_key_value
+    strips it to empty. The template still renders and decide handles the
+    resulting key gracefully without error.
+    """
+    assert render_key("invoice:{id}", {"id": "   "}) == "invoice:"
+    decision = decide(
+        {"send_invoice": {"key_template": "invoice:{id}"}},
+        "send_invoice",
+        {"id": "   "},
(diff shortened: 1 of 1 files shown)


Cyrax321 (approved):
Thank you @mhaye9545 for contributing to CONTINUUM and for taking on #512.

This is an important edge case from #361. After `normalize_key_value`, a whitespace-only value becomes an empty string, so `invoice:{id}` correctly renders as `invoice:`. `decide` must still return an unclaimed denial rather than raising, and the tests cover both the rendering and end-to-end `decide` paths.

I checked the placement alongside the existing #361 whitespace tests. The naming, docstring, and assertions are clear and focused on the smallest meaningful seam. The regression also correctly distinguishes the new behavior from the previous `invoice:   ` result.

CI is otherwise as expected. The three Windows `WinError 32` failures in `test_digest_auditable_and_briefing_consumption` are pre-existing failures on `main` related to file-backed storage cleanup and are unrelated to this change. They are already being addressed in #514.

LGTM. No blocking issues found. The change is narrow, correct, and fully addresses #512. Approved and ready to pull.

Issue #317, opened 2026-08-24
Cap dashboard POST body size to prevent unbounded reads


Cyrax321:
## Context
The enforcing gateway correctly limits request bodies. The dashboard does not, even though they share the same threat.

**Gateway** (`src/continuum/gateway.py:64`) caps at 10 MB and drains up to 256 MB before returning 413. **Dashboard** (`src/continuum/dashboard/app.py:118`) does:

```python
length = int(self.headers.get("Content-Length") or 0)
raw = self.rfile.read(length) if length else b""
```

No limit. A large POST to `/action/reconcile` can force the server to buffer an unbounded body. The dashboard binds to `127.0.0.1` by default, but hardening should be consistent.

## How to reproduce locally
1. Start the dashboard in one terminal:
```bash
python -m continuum dashboard --port 8001
```
2. In another terminal, send a large body:
```bash
python -c "print('x'*6000000)" > /tmp/big
curl -X POST http://127.0.0.1:8001/action/reconcile -H "Content-Length: 6000000" --data-binary @/tmp/big -i
# before fix: server tries to read all 6 MB
# after fix: 413
```

## Expected behavior
Form POSTs are tiny. Cap them at 1 MB, return `413 Body too large`, and drain with a bounded loop like the gateway does. Add a short docstring note on `make_dashboard_server`.

## Step by step
1. Open `src/continuum/dashboard/app.py` and find `do_POST`.
2. Add a constant `MAX_DASHBOARD_BODY = 1 * 1024 * 1024` near the top.
3. Before `self.rfile.read`, check `length > MAX_DASHBOARD_BODY`. If true, drain up to a sanity bound (reuse the gateway pattern) and return 413.
4. Copy the drain pattern from `src/continuum/gateway.py:212` to avoid duplicating bugs.
5. Add a test in `tests/test_dashboard.py` that posts a >1 MB body and expects 413.

## Files
- `src/continuum/dashboard/app.py`
- `tests/test_dashboard.py`
- Reference: `src/continuum/gateway.py`

Good for someone who likes security hardening with a clear reference implementation.

PR #525, opened 2026-08-30
fix(http): handle malformed Content-Length and chunked Transfer-Encoding in dashboard and gateway (#522)

### LINKED ISSUE #522: Follow-up to #317: handle malformed Content-Length and chunked Transfer-Encoding in dashboard and gateway
## Context

After #317 and PR #521, both `src/continuum/gateway.py` and `src/continuum/dashboard/app.py` cap POST bodies by `Content-Length` and drain up to 256 MB before returning 413. That matches the issue spec and the gateway reference implementation.

During verification of #317 we noticed two related hardening gaps that are out of scope for the 1 MB cap but share the same threat:

1. **Malformed `Content-Length`**: both handlers do `int(self.headers.get("Content-Length") or 0)` without a try/except. A header like `Content-Length: abc` raises `ValueError` which is not caught, so the handler crashes without a response and the connection closes. The gateway has the same pattern. A client that sends a malformed length should get a 400 instead of a dropped connection.

2. **Chunked `Transfer-Encoding`**: neither server checks `Transfer-Encoding: chunked`. When a client omits `Content-Length` and sends a chunked body, `length` is read as 0 and the body is treated as empty. An attacker can therefore send an unbounded chunked body that bypasses the 1 MB / 10 MB caps and the 256 MB drain limit.

Both gaps are present in gateway and dashboard, so the hardening should stay consistent, same rationale as #317.

## Repro

Malformed length (dashboard):
```bash
python -m continuum dashboard --port 8001 &
curl -X POST http://127.0.0.1:8001/action/reconcile -H "Content-Length: abc" --data-binary "x" -i
# before fix: connection closed, no HTTP response (ValueError in handler)
# after fix: 400 with a clear error
```

Chunked bypass (both servers):
```bash
# Gateway or dashboard: send a POST with Transfer-Encoding: chunked and no Content-Length.
# The server currently reads length 0 and does not limit the chunked stream.
```

## Expected behaviour

- If `Content-Length` is present but not a valid non-negative integer, answer `400 Body too large or malformed` (or similar) without crashing, same for gateway and dashboard.
- If `Transfer-Encoding: chunked` is present, either reject chunked POSTs with 411/400 or enforce the same 1 MB / 10 MB caps while reading the chunked stream, drain with the same 256 MB bound, and keep the two servers consistent.

## Search for duplicates

Checked `gh issue list --state open` and `--state closed` for duplicates before filing:
- Searched for "Content-Length", "chunked", "413", "drain", "cap", "malformed" across all open and closed issues and PRs.
- Found #317 (the cap itself, now fixed in #521) and #323 (gateway malformed JSON), but nothing about malformed Content-Length or chunked handling.
- No open PR covers this; PR #521 only addresses Content-Length caps.

## Scope

This is a follow-up, not part of #317. #317 is complete and verified (large body 413, small body 200, drain limits equal). This issue tracks the remaining hardening so the two servers fail closed on the next layer of malformed input.

## Files

- `src/continuum/dashboard/app.py` (do_POST, _html)
- `src/continuum/gateway.py` (_body, drain loop)
- Tests: `tests/test_dashboard.py`, `tests/test_gateway.py`

## References

- Follow-up to #317, PR #521
- Gateway reference: `src/continuum/gateway.py:69` (`MAX_BODY_BYTES`, `DRAIN_LIMIT_BYTES`) and `src/continuum/gateway.py:212` drain loop
- Dashboard fix: `MAX_DASHBOARD_BODY`, `DASHBOARD_DRAIN_LIMIT_BYTES` in `src/continuum/dashboard/app.py:18`

vjymisal0:
Fixes #522

### Summary
Hardens HTTP body ingestion across both \src/continuum/dashboard/app.py\ and \src/continuum/gateway.py\ to gracefully handle malformed \Content-Length\ headers and reject unsupported chunked \Transfer-Encoding\ streams with HTTP 400 Bad Request instead of dropping connections.

### Changes
- In \src/continuum/dashboard/app.py\:
  - Checked for \Transfer-Encoding: chunked\ and answered with 400 Bad Request.
  - Safely parsed \Content-Length\ and verified non-negative integers; returns 400 on \ValueError\ without crashing.
- In \src/continuum/gateway.py\:
  - Added early check for \Transfer-Encoding: chunked\ in \_body\ and answered with 400 JSON error.
  - Validated \Content-Length\ integer values and answered 400 on malformed values.
- In \	ests/test_dashboard.py\ & \	ests/test_gateway.py\:
  - Added unit test cases covering malformed \Content-Length\ and chunked \Transfer-Encoding\ requests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests with malformed, negative, or duplicate `Content-Length` headers now receive a 400 Bad Request response.
  * Requests using transfer encoding are rejected safely with a 400 response, and affected connections are closed.
  * Gateway routing now prevents cross-tenant access and denies memory claims already associated with another run.
  * Run details now include archived events in the total shown in the event-count hint.

* **Tests**
  * Added coverage for invalid and ambiguous request framing, connection closure, and cross-run memory claim rejection.

### DIFF
--- a/src/continuum/dashboard/app.py
+++ b/src/continuum/dashboard/app.py
@@ -281,7 +281,29 @@ def make_dashboard_server(
             self._html(render_dashboard_html(storage))
 
         def do_POST(self) -> None:  # noqa: N802
-            length = int(self.headers.get("Content-Length") or 0)
+            te = self.headers.get("Transfer-Encoding", "").lower()
+            if "chunked" in te:
+                self._html(
+                    "<h1>400 Bad Request</h1><p>chunked transfer encoding is not supported</p>",
+                    code=400,
+                )
+                return
+
+            cl_header = self.headers.get("Content-Length")
+            if cl_header is not None:
+                try:
+                    length = int(cl_header)
+                    if length < 0:
+                        raise ValueError("Content-Length must be non-negative")
+                except ValueError as exc:
+                    self._html(
+                        f"<h1>400 Bad Request</h1><p>malformed Content-Length: {html.escape(str(exc))}</p>",
+                        code=400,
+                    )
+                    return
+            else:
+                length = 0
+
             if length > MAX_DASHBOARD_BODY:
                 drained = 0
                 while drained < length:
--- a/src/continuum/dashboard/app.py
+++ b/src/continuum/dashboard/app.py
@@ -285,21 +285,32 @@ def make_dashboard_server(
             self._html(render_dashboard_html(storage))
 
         def do_POST(self) -> None:  # noqa: N802
-            te = self.headers.get("Transfer-Encoding", "").lower()
-            if "chunked" in te:
+            transfer_encodings = self.headers.get_all("Transfer-Encoding", [])
+            if transfer_encodings:
+                self.close_connection = True
                 self._html(
-                    "<h1>400 Bad Request</h1><p>chunked transfer encoding is not supported</p>",
+                    "<h1>400 Bad Request</h1><p>transfer encoding is not supported</p>",
                     code=400,
                 )
                 return
 
-            cl_header = self.headers.get("Content-Length")
+            content_lengths = self.headers.get_all("Content-Length", [])
+            if len(content_lengths) > 1:
+                self.close_connection = True
+                self._html(
+                    "<h1>400 Bad Request</h1><p>multiple Content-Length headers are not supported</p>",
+                    code=400,
+                )
+                return
+
+            cl_header = content_lengths[0] if content_lengths else None
             if cl_header is not None:
                 try:
                     length = int(cl_header)
                     if length < 0:
                         raise ValueError("Content-Length must be non-negative")
                 except ValueError as exc:
+                    self.close_connection = True
                     self._html(
                         f"<h1>400 Bad Request</h1><p>malformed Content-Length: {html.escape(str(exc))}</p>",
                         code=400,
--- a/src/continuum/gateway.py
+++ b/src/continuum/gateway.py
@@ -187,16 +187,11 @@ def match_route(
         for _v in body.values():
             if isinstance(_v, str) and _v in consumed_authorities:
                 ev = consumed_authorities[_v]
-                seq = (
-                    getattr(ev, "sequence", "?")
-                    if hasattr(ev, "sequence")
-                    else ev.get("sequence", "?")
-                )
-                payload = getattr(ev, "payload", {}) or {}
                 if hasattr(ev, "payload"):
                     seq = ev.sequence
-                    payload = ev.payload
+                    payload = ev.payload or {}
                 else:
+                    seq = ev.get("sequence", "?")
                     payload = ev.get("payload", {})
                 consumer = payload.get("consumer_run_id", "?")
                 return Decision(
@@ -268,13 +263,6 @@ def match_route(
                     )
         except Exception:
             action = None
-    else:
-        # action already from local, foreign stays None
-        pass
-    # If we already resolved action via foreign, we need to handle status below
-    # Reuse variable action; key already computed
-    # To avoid double lookup, we will have a flag
-
     if action is None or getattr(action, "action_type", None) != route.action_type:
         return Decision(
             False,
--- a/src/continuum/gateway.py
+++ b/src/continuum/gateway.py
@@ -240,7 +240,23 @@ class GatewayServer:
                 decode (issue #323). Callers catch both and return, since the
                 response is already on the wire.
                 """
-                length = int(self.headers.get("Content-Length") or 0)
+                te = self.headers.get("Transfer-Encoding", "").lower()
+                if "chunked" in te:
+                    self._respond(400, {"error": "chunked transfer encoding is not supported"})
+                    raise _MalformedBody
+
+                cl_header = self.headers.get("Content-Length")
+                if cl_header is not None:
+                    try:
+                        length = int(cl_header)
+                        if length < 0:
+                            raise ValueError("Content-Length must be non-negative")
+                    except ValueError as exc:
+                        self._respond(400, {"error": f"malformed Content-Length header: {exc}"})
+                        raise _MalformedBody from exc
+                else:
+                    length = 0
+
                 if length > max_bytes:
                     # Drain (without buffering) so the client can finish
                     # writing and read our 413, instead of dying on a broken
--- a/src/continuum/gateway.py
+++ b/src/continuum/gateway.py
@@ -259,7 +259,13 @@ def match_route(
             if getattr(storage, "supports_action_index", False):
                 foreign_action = storage.foreign_action(key, exclude_run=run_id)
                 if foreign_action is not None:
-                    action = foreign_action
+                    return Decision(
+                        False,
+                        f"side effect {route.action_type!r} key {rendered!r} "
+                        f"already has a claim in another run "
+                        f"({foreign_action.status.value}); reconcile it first",
+                        route=route,
+                    )
         except Exception:
             action = None
     else:
--- a/src/continuum/gateway.py
+++ b/src/continuum/gateway.py
@@ -337,18 +337,28 @@ class GatewayServer:
                 decode (issue #323). Callers catch both and return, since the
                 response is already on the wire.
                 """
-                te = self.headers.get("Transfer-Encoding", "").lower()
-                if "chunked" in te:
-                    self._respond(400, {"error": "chunked transfer encoding is not supported"})
+                transfer_encodings = self.headers.get_all("Transfer-Encoding", [])
+                if transfer_encodings:
+                    self.close_connection = True
+                    self._respond(400, {"error": "transfer encoding is not supported"})
                     raise _MalformedBody
 
-                cl_header = self.headers.get("Content-Length")
+                content_lengths = self.headers.get_all("Content-Length", [])
+                if len(content_lengths) > 1:
+                    self.close_connection = True
+                    self._respond(
+                        400, {"error": "multiple Content-Length headers are not supported"}
+                    )
+                    raise _MalformedBody
+
+                cl_header = content_lengths[0] if content_lengths else None
                 if cl_header is not None:
                     try:
                         length = int(cl_header)
                         if length < 0:
                             raise ValueError("Content-Length must be non-negative")
                     except ValueError as exc:
+                        self.close_connection = True
                         self._respond(400, {"error": f"malformed Content-Length header: {exc}"})
                         raise _MalformedBody from exc
                 else:
--- a/tests/test_dashboard.py
+++ b/tests/test_dashboard.py
@@ -118,6 +118,34 @@ def test_dashboard_post_body_too_large_returns_413(tmp_path: Path) -> None:
         conn.close()
         assert resp.status == 200
         assert "goal and progress confirmed" in data
+
+        # Malformed Content-Length returns 400 without crashing (#522)
+        conn = http.client.HTTPConnection(addr, timeout=10)
+        conn.request(
+            "POST",
+            "/action/confirm",
+            body=b"x",
+            headers={"Content-Length": "abc"},
+        )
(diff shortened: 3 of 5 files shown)


Cyrax321 (changes requested):
Hi @vjymisal0 ,  thank you for taking on #522, and for your first contribution to CONTINUUM. The follow-up consistency between gateway and dashboard is exactly what the issue asked for, and you have the core logic right.                                                                                                                                                                                                      
                                                                                                                                                                                                                   
   ### Assessment                                                                                                                                                                                                  
                                                                                                                                                                                                                   
   The hardening is in the right direction. Both `src/continuum/gateway.py` and `src/continuum/dashboard/app.py` now handle malformed `Content-Length` and `Transfer-Encoding: chunked` with a 400 instead of      
 crashing or accepting an empty body, and the added tests cover the two repros from #522. That closes the crash and the chunked bypass left after #317.                                                            
                                                                                                                                                                                                                   
   I verified your branch `470c358` live against ephemeral servers with raw sockets. Two blocking items remain before merge - both flagged by automated review and reproducible. They are small fixes on top of    
 correct work.                                                                                                                                                                                                     
                                                                                                                                                                                                                   
   ### Required changes                                                                                                                                                                                            
                                                                                                                                                                                                                   
   **1. Close the connection after 400 that leaves the body unread**                                                                                                                                               
                                                                                                                                                                                                                   
   `GatewayServer.Handler` is `HTTP/1.1` (keep-alive). The new 400 paths do not drain and do not set `self.close_connection`. Pipelining reproduces desync:                                                        
                                                                                                                                                                                                                   
 ```                                                                                                                                                                                                               
                                                                                                                                                                                                                   
 POST /test Content-Length: abc\r\n\r\nx  -> 400 without Connection: close                                                                                                                                         
 POST /test Content-Length: 0\r\n\r\n   -> 501 Unsupported method ('xPOST')                                                                                                                                        
                                                                                                                                                                                                                   
 ```                                                                                                                                                                                                               
                                                                                                                                                                                                                   
   Leftover `x` plus the next `POST` is parsed as `xPOST`. `http.client` also shows `400` with `Connection: None` for chunked. The existing 413 path already closes or drains for this reason. With untrusted      
 length we cannot drain, so we must close.                                                                                                                                                                         
                                                                                                                                                                                                                   
   Please set `self.close_connection = True` before each new 400 in both handlers. Dashboard is `HTTP/1.0` by default and already closes, but setting the flag explicitly is correct for `Connection: keep-alive`  
 clients.                                                                                                                                                                                                          
                                                                                                                                                                                                                   
   ```python                                                                                                                                                                                                       
   transfer_encodings = self.headers.get_all("Transfer-Encoding", [])                                                                                                                                              
   if any("chunked" in v.lower() for v in transfer_encodings):                                                                                                                                                     
       self.close_connection = True                                                                                                                                                                                
       self._respond(400, {"error": "chunked transfer encoding is not supported"})                                                                                                                                 
       raise _MalformedBody                                                                                                                                                                                        
                                                                                                                                                                                                                   
   content_lengths = self.headers.get_all("Content-Length", [])                                                                                                                                                    
   if len(content_lengths) > 1:                                                                                                                                                                                    
       self.close_connection = True                                                                                                                                                                                
       self._respond(400, {"error": "multiple Content-Length headers are not supported"})                                                                                                                          
       raise _MalformedBody                                                                                                                                                                                        
   cl_header = content_lengths[0] if content_lengths else None                                                                                                                                                     
   if cl_header is not None:                                                                                                                                                                                       
       try:                                                                                                                                                                                                        
           length = int(cl_header)                                                                                                                                                                                 
           if length < 0:                                                                                                                                                                                          
               raise ValueError("Content-Length must be non-negative")                                                                                                                                             
       except ValueError as exc:                                                                                                                                                                                   
           self.close_connection = True                                                                                                                                                                            
           self._respond(400, {"error": f"malformed Content-Length header: {exc}"})                                                                                                                                
           raise _MalformedBody from exc                                                                                                                                                                           
 ```                                                                                                                                                                                                               
                                                                                                                                                                                                                   
 Use _html + return in dashboard/app.py. After the patch the pipeline gives single 400 Connection: close and no second response.                                                                                   
                                                                                                                                                                                                                   
 2. Reject ambiguous framing via get_all                                                                                                                                                                           
                                                                                                                                                                                                                   
 HTTPMessage.get() returns only the first value, so duplicate headers bypass the current check:                                                                                                                    
                                                                                                                                                                                                                   
 ```                                                                                                                                                                                                               
   Transfer-Encoding: gzip + chunked -> get() = 'gzip' -> "chunked" not found -> got 403 not 400                                                                                                                   
   Content-Length: 4 + 9 -> only first parsed -> bypass                                                                                                                                                            
 ```                                                                                                                                                                                                               
                                                                                                                                                                                                                   
 The code above already fixes this by checking all values with get_all() and rejecting len(content_lengths) > 1. Please keep this. Both orders (gzip+chunked and chunked+gzip) then correctly return 400.          
                                                                                                                                                                                                                   
 3. Housekeeping                                                                                                                                                                                                   
                                                                                                                                                                                                                   
 CI is red on Lint & Type-check for ruff format --check in tests/test_gateway.py:290,308. Please run ruff format src/ tests/ examples/ and push. The branch is also out-of-date with main - please run Update      
 branch before pushing.                                                                                                                                                                                            
                                                                                                                                                                                                                   
 ### Verification                                                                                                                                                                                                  
                                                                                                                                                                                                                   
 Before fix: pipelined malformed CL gives 400 then 501 xPOST (smuggling), gzip+chunked gives 403 (bypass).                                                                                                         
 After fix: all malformed, negative, chunked, duplicate CL and duplicate TE cases give 400 Connection: close, single-header cases still pass, existing ubuntu/macos tests remain green.                            
                                                                                                                                                                                                                   
Please push to this same branch, no new PR needed. I will re-review promptly. The core work is done; these are five lines of hardening to land it.

Cyrax321:
Duplicate of #529 which was merged with thorough drain handling and comma-list support. Closing as duplicate of the fix for #522.

Cyrax321:
Reopened, since you were first on #522, I’ve reverted my #529 so this PR remains the landing PR for the fix.

Please address the two blocking review items and push the changes to this same branch. No new PR is needed. @vjymisal0

Cyrax321:
Hi @vjymisal0,  thank you for being first on #522.

I merged my own #529 to get the hardening in quickly, but you deserve the land. I have **reverted #529** () and **reopened this PR**,  it is now the sole PR for #522 and reserved for you.

**What to fix:  push to this same branch , no new PR needed:**

**1. Close the connection after 400 that leaves body unread (smuggling fix)**
Gateway is HTTP/1.1 keep-alive. Your new 400 paths did not set `self.close_connection = True`, so pipelined attack reproduces:
```
POST / Content-Length: abc -> 400
POST / Content-Length: 0 -> 501 xPOST (leftover x + next POST parsed as xPOST)
```
Please set `self.close_connection = True` before each new 400 in BOTH handlers. Dashboard is HTTP/1.0 but set it explicitly for keep-alive clients:
```python
transfer_encodings = self.headers.get_all("Transfer-Encoding", [])
if any("chunked" in v.lower() for v in transfer_encodings):
    self.close_connection = True
    self._respond(400, {"error": "chunked transfer encoding is not supported"})
    raise _MalformedBody

content_lengths = self.headers.get_all("Content-Length", [])
if len(content_lengths) > 1:
    self.close_connection = True
    self._respond(400, {"error": "multiple Content-Length headers are not supported"})
    raise _MalformedBody
cl_header = content_lengths[0] if content_lengths else None
if cl_header is not None:
    try:
        length = int(cl_header)
        if length < 0:
            raise ValueError("Content-Length must be non-negative")
    except ValueError as exc:
        self.close_connection = True
        self._respond(400, {"error": f"malformed Content-Length header: {exc}"})
        raise _MalformedBody from exc
```
Use `_html + return` in `dashboard/app.py`. After fix pipelining gives single 400 `Connection: close`.

**2. Reject ambiguous framing via get_all (CWE-444)**
`HTTPMessage.get()` returns only first value, so duplicate headers bypass your check:
`Transfer-Encoding: gzip + chunked -> get() = gzip -> got 403 not 400`
The snippet above already fixes this.

**Housekeeping:**
- `ruff format src/ tests/ examples/` :  CI was red at `tests/test_gateway.py:290,308`
- Click `Update branch` on this PR before pushing (branch is behind `main` after revert)

**Verification you can run locally:**
Before fix: malformed CL gives 400 then 501 xPOST, gzip+chunked gives 403
After fix: all malformed/negative/chunked/duplicate give 400 Connection: close, small body still 200

I will re-review promptly and merge this PR to close #522. Your work is exactly what the issue asked for, these are 5 lines of hardening to land it. Thanks again for your first contribution!

Cyrax321:
Revert commit is 956bc92e9609856b9e88cb086031644ff43e6c1e on main,  your branch is now directly mergeable after Update branch. Issue #522 is reserved for you, no competing PR will be opened.

Adhi1-2:
Heads-up from outside this PR, filed as #611 so it does not block you here.

The three refusal branches this PR adds to `gateway.py::_body` answer 400 without
draining and without closing. `Handler.protocol_version = "HTTP/1.1"`
(`gateway.py:320`), so the connection stays alive and whatever the client already
wrote is read as the next request line. I built a worktree at this PR's head
(`690abdb`) and put a complete HTTP request inside the body of a refused one, sent
in one write on one socket:

```
[malformed Content-Length]  400 Bad Request  then  403 Forbidden
[negative Content-Length]   400 Bad Request  then  403 Forbidden
[chunked, raw body bytes]   400 Bad Request  then  403 Forbidden
[control, sent alone]                              403 Forbidden
```

The second response matches the control exactly, so the body was dispatched
through the gate rather than discarded. Even without aligning the boundary the
desync shows: a body of `{"id":"Y2"}` yields
`501 Unsupported method ('{"id":"Y2"}POST')`, and a chunk-size line yields
`400 Bad request syntax ('5')`.

The dashboard hunk has the same shape but is not exposed today, because it sets no
`protocol_version` and answers HTTP/1.0 with `close_connection` left `True`.

`self.close_connection = True` before each of the three `_respond(400, ...)` calls
is the whole fix, matching what the drain-limit path in the same method already
does at `gateway.py:352`. Refusals whose boundary *is* known should stay
keep-alive; an oversize `Content-Length` drained in full is a correct 413 on a
live socket. That is the same line #577 drew for `src/continuum/serve/server.py`
after @Cyrax321's review raised it, and their review of #577 says outright that
the gateway hardening has the gap too.

Existing gateway tests all send `Connection: close`, which is why this is
invisible in CI. `tests/test_serve_http.py` has the pipelining shape and ports
across.

Entirely your call whether to fold it in here or take #611 as a follow-up. I have
no changes to this PR to request, and #611 has the reproduction if it is easier to
handle separately.

vjymisal0:
Addressed the remaining review cleanup on the same branch in e850823: simplified authority event extraction to one object/dict branch, removed the redundant no-op else and stale lookup comments. Ruff format/check and git diff --check pass. The full pytest suite is blocked locally by an unrelated environment dependency error (pytest_httpbin imports Flask, which is not installed). Please re-review the updated branch.

vjymisal0:
The requested malformed-request connection handling is now included in the latest branch commits (including e4fd6fc and follow-up refactors). Could you please re-review the updated head when convenient?

Cyrax321 (approved):

I reviewed the framing changes against #522 and the surrounding handler code. The implementation is correct: both servers validate with `get_all`, reject duplicate `Content-Length` and any `Transfer-Encoding`, return 400 on malformed values, and set `close_connection` before answering so an unread body cannot desync the connection. The added tests cover the malformed, duplicate, and pipelined cases, and `tests/test_gateway.py`, `tests/test_dashboard.py`, and `tests/test_memory_forensic.py` pass. The foreign claim denial fails closed with direct coverage. It is outside the original issue scope, so note it in the PR description for the record.

No blocking issues remain. Pending green CI, this is acceptable to merge. Appreciate the persistence through the revert and review rounds.

PR #542, opened 2026-08-30
docs(cli): show the --db default and a --json example that runs

abyyxhek:
## Description

Two docs-only fixes in `docs/api/cli.md`, both one-liners in the issues but one
of them turned out to document a command that does not work.

**#510, the `--db` default.** The header read `continuum --db continuum.db
<command> [args]`, which shows the value without saying it is the default, so
`--db` reads as required. It now mirrors the `--help` text and shows the form
that omits the flag:

```bash
continuum <command> [args]                    # storage defaults to ./continuum.db
continuum --db <url-or-path> <command>        # storage URL or path (default: continuum.db)
continuum --json <command>                    # machine-readable output
```

`continuum --help` prints `--db DB storage URL or path (default: continuum.db).`
and `src/continuum/cli/main.py:88` sets `_DEFAULT_DB = "continuum.db"`, so the
header now matches both. The CLI reads no environment variable for this (only
the MCP server and `serve` consult `CONTINUUM_DB`), so the doc does not claim
one.

**#508, the missing `--json` example.** The Examples block had no `--json`
invocation. The issue suggests `continuum runs --json | jq .`; that form does
not work, and the note under the block said something close to it:

> Most commands accept `--db` (storage path or URL) and `--json`
> (machine-readable output, also available as `continuum <command> --help` for
> each subcommand).

`--json` is a global flag, parsed before the subcommand. Verified against the
built CLI:

```console
$ continuum runs --json
usage: continuum [-h] [--version] [--db DB] [--json] [--color | --no-color] COMMAND ...
continuum: error: unrecognized arguments: --json

$ continuum --json runs
{
  "runs": []
}
```

So the example uses the working order, and the note now says which order is
which instead of implying the flag belongs to each subcommand:

```bash
# Same data, machine-readable: --json goes before the command, not after it
continuum --json runs | jq '.runs[] | {run_id, status}'
continuum --json resume run_42 | jq '{safe, mode}'
```

The note keeps "most commands emit JSON with it" rather than "every command",
because that is what the CLI does: `runs`, `inspect`, `history`, `events`,
`resume`, `budget`, `tree`, `actions`, `verify`, `validate`, `show-contract` and
`replay` emit JSON, while `status` and `health` print their human text either
way. Whether those two should also honour it is a separate question and not
touched here.

## Related issues

Fixes #510
Fixes #508

## Types of changes

- Type: `docs`

## Breaking changes

No. Documentation only; no source, test or CI file is touched.

## How has this been tested?

[...]

PR #544, opened 2026-08-30
fix(cli): honour tree --limit instead of accepting and ignoring it

### LINKED ISSUE #321: Document or unhide --limit on continuum tree
## Context
`continuum tree` has a flag that help hides.

`--limit` is defined with `help=argparse.SUPPRESS` in `src/continuum/cli/main.py:1957`, so `continuum tree --help` shows only `run_id`. But the flag is not actually wired through to the implementation. Open `cmd_tree` and you will see `args.limit` is never used.

## What a newcomer sees
They try to page large families, cannot find the flag, and think the feature is missing.

## Decision needed, then docs
Pick one and do it. Both are fine, just be consistent:

**Option A, wire it:** make `children_of` respect limit, add help like `help="limit children shown per parent (default: all)"`, and add a note to `docs/api/cli.md`.

**Option B, remove it:** delete the `add_argument("--limit", ...)` line if it was left from an experiment.

## How to verify
```bash
continuum tree --help        # should now show --limit with help, or not list it at all
continuum tree <parent> --limit 2  # should show at most 2 children if you kept it
```

## Files
- `src/continuum/cli/main.py`
- `docs/api/cli.md`

Small scoped choice, good for a first PR that involves reading CLI wiring.

abyyxhek:
## Description

`continuum tree` registered `--limit` with `help=argparse.SUPPRESS` and then
never read it, so the flag was accepted, hidden from `--help`, and printed every
child anyway. `continuum tree boss --limit 1` on `main` exits 0 and lists all
three children.

`cmd_tree` now truncates the printed child list to the newest `n` children and
says how many it hid:

```text
boss  [resume, safe=True]  supervise
  ok kid2  [resume, uncertain=0]  work 2
  ... 2 of 3 children hidden by --limit 1; run without it to see the whole family
```

The hidden count is the part that matters. A truncated tree that says nothing is
indistinguishable from a small family, which is exactly the wrong impression for
a command whose job is showing you what a supervisor is responsible for. With
`--json`, `children_total` and `children_hidden` carry the same fact for
scripts, added alongside the existing `children` list so nothing that reads the
payload today breaks. Output without `--limit` is unchanged.

A limit below `1` is refused with exit code 1 rather than clamped: `--limit 0`
would render a family that has children as childless, which is the one output
this command must never produce.

**One deliberate deviation from the issue.** The issue suggests wiring the limit
into `children_of`. I did not, because `roll_up_children` calls `children_of` to
decide whether a family blocks `resume` — truncating there would let an
uncertain child fall off the end of a safety decision and report a blocked
family as safe. The truncation is display-only, and there is a test that pins
this: an uncertain child hidden from the tree by `--limit 1` still forces
`resume` into `request_human`.

## Related issues

Fixes #321

## Types of changes

- Type: `bugfix`

## Breaking changes

No. Behaviour is unchanged without `--limit`; the JSON payload gains two keys
and loses none.

## How has this been tested?

Windows 11, Python 3.11.13 (project `.venv`), from the repo root:

```bash
python -m pytest tests/test_family.py -q          # 14 passed
python -m pytest -q -p no:randomly                # full suite green, exit 0
python -m ruff check src tests                    # All checks passed!
python -m ruff format --check src tests           # 246 files already formatted
python -m mypy src/continuum                      # no issues in 114 source files
npx markdownlint-cli2 CHANGELOG.md docs/api/cli.md # 0 issues
```

Seven new tests in `tests/test_family.py` cover truncation and its report, the
unlimited path hiding nothing, the JSON totals, `--limit 0` and `--limit -1`
being refused, family safety surviving truncation, and `--limit` appearing in
`tree --help`.

All seven were checked against the baseline before the fix: `upstream/main`
(`e129f91`) in a separate `git worktree` with `PYTHONPATH` pointed at it (
`continuum.__file__` confirmed to resolve into the worktree), where all seven
fail. The `--help` one fails with `assert '--limit' in 'usage: continuum tree
[-h] run_id ...'`, which is the suppressed-flag symptom itself.

Also exercised by hand end to end (`init`, `start` a parent and three children,
then `tree --limit 1`, `tree`, `--json tree --limit 2`, `tree --limit 0`) to
confirm the ordering claim in the help text: `list_runs` orders
`created_at DESC, run_id DESC`, so the children shown are genuinely the newest.

## Checklist

Tests added for the changed behaviour and verified to fail at the baseline
commit. `docs/api/cli.md` documents the flag, its display-only scope, the JSON
keys, and the sub-1 refusal; `CHANGELOG.md` has an entry under
`[Unreleased] / Fixed`. Lint, format, types and the full suite pass locally. No
security-relevant surface is touched; the one safety-adjacent question (whether
truncation can weaken the family roll-up) is answered by
`test_tree_limit_cannot_hide_a_child_from_family_safety`.

## Additional context

Branched from `upstream/main` at `e129f91`, so this merges cleanly and does not
touch the regions changed by the recently merged #536 or #542.

The alternative in the issue was removing the flag instead of wiring it. Wiring
it is the better trade here: the flag is already in released argv and a wide
fork tree is genuinely hard to read, so the capability is worth having as long
as the truncation announces itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,20 @@ All notable changes to this project are documented here. The format follows
   as before. A settings file that already holds duplicates is cleaned by
   `hooks remove` followed by `hooks install`.
 
+- **`continuum tree --limit` truncates the child list instead of being ignored
+  (#321).** The flag was registered with `help=argparse.SUPPRESS` and never
+  read, so `--limit 5` was accepted and printed every child anyway: an operator
+  with a wide family got the full wall of output and no error saying why. The
+  flag now shows the newest `n` children and reports what it hid
+  (`... 3 of 40 children hidden by --limit 5`), because a truncated tree that
+  says nothing is indistinguishable from a small family. `--json` gained
+  `children_total` and `children_hidden` alongside the truncated `children`
+  list; output without `--limit` is unchanged. The truncation is display-only:
+  `children_of` still returns every child, so the family safety roll-up behind
+  `resume` cannot lose an uncertain child that scrolled off the printed list.
+  A limit below `1` is refused with exit code 1 rather than clamped, since
+  `--limit 0` would render a family with children as childless.
+
 ## [0.1.0] - 2026-08-27
 
 ### Added
--- a/docs/api/cli.md
+++ b/docs/api/cli.md
@@ -27,7 +27,7 @@ continuum --json <command>                    # machine-readable output
 | `confirm <run_id>` | Confirm a human-approved recovery step. |
 | `complete <run_id>` | Close a run as done. Mutates storage. |
 | `budget <run_id>` | Retry-budget usage per action type. |
-| `tree <run_id>` | Show a parent run and its children. |
+| `tree <run_id> [--limit <n>]` | Show a parent run and its children. `--limit` shows only the newest `n` children. |
 | `fork <run_id> --reason <text>` | Approve a divergent continuation as a child run. Mutates storage. |
 | `compact <run_id>` | Archive the pre-anchor log prefix. Mutates storage. |
 | `checkpoint <run_id>` | Force a state checkpoint. |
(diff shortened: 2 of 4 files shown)


Cyrax321 (approved):
Hi @abyyxhek, thank you for your contribution to CONTINUUM.                                                                                                                                                    
                                                                                                                                                                                                                  
   Fix for #321 is precise and complete. The `--limit` flag was accepted but ignored and hidden from help. Your implementation makes it display only, shows the newest children, reports how many were hidden, and keeps the family safety roll up reading every child so a hidden uncertain child still blocks resume. JSON adds `children_total` and `children_hidden` without breaking existing consumers, and values below  
 1 are correctly refused. Documentation and tests are thorough.                                                                                                                                                   
                                                                                                                                                                                                                  
   Approved. Excellent work.

PR #577, opened 2026-09-01
fix(serve): fail closed on HTTP body framing (#533)

### LINKED ISSUE #533: Follow-up to #522: handle malformed Content-Length and chunked Transfer-Encoding in serve HTTP transport
## Context

After #522/PR #529, both `src/continuum/gateway.py` and `src/continuum/dashboard/app.py` validate `Content-Length` strictly and reject `Transfer-Encoding: chunked` consistently, draining up to 256MB before returning 400/413. The same hardening was not applied to the third HTTP surface in the repo.

## Root cause verified on main

- `src/continuum/serve/server.py:335` does `length = int(self.headers.get("Content-Length") or 0)` with no try/except. A header like `Content-Length: abc` raises `ValueError` and the handler crashes with no response, same pattern that #522 fixed in gateway and dashboard.
- `SidecarHTTP.Handler.do_POST` does not check `Transfer-Encoding: chunked`. When a client omits `Content-Length` and sends a chunked body, `length` is read as 0 and the body is treated as empty (`b"{}"`), so an unbounded chunked body bypasses any cap. The gateway and dashboard now reject chunked with 400 after draining up to 256MB, but the serve transport does not.

The serve transport is reachable at `127.0.0.1:8765` by default and exposes the same durability plane as the other servers, so fail-closed handling should be consistent.

## Repro

Serve HTTP transport (default bind 127.0.0.1):

```bash
# Start serve HTTP in one terminal
python -m continuum serve --http --port 8766 &
# Malformed length (before fix: connection closed, no HTTP response)
printf 'POST /record_progress HTTP/1.1\r\nHost: localhost\r\nContent-Length: abc\r\nContent-Type: application/json\r\nConnection: close\r\n\r\n{}' | nc 127.0.0.1 8766 -i
# Chunked bypass (before fix: treated as empty body, reaches dispatch)
printf 'POST /record_progress HTTP/1.1\r\nHost: localhost\r\nTransfer-Encoding: chunked\r\nContent-Type: application/json\r\nConnection: close\r\n\r\n5\r\nhello\r\n0\r\n\r\n' | nc 127.0.0.1 8766 -i
```

## Expected behaviour

Same as #522 for the other two servers, keep them consistent:

- If `Content-Length` is present but not a valid non-negative int, respond 400 with a clear error without crashing, and keep connection handling consistent.
- If `Transfer-Encoding` contains `chunked` (case-insensitive, comma list), enforce the same caps while reading the chunked stream or reject with 400, and respect a bounded drain (256MB, same as gateway/dashboard) even for chunked. Choose one behaviour but keep all three servers identical.

Do not change happy path (valid Content-Length 0 or small JSON body).

## Search for duplicates

Checked `gh issue list --state open` and `--state closed` for Content-Length, chunked, 413, drain, and serve. Found #317 (dashboard cap, now fixed), #522 (gateway and dashboard, now fixed in PR #529), and #323 (gateway malformed JSON), but nothing about malformed Content-Length or chunked handling in `src/continuum/serve/server.py`. No open PR covers the serve transport.

## Files

- `src/continuum/serve/server.py` (SidecarHTTP.Handler.do_POST, length parsing and Transfer-Encoding check)
- Tests: `tests/test_serve_http.py` (or similar, add coverage for malformed, negative, chunked, small body 200, large body 413)

## References

- Follow-up to #522, PR #529 (gateway and dashboard hardening)
- Serve reference: `src/continuum/serve/server.py:335` (`int(self.headers.get("Content-Length") or 0)`)
- Gateway reference: `src/continuum/gateway.py:68` (MAX_BODY_BYTES, DRAIN_LIMIT_BYTES) and dashboard fix in #317

Adhi1-2:
Closes #533.

## The two holes

`SidecarHTTP.Handler.do_POST` read the body as `int(self.headers.get("Content-Length") or 0)`:

1. **Malformed length crashes the handler unanswered.** `Content-Length: abc` raised `ValueError` out of `do_POST`, so the connection closed with no HTTP response at all. From the caller's side a refused request is indistinguishable from a dead sidecar. Only the one connection was lost (each is handled on its own thread), which is what kept it quiet.
2. **Chunked bodies bypassed the caps.** Nothing looked at `Transfer-Encoding`. A client that omitted `Content-Length` and sent a chunked body had `length` read as `0`, so the request was dispatched as if its body were empty (`b"{}"`) while the stream itself went unbounded. The serve transport also had no body cap of its own, unlike the other two HTTP surfaces.

The transport binds `127.0.0.1:8765` by default and exposes the same durability plane as the gateway and dashboard, so it should fail closed the same way.

## What changed

Body reading moved into `_read_body`, which answers rather than raises:

| input | before | after |
|---|---|---|
| `Content-Length: abc` / `-1` / `1.5` / `0x10` / blank / `12abc` | connection closed, no response | `400 invalid Content-Length: ...` |
| `Transfer-Encoding: chunked` (no length) | dispatched as an empty body, stream unbounded | `400 chunked Transfer-Encoding is not supported`, after a bounded drain |
| chunked body whose framing does not parse | as above | `400 invalid chunked Transfer-Encoding` |
| drain bound crossed | n/a | `413` + `Connection: close` |
| `Content-Length` over the cap | body read in full | `413 request body exceeds 1048576 bytes`, judged from the header |
| absent `Content-Length`, or `0` | dispatches `{}` | unchanged |
| valid small JSON body | 200 | unchanged |

Notes on the choices:

- **Reject chunked, do not decode it.** Decoding means reading unbounded framing into memory, which is the denial of service the cap exists to stop, and no sidecar client needs streaming to send a run id and two counters. `chunked` is matched case-insensitively anywhere in the comma list.
- **Drain before refusing**, up to `SIDECAR_DRAIN_LIMIT_BYTES` (256 MB, the gateway and dashboard's bound), so the client finishes writing and reads the refusal instead of dying on a broken pipe mid-send. #533 asks for the bounded drain explicitly.
- **`MAX_SIDECAR_BODY_BYTES` is 1 MB**, matching the dashboard rather than the gateway's 10 MB: neither surface forwards a caller's payload anywhere, and this one dispatches small JSON control messages.
- **Validated with `isdigit`, not by catching `int()`.** That also refuses `+5` and non-ASCII digit forms, which `int()` would have accepted as a length. A header that is present but blank is malformed too, not zero.
- **`_json` now sends `Connection: close`** when the handler has set it, which it did not before, so a 413 that gives up on draining is not followed by a keep-alive the server will not honour.

## Scope

Only `src/continuum/serve/server.py` and its tests. `gateway.py` and `dashboard/app.py` are deliberately untouched: #522 is reserved for @vjymisal0 via #525, per the maintainer note on that issue after #529 was reverted in 956bc92. This PR picks the behaviour #533 specifies so the three surfaces converge once #525 lands.

## Tests

Ten tests (22 cases with parametrisation) in `tests/test_serve_http.py`, spoken on a raw socket because urllib reframes the body and rewrites `Content-Length`:

- six malformed lengths are answered, not dropped
- the server still serves a normal request afterwards
- four `Transfer-Encoding` spellings are refused with 400 rather than answered `403 missing secret`, which is exactly what the bypass looked like
- unparseable chunk framing is 400
- oversize length is 413 from the header alone
- a valid small body still round-trips 200 on the socket, and no `Content-Length` at all still reaches dispatch

Verified the 12 new failure cases fail on the parent commit and pass with the fix.

```
1816 passed, 23 skipped
ruff check: All checks passed    ruff format: 2 files already formatted
mypy --strict: Success, no issues found in 114 source files
```

Refs #533, follows #317 and #522.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP request-body validation for invalid or missing framing information.
  * Oversized request bodies now receive a clear `413` response.
  * Unsupported chunked requests and malformed content lengths now return `400`.
  * Untrusted or incomplete request boundaries now close the connection to prevent request smuggling.
  * Invalid or incomplete request bodies no longer trigger unintended database writes.
  * The service continues operating reliably after malformed requests.


src/continuum/serve/server.py
```diff
@@ -325,15 +348,131 @@ def log_message(self, *args: Any) -> None:  # silence
...
+                    # run id and a counter (#533).
+                    try:
+                        outcome = self._drain_chunked()
+                    except OSError:
+                        outcome = "malformed"
+                    if outcome == "unbounded":
+                        self.close_connection = True
+                        self._json(413, {"error": "request body too large to drain"})
+                    elif outcome == "malformed":
+                        self._json(400, {"error": "invalid chunked Transfer-Encoding"})
```
Cyrax321, line 475: Security and correctness: malformed chunk path does not set close_connection. protocol_version is HTTP/1.1 so the connection stays alive unless closed. _drain_chunked has bailed mid-stream at unparseable framing, so the body boundary is unknown and remaining bytes will be parsed as the next request on a keep-alive connection, which is a desync and a request smuggling class issue. All current tests send Connection: close so this stays hidden. Set self.close_connection = True before the 400, matching the unbounded branch at line 432.

src/continuum/serve/server.py
```diff
@@ -325,15 +348,131 @@ def log_message(self, *args: Any) -> None:  # silence
...
+                raw_length = self.headers.get("Content-Length")
+                if raw_length is None:
+                    return b"{}"
+                # Answered, not raised: int() on a malformed header used to
+                # escape the handler and close the connection with no response
+                # at all, which reads to the caller as a network fault (#533).
+                # A header that is present but blank is malformed too, not zero.
+                token = raw_length.strip()
+                if not (token.isascii() and token.isdigit()):
+                    self._json(400, {"error": f"invalid Content-Length: {raw_length!r}"})
```
Cyrax321, line 497: Same keep-alive desync on invalid Content-Length. This branch answers 400 without draining or closing. Because the length cannot be trusted, the body cannot be discarded safely and any body bytes already sent will be treated as a pipelined request. Set self.close_connection = True before the 400, matching the unbounded handling in the oversize path. The gateway hardening in 6c953fb has the same gap, but for serve this should be closed since the framing is not trustable.

Cyrax321 (changes requested):
Ref: #577 / close #533
Surface: src/continuum/serve/server.py SidecarHTTP.Handler, tests/test_serve_http.py

What is the bug being fixed: SidecarHTTP.Handler.do_POST read Content-Length with int(headers.get(Content-Length) or 0). Malformed values raised ValueError and closed the connection with no HTTP response, indistinguishable from a dead sidecar. Transfer-Encoding was ignored, so a client omitting Content-Length and sending chunked had length 0, dispatched as empty JSON while the chunked stream went unbounded past any cap. The transport binds 127.0.0.1:8765 and exposes the same durability plane as gateway and dashboard, so it should fail closed the same way.

Can the root cause be identified: Yes. Direct code path in origin/main:335 int() with no try except, and no Transfer-Encoding check. Confidence clear, reproduced via raw socket and verified as fixed in this PR.

Provenance: The pattern was introduced with SidecarHTTP and carried forward. Hardening for the other two surfaces landed in 6c953fb and fb731cf as part of 522, then reverted in 956bc92. This PR correctly scopes only to serve per 533 and the maintainer note that 522 is reserved for 525, so no overlap with gateway or dashboard.

Is the fix the best available: Yes for the stated scope. Body reading moves to _read_body which answers rather than raises, with MAX_SIDECAR_BODY_BYTES at 1 MB matching the dashboard, and SIDECAR_DRAIN_LIMIT_BYTES at 256 MB matching gateway and dashboard, so all three surfaces converge once 525 lands. Rejecting chunked rather than decoding is correct, decoding would buffer unbounded framing and no sidecar client needs streaming for a run id and counters. Using isdigit with isascii correctly rejects blank, negative, float, hex, plus prefix and non ASCII digits that int() would accept. _json now honours close_connection, and _drain_chunked correctly handles hex, extensions after semicolon, zero chunk trailers and bounded draining.

Verified proof: Local checkout of pr-577, 22 tests in test_serve_http.py all pass. Six malformed Content-Length values now return 400 rather than dropping the connection, and the server still serves a subsequent request. Four Transfer-Encoding variants including case differences and comma lists now return 400 rather than 403 for missing secret, which was the bypass signal. Malformed chunk framing returns 400, oversize Content-Length returns 413 from the header with drain, valid small JSON still returns 200, absent Content-Length still dispatches as empty object and hits auth. CI shows 20 successful checks including all platform matrix, lint and mypy strict clean. Codecov patch at 68.23 percent is below target 87.77, the missing lines are the unbounded and error branches that are not exercised with Connection: close tests, not a functional gap but worth noting.

What remains risky or unverified: Two inline findings are blocking for a persistent connection. See inline comments on src/continuum/serve/server.py:436 and :450. On HTTP/1.1 the connection stays alive by default. The invalid Content-Length branch and the malformed chunk branch both answer 400 without setting close_connection and without fully consuming the body. Any bytes already sent remain in the socket buffer and are parsed as the next request, which is a desync and a request smuggling class issue. Existing tests all send Connection: close so they do not exercise this. Gateway hardening in 6c953fb has the same gap, but for serve where the length is not trustable the correct action is to close. Please set self.close_connection = True before the 400 in both branches, matching what the unbounded paths already do. A keep-alive regression test that pipelines two requests on one socket and asserts the second is not misframed would also lock this.

Architecture: No generic core pollution, changes isolated to serve. Larger refactor not needed. Scope deliberately excludes gateway and dashboard per 533 notes, which is the right call.

Security: No secret exposure or auth bypass introduced. Fix closes the chunked bypass. The remaining keep-alive desync should be treated as high severity for correctness even though the bind is localhost only, since a local multiplexer could pipeline.

Overall: Strong and well tested hardening PR that correctly closes the two holes in 533. Requesting changes only for the keep-alive close handling. Once the two close_connection lines are added, this is ready to approve. The current tests and documentation are sound.

src/continuum/serve/server.py
Adhi1-2: Done, `self.close_connection = True` now precedes the 400 here (`src/continuum/serve/server.py:454`), matching the unbounded branch above it.

Your read of the consequence was exact, and it turned out to be worse than a desync in the general case. The regression test pipelines a complete, valid, authenticated `record_progress` where the refused body's unread remainder sits, and on the old code that request was **executed**:

```
E   AssertionError: the unread remainder was answered too:
E   b'HTTP/1.1 400 Bad Request\r\n...\r\n{"error": "invalid chunked Transfer-Encoding"}
E     HTTP/1.1 200 OK\r\n...\r\n{"run_id": "smuggled-past-chunk", "completed": 1, "pending": 0,
E     "failed": 0, "total": 1, "source_sequence": 2}'
E   assert [400, 200] == [400]
```

`test_malformed_chunk_framing_closes_instead_of_smuggling` asserts one response, `Connection: close` on it, nothing trailing it on the wire, and that `smuggled-past-chunk` has no events in the database. It sends no `Connection` header at all, which is the gap you named in every existing test.

One thing your comment made me check: the `else` branch (a chunked body drained to its terminal chunk) is left keep-alive on purpose, because there the boundary really is known. That was only true by luck, though. `take(2)` consumed the chunk-terminating CRLF blindly, so a chunk without its terminator swallowed the front of the next chunk header and the drain reported a clean finish from the middle of the stream, which is the one chunked outcome that answers without closing. The drain now reads that CRLF with `readline` and returns `malformed` if it is not empty (`:421`), so `None` means the stream ended where the framing said it would. `test_a_chunk_without_its_terminating_crlf_is_malformed` covers it and also fails on the old code.

src/continuum/serve/server.py
Adhi1-2: Done, `self.close_connection = True` now precedes this 400 as well (`src/continuum/serve/server.py:473`).

Agreed on the reasoning: the length that would say how many bytes to discard is the thing that failed to parse, so there is nothing to drain by and the queued body is whatever the client chose to send. Same regression shape as the chunked one, and the same result on the old code, the smuggled request ran:

```
E   AssertionError: the unread body was answered too:
E   b'HTTP/1.1 400 Bad Request\r\n...\r\n{"error": "invalid Content-Length: \'abc\'"}
E     HTTP/1.1 200 OK\r\n...\r\n{"run_id": "smuggled-past-length", "completed": 1, "pending": 0,
E     "failed": 0, "total": 1, "source_sequence": 2}'
E   assert [400, 200] == [400]
```

`test_invalid_content_length_closes_instead_of_smuggling` pins it, with the same four assertions (one response, `Connection: close`, nothing trailing, no events for the run id).

I did not touch the gateway. You called it the same gap but scoped it out of `serve`, so it stays a separate issue rather than riding along in this diff; happy to open one if there is not one already.

The oversize path is deliberately left keep-alive, since there the declared length is drained in full and the boundary is known. `test_an_oversize_refusal_keeps_a_well_framed_connection` asserts the pipelined request behind a 413 is still served (`[413, 200]`) and reaches the database, so the fix is "close what cannot be framed", not "close everything refused".

src/continuum/serve/server.py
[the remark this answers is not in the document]
Adhi1-2: Fixed in d798ffe. Both refusals set `close_connection` before answering, so the socket is closed rather than left holding an unread body remainder.

Your assessment of severity was right, and the proof is stronger than "desync": pipelining a valid authenticated `record_progress` where the remainder sits got it dispatched on the old code (`[400, 200]` with the run written to the database, for both the malformed chunk and the invalid `Content-Length`). Three new tests fail on the pre-fix code, and they send no `Connection` header, which is why the existing suite never saw this.

The refusals that can prove where the body ends still keep the connection: an oversize `Content-Length` drained in full, and a chunked body drained to its terminal chunk. The second of those needed a real guarantee rather than an assumed one, so the drain no longer consumes the chunk-terminating CRLF blindly; a missing terminator is now `malformed` (and therefore closes) instead of resyncing onto the middle of the next chunk header.

Adhi1-2:
Review round addressed in d798ffe (plus 100f0e4 for the conflict). Summary of what changed since 1f666da.

**The blocking ask, both branches.** `self.close_connection = True` now precedes the 400 for unparseable chunk framing (`src/continuum/serve/server.py:454`) and for an untrustable `Content-Length` (`:473`), matching the unbounded drain path that already did it.

**The keep-alive regression test you asked for.** `tests/test_serve_http.py` gained a `raw_pipeline` helper that sends exact bytes on one socket with **no** `Connection` header (the default under `protocol_version = "HTTP/1.1"`) and reads every response that comes back. Behind each refusal it pipelines `pipelined_post(run_id)`: a complete, valid, authenticated `record_progress`. On the old code that request was not merely misframed, it ran:

```
E   AssertionError: the unread remainder was answered too:
E   b'HTTP/1.1 400 Bad Request\r\n...{"error": "invalid chunked Transfer-Encoding"}
E     HTTP/1.1 200 OK\r\n...{"run_id": "smuggled-past-chunk", "completed": 1, ...}'
E   assert [400, 200] == [400]
```

Each closing test asserts four things: exactly one response, `Connection: close` on it, nothing trailing it on the wire, and no events for the smuggled run id in the database. The "nothing trailing" assertion matters because a leftover line that the stdlib parser reads as HTTP/0.9 answers with **no status line at all**, so counting status lines cannot see it alone.

Verified red on the pre-fix code by stashing only `server.py` and rerunning: 3 failed, 2 passed (`test_malformed_chunk_framing_closes_instead_of_smuggling`, `test_invalid_content_length_closes_instead_of_smuggling`, `test_a_chunk_without_its_terminating_crlf_is_malformed`). The two that pass on old code are the patched-bound 413 tests, which cover branches that were already correct; they are coverage, not regression.

**One thing your comment led me to.** The `else` branch (chunked body drained to its terminal chunk) is left keep-alive on purpose, because there the boundary is known. That was only true by accident: `take(2)` consumed the chunk-terminating CRLF blindly, so a chunk with no terminator swallowed the front of the next chunk header and the drain reported a clean finish from mid-stream, which is the one chunked outcome that answers without closing. The drain now reads that CRLF with `readline` and returns `malformed` if it is not empty (`:421`). Also verified from the other side: `test_an_oversize_refusal_keeps_a_well_framed_connection` asserts a pipelined request behind a 413 is still served (`[413, 200]`) and reaches the database, so this is "close what cannot be framed", not "close everything refused".

**codecov/patch.** You read it correctly, the misses were the unbounded and error branches that `Connection: close` tests never reach. Now covered: a chunk that ends early, a blank line where a header belongs, a negative hex length (`int(b"-5", 16)` parses, so it reaches the size check rather than the `ValueError`), trailers past the bound, and chunk data past the bound (read granularity patched down with the bound so the loop takes more than one pass, rather than sending the real 256 MB). Measured locally over the diff lines of `server.py`: **88/90 = 97.8%**, up from 68.23%. The two left are the `except OSError` guard around the drain; forcing a reset mid-drain means `SO_LINGER 0` on the client and racing the server's read, which is platform-dependent, so I left it uncovered rather than add a flaky test for a two-line guard.

**Conflict.** Resolved by merging `main` (100f0e4), no rebase and no force-push. `main`'s CHANGELOG carries a stray committed `>>>>>>> 4e27e6f (fix(hooks): ...#580)` at line 146, inside the region I had to resolve, so it is gone in the merged file; that is an upstream artifact, not something this branch introduced (same one I flagged on #579).

**Gate:** 1926 passed, 23 skipped (`tests/test_serve_http.py` is 37 of them, up from 22). `ruff format --check`, `ruff check`, `mypy --strict src/` all clean. Diff is still the same three files.

Adhi1-2:
The gateway half of your review note is now filed as #611.

You wrote: *"Gateway hardening in 6c953fb has the same gap, but for serve where
the length is not trustable the correct action is to close."* I checked that
rather than take it on faith, and it is worse than the same gap. `6c953fb` was
reverted in `956bc92`, so those branches are not on `main` at all right now;
they come back with open PR #525. Against a worktree at #525's head I put a
complete HTTP request inside the body of a refused one and sent it in a single
write on one keep-alive socket:

```
[malformed Content-Length, body is a whole request]
    HTTP/1.1 400 Bad Request
    HTTP/1.1 403 Forbidden          <- the body, dispatched through the gate
[negative Content-Length, body is a whole request]
    HTTP/1.1 400 Bad Request
    HTTP/1.1 403 Forbidden
[chunked declared, body is raw request bytes]
    HTTP/1.1 400 Bad Request
    HTTP/1.1 403 Forbidden
[control: the same request sent on its own]
    HTTP/1.1 403 Forbidden
```

The second response is byte-identical to the control, so it is a dispatch and not
a parse error. Three branches rather than two: #525 adds a chunked refusal that
does not drain at all, on top of the malformed and negative `Content-Length`
paths.

The dashboard has the same two unguarded branches and is not exposed, because it
sets no `protocol_version` and so answers HTTP/1.0 with `close_connection` left
`True`. That is protection by accident. One line adopting HTTP/1.1 would make it
behave like the gateway, and #611 records that.

Filed separately rather than widened into this PR, since #533 scopes to `serve`
and #522 is reserved for #525. Both close-handling lines you asked for are in
`d798ffe`; this PR has no failing checks and is waiting only on re-review.

Cyrax321 (dismissed):
Thank you for contributing to Continuum. Re reviewing after the Sep 1 round: the two blocking desync findings are addressed. The malformed chunk and unparseable length refusals now set close_connection before answering, refusals with a known boundary stay keep alive, and the chunk terminator is read as a line so a missing CRLF reports malformed instead of resyncing mid stream. The pipelining tests prove a refused body can no longer dispatch as the next request. Serve suites pass locally and CI is green across the matrix with patch coverage now passing. No blocking issues remain. Acceptable to merge pending branch update.

Cyrax321 (approved):
Thank you for contributing to Continuum. Re reviewing after the main merge: the CHANGELOG conflict was resolved by keeping both entries with no other content changes. Serve suites pass on the resolved tree and CI is green across the matrix. No blocking issues remain. Acceptable to merge.

PR #581, opened 2026-09-01
fix(hooks): default the remove path to the client profile, like install (#580)

### LINKED ISSUE #580: hooks remove <client> crashes with TypeError when --settings is omitted
### Before reporting

- [x] I can reproduce this on the latest commit on `main` (d06fe56)

### Affected component

CLI

### Severity

### System information

```
OS: macOS (darwin 25.0.0)
Python: 3.13.5
continuum-agent: 0.1.0 from source at d06fe56
Installed via: uv pip install -e ".[dev]"
```

```shell
cd "$(mktemp -d)"
continuum hooks install claude-code   # exit 0, writes .claude/settings.json
continuum hooks remove claude-code    # traceback, exit 1
```

### Current behaviour

`hooks remove` dies with an uncaught `TypeError` before it reads anything. `--settings`
defaults to `None` for both subcommands (they share `hooks_client` in `build_parser`), and
only the installer falls back to the client profile:

```python
# cmd_hooks_install
settings_path = Path(args.settings or profile["settings"])
# cmd_hooks_remove
settings_path = Path(args.settings)
```

So the documented uninstall path is reachable only if the operator repeats by hand the path
that install worked out for them. `docs/guides/embed-claude-code.md` tells them
`continuum hooks remove claude-code` takes the hooks back out again, which is exactly the
invocation that crashes.

### Expected behaviour

Symmetry with install: `hooks remove <client>` with no `--settings` reads the same file
install wrote, that is `CLIENT_PROFILES[client]["settings"]` (`.claude/settings.json`,
`.gemini/settings.json`, `.codex/hooks.json`), removes the entries it recognises, and exits
0. A file that is not there reports that nothing was found rather than raising.

### Logs or CI run

```
  File ".../.venv/bin/continuum", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File ".../src/continuum/cli/main.py", line 3493, in main
    return int(args.func(args, None, out, err))
               ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File ".../src/continuum/cli/main.py", line 2368, in cmd_hooks_remove
    settings_path = Path(args.settings)
  File ".../pathlib/_local.py", line 503, in __init__
    super().__init__(*args)
TypeError: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'NoneType'
```

### Why the suite did not catch it

Every test of this command passes `--settings` explicitly
(`tests/test_client_installers.py`, `tests/test_cli_observe.py`,
`tests/test_cli_precompact.py`), so the default path has no coverage on the remove side.
The install side had this same bug fixed once already and pinned by
`test_default_settings_path_comes_from_the_profile`, whose docstring records the identical
cause: "an explicit --settings was tested everywhere, so a hardcoded CLI default silently
sent every client's hooks to Claude Code's settings file". The remover was never extended
with it. That makes this a sibling of #484 (installer and remover drifting apart), with the
same shape: a fact about hooks that lives in only one of the two halves.

### Adjacent, in the same function

`cmd_hooks_remove` reports "Removed observation hook from ...", but
`remove_claude_code_hook` removes every kind in `_INSTALLED_KINDS`, which is now four of
them (observe, gate, briefing, and precompact after #449). An operator who also installed
the gate or the compaction hook is told only the observation hook went, so the report
understates what the command just did to their settings. Worth correcting in the same
change.

Adhi1-2:
Fixes #580.

## The crash

```console
$ cd "$(mktemp -d)"
$ continuum hooks install claude-code   # exit 0, writes .claude/settings.json
$ continuum hooks remove claude-code
TypeError: argument should be a str or an os.PathLike object where __fspath__
returns a str, not 'NoneType'
```

Both hook subcommands share one `--settings` option, defined once in
`hooks_client`, defaulting to `None` and advertising "default: per client
profile". Only the installer honoured that promise:

```python
# cmd_hooks_install
settings_path = Path(args.settings or profile["settings"])
# cmd_hooks_remove
settings_path = Path(args.settings)
```

So `hooks install` was a one-way door for every operator who had not written
down the path install worked out for them, and `continuum hooks remove
claude-code`, the uninstall named in `docs/guides/embed-claude-code.md`, was the
invocation that crashed. The help text was already correct; the code was not.

## The fix

`cmd_hooks_remove` reads the same profile default, so the pair resolves one file
per client (`.claude/settings.json`, `.gemini/settings.json`,
`.codex/hooks.json`), and a directory that was never wired reports that nothing
was found rather than raising.

```console
$ continuum hooks remove claude-code
No CONTINUUM hooks found in .claude/settings.json
$ continuum hooks install claude-code --with-gate >/dev/null
$ continuum hooks remove claude-code
Removed CONTINUUM hooks from .claude/settings.json
```

## Second correction, same function

The report said "Removed observation hook" while `remove_claude_code_hook`
removes every kind in `_INSTALLED_KINDS`. An operator who had installed the
`--with-gate` PreToolUse hook was told only the observation hook went, and could
reasonably believe the gate still stood between their agent and an unclaimed
side effect. The text now reads `Removed CONTINUUM hooks from <path>` (`No
CONTINUUM hooks found in <path>` when there was nothing to do), and the
subcommand's `--help` line says what it removes. The `removed` boolean in
`--json` is unchanged, as is removal itself.

## Why the suite missed it

Every existing test of this command passes `--settings` explicitly
(`test_client_installers.py`, `test_cli_observe.py`), so the default path had no
coverage on the remove side. The install side had this same bug fixed once and
pinned by `test_default_settings_path_comes_from_the_profile`, whose docstring
records the identical cause: "an explicit --settings was tested everywhere, so a
hardcoded CLI default silently sent every client's hooks to Claude Code's
settings file". The remover was never extended with it, which makes this a
sibling of #484: a fact about hooks living in only one of the two halves.

The five new tests sit directly under that install test, so the two halves are
now pinned side by side:

- `test_remove_defaults_to_the_same_file_install_wrote` (parametrized over all
  three clients): install with no `--settings`, remove with no `--settings`, and
  the file install wrote is the file remove empties.
- `test_remove_without_settings_is_quiet_when_nothing_is_installed`: the absent
  file is the common case for the crash, so it has to be a reported no-op.
- `test_remove_reports_hooks_not_just_the_observation_hook`: pins the honest
  report with the gate installed.

All five fail on the unpatched command (three on the `TypeError`, two on the
text); `test_remove_cleans_each_client` passes either way because it passes
`--settings`. The kind-set assertion uses `<=` rather than `==` so it holds
whether or not a fourth kind is wired by the time this lands.

## Gate

- `ruff format --check src/ tests/`: 246 files already formatted
- `ruff check src/ tests/`: all checks passed
- `mypy --strict src/`: no issues in 114 source files
- `pytest`: **1805 passed, 23 skipped** (1800 on `main` plus 5)

No behaviour changes beyond the two above; the enforcement path, the settings
writer, and `remove_claude_code_hook` itself are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * `hooks remove` now automatically uses the selected client profile’s default settings path when none is specified.
  * Safely handles directories without installed hooks without reporting an error.
  * Removes all installed CONTINUUM hooks and accurately reports what was removed.
  * Updated help text and removal messages to reflect support for every hook type.
  * JSON removal status remains unchanged.

### DIFF
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -2185,13 +2185,28 @@ def _codex_feature_flag_hint() -> str:
 
 
 def cmd_hooks_remove(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -> int:
-    """Remove the observation hook previously installed for a coding CLI."""
-    settings_path = Path(args.settings)
+    """Remove every hook ``hooks install`` wired for a coding CLI (issue #580).
+
+    The settings path defaults to the client's profile, exactly as it does for
+    install: both subcommands share one ``--settings`` option that defaults to
+    ``None``, and only the installer used to fall back, so the documented
+    uninstall (``continuum hooks remove claude-code``, named in
+    docs/guides/embed-claude-code.md) died on ``Path(None)`` before it read
+    anything. An operator could only reach it by repeating by hand the path
+    install had already worked out for them.
+
+    The report names hooks rather than the observation hook because
+    :func:`remove_claude_code_hook` takes out every kind in ``_INSTALLED_KINDS``:
+    an operator who also installed the gate was told only the observation hook
+    went, which understates what just changed in the file that decides whether
+    their side effects are still guarded.
+    """
+    settings_path = Path(args.settings or CLIENT_PROFILES[args.client]["settings"])
     removed = remove_claude_code_hook(settings_path)
     text = (
-        f"Removed observation hook from {settings_path}"
+        f"Removed CONTINUUM hooks from {settings_path}"
         if removed
-        else f"No observation hook found in {settings_path}"
+        else f"No CONTINUUM hooks found in {settings_path}"
     )
     _emit(
         {"client": args.client, "settings": str(settings_path), "removed": removed},
@@ -3162,7 +3177,9 @@ def build_parser() -> argparse.ArgumentParser:
     )
     hooks_client(install, cmd_hooks_install)
 
-    remove = hooks_sub.add_parser("remove", help="Remove the observation hook.")
+    remove = hooks_sub.add_parser(
+        "remove", help="Remove every hook install wired. Mutates settings."
+    )
     hooks_client(remove, cmd_hooks_remove)
 
     verify = with_run(add("verify", cmd_verify, "Re-audit the event chain."))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,34 @@ All notable changes to this project are documented here. The format follows
   server declares and needed no change. `tests/test_mcp_docs.py` now checks the
   table against `tools/list`, so a tool registered without a row fails the
   suite instead of shipping undocumented.
+- **`hooks remove` resolves the settings path the same way `hooks install`
+  does (#580).** Both subcommands share one `--settings` option that defaults
+  to `None` and advertises "default: per client profile", but only the
+  installer fell back to `CLIENT_PROFILES[client]["settings"]`. The remover
+  passed the `None` straight to `Path`, so `continuum hooks remove
+  claude-code`, the uninstall named in `docs/guides/embed-claude-code.md`, died
+  with an uncaught `TypeError` before it read anything. Install was therefore a
+  one-way door for every operator who had not written down the path it worked
+  out for them: the only way back out was to repeat that path by hand. The
+  remover now reads the profile default, so the pair resolves one file per
+  client (`.claude/settings.json`, `.gemini/settings.json`,
+  `.codex/hooks.json`), and a directory that was never wired reports that
+  nothing was found instead of raising. The default path had no test on the
+  remove side because every case passed `--settings` explicitly, which is the
+  same reason the installer carried this bug once before; the profile default is
+  now pinned for both halves. Removal itself is unchanged, as is the `--json`
+  payload beyond the path it reports.
+
+- **The removal report names every hook it took out, not just the observation
+  hook (#580).** `remove_claude_code_hook` removes each kind in
+  `_INSTALLED_KINDS`, so an operator who had installed the `--with-gate`
+  PreToolUse hook was told "Removed observation hook" and could reasonably
+  believe the gate still stood between their agent and an unclaimed side
+  effect. The text now reads `Removed CONTINUUM hooks from <path>` (and `No
+  CONTINUUM hooks found in <path>` when there was nothing to do), and the
+  subcommand's `--help` line says what it removes. The `removed` boolean in
+  `--json` is unchanged.
+>>>>>>> 4e27e6f (fix(hooks): default the remove path to the client profile, like install (#580))
 
 ## [0.1.0] - 2026-08-27
 
--- a/tests/test_client_installers.py
+++ b/tests/test_client_installers.py
@@ -180,6 +180,85 @@ def test_default_settings_path_comes_from_the_profile(
     assert (tmp_path / expected).exists(), expected
 
 
+def _installed_kinds(settings: Path) -> set[str]:
+    """The last word of every hook command in the file, across all events."""
+    hooks = json.loads(settings.read_text()).get("hooks", {})
+    return {
+        str(h.get("command", "")).split()[-1]
+        for groups in hooks.values()
+        if isinstance(groups, list)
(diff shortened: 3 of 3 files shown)


src/continuum/cli/main.py
```diff
@@ -3162,7 +3177,9 @@ def hooks_client(p: argparse.ArgumentParser, func: Any) -> None:
     )
     hooks_client(install, cmd_hooks_install)
 
-    remove = hooks_sub.add_parser("remove", help="Remove the observation hook.")
+    remove = hooks_sub.add_parser(
+        "remove", help="Remove every hook install wired. Mutates settings."
```
Cyrax321, line 3181: Minor wording: help text Remove every hook install wired is not grammatical and does not name CONTINUUM hooks, so it reads as installer jargon. The function docstring and the runtime report both say CONTINUUM hooks, and the help should match. Consider Remove every CONTINUUM hook that install wired. Mutates settings. Not blocking, but worth fixing while the help line is being touched.

Cyrax321 (dismissed):
Ref: #581 / close #580
Surface: src/continuum/cli/main.py cmd_hooks_remove, tests/test_client_installers.py, CHANGELOG.md

What is the bug being fixed: continuum hooks remove claude-code crashed with TypeError on Path(None) when --settings was omitted. Both hook subcommands share one --settings option defaulting to None and advertising default per client profile, but only cmd_hooks_install fell back to CLIENT_PROFILES[client][settings]. The remover handed None straight to Path. Install was therefore a one way door for every operator who had not written down the path, and the uninstall named in docs/guides/embed-claude-code.md was exactly the invocation that crashed.

A second correction in the same function: the report said Removed observation hook while remove_claude_code_hook removes every kind in _INSTALLED_KINDS, so an operator who had installed --with-gate was told only the observation hook went and could reasonably believe the gate still stood.

Can the root cause be identified: Yes. Direct comparison of the two functions shows the missing fallback, and the suite missed it because every existing test passed --settings explicitly. The sibling test test_default_settings_path_comes_from_the_profile pinned the same bug on the install side in #484. Confidence clear.

Is the fix the best available: Yes. cmd_hooks_remove now resolves Path(args.settings or CLIENT_PROFILES[args.client][settings]), so the pair agrees on one file per client, .claude/settings.json, .gemini/settings.json and .codex/hooks.json. A directory never wired now reports No CONTINUUM hooks found instead of raising, which is the correct quiet no-op for a command operators run without arguments. The report now reads Removed CONTINUUM hooks from and the help line says what it removes, which matches the actual removal of every kind. The json payload keeps its shape with removed boolean and settings path.

Provenance: Sibling of #484. The bug existed since the remover was added and was never pinned on the remove side. Fixed in bd57e7a and Windows path comparison corrected in e8b72c1.

Proof: Five new tests sit next to the install pin. test_remove_defaults_to_the_same_file_install_wrote is parametrized over all three clients and asserts install with no --settings and remove with no --settings touch the same file and empty it. test_remove_without_settings_is_quiet_when_nothing_is_installed covers the crash case. test_remove_reports_hooks_not_just_the_observation_hook pins the honest report with gate installed. Verified that all five fail on the unpatched command and pass now. The kind set assertion uses <= so it holds whether or not a fourth kind is added later. Second commit e8b72c1 fixes Windows comparison by using Path equality rather than raw string equality, which previously would have failed on windows-latest where str(Path) yields backslashes. CI now shows 19 successful, 2 skipped, 1805 passed 23 skipped, ruff and mypy strict clean, codecov patch at 100 percent, Gitar approved, CodeRabbit checks 5 of 5.

What remains risky: Low. One inline nit on the help text wording at src/continuum/cli/main.py:3181, Remove every hook install wired is not grammatical. Suggested Remove every CONTINUUM hook that install wired. Not blocking. No security, performance or compatibility risk.

Review outcome: No blocking correctness or security issues. The implementation satisfies 580, restores symmetry between install and remove, and is well pinned. Ready to merge once the minor help wording is considered.

Cyrax321:
Thank you for this contribution. Restoring symmetry between install and remove so the documented uninstall no longer crashes, and correcting the report to name all hooks, is exactly the right fix. The parametrized coverage across all three clients is appreciated. LGTM, merging shortly.

PR #578, opened 2026-09-01
chore(pre-commit): ship the ruff hook config, pinned to the ruff CI runs (#537)

abyyxhek:
## Description

`CONTRIBUTING.md` described `.pre-commit-config.yaml` instead of shipping it, so `pre-commit install` on a fresh clone installed a hook that ran nothing, and a first PR still met a red `ruff format --check src/ tests/ examples/` in the `Lint & Type-check` job. This commits the config.

Two details are deliberate, and both are the difference between a hook that helps and one that misleads:

- **`rev` names the ruff the `dev` extra pins** (`ruff==0.16.4`, so `rev: v0.16.4`). A hook running a different ruff than CI is exactly how a locally formatted file still fails the format check on the PR. The snippet in `CONTRIBUTING.md` had already drifted here: it documented `rev: v0.16.3` against a `0.16.4` pin.
- **Both hooks are scoped to `src/`, `tests/`, `examples/`**, the three directories the CI lint job checks. `benchmarks/`, `demo-run/` and `_bugaudit/` are not ruff-clean today (verified: `4 files would be reformatted`, `Found 9 errors`) and no gate checks them, so leaving them in scope would fail `pre-commit run --all-files` on a tree nobody touched and rewrite files no gate asks about.

`mypy` is left out, for the reason `CONTRIBUTING.md` already gives: pre-commit hooks run in isolated environments without the project's installed dependencies, so its results there are not trustworthy. CI keeps running `mypy src/continuum` against a full environment.

Nothing under `src/` changes, and no dependency is added: `pre-commit` stays a `pip install pre-commit` for contributors who want it, exactly as `CONTRIBUTING.md` says.

## Related issues

Fixes #537

## Types of changes

- Type: `ci`

## Breaking changes

No.

## How has this been tested?

Environment: Windows 11, Python 3.11.15, `pre-commit` 4.6.2. The local `.venv` carries ruff 0.16.3, so every ruff gate below was run through `uvx ruff@0.16.4`, the version `pyproject.toml` pins and CI installs.

Acceptance criteria from the issue, checked in order:

```console
$ pre-commit run --all-files
ruff check...............................................................Passed
ruff format..............................................................Passed
```

A deliberately unformatted in-scope file is flagged and fixed before the commit lands:

```console
$ printf 'x   =    1\nimport os\n' > tests/test_zz_scratch_fmt.py
$ pre-commit run --files tests/test_zz_scratch_fmt.py
ruff check...............................................................Failed
- hook id: ruff-check
[...]

PR #573, opened 2026-08-31
docs(mcp): document continuum_record_plan and audit the tool table (#271)

abyyxhek:
## Description

`docs/api/mcp.md` listed eleven tools and summarised them as "three read-only,
eight mutating", but the server has served twelve since `continuum_record_plan`
was added in ba7b5a9. The plan tool was undocumented: a client author reading
the reference had no way to learn it exists.

This adds the missing `continuum_record_plan` row, corrects the totals under the
table to twelve tools (three read-only, nine mutating), and adds
`tests/test_mcp_docs.py` so the audit is mechanical from now on: it compares the
documented rows against `await server.list_tools()` rather than a hardcoded
list, so a tool registered without a row fails the suite instead of shipping
undocumented.

Audit results for the rest of the issue:

- Every documented `read`/`mutate` kind matches the `read_only_hint` the server
  declares. No row needed correcting.
- The em dash reported in the `continuum_record_summary` row is already gone
  from `main` (removed in 62b1634). The test pins that too, so it cannot come
  back.
- The stale "eleven tools" claim also appears in `README.md` (lines 187, 239),
  `references/architecture-data.md`, `references/auto-resume-integration.md`,
  `references/mcp.md`, `references/testing.md`, and
  `docs/research/token_floor.md`. Those are outside the file this issue names,
  and `docs/research/token_floor.md` reads as a measurement taken at the time,
  so I left them alone rather than widening a docs PR. Happy to open a follow-up
  issue or extend this PR if you would prefer them fixed here.

Verified against the live server rather than by reading the source:

```
$ python -c "..."   # build_server + await server.list_tools()
mutate continuum_record_progress
mutate continuum_checkpoint
mutate continuum_record_summary
mutate continuum_record_plan
read   continuum_validate
read   continuum_resume
mutate continuum_confirm
mutate continuum_intercept_action
mutate continuum_complete_action
mutate continuum_fail_action
mutate continuum_reconcile_action
read   continuum_list_actions
total: 12
```

## Related issues

Fixes #271

## Types of changes

- Type: `docs`

## Breaking changes

No.

## How has this been tested?

Environment: Windows 11, Python 3.11.15, `pip install -e ".[dev]"`.

The new tests fail on the documentation as it stands on `main` (docs stashed,
test file kept):

```
$ pytest tests/test_mcp_docs.py --no-cov --tb=line
E   AssertionError: assert {'continuum_r...'mutate', ...} == {'continuum_r...'mutate', ...}
      Right contains 1 more item:
[...]

Issue #316, opened 2026-08-24
Fix documentation test count drift between README, CONTRIBUTING_ONBOARDING, and CHANGELOG


Cyrax321:
## Context
Docs quote three different test counts. They drifted after recent merges.

- `README.md` says `~1,370 collected`
- `docs/CONTRIBUTING_ONBOARDING.md` says `Expect 1020 passed, 9 skipped`
- `CHANGELOG.md` says `1224`

A new contributor cannot tell which is current.

## Why this matters
The onboarding guide is the first file a newcomer reads. Wrong counts erode trust.

## How to reproduce
```bash
grep -rn "passed\|collected\|1224\|1020\|1370" README.md docs/CONTRIBUTING_ONBOARDING.md CHANGELOG.md
```

## Expected outcome
All three files agree and point to how the number was generated.

## Step by step for the fix
1. On `main`, run:
```bash
pytest --collect-only -q 2>&1 | tail -1
# and also
pytest -q 2>&1 | tail -1
```
Note that skips vary by environment (Postgres without `CONTINUUM_TEST_POSTGRES_DSN`, adapter tests without `langgraph` or `openai-agents`). That is expected.
2. Update all three files to the same number. Keep the wording honest like `~1,260 collected, about 1,240 passed on a minimal env`.
3. Add a one line HTML comment in each file near the count, for example `<!-- generated via: pytest --collect-only -q -->`, so the next person knows how to refresh it.
4. Run `grep` again to confirm only one count style remains.

## Files to touch
- `README.md`
- `docs/CONTRIBUTING_ONBOARDING.md`
- `CHANGELOG.md`

Very small docs PR, ideal for a first contribution.

Amiirhosseini:
Hi — I'd like to work on this.

**Plan:**
1. Run `pytest --collect-only -q` and `pytest -q` on current `main` to get honest counts.
2. Update `README.md`, `docs/CONTRIBUTING_ONBOARDING.md`, and `CHANGELOG.md` to the same numbers with environment caveats.
3. Add the `<!-- generated via: pytest --collect-only -q -->` HTML comment near each count so the next refresh is obvious.

Docs-only PR. Happy to open once assigned.

PR #583, opened 2026-09-01
fix(mypy): type-check a core install without the extras (#315)

### LINKED ISSUE #315: Add missing mypy overrides for optional dependencies (cryptography, mcp, langgraph)
## Background
Local type checking should be quiet on a fresh clone. Right now it is not.

If you install only the core dependency (`pip install -e .`) and run `mypy src/continuum`, you get about 25 errors like:

```
src/continuum/security/attestation.py:44: error: Cannot find implementation or library stub for module named "cryptography.hazmat.primitives"  [import-not-found]
src/continuum/mcp/server.py:89: error: Cannot find implementation or library stub for module named "mcp.server"  [import-not-found]
src/continuum/adapters/langgraph_store.py:46: error: Cannot find implementation or library stub for module named "langgraph.checkpoint.base"  [import-not-found]
src/continuum/adapters/openai.py:74: error: Cannot find implementation or library stub for module named "agents"  [import-not-found]
```

`ruff check src/ tests/` passes, so this is the last failing signal for contributors. The project already handles this for `psycopg`, `playwright`, and `kubernetes` via `[[tool.mypy.overrides]]` in `pyproject.toml`, but the other optional families are missing.

## Why this matters
Optional deps are not installed by default on purpose. A newcomer should be able to run the three required checks (`pytest`, `ruff`, `mypy`) without installing every extra.

## Current behavior
`mypy` fails unless you installed `[dev]` which pulls `cryptography`, `mcp`, `langgraph`, etc.

## Expected behavior
`mypy src/continuum` is clean with only `pydantic` installed. All optional imports are marked as `ignore_missing_imports = true`.

## How to reproduce (copy paste)
```bash
python -m venv /tmp/clean
/tmp/clean/bin/pip install -e .
/tmp/clean/bin/pip install mypy pydantic
/tmp/clean/bin/mypy src/continuum
# expect 25 errors before fix, 0 after
```

## How to fix, step by step
1. Open `pyproject.toml` and find the existing overrides for `psycopg` and `playwright`.
2. Add consolidated overrides for:
   - `cryptography` and `cryptography.*`
   - `mcp` and `mcp.*` and `mcp.server.*` and `mcp.types`
   - `langgraph` and `langgraph.*`
   - `agents` and `agents.*` (openai-agents package)
   - `opentelemetry` and `opentelemetry.*` (there is a late entry at the bottom, merge it into one block)
3. Each block should be:
```toml
[[tool.mypy.overrides]]
module = ["cryptography", "cryptography.*"]
ignore_missing_imports = true
```
4. Run `mypy src/continuum` again in the clean venv and also with `pip install -e ".[dev]"` to confirm CI still passes.

## Files to touch
- `pyproject.toml` only

## How we will review
- No runtime code change
- Keep `strict = true` for all non optional code
- `mypy src/continuum` clean in both venvs

This is a great first PR if you like packaging and tooling.

abyyxhek:
Fixes #315.

`CONTRIBUTING.md` asks for three checks before a PR. On a fresh `pip install -e .` clone the third one, `mypy src/continuum`, opened with **26 errors**, none of them about the code: `cryptography`, `mcp`, `langgraph`, `agents` and `langchain_core` had no `[[tool.mypy.overrides]]` family, so every lazy import of an optional package read as a missing library. The only route to a quiet run was to install the extras the core is built not to need.

## What changed

**`pyproject.toml`**

- Five families excused with `ignore_missing_imports`, the way `psycopg`, `playwright` and `kubernetes` already were: `cryptography.*`, `mcp.*`, `langgraph.*`, `agents` / `agents.*`, `langchain_core`. Patterns are the ones mypy actually resolves, so none of them lands as a new `unused section` note.
- The `opentelemetry` and `crewai` section that had drifted below the coverage tables moved back under `[tool.mypy]`, module list byte-identical, so all the mypy config is in one place again.
- Excusing the import is half the job. The names it supplied are now `Any`, and strict mode rejects the `BaseCheckpointSaver` and `RunHooks` subclasses and the twelve MCP tools registered through the server's own decorator, so the issue's steps taken literally leave **14** errors. Those are answered by scoping two checks off for the three modules that hold an optional seam: `disallow_subclassing_any` for `continuum.adapters.langgraph_store` and `continuum.adapters.openai`, `disallow_untyped_decorators` for `continuum.mcp.server`. `strict = true` is untouched everywhere else, and CI installs those packages, so the real signatures are what gets checked there. An inline `type: ignore` would not work here: with the packages installed, `warn_unused_ignores` turns each one into a CI failure.

**`tests/test_mypy_overrides.py`** (new, 3 tests)

Walks the package's own imports with `ast` instead of a list someone has to remember to edit, so the next optional dependency added without a family fails the suite rather than the next fresh clone. It also holds `strict = true` on, keeps the two relaxations pointed at first-party modules that exist, and keeps the mypy config from splitting across the file again. Both tests that reproduce #315 fail against the config as it stands on `main`.

## Verification

| check | before | after |
| --- | --- | --- |
| `mypy src/continuum`, core install (pydantic only) | 26 errors in 6 files | clean |
| `mypy src/continuum`, `[dev]` extra (what CI runs) | clean | clean |
| `pytest` | 1823 passed | 1826 passed |
| `ruff check src/ tests/ examples/` | pass | pass |
| `ruff format --check src/ tests/ examples/` | pass | pass |

This machine is offline, so instead of the issue's `/tmp/clean` venv the core-only run was reproduced by pointing mypy at a venv holding only pydantic and its dependency closure: `mypy --python-executable <core-venv>/python src/continuum`. That is the same package resolution the recipe in the issue produces, and it gives the same 26 errors the issue predicts.

One line of mypy output survives, and it predates this PR:

```
pyproject.toml: note: unused section(s): module = ['crewai.*', 'crewai.hooks']
```

Both patterns match nothing while crewai is absent, which is true of a core install and of the `[dev]` extra. `crewai.hooks` is redundant with `crewai.*` either way, but pruning them is a behaviour question for someone with crewai installed, so I moved the list without editing it. Say the word and I will take it in this PR.

## Notes for review

- Config and one test file. Nothing under `src/continuum` is touched, so there is no runtime change.
- Merges cleanly on `main` and against all eleven open PRs (checked with `git merge-tree`). The changelog entry sits mid-list in `### Fixed` on purpose: #578 and #579 both open an `### Added` at the top of `[Unreleased]` and #573, #577 and #581 all append at the bottom, so either end would have conflicted with two or three of them. Happy to move it wherever you prefer once those land.
- README's "roughly 1,380 tests" is left alone as #316's business.

## Summary by CodeRabbit

* **New Features**
  * Documented the `continuum_record_plan` MCP tool and expanded MCP tool coverage.

* **Documentation**
  * Corrected MCP tool totals and classifications.
  * Added validation to help keep MCP documentation complete and accurate.

* **Bug Fixes**
  * Improved static analysis reliability for installations without optional dependencies.
  * Added safeguards to keep type-checking configuration aligned with optional integrations.

* **Tests**
  * Added automated checks for optional imports and correctly scoped configuration rules.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,6 +104,13 @@ warn_unused_configs = true
 plugins = ["pydantic.mypy"]
 files = ["src/continuum"]
 
+# Every optional dependency is imported lazily, behind a `try` or under
+# `TYPE_CHECKING`, so each module that reaches for one still has to type-check on
+# an install that does not have it. A plain `pip install -e .` clone has only
+# pydantic, and the families below are what stand between that clone and 26
+# errors from `mypy src/continuum` that say nothing about the code (#315). An
+# optional import added without a family here brings the errors back.
+
 # psycopg is an optional runtime dependency (the [postgres] extra). It is
 # imported lazily, so the module must type-check without it installed.
 [[tool.mypy.overrides]]
@@ -116,6 +123,47 @@ ignore_missing_imports = true
 [[tool.mypy.overrides]]
 module = ["playwright", "playwright.*", "kubernetes"]
 ignore_missing_imports = true
+
+# opentelemetry backs the [otel] span bridge and crewai the thin hook adapter.
+[[tool.mypy.overrides]]
+module = ["opentelemetry.*", "crewai", "crewai.*", "crewai.hooks"]
+ignore_missing_imports = true
+
+# cryptography signs the event chain for `continuum attest` (the [attest]
+# extra). The recovery path must import with no crypto installed, so the import
+# sits inside the signing call.
+[[tool.mypy.overrides]]
+module = ["cryptography.*"]
+ignore_missing_imports = true
+
+# The MCP server stack (the [mcp] extra). The core library and CLI never import
+# it, so a dependency-free install must still type-check the server module.
+[[tool.mypy.overrides]]
+module = ["mcp.*"]
+ignore_missing_imports = true
+
+# The framework adapters: langgraph, openai-agents (imported as `agents`) and
+# langchain-core. Each has its own extra and its own availability guard, so none
+# of the three is present in a core install.
+[[tool.mypy.overrides]]
+module = ["langgraph.*", "agents", "agents.*", "langchain_core"]
+ignore_missing_imports = true
+
+# Silencing the missing import is half the job: the names it supplied are now
+# `Any`, and two strict checks refuse `Any` where these three modules need it.
+# `langgraph_store` subclasses LangGraph's `BaseCheckpointSaver` and the openai
+# adapter subclasses `RunHooks`, which `disallow_subclassing_any` rejects; every
+# MCP tool is registered through the server's own decorator, which
+# `disallow_untyped_decorators` rejects. Both are checked in CI, where the
+# packages are installed and the real signatures apply. Scoped to the three
+# modules that hold an optional seam so strict stays whole everywhere else.
+[[tool.mypy.overrides]]
+module = ["continuum.adapters.langgraph_store", "continuum.adapters.openai"]
+disallow_subclassing_any = false
+
+[[tool.mypy.overrides]]
+module = ["continuum.mcp.server"]
+disallow_untyped_decorators = false
 
 [tool.pydantic.mypy]
 init_forbid_extra = true
@@ -138,6 +186,3 @@ branch = true
 [tool.coverage.report]
 show_missing = true
 skip_covered = false
-[[tool.mypy.overrides]]
-module = ["opentelemetry.*", "crewai", "crewai.*", "crewai.hooks"]
-ignore_missing_imports = true
--- a/tests/test_mypy_overrides.py
+++ b/tests/test_mypy_overrides.py
@@ -1,0 +1,168 @@
+"""Tests for the mypy overrides that keep a core install type-checkable (#315).
+
+`mypy src/continuum` is one of the three checks `CONTRIBUTING.md` asks of a
+contributor, and the core library declares exactly one dependency: pydantic.
+Every optional package is imported lazily, behind a `try` or inside the call that
+needs it, so a recovery tool never fails to import when it is needed most. That
+leaves the type checker reading imports which resolve to nothing on a plain
+`pip install -e .` clone, and without the per-module sections in `pyproject.toml`
+that clone opens with 26 errors, none of them about the code.
+
+These tests hold the config to the imports rather than to a list someone has to
+remember to edit:
+
+* Every third-party module `src/continuum` imports is covered by an
+  `ignore_missing_imports` family, so a new optional import cannot land without
+  one and quietly re-break a fresh clone.
+* The strict relaxations that follow from those imports being `Any` stay scoped
+  to modules that exist and hold an optional seam, so `strict = true` cannot be
+  widened away under cover of this fix.
+* All of the mypy config stays under `[tool.mypy]`, because the override that
+  drifted below the coverage tables is how the missing families went unnoticed.
+"""
+
+from __future__ import annotations
+
+import ast
+import sys
+import tomllib
+from fnmatch import fnmatch
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+PYPROJECT = REPO_ROOT / "pyproject.toml"
+PACKAGE = REPO_ROOT / "src" / "continuum"
+
+# pydantic is the one hard runtime dependency: always importable, and it ships
+# its own types. Every other third-party import is optional by design.
+ALWAYS_INSTALLED = {"pydantic", "pydantic_core"}
+
+# The only strict checks an absent optional package is allowed to switch off.
+# Both fire on `Any` that is only `Any` because the package is not installed.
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -117,6 +124,47 @@ ignore_missing_imports = true
...
+# it, so a dependency-free install must still type-check the server module.
+[[tool.mypy.overrides]]
+module = ["mcp.*"]
+ignore_missing_imports = true
+
+# The framework adapters: langgraph, openai-agents (imported as `agents`) and
+# langchain-core. Each has its own extra and its own availability guard, so none
+# of the three is present in a core install.
+[[tool.mypy.overrides]]
+module = ["langgraph.*", "agents", "agents.*", "langchain_core"]
```
Cyrax321, line 149: Minor consistency: module list uses langchain_core without langchain_core.*. In current src the only real import is bare langchain_core inside a try, so this is enough and mypy is clean. For consistency with langgraph.* and agents.* and to future proof against a direct from langchain_core.runnables import, consider adding langchain_core.* as well. Not blocking.

tests/test_mypy_overrides.py
```diff
@@ -0,0 +1,168 @@
...
+
+    for override in _overrides():
+        relaxed = {key for key, value in override.items() if key != "module" and value is False}
+        if not relaxed:
+            continue
+        assert relaxed <= RELAXABLE, (
+            f"{sorted(relaxed)} switches off more than the checks an absent optional "
+            f"package can account for ({sorted(RELAXABLE)})"
+        )
+        for module in _module_patterns(override):
```
Cyrax321, line 143: Maintainability note: this check validates that relaxed modules are first party and exist, but does not pin the exact allowlist of three seams. An unrelated first party module could add disallow_subclassing_any false or disallow_untyped_decorators false and still pass. Consider adding an explicit EXPECTED_RELAXATIONS mapping like {continuum.adapters.langgraph_store: disallow_subclassing_any, continuum.adapters.openai: disallow_subclassing_any, continuum.mcp.server: disallow_untyped_decorators} and assert equality. Low risk, follow up is fine.

Cyrax321 (approved):
Ref: #583 / close #315
Surface: pyproject.toml mypy overrides, tests/test_mypy_overrides.py, CHANGELOG.md

What is the bug being fixed: On a plain pip install -e . clone with only pydantic, mypy src/continuum reports about 26 errors for optional families cryptography, mcp, langgraph, agents and langchain_core that are imported lazily behind try or TYPE_CHECKING. The project handles psycopg, playwright and kubernetes this way already, but the five families were missing so the only way to a clean run was to install extras the core is built not to need.

Can the root cause be identified: Yes. Missing ignore_missing_imports families in pyproject.toml and missing scoped relaxations for strict checks that fire on Any. Confidence clear, reproduced by the tests and by mypy before and after.

Is the fix the best available: Yes. Five families are added with ignore_missing_imports, using the same shape as existing overrides: cryptography.*, mcp.*, langgraph.* with agents and agents.* and langchain_core. The opentelemetry and crewai block that had drifted below coverage tables is moved back under tool.mypy so all mypy config is together again. Silencing the import makes the names Any, which then triggers 14 strict errors on BaseCheckpointSaver and RunHooks subclasses and on the twelve MCP tools registered through the server decorator. Those are scoped off only where the seam lives: disallow_subclassing_any false for continuum.adapters.langgraph_store and continuum.adapters.openai, disallow_untyped_decorators false for continuum.mcp.server. Strict stays true everywhere else and CI installs those packages so real signatures are checked there. An inline type ignore would become unused when packages are installed and would then fail with warn_unused_ignores, so a per module override is the correct mechanism.

Provenance: The missing families were absent since those adapters were added. The scoped relaxations are new and minimal for the Any case. No runtime code under src is touched.

Proof: Checked out pr-583, tests/test_mypy_overrides.py passes 3 of 3. Before the fix the same tests fail with FileNotFoundError for the missing config, and on main mypy on a core install reports 26 errors in 6 files. With the fix mypy is clean both on a core install with only pydantic and on the dev extra that CI uses. The tests walk imports with ast rather than a hard coded list, so a new optional import without a family will fail the suite instead of silently re-breaking a fresh clone. They also hold strict true, keep relaxations pointed at real first party modules, and keep mypy config from splitting across the file. CI shows 22 successful, 2 skipped, lint and ruff clean, codecov patch covered, Gitar suggests one pattern addition and CodeRabbit one maintainability check, both addressed inline as non blocking.

What remains risky: Low. Two minor follow ups noted inline. One, for consistency consider adding langchain_core.* alongside langchain_core to mirror langgraph.* and agents.* and to cover a future direct submodule import, current code only imports bare langchain_core so this is not required for clean. Two, the strict relaxation test could pin the exact allowlist of three seams rather than just checking prefix and existence, to prevent an unrelated module from relaxing strict without notice. Both are low risk and suitable for a follow up.

Review outcome: No blocking correctness, security or compatibility issues. The implementation satisfies 315, keeps strict mode intact outside the three seams, and adds durable guards. Ready to merge.

Cyrax321:
Thank you for this contribution. Making mypy clean on a core install withoutExtras while keeping strict mode intact everywhere else is a real improvement for first time contributors. The structured overrides and the guard tests that will catch the next missing family are well done. LGTM, merging shortly.

PR #603, opened 2026-09-02
docs: clarify observe_command hook portability

### LINKED ISSUE #325: Clarify observe_command portability when continuum binary not on PATH
## Context
Hook install is confusing when `continuum` is not on PATH.

`observe_command` in `src/continuum/clienthooks.py` bakes either the absolute `continuum` path or `python -m continuum.cli observe` when the binary is not found (editable installs, `uvx`, `pipx`). Both forms work, but the docs do not explain when each appears, or when `--db` is baked in.

Newcomers see `python -m continuum.cli observe` in `.claude/settings.json` and think the install failed.

## What to do
Expand docs, no code change needed.

1. Open `src/continuum/clienthooks.py` and expand the docstring for `observe_command` with an `Examples:` section showing both forms and the `db` case:
```python
continuum  # on PATH -> "/usr/local/bin/continuum observe"
python -m continuum.cli observe  # not on PATH
continuum --db /abs/path.db observe  # when --db is passed
```
2. Add a short paragraph to `docs/api/cli.md` under `continuum hooks install` explaining the two forms and how to debug with `cat .claude/settings.json` or `cat .gemini/settings.json` etc.

## How to verify
Run `continuum hooks install claude-code --db /tmp/test.db` and inspect the settings file. Docs should now answer why the command looks the way it does.

## Files
- `src/continuum/clienthooks.py`
- `docs/api/cli.md`

Docs only.

Amiirhosseini:
## Summary

- Expands the `observe_command` docstring with examples for on-PATH, interpreter fallback, and `--db` baked-in forms.
- Adds a `hooks` section to `docs/api/cli.md` explaining installed command shapes and how to inspect settings files after `hooks install`.

Fixes #325

## Test plan

- [x] `markdownlint-cli2 docs/api/cli.md` passes
- [x] `pytest tests/test_cli_observe.py` — observe_command tests pass
- [ ] Maintainer review for doc accuracy

## Summary by CodeRabbit

* **Documentation**
  * Added CLI documentation for installing and inspecting hooks, supported command invocation forms, database configuration, and reinstalling after virtual environment relocation.
  * Added usage examples for observing commands with different executable and interpreter configurations.

### DIFF
--- a/docs/api/cli.md
+++ b/docs/api/cli.md
@@ -98,9 +98,9 @@ settings files (for example `.claude/settings.json` or `.gemini/settings.json`).
 The installed `observe` command is baked in at install time and may take one of
 two shapes:
 
-- **`continuum` on PATH** — an absolute path to the resolved executable, for
+- **`continuum` on PATH**: an absolute path to the resolved executable, for
   example `/usr/local/bin/continuum observe`.
-- **Editable / interpreter-only installs** — `python -m continuum.cli observe`
+- **Editable / interpreter-only installs**: `/path/to/python -m continuum.cli observe`
   when no `continuum` executable is found on PATH.
 
 If you pass `--db`, that path is baked into the command too (for example
@@ -110,7 +110,7 @@ project root as cwd and the default database path would otherwise be ambiguous.
 To see what was installed, inspect the settings file after install:
 
 ```bash
-continuum hooks install claude-code --db /tmp/test.db
+continuum --db /tmp/test.db hooks install claude-code
 cat .claude/settings.json
 ```
 
--- a/docs/api/cli.md
+++ b/docs/api/cli.md
@@ -110,9 +110,11 @@ project root as cwd and the default database path would otherwise be ambiguous.
 To see what was installed, inspect the settings file after install:
 
 ```bash
-continuum --db /tmp/test.db hooks install claude-code
+continuum hooks install claude-code --db /tmp/test.db
 cat .claude/settings.json
 ```
 
-If the command looks unexpected after moving a virtualenv, re-run
-`continuum hooks install` for that host; it rewrites the baked command path.
+If the command looks unexpected after moving a virtualenv, re-run the
+same install command for that host, including the original `--db` value
+when one was used (for example `continuum hooks install claude-code --db /tmp/test.db`);
+it rewrites the baked command path without changing the database target.
--- a/docs/api/cli.md
+++ b/docs/api/cli.md
@@ -90,3 +90,29 @@ output.
 with per-pin status (`present`, `absent`, `unverifiable`), grace deadline, and
 flagged set. Flagged pins render prominently in human text as `[!!]` lines
 coloured on TTY and plain when piped, byte-identical modulo colour (issue #419).
+
+## hooks
+
+`continuum hooks install` writes host-side observation hooks into agent
+settings files (for example `.claude/settings.json` or `.gemini/settings.json`).
+The installed `observe` command is baked in at install time and may take one of
+two shapes:
+
+- **`continuum` on PATH** — an absolute path to the resolved executable, for
+  example `/usr/local/bin/continuum observe`.
+- **Editable / interpreter-only installs** — `python -m continuum.cli observe`
+  when no `continuum` executable is found on PATH.
+
+If you pass `--db`, that path is baked into the command too (for example
+`continuum --db /abs/path.db observe`), because hook processes run with the
+project root as cwd and the default database path would otherwise be ambiguous.
+
+To see what was installed, inspect the settings file after install:
+
+```bash
+continuum hooks install claude-code --db /tmp/test.db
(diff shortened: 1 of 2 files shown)


anya-research:
**Code Review**

This PR is docs-only and adds useful portability guidance for `observe_command`: it documents the two installed command shapes (resolved `continuum` executable vs `sys.executable -m continuum.cli` fallback) and how `--db` is baked in, plus a `hooks` section in `docs/api/cli.md` describing inspection and re-install after a virtualenv move. The docstring examples in `src/continuum/clienthooks.py:203` accurately reflect `observe_command` at `src/continuum/clienthooks.py:216`. Overall quality is good and risk is low, but one copied command in the CLI guide will fail as written.

**medium**
`docs/api/cli.md:113`
The `hooks install` example places the global `--db` flag after the subcommand (`continuum hooks install claude-code --db /tmp/test.db`), which contradicts the contract documented at `docs/api/cli.md:81` that `--db` must precede the command and is rejected as an unrecognised argument otherwise. Users copy-pasting the example will get a hard failure, undermining the new portability docs.
```suggestion
continuum --db /tmp/test.db hooks install claude-code
```

**low**
`docs/api/cli.md:103`
The interpreter-fallback bullet shows a bare `python -m continuum.cli observe`, but `src/continuum/clienthooks.py:217` actually emits `[sys.executable, "-m", "continuum.cli"]`, which is an absolute path like `/path/to/python -m continuum.cli observe` as correctly shown in `src/continuum/clienthooks.py:210`. The bare `python` form misrepresents what users will see in `settings.json` and can cause confusion when verifying the installed hook.
```suggestion
- **Editable / interpreter-only installs**: `/path/to/python -m continuum.cli observe`
  when no `continuum` executable is found on PATH.
```


[github run](/Cyrax321/CONTINUUM/actions/runs/33658341120)

docs/api/cli.md
```diff
@@ -90,3 +90,29 @@ output.
...
+  when no `continuum` executable is found on PATH.
+
+If you pass `--db`, that path is baked into the command too (for example
+`continuum --db /abs/path.db observe`), because hook processes run with the
+project root as cwd and the default database path would otherwise be ambiguous.
+
+To see what was installed, inspect the settings file after install:
+
+```bash
+continuum --db /tmp/test.db hooks install claude-code
```
Cyrax321: This pre-command form does not bake the database. `hooks install` has its own `--db` (see `continuum hooks install --help`) which shadows the global `--db`, so this parses to `db=None` and the installed command loses `--db` silently. I verified both orderings live: post-subcommand bakes `continuum --db /tmp/test.db observe`, pre-command bakes plain `continuum observe`.

Please revert to the install-local form:

```suggestion
continuum hooks install claude-code --db /tmp/test.db
```

docs/api/cli.md
```diff
@@ -90,3 +90,29 @@ output.
...
+cat .claude/settings.json
+```
+
+If the command looks unexpected after moving a virtualenv, re-run
+`continuum hooks install` for that host; it rewrites the baked command path.
```
Cyrax321: This reinstall drops the database target and is missing the required client argument. `observe_command` only adds `--db` when a value is passed, so re-running without it repoints the hook at the default database.

```suggestion
If the command looks unexpected after moving a virtualenv, re-run the
same install command for that host, including the original `--db` value
when one was used (for example `continuum hooks install claude-code --db /tmp/test.db`); it rewrites the baked command path without changing the database target.
```

Cyrax321 (changes requested):
Thank you for documenting the two installed command shapes. The `observe_command` docstring additions match the implementation in `clienthooks.py:216`, and the em dash and interpreter path fixes in c690cf1 are correct.

I verified the parser and the installed output live and found one blocking docs bug: `hooks install` owns its own `--db`, so the global position (`continuum --db ... hooks install`) parses to `None` and silently drops the baked path. The example at line 113 and the reinstall note at 117-118 need to use the install-local form `continuum hooks install <client> --db <path>`.

Once those two spots are corrected there are no further blocking issues. CI is green and the change is docs only, so no regression test is needed. Marking as requesting changes for the `--db` ordering only.

Amiirhosseini:
﻿Addressed your review:

- Install example is back to `continuum hooks install claude-code --db /tmp/test.db` (install-local `--db`, not the global pre-command form).
- Reinstall note now keeps the client argument and original `--db` value so the baked database target is preserved.

Cyrax321 (approved):
Thank you for contributing to Continuum. This addresses #325 clearly and the docs now reflect the implementation.

I reviewed `observe_command` and the `hooks install` parser against the new text, and verified both `--db` orderings with live installs. The install local form used in this head is the correct one, since the global position leaves the baked command without the database path. The two command shapes, the inspection steps, and the reinstall guidance all match observed behavior. The style fixes from the earlier round are also in good shape.

No blocking issues remain. Appreciate the thorough revisions.

PR #606, opened 2026-09-02
docs(adapters): document the thin hook adapters and the transport seams (#267)

### LINKED ISSUE #267: references/adapters.md is missing the CrewAI / AutoGen / Pydantic AI thin adapters (and the gateway + OTel seams)
### Affected component

Documentation

### What is stale

`references/adapters.md` documents the nine class-based adapters but predates three shipped integrations and two transport seams:

1. **Thin hooks for CrewAI / AutoGen core / Pydantic AI** in `src/continuum/adapters/thin.py`:
   - `install_crewai_hooks(storage, run_id)` via the global before/after tool-call registry
   - `wrap_autogen_tool(tool, storage, run_id)` wrapping `FunctionTool.run_json`
   - `wrap_pydantic_ai_hooks(storage, run_id)` returning the async Hooks capability
   All three route through the shared claim/complete ledger with stable keys (`key_fn`) and EXTERNAL_AGENT provenance. Tests: `tests/test_adapters_thin.py`.

2. **Transport-level seams** that cover stacks no adapter reaches:
   - `continuum gateway --port N` enforcing proxy (routes in `.continuum/gateway.json`)
   - `continuum.otel.make_span_processor(storage)` turning OTel tool spans into evidence (`[otel]` extra)

### Suggested content

Add an "## Thin hook adapters" section after the adapter table mirroring the README table (framework / surface / entry point), plus a "## Transport seams" section pointing at the gateway and OTel bridge with the same code snippets used in README's wiring section. Keep the existing tone: verified behaviour first, limitations stated.

### Verification

- Every entry point named in the doc exists in code and has a test file reference.
- `grep -n "thin.py" references/adapters.md` finds at least one mention.
- No em dashes (see #266 style sweep).

Yes, with guidance

Adhi1-2:
Closes #267.

`references/adapters.md` documented the class-based adapters and stopped there, so three shipped integrations and two transport seams lived only in README bullets and source docstrings. Someone who opened the adapter reference to wire CrewAI found no mention of it.

## What is new

**`## Thin hook adapters`**, placed after the class-based adapter sections. Mirrors the README table (framework / surface / entry point), then one subsection per surface with a snippet: `install_crewai_hooks` (process-global registry, so the returned uninstaller is how you scope it, and `action_types` lets untracked tools pass through), `wrap_autogen_tool` (`run_json` replaced in place, same instance back, exception recorded as a certain failure then re-raised), and `wrap_pydantic_ai_hooks` (async capability that returns args and result unchanged). Plus a `key_fn` subsection, because the signature is deliberately different from `wrap_tool`'s: `key_fn(tool_name, args_dict)` rather than `(*args, **kwargs)`, and the argument normalisation in front of it (dict through, `model_dump()` dumped, bare string to `{"input": ...}`) is what decides whether your key is stable. `ContinuumToolGuard` is documented for the case none of the three shapes fits.

**`## Transport seams`**, numbered to match the README's five. The gateway section carries the config shape, the `gate`-identical `key_template` substitution, the fail-closed unknown-host rule, and the part that distinguishes it from the in-process adapters: it settles the claim from the real response rather than trusting the caller (COMPLETED on 2xx/3xx, FAILED-certain on 4xx, FAILED-uncertain on 5xx and timeouts), stamps the external id with `METHOD path -> status`, and answers `403` with a reason naming `continuum_intercept_action` so an unclaimed caller is told how to comply. The OTel section carries the five recognised attribute keys, the observe-identical payload shape, and two limits stated plainly: it never blocks, and `[otel]` pins `opentelemetry-api` while `make_span_processor` imports `SpanProcessor` from `opentelemetry.sdk.trace`, so a bare `[otel]` install hits the `RuntimeError` hint.

Also: the `[otel]` extra added to the Installation list, a note that the thin adapters need no extra of their own, and the three seam test files added to the Integration tests section (with the one OTel test that does skip without the SDK called out).

## Verification, against the issue's own criteria

- **Every entry point exists in code.** Checked each named symbol against `src/`: `install_crewai_hooks`, `wrap_autogen_tool`, `wrap_pydantic_ai_hooks`, `ContinuumToolGuard`, the three `*_available()` probes, `make_span_processor`, `observation_from_span`, `record_span`, `load_gateway_config`. The gateway config shape and settlement rules come from `src/continuum/gateway.py`, the `403` reason and the `502`-leaves-it-uncertain behaviour from `tests/test_gateway.py`.
- **Each section names its test file.** `tests/test_adapters_thin.py`, `tests/test_gateway.py`, `tests/test_otel.py`. All three run: `30 passed, 1 skipped` (the skip is `test_processor_end_to_end_when_sdk_present`, no OTel SDK here, and the doc says so).
- **`grep -n "thin.py" references/adapters.md`** finds 4 mentions.
- **No em dashes** in either changed file.
- MD031 / MD040 / MD047 clean (the rules `.markdownlint-cli2.jsonc` enables; CI's markdownlint globs the four root docs, not `references/`).
- Full gate on this branch: `1895 passed, 23 skipped`, `ruff check` and `ruff format --check` clean, `mypy --strict src/continuum` clean. Docs-only, so no runtime change to gate.

## One thing I found and did not fix

`thin.py`'s module docstring says "Provenance is EXTERNAL_AGENT because the framework asserts the facts", but `ActionLedger` never passes `source=`, so `append_event`'s `Origin.DETERMINISTIC` default is what lands. Verified by reading the log back after a claim and completion:

```text
1  RUN_STARTED      source=deterministic
2  ACTION_RECORDED  source=deterministic
3  ACTION_RECORDED  source=deterministic
```

The practical consequence is that a thin-adapter run is not held for review the way an MCP-reported or OpenAI-adapter run is. I documented the behaviour as it is, with the mismatch stated, rather than repeating the docstring's claim. Changing the ledger to stamp `EXTERNAL_AGENT` would flip these runs to `request_human`, which is a behaviour change and not a docs PR's call. Happy to open a separate issue for it if you agree it is one, or to drop that paragraph if you would rather the doc not carry it.

## Summary by CodeRabbit

- **Documentation**
  - Expanded adapter reference documentation for CrewAI, AutoGen, and Pydantic AI integrations.
  - Documented hook configuration options, availability checks, and installation details for OpenTelemetry support.
  - Added coverage of the enforcing gateway and OpenTelemetry integration.
  - Clarified adapter provenance behavior and included testing guidance for adapter and transport integrations.


Cyrax321 (approved):

I checked the new reference sections against src/continuum/adapters/thin.py. The entry points, lazy import behavior, key_fn shape, and gateway versus OTel distinction are accurate. Documenting the EXTERNAL_AGENT versus DETERMINISTIC discrepancy rather than papering over it is the right call for a docs PR. CI is green.

Issue #280, opened 2026-08-23
Docker publish workflow should smoke-test the demo path, not just continuum --version


Cyrax321:
## Affected component

Other / not sure (CI)

## What problem does this solve?

The Docker publish workflow (`.github/workflows/docker-publish.yml`) smoke-tests the image with `continuum --version` only. But the image's default CMD runs the crash-recovery demo, and that path (working directory writability, example assets present, subprocess kill/resume inside the container) is not exercised until someone pulls the image and runs it with no arguments.

## Proposed solution

Extend the smoke-test step to also run the demo non-interactively in a throwaway container and check its final verdict line.

### Acceptance criteria

- CI runs something like `docker run --rm "$IMAGE:latest"` and asserts the output contains "No work repeated. No side effect duplicated.".
- Failure of the demo run fails the workflow.
- Total added runtime stays reasonable (< ~2 minutes).

## Alternatives considered

A separate nightly image test; rejected because the demo is the image's default entrypoint and belongs in the publish gate itself.

## Is this a breaking change?

No

## Would you like to work on this?

Yes, with guidance

abyyxhek:
Working on this. Opened #608 for one half of it first.

While building the image to test the new smoke step, I found the thing this issue
suspects is not hypothetical: the image's default command is broken on `main`
right now. `Dockerfile` copies the sources in root-owned under `/opt/continuum`
and then drops to `USER continuum`, but the demo pinned its `demo-run` workspace
to the source tree, so a bare `docker run` dies immediately:

```text
$ docker run --rm <image built from 2e0ef3e>
PermissionError: [Errno 13] Permission denied: '/opt/continuum/demo-run/worker.py'
exit 1

$ docker run --rm <same image> continuum --version
continuum 0.1.0
exit 0
```

That second block is exactly the gap described here: the current smoke test
passes on an image whose default command is dead.

#608 is the one-line fix plus a CHANGELOG entry, kept separate so it is
reviewable on its own. The workflow step itself is written and verified locally
against the same images, and I will send it as a follow-up PR:

```yaml
      - name: Smoke-test the demo path (the image's default command)
        timeout-minutes: 3
        run: |
          set -o pipefail
          docker run --rm ${{ env.IMAGE }}:latest | tee "$RUNNER_TEMP/demo.txt"
          grep -qF "No work repeated. No side effect duplicated." "$RUNNER_TEMP/demo.txt"
```

[...]

PR #608, opened 2026-09-02
fix(examples): unbreak the published image's default command

abyyxhek:
## Description

The published image's default command is broken on `main`. `Dockerfile` copies
the sources in root-owned under `/opt/continuum` and then drops to
`USER continuum`, but `examples/crash_recovery_agent.py` pinned its `demo-run`
workspace to the source tree, so the documented bare invocation dies before
printing a line:

```text
$ docker run --rm continuum-280-base:latest        # built from 2e0ef3e, unmodified
  File "/opt/continuum/examples/crash_recovery_agent.py", line 236, in <module>
    raise SystemExit(main())
  File "/opt/continuum/examples/crash_recovery_agent.py", line 94, in main
    worker.write_text(WORKER)
PermissionError: [Errno 13] Permission denied: '/opt/continuum/demo-run/worker.py'
exit 1
```

The fix is one line: resolve the workspace against the working directory, which
`Dockerfile` already sets to a writable `/home/continuum` and says so in a
comment ("Default working directory is writable so the demo and CLI can create
continuum.db"). The comment above the line records why, so it does not get put
back.

I found this while working on #280, which asks the publish workflow to
smoke-test the demo path instead of only `continuum --version`. The bug is why
that matters: `continuum --version` exits 0 on exactly the image whose default
command is dead.

```text
$ docker run --rm continuum-280-base:latest continuum --version
continuum 0.1.0
exit 0
```

This PR is the bug half only, so it stays reviewable on its own. The workflow
step follows separately.

## Related issues

Refs #280

## Types of changes

- Type: `bugfix`

## Breaking changes

No.

Every documented entry point runs from the repository root, where the two
expressions name the same directory:

```text
$ python -c "from pathlib import Path; a = Path.cwd() / 'demo-run'; b = Path('examples/crash_recovery_agent.py').resolve().parents[1] / 'demo-run'; print(a); print(b); print('identical:', a == b)"
C:\...\CONTINUUM\demo-run
C:\...\CONTINUUM\demo-run
identical: True
```

`try-it.sh` does `cd "$(dirname "$0")"` and `try-it.ps1` does
`Set-Location $Root` before invoking the demo, and the README documents
`python examples/crash_recovery_agent.py` from the root. The script's own footer
already assumed this: it prints `rm -rf demo-run` and computes
`db.relative_to(Path.cwd())`.

## How has this been tested?

Environment: Docker 29.6.2, Linux containers, Windows 11 host. Images built from
a pristine `git worktree` checkout so the build context matches
`actions/checkout` rather than a dirty local tree.

Built from this branch's head and run with no arguments, which is the default
command:

```text
$ docker run --rm continuum-280-fixonly:latest
...
7. Did we do anything twice?
====================================================================
    documents processed      1000
    duplicates               0
    GitHub issues created    1
    progress recovered       1000/1000
    event chain verified     True

    No work repeated. No side effect duplicated.
exit 0
real    0m22.499s
```

So the whole demo runs inside the image: the real `os._exit(9)` kill, the
reconcile, the resume, and a verdict of one side effect and zero duplicates.

Locally, from the repository root, unchanged: exit 0, same verdict line,
workspace still at `<repo root>/demo-run`.

Repo gates on this branch:

```text
$ .venv/Scripts/python.exe -m pytest --no-cov
1881 passed, 37 skipped in 302.69s (0:05:02)

$ uvx ruff@0.16.4 check src/ tests/ examples/
All checks passed!

$ uvx ruff@0.16.4 format --check src/ tests/ examples/
279 files already formatted

$ .venv/Scripts/mypy.exe src/continuum
Success: no issues found in 118 source files
```

## Checklist

Tests added or updated: none. The behaviour is "the published image's default
command runs", which is only observable by building and running the image, and
no existing test imports this example. The publish-workflow step in #280 is the
assertion that belongs here, which is the other half of that issue.

Documentation updated: `CHANGELOG.md`. `docs/ARCHITECTURE_EVOLUTION.md` section
19.2 also describes the `continuum --version` smoke test, but that file is a
dated record of past stages, so I left it as written rather than editing history.

CI passes on the latest commit: gates above are green locally. Note that
`docker-publish.yml` only runs on pushes to `main` in this repository, so no CI
run on a PR builds the image; the local runs above are the evidence.

Security-relevant changes: none. The image still runs as the unprivileged
`continuum` user and nothing is chowned to make the source tree writable.

## Additional context

**The alternative I rejected.** A `Dockerfile`-only fix
(`chown continuum /opt/continuum /opt/continuum/demo-run`) also turns the demo
green and leaves `examples/` untouched. I did not take it for two reasons: it
would make the image's own source tree writable by the user executing it, and it
depends on `demo-run/` staying out of `.dockerignore`, because the demo's
`shutil.rmtree` swallows its own failure and only survives today thanks to
`mkdir(exist_ok=True)` finding the copied directory already there. Resolving
against the working directory has neither coupling.

**Noticed, not fixed.** Running the demo from the repository root deletes tracked
files. `.gitignore` ignores `demo-run/*` but un-ignores
`demo-run/generate_crash_visual.py` and `demo-run/worker.py`, and `main()` opens
with `shutil.rmtree(workspace, ignore_errors=True)`, so `./try-it.sh demo`
removes `generate_crash_visual.py` and overwrites `worker.py`. I hit it while
verifying: after a run, `ls demo-run/` shows only `agent.db`,
`github-issues.log` and a regenerated `worker.py`. It predates this PR and the
change neither causes nor worsens it, since from the repository root the path is
identical, so I left it alone rather than widening the diff. Happy to open a
separate issue if you want it tracked.

One incidental improvement: `docs/recipes/adapter-funnel.md` suggests
`cp examples/crash_recovery_agent.py demo.py`. Under the old source-relative
path that wrote `demo-run` into the repository's parent directory; it now lands
in the working directory.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the published Docker image’s default command so the demo runs successfully when launched by an unprivileged user.
  - Prevented a workspace permission error that previously caused the container to exit before producing output.

- **Documentation**
  - Added an Unreleased changelog entry describing the Docker demo startup fix.


Cyrax321 (changes requested):

The workspace fix is correct: resolving demo-run against the working directory matches the Dockerfile writable dir and leaves the documented repo root entry points unchanged. The CHANGELOG entry describes the failure and scope accurately.

One blocking item before merge: this branch currently reports mergeable_state dirty, indicating a merge conflict, most likely in CHANGELOG against main. Please update the branch and resolve the conflict, then push. I will re review promptly once it is mergeable.

Cyrax321 (approved):
Thank you for contributing to Continuum. Re reviewing after the conflict resolutions: the CHANGELOG now carries this fix entry alongside the newer ones with no other content changes, and the workspace fix itself is unchanged and previously verified. CI is green across the matrix. No blocking issues remain. Acceptable to merge.

PR #605, opened 2026-09-02
docs: align test counts across contributor docs

Amiirhosseini:
## Summary

- Syncs test counts in README, CONTRIBUTING_ONBOARDING, and CHANGELOG to ~1,918 collected / ~1,880 passed / ~38 skipped on a minimal env.
- Adds HTML comments noting how to regenerate via pytest.

Fixes #316

## Test plan

- [x] \pytest --collect-only -q\ → 1918 tests collected (local, dev extras)
- [x] markdownlint passes on README.md and CONTRIBUTING_ONBOARDING.md
- [x] grep confirms old drift values (1020/1224/1380) removed from the three files

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

- **Documentation**
  - Updated test-suite statistics across the changelog, Quick Start guide, and contributor onboarding documentation.
  - Documented approximately 1,918 collected tests, including 1,880 passing and 38 skipped tests.
  - Added the commands used to generate these figures and clarified environment-dependent skips.

PR #610, opened 2026-09-02
docs(examples): add the crash-recovery walkthrough as a notebook

### LINKED ISSUE #283: Interactive notebook demo (examples/demo.ipynb) with Colab and Binder badges for zero-install evaluation
## Affected component

Documentation

There is no notebook demo. A runnable `examples/demo.ipynb` would give the lowest-friction possible evaluation path (one click from the README on Colab, no clone, no install, no Docker), which matters for social/launch traffic where every setup step loses people.

## Proposed solution

Create an interactive notebook version of the crash-recovery walkthrough: markdown cells explaining each phase (run start, kill at doc 400, dataset drift v3 to v4, verified resume), code cells calling into installed CONTINUUM APIs, and Colab + Binder badges in README Quick Start.

### Acceptance criteria

- Notebook executes top-to-bottom on Colab after a single `!pip install git+https://github.com/Cyrax321/CONTINUUM.git` cell (or equivalent).
- Ends with the same integrity verdict as the CLI demo.
- Badges render correctly in the README table.

## Alternatives considered

Linking the existing examples as plain scripts; rejected because notebooks carry narrative and can run without local setup.

## Is this a breaking change?

No

Yes, I have a plan

abyyxhek:
## Description

There is no notebook demo, so the lowest-friction way to watch a recovery is
`docker run`, which still wants a daemon and a pull. `examples/demo.ipynb` wants
a browser: the Colab and Binder badges added to the Quick Start table open the
crash-recovery walkthrough with no clone, no install and no Docker.

It is the same walkthrough `examples/crash_recovery_agent.py` prints, driven
through the library instead of the CLI, in narrated steps:

1. the agent starts on dataset v3 and is killed at document 400 by a real
   `os._exit(9)` in a real child process
2. what survived: `project()` replays the event log, `verify_events()` recomputes
   the hash chain over it
3. the dataset moves v3 to v4, `RecoveryEngine.assess` answers `REQUEST_HUMAN`,
   and `exit_code_for` shows the 20 that makes
   `continuum resume "$RUN" && ./start-agent.sh` short-circuit
4. `ProbeReconciler` asks the real system about the side effect nobody could
   vouch for
5. the run resumes at 400 documents, and a second claim of the same action comes
   back `fresh=False` carrying the external id the probe found
6. the verdict: 1,000 documents, 0 duplicates, 1 GitHub issue, chain verified

Three decisions worth a reviewer's attention:

- **The kill is a child process, everything else runs in the kernel.**
  `os._exit(9)` would take the kernel down with it, so the worker is spawned the
  way the script spawns it. Every other phase calls the installed API directly,
  which is what #283 asked for and reads better in a notebook than scraped CLI
  text.
- **The install cell uses `subprocess`, not `%pip`.** That keeps the notebook
  plain Python, which is what lets `tests/test_demo_notebook.py` execute the
  cells without adding Jupyter to the dev extra. It installs only when
  `importlib.util.find_spec("continuum")` is None, so opened from a clone that
  already has the package, the cell does nothing.
- **The workspace is a `tempfile.mkdtemp()`, not `demo-run/`.** `demo-run/` holds
  tracked files (`.gitignore` un-ignores `worker.py` and
  `generate_crash_visual.py`) and the demo opens by deleting its workspace. The
  last cell removes what the notebook made.

The notebook ships without stored outputs. A stored run embeds the absolute temp
path of whoever produced it, and re-running would drop a 400 line diff into every
future PR that touches the file. The markdown cells say what a clean run ends
with instead.

## Related issues

Closes #283

## Types of changes

- Type: `docs`

One test comes with it (`tests/test_demo_notebook.py`): nothing in CI executed
the notebook, so a renamed API would break the first thing a new reader runs, and
break it silently.

## Breaking changes

No. Nothing under `src/` changed. `.gitignore` gains `.ipynb_checkpoints/`, which
Jupyter writes beside any notebook it opens.

## How has this been tested?

Environment: Windows 11, Python 3.11.15, `uv` 0.12.5.

**It runs top to bottom in an environment with no CONTINUUM installed**, which is
the first acceptance criterion (the Colab path). The ephemeral env holds only
nbclient, ipykernel and pip, so the notebook's own first cell is what installs
CONTINUUM, from git. nbclient stops at the first cell that raises unless told not
to, so a clean exit means every cell ran:

```text
$ uvx --from nbclient --with ipykernel --with pip jupyter-execute --timeout=600 examples/demo.ipynb
[NbClientApp] Executing examples/demo.ipynb
[NbClientApp] Executing notebook with kernel: python3
exit 0
real    0m39.315s
```

Executing the same file again with `--inplace` into a copy under `/tmp` captures
what the cells actually printed. The decision and the verdict, verbatim:

```text
Recovery decision: REQUEST_HUMAN
  because 1 external side effect(s) have unknown outcomes
  because at least one repair needs a person
...
mode                 request_human
safe to resume       False
next allowed action  reconcile_action:action_c97fa87e3cd2c12030ff627c3f540256
continuum resume     would exit 20

unresolved before  1
probe says         confirmed as performed: github.create_issue
unresolved after   0

resumed at 400 documents, nothing reprocessed
claimed again: fresh=False, external_id='481'
finished at 1000 documents

documents processed      1000
duplicates               0
GitHub issues created    1
progress recovered       1000/1000
event chain verified     True

No work repeated. No side effect duplicated.
```

That last line is the verdict `examples/crash_recovery_agent.py` ends with, which
is the second acceptance criterion.

**The rot guard**, which is also how the notebook is tested against this working
tree rather than against `main`:

```text
$ .venv/Scripts/python.exe -m pytest tests/test_demo_notebook.py --no-cov -v
tests\test_demo_notebook.py ...                                          [100%]
3 passed in 11.23s
```

**Repo gates on this branch:**

```text
$ .venv/Scripts/python.exe -m pytest --no-cov
1884 passed, 37 skipped in 288.85s (0:04:48)

$ uvx ruff@0.16.4 check src/ tests/ examples/
All checks passed!

$ uvx ruff@0.16.4 format --check src/ tests/ examples/
281 files already formatted

$ .venv/Scripts/mypy.exe src/continuum
Success: no issues found in 118 source files
```

1884 is 1881 plus the three tests in this PR. `ruff` reads `.ipynb` as well as
`.py`, so the notebook's cells are linted and format-checked by the existing lint
job; `mypy` is scoped to `src/continuum` and does not see it.

**Schema and badges.** The file validates against the nbformat schema, which is
what GitHub and Colab render from, and both badge images resolve:

```text
$ uvx --from nbformat python -c "import nbformat; nb=nbformat.read('examples/demo.ipynb', as_version=4); nbformat.validate(nb); print(nb.nbformat, nb.nbformat_minor, len(nb.cells))"
4 5 19

$ curl -sI -o /dev/null -w "%{http_code} %{content_type}\n" https://colab.research.google.com/assets/colab-badge.svg
200 image/svg+xml

$ curl -sL -o /dev/null -w "%{http_code} %{content_type} %{size_download}\n" https://mybinder.org/badge_logo.svg
200 image/svg+xml 3378
```

What I could not verify from here: an actual hosted run. Both badge links point at
`Cyrax321/CONTINUUM` on `main`, so they resolve once this merges and 404 until
then. I checked the equivalent paths on the pushed branch instead
(`raw.githubusercontent.com/.../docs/notebook-demo-283/examples/demo.ipynb` and
its GitHub blob page both return 200), and the table renders on this branch's
README if you want to eyeball the third acceptance criterion before merging.

## Checklist

Tests added or updated: `tests/test_demo_notebook.py`, three tests. It reads the
`.ipynb`, concatenates the code cells and runs them as a script, then asserts the
kill (`exit code 9`), the surviving checkpoint (`400/1000`), the refusal
(`request_human`, exit 20), the reconciliation (1 pending to 0), the idempotent
re-claim (`fresh=False`) and the verdict line. A second test asserts the notebook
stays free of line magics, because that is the assumption the first one rests on.

Documentation updated: `README.md` (two rows in the Quick Start table plus one
sentence) and `CHANGELOG.md` under `[Unreleased] / Added`.

CI passes on the latest commit: gates above are green locally on `30ddd6b`.

Security-relevant changes: worth one line, because the notebook contains a
`pip install`. It runs only when `continuum` cannot be imported, and the guard
test injects the parent's `PYTHONPATH` into the child so CI always takes the
import path. No test in this PR reaches the network. On Colab and Binder the
install is the point, and it installs this repository over HTTPS.

## Additional context

**The alternative I rejected.** `binder/requirements.txt` with `-e .` (plus a
`runtime.txt` to pin the interpreter) would preinstall CONTINUUM in the Binder
image so the first cell became a no-op there. I left it out because I cannot run
a repo2docker build here to prove it: with a `binder/` directory present
repo2docker stops considering root config, and if its default interpreter came up
older than 3.11 the build would fail outright, which is worse than the runtime
install this notebook already does. Happy to add both files if you would rather
have the faster Binder start, but I would want the build watched once.

**Left alone deliberately.** The six translated READMEs carry the same Quick Start
table. I did not touch them: I cannot write those six languages to the standard of
the rest of the file, and #604 sets the same precedent for a badge change. Say the
word and I will open a follow-up with the rows in English inside each.

**Overlap.** Nothing here touches a file that another open PR touches, except
`README.md` and `CHANGELOG.md`. #604 and #605 edit the badge block at the top and
the `pytest -q` line under Verify; this edits the table between them, and it adds
to `### Added` where my #608 adds to `### Fixed`. All three merge clean in either
order:

```text
$ for p in 604 605 608; do git merge-tree --write-tree HEAD refs/pr/$p >/dev/null; echo "pr/$p -> exit $?"; done
pr/604 -> exit 0
pr/605 -> exit 0
pr/608 -> exit 0
```


Cyrax321 (approved):

I reviewed the notebook and its guard tests. Keeping the cells plain Python so they run without a kernel, using a temporary workspace cleaned up by the last cell, and asserting the same decision points as the CLI walkthrough is the right way to keep this from rotting. CI is green across the matrix.

PR #614, opened 2026-09-02
fix(serve): answer a valid-JSON non-object request on both transports

### LINKED ISSUE #582: serve: a valid-JSON non-object request kills the stdio loop and closes HTTP unanswered
## Context

Both sidecar transports in `src/continuum/serve/server.py` handle a body that is
not valid JSON: stdio answers `{"error": {"type": "bad_request", ...}}` and HTTP
answers `400 {"error": "invalid JSON body: ..."}`. Neither handles a body that
*is* valid JSON but is not a JSON object (`[]`, `null`, `5`, `"resume"`). Both
then fail in the way each was explicitly written not to.

This is the shape #522/#533 fixed for body framing and #323 fixed for the
gateway ("escaped the handler and the connection closed with no response at
all"), one layer up: the framing is now fail-closed, the payload *shape* is not.

## Root cause verified on main (d06fe56)

**stdio, `SidecarServer.serve_stdio`.** `src/continuum/serve/server.py:282` is
`rid = req.get("id")`, one line above the try whose own comment states the
contract:

```python
rid = req.get("id")                       # line 282, outside the guard
try:
    result = self.dispatch(str(req.get("method")), dict(req.get("params") or {}))
except SidecarError as exc:
    ...
except Exception as exc:  # noqa: BLE001 - report, never crash the loop
```

A request of `[]` or `null` raises `AttributeError` before the guard can see it,
so the loop the comment protects is exactly what dies. The process exits 1, and
because stdio is a long-lived session, every later request on that connection is
lost with it. One malformed line from any client takes the durability plane down.

**HTTP, `SidecarHTTP.Handler.do_POST`.** `src/continuum/serve/server.py:340`
already computes the right answer, but raises it inside a try that catches only
`json.JSONDecodeError`:

```python
try:
    params = json.loads(raw)
    if not isinstance(params, dict):
        raise BadParams("body must be a JSON object")   # escapes do_POST
except json.JSONDecodeError as exc:
    self._json(400, {"error": f"invalid JSON body: {exc}"})
    return
```

`BadParams` is not a `JSONDecodeError`, so it propagates out of the handler,
`ThreadingHTTPServer` prints a traceback, and the connection closes with no
response. The client sees `RemoteDisconnected`, indistinguishable from a crashed
sidecar, even though the code three lines up knows the request was a 400. The
`except BadParams` arm below it never sees this one, because it guards the
`dispatch` call in a separate statement.

## Repro

stdio, one line, and the process is gone:

```console
$ echo '[]' | continuum serve
  ...
  File "src/continuum/serve/server.py", line 282, in serve_stdio
    rid = req.get("id")
AttributeError: 'list' object has no attribute 'get'
$ echo "exit $?"
exit 1
```

HTTP, against a live sidecar on the default bind:

```console
$ continuum serve --transport http --port 8766 &
$ curl -sS -X POST localhost:8766/resume -d 'not json'
{"error": "invalid JSON body: Expecting value: line 1 column 1 (char 0)"}   # 400, correct
$ curl -sS -X POST localhost:8766/resume -d '[]'
curl: (52) Empty reply from server                                          # no response
$ curl -sS -X POST localhost:8766/resume -d 'null'
curl: (52) Empty reply from server
```

Measured on d06fe56: object bodies answer 200/500 normally before and after, so
the happy path is not involved.

## Expected behaviour

A request that parses but is not a JSON object is a client error and must be
answered as one, on both transports, without taking anything down:

- stdio: one `{"id": null, "error": {"type": "bad_request", "message": ...}}`
  line, then the loop continues to the next request. `id` is `null` because a
  non-object request carries nowhere to put one.
- HTTP: `400 {"error": "body must be a JSON object"}`, the answer the code
  already spells out, delivered instead of escaping.
- The two transports should reach that answer through one shared check rather
  than two, since they have already drifted once here.

Unchanged: object requests, absent or empty bodies (still the empty params
object, so a method needing no params stays callable), and the existing
`bad_request` answer for genuinely malformed JSON.

## Search for duplicates

`gh issue list --state open|closed` for serve, sidecar, stdio, JSON object,
bad_request, AttributeError. Related but distinct: #533 (Content-Length and
chunked framing on the serve HTTP transport, open PR #577), #522/#317 (the same
framing for gateway and dashboard), #323 (gateway answers malformed JSON with
400), #91 (absent `params` on a resume request, fixed). None covers a
valid-JSON-wrong-shape request on either transport. PR #577 leaves both lines
above untouched: `if not isinstance(params, dict)` appears in its diff only as
context.

## Files

- `src/continuum/serve/server.py` (`serve_stdio`, `SidecarHTTP.Handler.do_POST`)
- `tests/test_serve.py`, `tests/test_serve_http.py`

Adhi1-2:
## Description

Body framing on the `continuum serve` sidecar was already fail-closed on both
transports. The payload *shape* was not, and each transport then failed in the
exact way it had been written not to.

**stdio.** `serve_stdio` reads `rid = req.get("id")` one line above the guard
whose own comment reads "report, never crash the loop". A line that is valid
JSON but not an object reaches `.get` on a list, raises `AttributeError` where
nothing is catching, and exits the process 1. stdio is a long-lived session, so
one malformed line from any client also takes down every request still queued
behind it.

**HTTP.** `do_POST` already spelled out the right refusal, but raised it as
`BadParams` inside a `try` that catches only `json.JSONDecodeError`. The refusal
escaped the handler and the connection closed with no response at all. The
caller sees `RemoteDisconnected`, which is what a crashed sidecar looks like,
three lines below code that knew the answer was 400.

Reproduced on `main` at `2e0ef3e`, sending `[]` followed by a valid request on
the same session:

```
### stdio: a `[]` line, then a real request on the same session
stdout: (empty)
stderr tail: AttributeError: 'list' object has no attribute 'get'
returncode: 1

### http: POST a `[]` body
RemoteDisconnected: Remote end closed connection without response
```

The HTTP server logs the answer it failed to send:

```
  File "/tmp/pre582/src/continuum/serve/server.py", line 340, in do_POST
    raise BadParams("body must be a JSON object")
continuum.serve.server.BadParams: body must be a JSON object
```

Note the stdio `stdout: (empty)`: the queued `id: 1` request was never answered
either.

### The fix

One shared check, reached by both transports, because these two had already
drifted apart on the same question:

```python
def _request_object(parsed: Any) -> dict[str, Any]:
    if not isinstance(parsed, dict):
        raise BadRequest("body must be a JSON object")
    return parsed
```

- stdio answers `{"id": null, "error": {"type": "bad_request", "message": "body
  must be a JSON object"}}` and reads the next request.
- HTTP answers `400 {"error": "body must be a JSON object"}` and keeps the
  connection usable.

The id is `null` because a non-object request has nowhere to put one. That is
also why this raises a new `BadRequest` rather than reusing `BadParams`:
`BadParams` answers a request that *was* read and dispatched, so its answer
carries that request's id. On the stdio wire `BadRequest` reuses the
`bad_request` type already answered for a line that is not JSON at all, so
clients gain no new error type to handle. On stdio the two arms are caught
together, since neither an unparseable line nor a non-object one is a request
and neither answer can name an id. On HTTP they stay separate so each keeps its
own message.

Unchanged: object requests, absent and empty bodies (still the empty params
object, so a method needing no params stays callable), and both existing answers
for genuinely malformed JSON.

## Related issues

Fixes #582

## Types of changes

- Type: `bugfix`

## Breaking changes

No. Every input that produced a useful answer before produces the same answer
now. The two inputs whose behaviour changed produced a dead process and a dead
connection.

## How has this been tested?

Rebased onto `main` at `2e0ef3e` and run there. Python 3.11, macOS, project venv:

```
./.venv/bin/python -m pytest                                              # 1911 passed, 23 skipped
./.venv/bin/python -m pytest tests/test_serve.py tests/test_serve_http.py # 54 passed
./.venv/bin/python -m ruff check src/ tests/ examples/                    # All checks passed
./.venv/bin/python -m ruff format --check src/ tests/ examples/           # 279 files already formatted
./.venv/bin/python -m mypy --strict src/continuum                         # no issues, 118 source files
npx markdownlint-cli2 CHANGELOG.md                                        # 0 issues
```

Six new tests, sixteen cases, three per transport:

- **The shape refusal**, parametrized over `[]`, `null`, `5`, `"resume"`, `true`
  and `["resume"]`, so the check cannot pass by special-casing lists. On stdio it
  asserts the `bad_request` answer with a null id **and** that the request behind
  it on the same stream still returns its result. That second half is the part
  that actually failed: a fix that answered the bad line but still exited would
  satisfy a weaker test.
- **The framing refusal, pinned.** A line and a body that no parser can read keep
  the decoder's own message, not the new one. Same status and type, different
  text, which is what lets a client tell "I cannot parse these bytes" from "I
  parsed them, and that is not a request". Without this the shared arm could
  quietly swallow the older answer.
- **Survivability.** On stdio, a run of `[]`, `not json`, `null` and then a real
  request produces four answers in order with ids `[None, None, None, 9]`. On
  HTTP, a refused `[]` is followed by a 200 on the same server.

The HTTP shape tests send no token and still expect 400 rather than 403: a body
that is not a request cannot carry credentials to check, and the message is a
fixed protocol string, so refusing on shape first discloses nothing.

## Checklist

Tests added for the changed behaviour on both transports, including the
next-request-survives assertions. CHANGELOG `Unreleased` entry added per
CONTRIBUTING step 4. No user-facing docs change: the error shapes are the ones
already documented for a malformed line.

## Additional context

**One unrelated line removed.** `9111a06` (#581) committed a stray
`>>>>>>> 4e27e6f (fix(hooks): default the remove path ...)` conflict marker into
the `Unreleased` section of `CHANGELOG.md`. It is still on `main`, and it is the
line my entry lands directly after, so the alternative was writing new text
around known garbage inside the same hunk. Removed here as a one-line deletion,
called out so it is not read as something this branch introduced. Happy to split
it out if you would rather keep this PR to a single subject.

**Expect a conflict with #577.** That PR refactors the same `do_POST` body read
into a `_read_body()` method; this one edits the JSON-parse block just below it.
Both are mine and both are open, so whichever lands first I will rebase the other
and re-run the suite. No behavioural overlap: #577 changes how the bytes are
read, this changes what happens once they parse.

**Not in scope.** A body that *is* an object but carries an unknown `method` or a
bad `params` still takes the existing 404 and `BadParams` paths, with its own id,
which is correct: those were read and dispatched. This PR is only about the case
where there was never a request to dispatch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Sidecar transports now gracefully reject valid JSON values that aren’t objects.
  - Stdio returns a structured bad-request response and continues serving subsequent requests instead of terminating.
  - HTTP returns a clear `400` response with an explanatory error message.
  - Malformed JSON continues to receive the existing error response.
- **Tests**
  - Added coverage for invalid request shapes, malformed JSON, and continued service after rejected requests.


Cyrax321 (approved):

I reviewed the shared _request_object check against both transports. Using BadRequest with a null id for a parsed non object, keeping BadParams for dispatched requests, and keeping the malformed JSON answers unchanged is the right split. The stdio session survival and HTTP 400 paths are covered by direct tests, and tests/test_serve.py with tests/test_serve_http.py passes locally. CI is green. The stray conflict marker removal in CHANGELOG is also correct.

PR #617, opened 2026-09-03
docs: align resolve_authorization_id token wording with the predicate

### LINKED ISSUE #613: resolve_authorization_id docstring says status words are weak tokens, but only 18 fixed words are
> **Labels to apply: `documentation`, `good first issue`.**
> Filed with `gh`, so the template's own `labels:` did not attach, and my
> permission on this repository is `pull` only, which GitHub does not allow
> to set labels. Applying them needs triage rights.

### Before reporting

### Affected component

Core

### Severity

Minor or cosmetic

### Regression check

The docstring has been inaccurate since `resolve_authorization_id` landed in
PR #430.

### System information

```
CONTINUUM   : main at 2e0ef3e
Python      : 3.13.5
OS          : macOS (Darwin 25.0.0)
Surface     : src/continuum/actions/idempotency.py:557 (docstring only)
```

```python
from continuum.actions.idempotency import (
    _STOPWORDS,
    _WEAK_TOKENS,
    resolve_authorization_id,
)

for args in (
    {"status": "sent", "id": 1},
    {"count": 3, "status": "pending"},
    {"status": "queued"},
    {"count": 7},
    {"status": "true"},
):
    print(args, "->", resolve_authorization_id("api.call", None, args))

print("'pending' in _WEAK_TOKENS:", "pending" in _WEAK_TOKENS)
print("'pending' in _STOPWORDS  :", "pending" in _STOPWORDS)
```

### Current behaviour

Some status words produce an id and some do not:

```
{'status': 'sent', 'id': 1}        -> None
{'count': 3, 'status': 'pending'}  -> bbc36d982c1a350b
{'status': 'queued'}               -> 1972c274b75ae044
{'count': 7}                       -> None
{'status': 'true'}                 -> None

'pending' in _WEAK_TOKENS: False
'pending' in _STOPWORDS  : False
```

The docstring at `src/continuum/actions/idempotency.py:557` promises they never
do:

> Weak tokens (counts, status words, stopwords) never produce an id.

"counts" and "stopwords" hold up. "status words" does not. There is no
status-word category in the code. The actual rule is at
`src/continuum/actions/idempotency.py:294`:

```python
len(token) >= 3 and lowered not in _WEAK_TOKENS and lowered not in _STOPWORDS
```

`_WEAK_TOKENS` (`:156-177`) is a fixed frozenset of 18 words: `true`, `false`,
`null`, `none`, `tmp`, `var`, `home`, `user`, `bin`, `etc`, `usr`, `log`, `logs`,
`out`, `in`, `on`, `sent`, `file`. `_STOPWORDS` (`:182` onward) is grammatical:
`the`, `and`, `for`, `with`, `from`, and so on.

So `sent` is dropped because it happens to be one of those 18 words, and `true`
because it is another. `pending`, `queued`, `failed`, `retrying` and every other
status word an agent might send is three characters or more and on neither list,
so it becomes an identity token and resolves to a real authorization id. The
`{"status": "sent"}` case reads like the rule working as documented, when it is a
coincidence of vocabulary.

This is a documentation defect, not a behaviour defect. The code is doing what
its own predicate says. Filing it because the docstring is the natural place a
caller looks to decide whether their arguments will bind a budget, and it
currently gives an answer that testing contradicts.

Confirmed as a real trap rather than a theoretical one: while writing the
CHANGELOG entry for #412 in PR #609 I copied this sentence into the changelog,
and the verification script rejected it. That is one reviewer catch away from
shipping the same wrong claim in user-facing release notes.

### Expected behaviour

The sentence describes the predicate. Something in the shape of:

> A token is dropped when it is shorter than three characters or appears in the
> fixed weak-token or stopword lists. Arguments made only of such tokens produce
> no id.

That removes the false "status words" category and points at the two lists as
the thing to check. Naming them makes the docstring verifiable against
`:156-177` and `:182`, which the current wording is not.

Optionally, and worth a separate opinion: if dropping status words *was* the
intent, then the fix is in `_WEAK_TOKENS` rather than the prose. I have not
proposed that here, because widening the list changes which arguments bind a
budget, and that is a behaviour change with a blast radius that a wording fix
does not have. It also cannot be made complete: status vocabularies are
open-ended, and an explicit `key` is the supported answer for callers who need a
guaranteed identity. Happy to file that separately if a maintainer wants it
considered.

### Logs or CI run

Reproduced locally on `main` at `2e0ef3e`. Output quoted verbatim above. The
two membership checks are printed rather than asserted so the result is visible
either way.

Good first issue notes: the change is one sentence in one docstring, with no
runtime effect. The only care needed is reading `:294`, `:156-177` and `:182`
first and describing what they actually do, rather than rewording the existing
sentence.

Amiirhosseini:
﻿## Summary

- Corrects `resolve_authorization_id` so it describes the real token predicate (`len >= 3`, not in `_WEAK_TOKENS` / `_STOPWORDS`) instead of a non-existent "status words" category.
- Records the wording fix in the unreleased changelog.

Fixes #613

## Test plan

- [x] Compared docstring to `_is_strong_token` at `src/continuum/actions/idempotency.py`
- [x] No runtime change (docstring + changelog only)
- [ ] Maintainer review that the new sentence matches the lists at `_WEAK_TOKENS` / `_STOPWORDS`

## Summary by CodeRabbit

* **Documentation**
  * Clarified authorization ID resolution documentation to accurately describe when tokens are ignored.
  * Documented that inputs containing only short, weak, or stopword tokens do not produce an ID.
  * No runtime behavior changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,13 @@ All notable changes to this project are documented here. The format follows
 
 ### Fixed
 
+- **`resolve_authorization_id` docstring matches the token predicate (#613).**
+  The function claimed that "status words" never produce an id. There is no
+  status-word category: a token is dropped only when it is shorter than three
+  characters or appears in `_WEAK_TOKENS` or `_STOPWORDS`. Words such as
+  `pending` and `queued` therefore bind an id, and the old sentence contradicted
+  testing. Wording only; no runtime change.
+
 - Make `continuum complete` idempotent for runs that are already completed (#356).
 
 - **Derived keys ignore surrounding whitespace in argument values (#361).**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,8 @@ All notable changes to this project are documented here. The format follows
   status-word category: a token is dropped only when it is shorter than three
   characters or appears in `_WEAK_TOKENS` or `_STOPWORDS`. Words such as
   `pending` and `queued` therefore bind an id, and the old sentence contradicted
-  testing. Wording only; no runtime change.
+  testing. The same false "status words" claim in `identity_tokens` is corrected
+  for consistency. Wording only; no runtime change.
 
 - Make `continuum complete` idempotent for runs that are already completed (#356).
 
--- a/src/continuum/actions/idempotency.py
+++ b/src/continuum/actions/idempotency.py
@@ -553,8 +553,9 @@ def resolve_authorization_id(
        by shared identity tokens anchors the id, reusing the same containment,
        derivation and location-agreement checks as ``ActionLedger.claim``'s
        fallback; without a ledger the id is the hash of the canonical leaf
-       tokens of the incoming arguments. Weak tokens (counts, status words,
-       stopwords) never produce an id.
+       tokens of the incoming arguments. A token is dropped when it is
+       shorter than three characters or appears in the fixed weak-token or
+       stopword lists. Arguments made only of such tokens produce no id.
     3. Unbound (neither key nor distinctive tokens, or an ambiguous
        multi-match): ``None`` is returned. The caller keeps today's
        behaviour (no authorization-bound budget) byte-identical.
--- a/src/continuum/actions/idempotency.py
+++ b/src/continuum/actions/idempotency.py
@@ -308,8 +308,9 @@ def identity_tokens(
     rendered as one, since a row id of ``4821`` identifies a row as well as
     ``INV-001`` identifies an invoice (issue #36) -- plus the basename and
     basename-stem of any path-like value, and the same for ``external_id``.
-    Weak tokens are dropped so the fallback never matches on incidental values
-    like counts or status words.
+    Weak tokens are dropped so the fallback never matches on incidental values:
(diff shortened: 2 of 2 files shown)


Amiirhosseini:
﻿Thanks Copilot — addressed in the latest commit: `identity_tokens` now describes the same predicate (`len >= 3`, `_WEAK_TOKENS` / `_STOPWORDS`) instead of the nonexistent "status words" category.

Cyrax321 (approved):

This PR corrects the resolve_authorization_id docstring to describe the real token predicate and applies the same correction to identity_tokens, addressing #613. I read _is_strong_token, _WEAK_TOKENS and _STOPWORDS, ran the issue's reproduction against the PR head and confirmed the reported behavior, and verified no "status words" wording remains in the module.

The Copilot consistency finding from the first revision is resolved by the fixup commit, and the changelog entry covers both docstrings. Targeted ledger and index tests pass and the change is wording only with no runtime effect.

I did not find any blocking issues. Ready to merge once required checks are green.

Thank you again for the careful documentation fix.

Issue #630, opened 2026-09-05
Docs test counts drifted again (1918 collected vs 2044 on main); add a guard so it cannot drift silently


Cyrax321:
## Context

#316 fixed the drift between README, CONTRIBUTING_ONBOARDING, and CHANGELOG, and #605 synced all three to ~1,918 collected. Since then the suite has grown again:

```bash
pytest --collect-only -q 2>&1 | tail -1
# 2044 tests collected
```

while the three files still say ~1,918. Same class of drift, new numbers.

## Why this matters

The onboarding guide is the first file a newcomer reads. Stale counts erode trust, and re-syncing by hand every few weeks does not scale.

## How to reproduce

```bash
pytest --collect-only -q 2>&1 | tail -1
grep -rn "1,918" README.md docs/CONTRIBUTING_ONBOARDING.md CHANGELOG.md
```

## Expected outcome

The durable fix has two parts:

1. Re-sync the three files to the current numbers, following the same honest wording and regeneration comments #605 established.
2. Add a small test (for example `tests/test_docs_counts.py`) that reads the documented counts and fails when `pytest --collect-only` disagrees beyond a stated tolerance, so the next drift breaks CI instead of silently landing. Note skips vary by environment (Postgres without CONTINUUM_TEST_POSTGRES_DSN, adapter tests without langgraph or openai-agents), so the guard must compare collected totals, not passed/skipped splits.

## Step by step for the fix

1. On `main`, run `pytest --collect-only -q 2>&1 | tail -1` and `pytest -q 2>&1 | tail -1`.
2. Update README.md, docs/CONTRIBUTING_ONBOARDING.md, and CHANGELOG.md to the same numbers, keeping the existing regeneration comments accurate.
3. Add the guard test and confirm it passes on the updated tree and fails when a count is edited back.
4. Run `grep` again to confirm one consistent figure.

## Files to touch

- `README.md`
- `docs/CONTRIBUTING_ONBOARDING.md`
- `CHANGELOG.md`
- `tests/test_docs_counts.py` (new)

Follows #316 and #605.

PR #644, opened 2026-09-06
docs: refresh README scale figures

### LINKED ISSUE #638: README scale figures stale: module and test-file counts no longer match the tree
## Context

`README.md` describes the project as "one library (`src/continuum`, 104 modules) plus a large test suite (98 test files, ~1,918 tests)". Measured on current `main`:

```bash
find src/continuum -name "*.py" | wc -l          # 121 (100 excluding __init__.py)
ls tests/test_*.py | wc -l                        # 151
pytest --collect-only -q 2>&1 | tail -1           # 2044 collected (see #630)
```

The module and test-file figures no longer match any counting of the tree (the test count drift is tracked separately in #630).

## Why this matters

Same reason as #316: the project description is the most-read paragraph in the repo, and wrong scale figures erode trust.

## How to reproduce

Run the three commands above on `main` and compare with the README sentence.

## Expected outcome

The sentence carries current figures with the counting convention stated (for example, whether `__init__.py` files count as modules), so the next refresh is unambiguous. Keep the existing regeneration comments accurate.

## Step by step for the fix

1. Agree the counting convention with the maintainer in a comment (total `.py` files vs non-`__init__` modules).
2. Update the sentence to the agreed figures.
3. Run markdownlint on README.md.

## Files to touch

- `README.md` (one sentence)

Tiny, verifiable, ideal for a first contribution. Related: #630 covers the test-count half.

Cyrax321:
## Description

Refreshes the stale scale sentence in the Module map section. All three figures were re-measured on current main.

## Related issues

Fixes #638

## Types of changes

- Type: `docs`

## Breaking changes

No.

## How has this been tested?

Measured on current main:

```bash
find src/continuum -name "*.py" | wc -l   # 121
ls tests/test_*.py | wc -l                   # 151
pytest --collect-only -q (summed per-file)   # 2044
```

`npx markdownlint-cli2 --config .markdownlint-cli2.jsonc README.md`: 0 errors.

## Checklist

Docs-only change, no behaviour change and no tests added. No security-relevant changes.

## Additional context

Counting convention (agreed on the issue): total `.py` files under `src/continuum`, matching the old 104 source-files usage. The other test-count mentions in README stay untouched for #630, which owns the guard mechanism. No translated README carries this sentence, so nothing drifts.

## Summary by CodeRabbit

- **Documentation**
  - Updated the README’s module map to reflect the library’s current size of 124 modules.
  - Confirmed that the documented test-suite counts remain unchanged.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,10 @@
 version: 2
+# The pre-commit ecosystem is deliberately not watched (#689). Dependabot
+# groups cannot cross ecosystems, so a pre-commit update would land as its own
+# PR bumping only .pre-commit-config.yaml, and the three-way ruff rev lockstep
+# (pyproject.toml, .pre-commit-config.yaml, CONTRIBUTING.md, enforced by
+# tests/test_precommit_config.py) would fail CI in that PR. Bump ruff in all
+# three places by hand in one PR.
 updates:
   - package-ecosystem: "pip"
     directory: "/"
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ demo-run/*
 
 # opencode local state
 .opencode/
+
+# Benchmark artifacts
+benchmarks/out/
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,7 +1,7 @@
 {
   "mcpServers": {
     "continuum-mcp": {
-      "command": "continuum-mcp",
+      "command": "${CLAUDE_PROJECT_DIR:-.}/.venv/bin/continuum-mcp",
       "args": ["--db", "continuum.db"],
       "env": {
         "CONTINUUM_MCP_MUTATING_CLIENTS": "claude-code"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,7 +13,7 @@ Guidance for AI coding agents (and their operators) working in this repository.
 
 ## Commit and push conventions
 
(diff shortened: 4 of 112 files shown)


Cyrax321 (commented):
Review. The figures check out within variance (I measure 121 modules, 151 test files, ~2,05x collected on current main; exact totals move with every merged test PR, and the honest-variance wording already says so).

One coordination problem: this edits the exact same lines as the already-open #664 (README verify block, onboarding counts, CHANGELOG entry), which re-syncs to 2,053/23 and adds a guard test asserting the three files agree. Whichever lands second will conflict, and landing #664 first then this would trip its own guard only if the gap exceeds tolerance 30 (2,056 versus 2,053 does not, so no false alarm, but the files would then disagree with the guard's pinned figure until re-synced). Cleanest order: land this first for the freshest numbers, then rebase #664 onto it so the guard pins 2,056. Not a defect in either PR.

Cyrax321 (commented):
Correction to my counts above, measured on pristine origin/main instead of my worktree: 121 modules (matches the PR), but 154 test files, not 151, so this PR's file count is already stale by three files from merges that landed after it was written. Same for the collected total, which moves with every merged test PR. This reinforces the sequencing point: rebase this onto current main and re-measure all three figures before merge, then land #664's guard right after so the numbers cannot rot silently again.

PR #646, opened 2026-09-06
fix: preserve archived history in authority and memory checks

### LINKED ISSUE #615: continuum compact makes spent grants and consumed authorities reusable again
#### Problem

`ActionLedger.claim` enforces the Authority Resurrection defence (#269, #289 via #555/#556) with two scans of the run's event log: one for spent single-use `grant` refs, one for `AUTHORITY_CONSUMED` events. Both read the **live** log only. `continuum compact` (#239) moves the pre-anchor prefix out of `events` into `events_archive`, so after compaction both scans see an empty history and the defence fails **open** — a spent grant or a consumed authority is accepted again under a fresh key.

The same function already handles this correctly one screen earlier. `_replay()` merges archived and live events, with a docstring explaining exactly why: "a claim settled before compaction must keep protecting afterwards, or exactly-once would quietly reset at the anchor boundary and a month-old side effect could fire a second time." The archive fold was applied to the exactly-once path and not to the two authority paths beside it.

#### Evidence

- `src/continuum/actions/ledger.py:860` — `spent, grants_by_key = scan_grants(self.storage.read_events(self.run_id))`
- `src/continuum/actions/ledger.py:899` — `consumed = collect_consumed_authorities(self.storage.read_events(self.run_id))`
- contrast `src/continuum/actions/ledger.py:518-532` — `_replay()` uses `merge(self.storage.read_archived_events(...), self.storage.read_events(...))`
- `src/continuum/storage/base.py:134` — `read_all_events()` is documented as the full-history read for precisely this hazard: "After compaction the live log holds only the anchor and tail; any provenance calculation that reads only live events would miss an archived `EXTERNAL_AGENT` fact and launder it to `DETERMINISTIC`."
- the other two enforcement seams read live-only the same way: `src/continuum/gateway.py:425` (enforcing HTTP gateway) and `src/continuum/cli/main.py:2376` (the `continuum gate` PreToolUse hook)
- `tests/test_compaction.py:258` (`test_completed_action_from_the_archived_prefix_does_not_re_run`) pins the archived-claim case for the **same** key, and `tests/test_gate_resurrection.py` pins resurrection across a restore — but no test compacts and then re-presents a spent grant or a consumed authority, which is why this is green.

Verified against `main` at 2e0ef3e.

#### Impact

`continuum compact` is presented as a maintenance operation that bounds replay cost while preserving history verbatim; nothing warns that it disables a security control. After it runs, the ACRFence Authority Resurrection attack the ledger was built to refuse succeeds: an agent re-presents a consumed approval token or one-time credential with drifted arguments, the ledger opens a fresh `STARTED` slot, and the gate and enforcing gateway — which only require a live claim — then let the call through.

Exactly-once for the *same* key still holds, because `_replay()` folds the archive. The hole is specific to the fresh-key / drifted-argument path, which is exactly the case `scan_grants` and `collect_consumed_authorities` exist to cover ("Idempotency keys help with exact replays but not with intent-level duplicates where fields change", #289).

#### Reproduction

Spent single-use grant, driven through the user-facing CLI:

```python
# claim + complete a refund carrying grant tok_9, then re-present tok_9 under a fresh key
led = ActionLedger(store, "run_1")
o = led.claim("issue_refund", {"order": "o-1"}, key="refund:o-1",
              grant={"id": "tok_9", "scope": "refund"})
led.complete(o.key, external_id="rf-1", result={})

def reuse(store, k):
    return ActionLedger(store, "run_1").claim(
        "issue_refund", {"orderId": "O-1", "cents": 4200}, key=k,
        grant={"id": "tok_9", "scope": "refund"})
```

```
BEFORE COMPACT: denied -> GrantDenied: grant 'tok_9' was already consumed by action action_a725c4f (completed); single-use authority does not survive a restore. Obtain a fresh grant from the issuer.
$ continuum compact run_1 --force  ->  exit 0 | Archived 5 event(s); the live log now starts at the anchor.
AFTER COMPACT: ALLOWED fresh=True status=started  <-- SPENT GRANT ACCEPTED
```

Consumed authority, same shape (`record_authority_consumed` then re-claim with the same `authority_id`):

```
BEFORE COMPACT: denied -> LedgerError: Authority 'tok_u' consumed at seq 4 by run 'run_1'. Obtain a fresh authority.
compact: {'archived': 5}
consumed (live only) : []
consumed (all events): ['tok_u']
AFTER COMPACT: ALLOWED fresh=True status=started <-- CONSUMED AUTHORITY ACCEPTED
```

The history is intact and the evidence is still on disk — only the scans stop looking at it:

```
scan_grants(live only)  spent ids: []
scan_grants(all events) spent ids: ['tok_9']
```

#### Expected behavior

Both scans cover the whole recorded history, so a grant spent or an authority consumed before the anchor is still refused after compaction, with the same messages naming the original consuming action.

#### Current behavior

After `continuum compact`, both scans return empty and the claim is accepted (`fresh=True`, `status=started`), while the same scans over `read_all_events` still report `tok_9` / `tok_u`.

#### Suggested direction

Point the two scans at the full history — `storage.read_all_events(run_id)`, or the same `merge(read_archived_events, read_events)` shape `_replay()` already uses on this hot path — and do the same at `gateway.py:425` and `cli/main.py:2376`, since all four are layers of the same defence.

`cli/main.py:2374`'s `fold_action_events(storage.read_events(run_id))` in the gate is a related live-only read, but it fails **closed** (a key missing from the fold denies as unclaimed and the agent is told to claim, where the ledger's archive-aware `_replay` catches it). Worth a look for verdict quality, not part of this hole.

#### Acceptance criteria

- A spent single-use grant is refused after `continuum compact`, with `GrantDenied` naming the archived consuming action.
- A consumed `AUTHORITY_CONSUMED` id is refused after `continuum compact` at all three enforcement points: `ActionLedger.claim`, `continuum gate`, and the enforcing gateway.
- Regression coverage that compacts and then re-presents both a spent grant and a consumed authority under a fresh key.
- A live mid-flight retry under the same key and grant still passes — the existing `live_match` / `live_auth_match` rules are unchanged.

### LINKED ISSUE #616: continuum forget --tenant misses archived memory writes after compaction
#### Problem

`cmd_forget` enumerates a tenant's memory writes by scanning `storage.read_events(run.run_id)` — the live log only. After `continuum compact` (#239) those `ACTION_RECORDED` rows live in `events_archive`, so `continuum forget --tenant X` reports zero records, exits 0, and writes no tombstone, while the writes are still recorded in the log. The command's own docstring and `docs/guides/memory_governance.md` promise the opposite: it "lists exactly what to delete externally".

The same live-only read is in the poisoning-forensics joins, so `forensic_lookup` / `forensic_join_across_runs` return no hits for records whose write events are still on disk.

#### Evidence

- `src/continuum/cli/main.py:2661` — `events = storage.read_events(run.run_id)` inside `cmd_forget`, whose docstring at `:2615-2629` says "The command lists exactly what to delete externally".
- `docs/guides/memory_governance.md:153` — "`continuum forget --tenant X` builds on this enumeration to list exactly what to delete externally and to append a tombstone event".
- `src/continuum/actions/ledger.py:1242`, `:1255` (`forensic_lookup`) and `:1339`, `:1352` (`forensic_join_across_runs`) — same live-only read. `docs/guides/memory_governance.md:164` presents these as how "sibling writes from the same contaminated origin can be enumerated".
- `src/continuum/storage/base.py:134` — `read_all_events()` is the documented full-history read, and is already used for the provenance DAG after #554 (`src/continuum/cli/main.py:686`, `:728`).
- `tests/test_forget.py` covers enumeration, tombstoning and `verify()` afterwards, but never compacts — which is why this is green.

Verified against `main` at 2e0ef3e.

#### Impact

Right-to-erasure is one of #304's four motivating failure classes: "erasure requests that cannot be honored because nobody can enumerate what the agent wrote." The runs where compaction is worth running are exactly the long-lived ones where the writes to erase are old, so the records most likely to be missed are the ones most likely to be requested.

The failure is silent and reads as a clean result: `Tenant 'acme': 0 record(s) found`, `(no matching memory records)`, exit code 0, no warning that part of the history was not searched. An operator honouring a deletion request concludes there is nothing to delete in the external store. Poisoning forensics degrades the same way — the blast radius of a contaminated origin looks empty.

#### Reproduction

```python
# three memory writes, two for tenant acme
ledger.claim("mem_write", {}, key="mem:pgvector_main:acme:rec-1", scoped_to_run=False)
ledger.claim("mem_write", {}, key="mem:pgvector_main:acme:rec-2", scoped_to_run=False)
ledger.claim("mem_write", {}, key="mem:pgvector_main:globex:rec-3", scoped_to_run=False)
```

```
BEFORE COMPACT: exit 0 -> ['rec-1', 'rec-2']
compact: exit 0 Archived 5 event(s); the live log now starts at the anchor.
after compact -> live: 2 archived: 5
mem writes visible in live log : 0
mem writes visible in full log : 3
AFTER COMPACT: exit 0 -> [] hits: 0
AFTER COMPACT (real): exit 0 -> tombstoned keys: []
```

Same run, the forensic joins:

```
forensic_lookup('rec-1')           -> 0 hit(s)
forensic_join_across_runs('rec-1') -> 0 hit(s)
...but the full log still holds    -> 1 matching action event(s)
```

#### Expected behavior

Enumeration covers archived plus live history, so `continuum forget --tenant acme` lists and tombstones `rec-1` and `rec-2` whether or not the run has been compacted, and the forensic joins return the archived hits.

#### Current behavior

After `continuum compact`, enumeration returns an empty record set and exits 0 with no tombstone written; the forensic joins return no hits.

#### Suggested direction

`storage.read_all_events(run.run_id)` in `cmd_forget` and in the two forensic joins — the same change #554 made for the provenance DAG. If a maintainer would rather keep `forget` scoped to the live log, then at minimum it should report that archived history was skipped instead of reporting zero, because the current output is indistinguishable from "nothing was ever written".

Filed separately from #615 (the authority scans have the same root cause) because the consequence, severity and acceptance criteria differ — matching the precedent set by #520 and #532 for the dashboard and CLI.

#### Acceptance criteria

- `continuum forget --tenant X` lists and tombstones the same record set before and after `continuum compact`, for both `--run-id` and store-wide enumeration.
- `ActionLedger.forensic_lookup` and `forensic_join_across_runs` return hits from the archived prefix, with `origin_digest` joins intact.
- Regression coverage in `tests/test_forget.py` that compacts and then runs `forget --dry-run` plus a forensic lookup.
- `verify()` still passes after the tombstone, keeping the #567 invariant.

be-student:
## Description

Compaction moves old events into the archive, but authority checks and memory-history consumers were still reading only the live tail. A completed grant could therefore be accepted under a fresh key after compaction, and archived memory writes disappeared from forget/forensic results.

Use the existing `Storage.read_all_events` contract for ledger grant and authority checks, CLI and HTTP gateway gate inputs, memory enumeration and forensic observation joins, and the non-indexed cross-run action fallback. Keep checkpoint-tail projection unchanged. Document the distinction in the storage contract and changelog.

## Related issues

Fixes #615
Fixes #616

## Types of changes

- Type: `bugfix`

## Breaking changes

No API or storage schema change. Previously missed archived facts now affect the same checks as live facts. Unfinished same-key retries retain the existing `UnknownSideEffect` behavior.

## How has this been tested?

Python 3.12.13 on macOS, editable `.[dev]` installation.

- `env -u NO_COLOR TERM=xterm .venv/bin/pytest`: 2,029 passed, 24 skipped.
- `.venv/bin/ruff check src/ tests/ examples/`: passed.
- `.venv/bin/ruff format --check src/ tests/ examples/`: passed.
- `.venv/bin/mypy src/continuum`: passed, 121 source files.
- `git diff --check`: passed.

Seven regression cases fail against unchanged upstream production code. Coverage uses real SQLite compaction and CLI dispatch, an actual local HTTP gateway, both grant and explicit-authority consumption, preserved unfinished-retry uncertainty, run-specific/store-wide forget, archived observation joins, non-indexed cross-run lookup, and hash-chain verification after denials and tombstones. The initial full run had two terminal-color failures with ambient color settings; the final command above controls those settings.

## Checklist

Tests added for changed behavior. Storage contract and changelog updated. Local validation is reported above; GitHub CI has not yet run on this PR. Security limitations are stated below.

## Additional context

The authority defense still depends on recorded consumption; this does not replace issuer-side credential validation. `forget` enumerates external deletion targets and appends a tombstone, but does not delete external data. Full-history scans include the archive and therefore scale with retained history. PostgreSQL integration was not run locally. A separate second-compaction failure (missing RUN_STARTED during anchoring) observed during testing remains outside this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Archived action history and memory records remain available after compaction.
  - Authority and grant enforcement continue to reject previously consumed credentials.
  - CLI and gateway decisions now account for complete event histories.
  - Forensics, cross-run scans, and memory enumeration include archived records.
  - Forgetting records continue to identify external deletion targets without deleting external data.
  - Unfinished retries retain existing uncertainty checks.

- **Documentation**
  - Clarified when complete event histories are required for enforcement and forensic operations.

- **Tests**
  - Added coverage for archived history, compaction, gateway enforcement, CLI behavior, and forensic lookups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -166,6 +166,13 @@ All notable changes to this project are documented here. The format follows
 
 ### Fixed
 
+- Preserve archived action history in grant and authority enforcement, CLI and
+  gateway gate decisions, cross-run action scans, and memory enumeration and
+  forensic joins (#615, #616). Compaction no longer hides spent authority or
+  memory records from those consumers. Existing uncertainty checks still govern
+  unfinished retries; forgetting records continues to enumerate external deletion
+  targets and append a tombstone, without deleting external data itself.
+
 - **The published image's default command runs the demo again (#280).**
   `Dockerfile` installs the sources root-owned under `/opt/continuum` and then
   drops to an unprivileged user, while `examples/crash_recovery_agent.py` pinned
--- a/src/continuum/actions/ledger.py
+++ b/src/continuum/actions/ledger.py
@@ -72,7 +72,7 @@ from continuum.budgets import (
     would_refuse,
 )
 from continuum.concurrency.lease import LeaseCoordinator
-from continuum.events import EventType
+from continuum.events import Event, EventType
 from continuum.models import Action, ActionStatus, ConsumedInputs, UnknownSideEffect, utcnow
 from continuum.security.hashing import stable_hash
 from continuum.storage.base import Storage
@@ -856,8 +856,10 @@ class ActionLedger:
         # before anything fires. A live attempt carrying the same grant under
         # the same key is an ordinary mid-flight retry and passes through.
         grant_clean = normalize_grant(grant)
+        authority_history: Sequence[Event] | None = None
         if grant_clean is not None:
-            spent, grants_by_key = scan_grants(self.storage.read_all_events(self.run_id))
+            authority_history = self.storage.read_all_events(self.run_id)
+            spent, grants_by_key = scan_grants(authority_history)
             prior = spent.get(grant_clean["id"])
             live_match = (
                 existing is not None
@@ -896,7 +898,9 @@ class ActionLedger:
         if authority_id is not None:
             from continuum.gate import collect_consumed_authorities
 
-            consumed = collect_consumed_authorities(self.storage.read_all_events(self.run_id))
+            if authority_history is None:
+                authority_history = self.storage.read_all_events(self.run_id)
+            consumed = collect_consumed_authorities(authority_history)
             prior_ev = consumed.get(authority_id)
             # Allow live retry under same key with same authority
             live_auth_match = (
--- a/src/continuum/actions/ledger.py
+++ b/src/continuum/actions/ledger.py
@@ -556,7 +556,7 @@ class ActionLedger:
         for run in self.storage.list_runs():
             if run.run_id == self.run_id:
                 continue
-            folded = fold_action_events(self.storage.read_events(run.run_id))
+            folded = fold_action_events(self.storage.read_all_events(run.run_id))
             candidate = folded.get(key)
             if candidate is not None:
                 found = candidate
@@ -857,7 +857,7 @@ class ActionLedger:
         # the same key is an ordinary mid-flight retry and passes through.
         grant_clean = normalize_grant(grant)
         if grant_clean is not None:
-            spent, grants_by_key = scan_grants(self.storage.read_events(self.run_id))
+            spent, grants_by_key = scan_grants(self.storage.read_all_events(self.run_id))
             prior = spent.get(grant_clean["id"])
             live_match = (
                 existing is not None
@@ -896,7 +896,7 @@ class ActionLedger:
         if authority_id is not None:
             from continuum.gate import collect_consumed_authorities
 
-            consumed = collect_consumed_authorities(self.storage.read_events(self.run_id))
+            consumed = collect_consumed_authorities(self.storage.read_all_events(self.run_id))
             prior_ev = consumed.get(authority_id)
             # Allow live retry under same key with same authority
             live_auth_match = (
@@ -1239,7 +1239,7 @@ class ActionLedger:
         try:
             from continuum.security.hashing import stable_hash as _sh
 
-            for ev in self.storage.read_events(self.run_id):
+            for ev in self.storage.read_all_events(self.run_id):
                 if ev.type == EventType.PERCEPTION_OBSERVED:
                     try:
                         digest = _sh(dict(ev.payload))
@@ -1252,7 +1252,7 @@ class ActionLedger:
                         obs_by_digest[ch] = ev
         except Exception:
             obs_by_digest = {}
-        for ev in self.storage.read_events(self.run_id):
+        for ev in self.storage.read_all_events(self.run_id):
             if ev.type not in _ACTION_EVENT_TYPES:
                 continue
             payload = dict(ev.payload)
@@ -1336,7 +1336,7 @@ def forensic_join_across_runs(storage: Storage, record_key: str) -> list[dict[st
         # Build per-run observation index
         obs_by_digest: dict[str, Any] = {}
         try:
-            for ev in storage.read_events(run_id):
+            for ev in storage.read_all_events(run_id):
                 if ev.type == EventType.PERCEPTION_OBSERVED:
                     try:
                         digest = _sh(dict(ev.payload))
@@ -1349,7 +1349,7 @@ def forensic_join_across_runs(storage: Storage, record_key: str) -> list[dict[st
         except Exception:
             obs_by_digest = {}
         try:
-            events = storage.read_events(run_id)
+            events = storage.read_all_events(run_id)
         except Exception:
             continue
         for ev in events:
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -2782,9 +2782,11 @@ def cmd_gate(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -> 
         )
         return 2
 
-    actions_by_key = fold_action_events(storage.read_events(run_id)) if run_id is not None else {}
+    actions_by_key = (
+        fold_action_events(storage.read_all_events(run_id)) if run_id is not None else {}
+    )
     consumed = (
-        collect_consumed_authorities(storage.read_events(run_id)) if run_id is not None else {}
+        collect_consumed_authorities(storage.read_all_events(run_id)) if run_id is not None else {}
     )
     decision = gate_decide(
         config,
@@ -3096,7 +3098,7 @@ def cmd_forget(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
 
     for run in runs_to_scan:
         try:
-            events = storage.read_events(run.run_id)
+            events = storage.read_all_events(run.run_id)
         except Exception:
             continue
         for ev in events:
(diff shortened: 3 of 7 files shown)


be-student:
Reused one full-history sequence for the gateway folds, CLI gate folds, and grant/authority checks, avoiding duplicate archive scans within those operations. Updated the storage contract to document reuse.

I retained the non-indexed cross-run fallback. `Storage.foreign_action()` explicitly returns `None` for backends without an index and requires callers to check `supports_action_index`; calling it unconditionally would lose duplicate protection for those backends. Indexed backends already take that path unchanged. The fallback's ordering across multiple runs is inherited behavior, and imposing an index on every backend would be a separate compatibility change. This PR adds archived events to that existing fallback and tests that a compacted foreign action still prevents a fresh claim.

Validation: 87 focused tests passed; full suite passed with 2,029 tests and 24 skips (`env -u NO_COLOR TERM=xterm .venv/bin/pytest`). Ruff lint/format and strict mypy pass.

Cyrax321 (approved):
Thank you for contributing to CONTINUUM. This PR corrects the archived-history handling across all affected enforcement and enumeration paths: ledger grant and authority checks, CLI and gateway gate inputs, forget enumeration, both forensic joins, and the non-indexed cross-run fallback now consume the full-history contract, with checkpoint-tail projection unchanged as documented.

I reviewed the complete diff against the base tree, including the Storage foreign_action contract, the scan and gate consumers, and the surrounding checkpoint logic. The snapshot reuse in the second commit addresses the earlier duplicate-scan feedback, and retaining the capability-gated fallback is consistent with the established storage interface. Verification is complete: focused suites pass, the full local suite reports zero failures, lint, format, and type checks are clean, and required CI, including the Postgres backend, is green. The new regression coverage fails on the unpatched tree as expected while preserving the live-retry guarantees.

No blocking issues remain. This PR is approved, and I appreciate the careful work.

Issue #611, opened 2026-09-02
Follow-up to #577: gateway refuses an unframed body on a live HTTP/1.1 connection, so the body is dispatched as the next request


Adhi1-2:
> **Labels to apply: `bug`, `security`.**
> Filed with `gh`, so the template's own `labels:` did not attach, and my
> permission on this repository is `pull` only, which GitHub does not allow
> to set labels. Applying them needs triage rights.

### Before reporting

- [x] I searched existing issues for a duplicate
- [x] I am reporting this against the correct repository
- [x] I am using a supported Python version (3.11 or later)
- [x] I can reproduce this on the latest commit on `main`

### Affected component

Core

### Severity

Major, workaround exists

### Regression check

- [ ] This worked in an earlier version or commit

The gap is as old as the branches themselves. It is not currently reachable on
`main`, because the hardening that introduces those branches was reverted in
`956bc92`. It becomes reachable the moment open PR #525 lands.

### System information

```
CONTINUUM   : PR #525 head, 690abdb (fix/http-body-hardening-522)
main        : 2e0ef3e
Python      : 3.13.5
OS          : macOS (Darwin 25.0.0)
Surface     : src/continuum/gateway.py, GatewayServer.Handler._body
```

### Command or code to reproduce

```python
# Against a gateway built from PR #525's head. Full script:
#   the refused request's body is itself a complete HTTP request,
#   sent in one write on one keep-alive socket.
import socket

SMUGGLED = (
    b"POST /v1/invoices HTTP/1.1\r\n"
    b"Host: api.example.com\r\n"
    b"Content-Length: 11\r\n"
    b"\r\n"
    b'{"id":"X1"}'
)

s = socket.create_connection(("127.0.0.1", gateway_port), timeout=4)
s.sendall(
    b"POST /v1/invoices HTTP/1.1\r\n"
    b"Host: api.example.com\r\n"
    b"Content-Length: abc\r\n"   # unparseable, so nothing is read
    b"\r\n" + SMUGGLED
)
# Count the status lines that come back.
```

### Current behaviour

Two responses come back on one socket. The refused request's body was parsed as
the next request and dispatched through the gate:

```
[malformed Content-Length, body is a whole request]
    HTTP/1.1 400 Bad Request
    HTTP/1.1 403 Forbidden
  responses: 2   SMUGGLED REQUEST DISPATCHED: True

[negative Content-Length, body is a whole request]
    HTTP/1.1 400 Bad Request
[...]

PR #651, opened 2026-09-06
test(gateway): smuggled body after malformed length is never dispatched

Cyrax321:
## Description

Regression test for #611: a refused head with an unparseable Content-Length carries no trustable body boundary, so the gateway must answer once and close. Sends malformed and negative Content-Length heads with a whole smuggled request as the body on one keep-alive socket and asserts exactly one response (400, Connection: close).

## Related issues

Refs #611 (the defect itself was fixed by the close-on-refusal in a72100e; this pins the attack end to end so it cannot regress silently)

## Types of changes

- Type: tests

## Breaking changes

No. Test-only, no runtime change.

## How has this been tested?

- New test_malformed_length_smuggled_body_is_never_dispatched (malformed and negative lengths): passes on main.
- With the close_connection line temporarily removed the new test fails with 2 responses (smuggled request dispatched), then passes again after restore, so it pins the fix.
- tests/test_gateway.py: 20 passed.
- .venv/bin/ruff check and format --check on the touched file: passed.

## Checklist

Tests added for changed behavior. No changelog entry (test-only, no user-visible change). CI has not yet run on this PR.

### DIFF
--- a/tests/test_gateway.py
+++ b/tests/test_gateway.py
@@ -383,3 +383,37 @@ def test_compacted_consumed_authority_still_denies(db: str, gateway: str) -> Non
     status, body = post(gateway, "/v1/invoices", {"id": "new", "authority_id": "spent"})
     assert status == 403
     assert "spent" in body["reason"] and "consumed at seq" in body["reason"]
+
+
+@pytest.mark.parametrize("bad_length", ["abc", "-5"])
+def test_malformed_length_smuggled_body_is_never_dispatched(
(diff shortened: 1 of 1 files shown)


Cyrax321 (commented):
Self-review (author). Smuggling regression test for malformed and negative lengths: passes on main, fails with 2 responses when the close line is removed. Gateway suite green. Test-only. Mergeable, no conflicts. Needs an independent approver.

PR #652, opened 2026-09-06
feat(adapters): thin-adapter ledger writes carry external-agent provenance

### LINKED ISSUE #612: thin.py docstring promises EXTERNAL_AGENT provenance, but thin-adapter runs record deterministic and skip the human gate
> **Labels to apply: `bug`, `adapter`, `provenance`.**
> Filed with `gh`, so the template's own `labels:` did not attach, and my
> permission on this repository is `pull` only, which GitHub does not allow
> to set labels. Applying them needs triage rights.

### Before reporting

### Affected component

Adapters

### Severity

Major, workaround exists

### Regression check

The docstring and the behaviour have disagreed since the thin adapters landed.

### System information

```
CONTINUUM   : main at 2e0ef3e
Python      : 3.13.5
OS          : macOS (Darwin 25.0.0)
Surface     : src/continuum/adapters/thin.py (ContinuumToolGuard)
```

```python
import tempfile
from pathlib import Path

from continuum.adapters.thin import ContinuumToolGuard

with tempfile.TemporaryDirectory() as td:
    with SQLiteStorage(str(Path(td) / "t.db")) as store:

        guard = ContinuumToolGuard(store, "r1")
        guard.complete(guard.claim("notify.customer", {"order_id": "O-9"}))

        for e in store.read_events("r1"):
            print(e.type.value, e.source.value)

        d = RecoveryEngine(store).assess("r1")
        print("mode:", d.mode.value, "safe:", d.safe)
```

### Current behaviour

Every event the thin adapter writes is recorded `deterministic`, and the run
resumes with no human in the loop:

```
RUN_STARTED       deterministic
ACTION_RECORDED   deterministic
ACTION_RECORDED   deterministic
mode: resume safe: True
```

The module docstring at `src/continuum/adapters/thin.py:16` says the opposite:

> All three route through one shared engine: claim before, evidence after,
> exactly-once under argument drift via optional stable keys. **Provenance is
> EXTERNAL_AGENT because the framework asserts the facts.**

The cause is a defaulted keyword, not a wrong constant. `ContinuumToolGuard`
writes through `ActionLedger`, and `ActionLedger` never passes `source=` on any
of its three `append_event` calls (`src/continuum/actions/ledger.py:762`, `:869`,
`:909`). `Storage.append_event` declares `source: Origin = Origin.DETERMINISTIC`
(`src/continuum/storage/base.py:262`, mirrored at
`src/continuum/storage/sqlite.py:323`), so the default lands on every event.

`grep -rn EXTERNAL_AGENT src/continuum/` shows `thin.py:16` is the only
occurrence in that file: prose, never an argument. The sibling adapters do pass
it. `adapters/generic.py:133`, `adapters/openai.py:430` and
`adapters/langgraph_store.py:93` all call `append_event(..., source=Origin.EXTERNAL_AGENT)`.

The consequence is a missing gate, not just a stale comment.
`references/adapters.md` states the invariant plainly:

> State reported over MCP, or through the OpenAI adapter, uses
> `Origin.EXTERNAL_AGENT` provenance, which the validator marks
> `REQUIRES_REVIEW`. That is intentional: an agent must not validate its own
> unverified work.

A run driven entirely by CrewAI, AutoGen or Pydantic AI through
`ContinuumToolGuard` is therefore indistinguishable from one CONTINUUM
orchestrated itself. It resolves to `resume` / `safe=True` and skips the review
the docstring, the reference doc and the OpenAI adapter all promise.

`references/adapters.md` also does not yet list the thin adapters at all, which
is tracked separately as #267, so there is no second place where the intended
provenance for these three frameworks is written down.

### Expected behaviour

One of two, and the choice is a maintainer call because they differ in blast
radius:

**A. Make the behaviour match the docstring.** Give `ContinuumToolGuard` a way
to stamp its ledger writes `Origin.EXTERNAL_AGENT`, so a CrewAI or AutoGen run
resolves to `request_human` like the OpenAI adapter does. This is the reading the
docstring and `references/adapters.md` support, and it closes a real gate. It
changes the recovery verdict for existing thin-adapter runs, so it needs a
CHANGELOG entry and probably a note in the reference doc. `ActionLedger` would
need to accept and forward a `source`, since it currently has no parameter for
it. That is the wider change of the two.

**B. Correct the docstring.** If ledger events are deliberately `deterministic`
because the *claim* is CONTINUUM's own bookkeeping rather than a framework
assertion, then line 16 is simply wrong and should say so, and the reference doc
should record that the thin adapters resume without a human gate. This is a
one-line change with no runtime effect.

I lean towards A, because the docstring's stated reason ("because the framework
asserts the facts") is the same reason the OpenAI adapter uses
`EXTERNAL_AGENT`, and a tool call executed by CrewAI is no more self-verified
than one executed by the Agents SDK. But B is defensible if the intent was
narrower than the sentence, and I would rather ask than assume.

Whichever is chosen, a test should pin it, so the two cannot drift apart again.
There is precedent for that being worth doing: #182 was the same class of bug in
`recovery ledger` ("module docstring promises append-only but compact rewrites
entries in place").

### Logs or CI run

Reproduced locally on `main` at `2e0ef3e` with the script above. Output quoted
verbatim. The contrast with `adapters/generic.py:133` was confirmed by grep, not
inferred.

Cyrax321:
## Description

Option A from #612: make the behavior match the docstring. ContinuumToolGuard claims about framework-executed tools were recorded deterministic, so agent-asserted effects laundered to trusted and derived provenance hid the writer. ActionLedger now accepts a source (default deterministic, used only for action records; denial records stay deterministic as ledger verdicts) and the guard stamps EXTERNAL_AGENT, matching its module docstring and the OpenAI adapter. The DETERMINISTIC enum doc is clarified so in-process framework execution is not misread as trusted local orchestration.

## Related issues

Fixes #612

## Types of changes

- Type: bugfix

## Breaking changes

No API break (new optional parameters). Behavior change scoped to thin-adapter runs: their action records now carry EXTERNAL_AGENT, so derived provenance reports the true writer. All other ledger writers keep byte-identical output.

## How has this been tested?

- 3 new tests in tests/test_adapters_thin.py: guard writes carry EXTERNAL_AGENT with derived provenance to match, default ledger stays deterministic, source override respected. The 2 behavior tests fail on unpatched code.
- Reproduced the issue script before (all deterministic, derived deterministic) and after (action records external_agent, derived external_agent).
- env -u NO_COLOR TERM=xterm .venv/bin/pytest: 2,033 passed, 23 skipped.
- .venv/bin/ruff check src/ tests/ examples/: passed.
- .venv/bin/ruff format --check src/ tests/ examples/: passed.
- git diff --check: passed.

## Checklist

Tests added for changed behavior. Changelog updated. CI has not yet run on this PR.

## Additional context

One honest boundary: the run-level human gate keys off goal and progress provenance, which still depends on who asserted those, so the issue's minimal repro script (operator-owned goal) still resumes. What changes is that framework-asserted effects can no longer launder to trusted anywhere derivations consume them. MCP and sidecar ledger construction still uses the default; widening that is a deliberate verdict change for every MCP run and is left as a follow-up, not smuggled in here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -166,6 +166,13 @@ All notable changes to this project are documented here. The format follows
 
 ### Fixed
 
+- Thin-adapter ledger writes carry EXTERNAL_AGENT provenance (#612).
+  `ContinuumToolGuard` claims about framework-executed tools were recorded
+  `deterministic`, so agent-asserted effects laundered to trusted and derived
+  provenance hid the writer. `ActionLedger` accepts a `source` (default
+  unchanged) and the guard stamps `EXTERNAL_AGENT`, matching its docstring
+  and the OpenAI adapter. Denial records stay deterministic as ledger verdicts.
+
 - Preserve archived action history in grant and authority enforcement, CLI and
   gateway gate decisions, cross-run action scans, and memory enumeration and
   forensic joins (#615, #616). Compaction no longer hides spent authority or
--- a/src/continuum/actions/ledger.py
+++ b/src/continuum/actions/ledger.py
@@ -73,7 +73,7 @@ from continuum.budgets import (
 )
 from continuum.concurrency.lease import LeaseCoordinator
 from continuum.events import Event, EventType
-from continuum.models import Action, ActionStatus, ConsumedInputs, UnknownSideEffect, utcnow
+from continuum.models import Action, ActionStatus, ConsumedInputs, Origin, UnknownSideEffect, utcnow
 from continuum.security.hashing import stable_hash
 from continuum.storage.base import Storage
 
@@ -360,7 +360,17 @@ class ActionLedger:
         lease: LeaseCoordinator | None = None,
         holder_id: str | None = None,
         ttl: timedelta | None = None,
+        source: Origin = Origin.DETERMINISTIC,
     ) -> None:
+        """Bind this ledger to a run.
+
+        ``source`` stamps the action records this ledger writes
+        (issue #612): the writer, not the derivation. Callers whose
+        claims are asserted by an autonomous agent about its own work
+        pass ``Origin.EXTERNAL_AGENT`` so the validator holds the run
+        for review. Denial records stay ``DETERMINISTIC``: they are the
+        ledger's own verdicts, not caller assertions.
+        """
         if lease is not None and not holder_id:
             raise ValueError(
                 "holder_id is required when a lease is supplied: a shared default "
@@ -369,6 +379,7 @@ class ActionLedger:
             )
         self.storage = storage
         self.run_id = run_id
+        self._source = source
         self._lease = lease
         self._holder_id = holder_id or ""
         self._ttl = ttl
@@ -759,7 +770,7 @@ class ActionLedger:
             # rather than explicit key; keep forensic searchable.
             with suppress(Exception):
                 payload["rendered_key"] = str(action.arguments.get("record_key"))
-        self.storage.append_event(self.run_id, event_type, payload)
+        self.storage.append_event(self.run_id, event_type, payload, source=self._source)
         return action
 
     @_single_writer
--- a/src/continuum/adapters/thin.py
+++ b/src/continuum/adapters/thin.py
@@ -24,6 +24,8 @@ from __future__ import annotations
 from collections.abc import Callable
 from typing import Any
 
+from continuum.models import Origin
+
 __all__ = [
     "crewai_available",
     "pydantic_ai_available",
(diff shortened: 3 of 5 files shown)


Cyrax321 (commented):
Self-review (author). ActionLedger source parameter defaulting deterministic, thin guard stamping external-agent, denial records unchanged, enum doc clarified. Three tests: two fail without, default-guard passes both ways. Derived provenance verified honest end to end. Full suite green at push. Mergeable, no conflicts. Needs an independent approver. Follow-up #653/#676 widens this to MCP and sidecar ledgers deliberately.

PR #663, opened 2026-09-06
feat(provenance): validate finding caused_by on every append path

Cyrax321:
## Description

Step 2 of #597 (findings half): FINDING_ADDED payloads accept caused_by causal links under the same 32-id, 1-128-char caps and unknown-id refusal as decisions and actions. One shared CAUSED_BY_TYPES constant in continuum.events replaces the repeated type tuples on all six validation sites (memory log, SQLite append plus sealed paths, Postgres equivalents). The graph fold already edges any node type, so finding-to-evidence edges appear with no projection change. PLAN_UPSERT nodes stay a separate design question, stated in the changelog.

## Related issues

Refs #597 (pagination landed in #654; Postgres live verification still open)

## Types of changes

- Type: feature

## Breaking changes

No. Additive validation: payloads that were silently accepted with dangling links are now refused, matching decision and action behavior since #551.

## How has this been tested?

Python 3.14, editable installation.

- 3 new tests: finding-to-evidence edge in the built graph, unknown-id refusal, caps enforcement. The refusal and caps tests fail on unpatched code.
- env -u NO_COLOR TERM=xterm .venv/bin/pytest: 2,033 passed, 23 skipped.
- .venv/bin/ruff check and format --check: passed.
- .venv/bin/mypy src/continuum: passed, 121 source files.
- git diff --check: passed.

## Checklist

Tests added for changed behavior. Changelog updated. CI has not yet run on this PR. Postgres path changed identically but not run live locally; CI covers it.

PR #654, opened 2026-09-06
feat(cli): paginate provenance and impact listings display-only

Cyrax321:
## Description

Step 1 of #597: continuum provenance and continuum impact accept --limit and --offset. Paging truncates the rendered node lists only; the in-memory graph behind staleness and validation stays whole, mirroring the tree --limit contract (#321). JSON carries nodes_total/nodes_hidden (provenance) and downstream_total/downstream_hidden (impact). --limit 0 and negative --offset are refused rather than clamped. The --dot path keeps full fidelity.

## Related issues

Refs #597 (pagination step; caused_by extension and Postgres live test stay separate)

## Types of changes

- Type: feature

## Breaking changes

No. Default output is unchanged except additive JSON total fields.

## How has this been tested?

Python 3.14, editable installation.

- New tests/test_provenance_paging.py (6 tests): limit truncates with hidden line, offset skips from the front, JSON totals reconcile, limit 0 refused, impact limit plus JSON totals, unflagged output shape unchanged.
- env -u NO_COLOR TERM=xterm .venv/bin/pytest: 2,036 passed, 23 skipped.
- .venv/bin/ruff check src/ tests/ examples/: passed.
- .venv/bin/ruff format --check src/ tests/ examples/: passed.
- .venv/bin/mypy src/continuum: passed, 121 source files.
- git diff --check: passed.

## Checklist

Tests added for changed behavior. Changelog updated. CI has not yet run on this PR.

Issue #597, opened 2026-09-02
Follow-up to #288: provenance graph pagination and scale for large runs


Cyrax321:
Parent epic #288 landed caused_by payload (#551), DAG projector with Origin (#552), N-hop staleness (#553), and DOT plus compaction survival (#554). All four sub-issues are closed and merged to main (cab5b5f, 4047e9f, 35f4fe9, 318542e).

This follow-up tracks what was intentionally out of scope for the initial wave and should be handled separately.

**What remains**

- The current projector builds the full graph in O(n) and renders all nodes: continuum provenance <run> prints every node, and to_dot emits every edge. For a run with 10k+ evidence and decisions, the text output is unwieldy and the DOT file is large. The existing tree --limit pattern shows that truncation with an explicit children_hidden count is the preferred UX, but the provenance CLI has no --limit or pagination yet.
- caused_by is currently only on DECISION_CREATED and ACTION_RECORDED. FINDING_ADDED, EVIDENCE_ADDED and PLAN_UPSERT could also carry causal links (e.g., finding derived from evidence, plan step derived from finding), but the initial scope guardrail kept the surface minimal to two types. Extending to additional types needs payload models and validation, but is not required for the core evidence -> decision -> action chain.
- Compaction survival is covered for SQLite via read_all_events and tested, but Postgres is only stubbed (psycopg not present in CI for this run). A dedicated Postgres test with a live service container should verify the same.

**Why not in this wave**

Board #550 sized each sub-issue to be reviewable and collision-free. Adding pagination, extra payload types, and Postgres live verification in the same wave would have grown the PRs beyond the house rule of small focused commits and would have collided with the parallel authority and admissibility tracks.

**Proposed next steps**

1. Add --limit and --offset to continuum provenance and continuum impact, with JSON fields nodes_total and nodes_hidden similar to tree.
2. Consider caused_by on FINDING_ADDED and PLAN_UPSERT behind the same 32-id, 1-128 caps and validation.
3. Add a Postgres live test for compaction survival (requires service container, skip locally).

**Acceptance for this follow-up**

- Pagination truncates display but not the underlying graph used for staleness (the validator still sees the full graph).
- No new required fields; old events without caused_by still project as empty.
- No em dashes, no AI fingerprints, house style preserved.

Refs #288
Refs #550

PR #673, opened 2026-09-06
test(postgres): provenance graph survives compaction via archive

Cyrax321:
## Description

Step 3 of #597: dedicated Postgres compaction-survival test for the provenance graph. Mirrors the SQLite archive test against a live Postgres: evidence, finding (with caused_by link), and decision events, checkpoint, compact, then rebuild via read_all_events and assert identical nodes and edges plus verify() clean. DSN-gated skip pattern copied from tests/test_storage_postgres.py, so local runs skip and CI exercises it via the service container.

## Related issues

Refs #597 (pagination in #654, finding links in #663 pending merge; this is the live-Postgres step)

## Types of changes

- Type: tests

## Breaking changes

No. Test-only.

## How has this been tested?

- Locally skips cleanly without CONTINUUM_TEST_POSTGRES_DSN (1 skipped).
- Identical step sequence validated against SQLite: 2 edges and 3 nodes before and after compact, verify clean.
- read_all_events is the shared base implementation over the Postgres read_archived_events plus read_events; compact_run, put_version, and verify_events signatures checked against the Postgres backend.
- CI (including the Postgres backend job) runs the real thing.

## Checklist

Test-only, DSN-gated like the existing Postgres contract tests. CI has not yet run on this PR.

### DIFF
--- a/tests/test_provenance_postgres.py
+++ b/tests/test_provenance_postgres.py
@@ -1,0 +1,74 @@
+"""Provenance graph compaction survival on PostgreSQL (#597, step 3).
+
+Mirrors the SQLite archive test against a live Postgres: compact moves the
+prefix to events_archive and the graph rebuilt via read_all_events keeps
+every node and edge, including finding-to-evidence links. Skips cleanly
+without CONTINUUM_TEST_POSTGRES_DSN; CI exercises it via a service
+container, following tests/test_storage_postgres.py.
+"""
+
+from __future__ import annotations
+
+import os
+
+import pytest
+
+from continuum.events import EventType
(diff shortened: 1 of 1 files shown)


Cyrax321 (commented):
Self-review (author). Single DSN-gated Postgres test mirroring the SQLite archive case plus finding edges. Skips cleanly locally; identical step sequence validated against SQLite (2 edges, 3 nodes, verify clean). Base read_all_events serves both backends. Test-only. Mergeable, no conflicts. Needs an independent approver. The real verdict arrives from the Postgres CI job.

PR #675, opened 2026-09-06
perf(storage): merge archive and live streams in linear time

Cyrax321:
## Description

Answers the performance thread on #650: read_all_events sorted the concatenated archive plus live tail on every call, so each full-history fold on a long compacted run paid O(n log n) per invocation. Both streams arrive sequence-ordered from their queries, so heapq.merge folds them in O(n). heapq.merge is stable, matching sorted for equal sequences, and the compact-twice ordering test pins sequence order plus uniqueness across the boundary.

## Related issues

Refs #650 (performance follow-up; correctness fix stays untouched)

## Types of changes

- Type: performance

## Breaking changes

No. Identical output ordering, linear instead of linearithmic.

## How has this been tested?

- New test_read_all_events_merges_in_sequence_order: compact plus tail appends, then asserts global order, uniqueness, and payload order.
- env -u NO_COLOR TERM=xterm .venv/bin/pytest: 2,031 passed, 23 skipped.
- git diff --check: passed.

## Checklist

No behavior change by construction (same order, cheaper). CI has not yet run on this PR.

### DIFF
--- a/src/continuum/storage/base.py
+++ b/src/continuum/storage/base.py
@@ -31,6 +31,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Iterable, Mapping, Sequence
+from heapq import merge
 from types import TracebackType
 from typing import Any, ClassVar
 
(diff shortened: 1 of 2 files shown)


Cyrax321 (commented):
Self-review (author). Linear merge of the two ordered streams replacing re-sort; ordering and uniqueness pinned across the archive boundary. No behavior change by construction. Full suite green at push. Answers the perf thread on #650. Mergeable, no conflicts. Needs an independent approver.

Issue #305, opened 2026-08-24
Webhook-out: notify a human when a run blocks on request_human


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

### What problem does this solve?

When the recovery engine returns `request_human` and `safe: false`, the only ways to learn about it are polling (`continuum resume`, `continuum status`) or watching the dashboard. Nothing pushes the verdict out to a human. A long-running unattended agent can sit blocked for hours because nobody looked.

The README lists this under Honest limitations as "No webhook-out for request_human notifications yet", so the gap is acknowledged but untracked. The human-in-the-loop door itself exists (#242 added confirm/reconcile/complete buttons on the dashboard with audit rows); what is missing is the bell next to the door.

This matters more as multi-agent families land (#243): a parent run composes the worst child state, so one blocked child silently parks an entire hierarchy until someone polls.

### Proposed solution

A small opt-in notifier invoked whenever `assess` produces `request_human` (and optionally on `requires_review`), reading its configuration from the existing registry convention (data beside code, not code):

1. `.continuum/webhooks.json` declares endpoints (URL, secret or HMAC key, event filter, retry policy).
2. The recovery path emits one signed POST per transition into a blocked state, not per re-assessment, so a cron that calls `resume` every minute does not spam. Dedup on (run_id, mode, contract hash) with a re-notify interval.
3. Payload mirrors the resume JSON contract (mode, safe, family_rationale, children) plus a dashboard deep link, so the recipient can act from the message.
4. Delivery is fire-and-forget with bounded retries and a dead-letter row in the event log (new event type, EXTERNAL origin provenance). Notification failure must never flip a recovery verdict; the safety decision and the bell stay decoupled.
5. CLI surfacing via `continuum health` or a `notify-test` command so operators can verify wiring without manufacturing a real blockage.

HMAC-signing follows the same operator-secret discipline as `CONTINUUM_MCP_CONFIRM_TOKEN`: the receiving side must be able to trust that the notification came from CONTINUUM and was not forged by the very agent being reported on. That distinction is load-bearing given the self-certification fix (`9738b9e`).

### Alternatives considered

- Dashboard-only polling (#242): already shipped, but requires a human to look, which is the failure mode above.
- OTel bridge events: observability-grade export exists, but consuming spans is too heavy a requirement for a small team that just wants a Slack ping or an email.
[...]

PR #691, opened 2026-09-07
feat(notify): non-projecting dead-letter event plus recorder

Cyrax321:
## Description

Unit 4a of #305, stacked on the registry unit: NOTIFY_FAILED event type registered as non-projecting (audit only, never state), plus record_delivery_failure helper that appends url, event, and truncated error best-effort. One deliberate deviation for review: the issue text suggests EXTERNAL origin, but the failure is observed by local code, so the recorder uses the append default DETERMINISTIC like other system bookkeeping; say the word and it changes.

## Related issues

Refs #305 (assess hookup remains)
Depends on #690 (merge that first, then retarget this to main)

## Types of changes

- Type: feature

## Breaking changes

No. New event type and helper; projection explicitly ignores the type.

## How has this been tested?

- New audit-only test: dead letter present with url, projection progress and goal unchanged, verify clean.
- Ruff, format, and mypy clean on touched files. CI runs the full gate.

## Checklist

Tests added. No changelog entry: covered by the unit-1 entry at merge time.

### DIFF
--- a/src/continuum/events.py
+++ b/src/continuum/events.py
@@ -145,6 +145,11 @@ class EventType(StrEnum):
     # memory governance (issue #304, #567): per-tenant tombstone for erasure
     MEMORY_TOMBSTONED = "MEMORY_TOMBSTONED"
 
+    # human notification (issue #305): dead letter for a failed webhook
+    # delivery. Audit only, never state; DETERMINISTIC because the failure
+    # is observed by local code, not asserted by an agent.
+    NOTIFY_FAILED = "NOTIFY_FAILED"
+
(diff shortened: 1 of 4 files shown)


Cyrax321 (commented):
Self-review (author). Stacked on #690: NOTIFY_FAILED plus recorder, audit-only proven by unchanged projection and clean verify. Open design question stated in the body (DETERMINISTIC versus EXTERNAL origin for dead letters). Mergeable, no conflicts. Merge #690 first, then retarget. Needs an independent approver.

PR #708, opened 2026-09-07
fix(state): read an offset-less expires_at as UTC instead of crashing

### LINKED ISSUE #704: Naive expires_at payload bricks validation: TypeError comparing offset-naive and offset-aware datetimes
## Problem

`APPROVAL_GRANTED` accepts an `expires_at` payload that `datetime.fromisoformat` parses to a **naive** datetime (e.g. `"2027-01-01"`, or any ISO string without a UTC offset):

- `src/continuum/state/semantic.py:470` — `datetime.fromisoformat(str(expires_raw))` stores the naive value in `Approval.expires_at` (the model field has no tz normalization, `models.py:388`);
- `src/continuum/state/validator.py:874` — `approval.expires_at <= now` compares it against the tz-aware `utcnow()`.

A naive datetime is not an invalid approval — but the comparison raises `TypeError: can't compare offset-naive and offset-aware datetimes`, and neither `RecoveryEngine.assess` (`engine.py:260`) nor `validate_state` guards `validate` with a try/except. Every future recovery/validate for the run dies with an opaque TypeError: **the run becomes unrecoverable** because its persisted state can no longer be graded. This violates the validator's own contract that an unverifiable resource yields UNKNOWN rather than failing the recovery path.

## Repro (against `upstream/main` at `f5cb8d0`)

```python
from datetime import datetime
from continuum.models import Approval, ApprovalStatus, Goal, SemanticState
from continuum.state.validator import validate_state

state = SemanticState(
    run_id="r",
    goal=Goal(description="g"),
    approvals=[Approval(subject="s", status=ApprovalStatus.GRANTED,
                        expires_at=datetime(2027, 1, 1))],  # naive
)
validate_state(state)   # TypeError: can't compare offset-naive and offset-aware datetimes
```

Reproduced exactly. The same root cause exists at `semantic.py:942` (`now - pin.pinned_at` with a naive event timestamp) and potentially in `state/diff.py` when the two sides mix awareness.

## Why tests miss it

Every test datum is tz-aware (`test_validator.py:295,312`, `test_recovery_engine.py:194` uses `utcnow().isoformat()`). No test feeds a payload string without an offset — which is the single most likely shape an external caller emits.

## Suggested fix

Normalize at the boundary: when `fromisoformat` yields a naive datetime in the semantic fold, assume UTC (matching the project's everywhere-UTC convention), and/or make the validator's comparison defensively aware. Happy to send a PR if the maintainers prefer one approach over the other.

Adhi1-2:
Fixes #704

## Problem

An `APPROVAL_GRANTED` payload whose `expires_at` ISO string carries no UTC offset (e.g. `"2027-01-01"` — `fromisoformat` accepts date-only strings) folded to a **naive** datetime, while every downstream comparison uses the tz-aware `utcnow()`. The next `validate`/`assess` raised `TypeError: can't compare offset-naive and offset-aware datetimes`, taking down the whole validation pass and **bricking recovery for the run** — its persisted state could no longer be graded.

## Change

Two layers, matching the issue's suggestion:

1. **Boundary (semantic fold, `semantic.py`)**: a new `_as_utc()` helper normalizes a naive `expires_at` to UTC at the fold — the project convention is UTC everywhere, so a missing offset is read as UTC.
2. **Defense in depth (`_check_approvals`, `validator.py`)**: states persisted by older versions (or constructed directly) can already carry naive values; the validator normalizes before comparing so they **grade** (EXPIRED/VALID) instead of raising.

## Verification

- Repro from the issue no longer raises; a naive past `expires_at` grades EXPIRED
- New regression tests:
  - `test_naive_expires_at_is_folded_as_utc` (fold produces a tz-aware UTC datetime from an offset-less payload)
  - `test_a_naive_expires_at_grades_instead_of_raising` (validator grades instead of TypeError)
- Full suite green (`pytest --no-cov -q`, 100%), `ruff check` + `ruff format` clean

### DIFF
--- a/src/continuum/state/semantic.py
+++ b/src/continuum/state/semantic.py
@@ -198,7 +198,7 @@ def _as_utc(value: datetime) -> datetime:
 
     Event payloads are external input: an ``expires_at`` ISO string without a
     UTC offset (``"2027-01-01"``) parses to a naive datetime, and everything
-    downstream compares against the tz-aware ``utcnow()`` — which would raise
+    downstream compares against the tz-aware ``utcnow()``, which would raise
     TypeError and brick validation for the run (issue #704). The project
     convention is UTC everywhere, so a missing offset is read as UTC.
     """
--- a/src/continuum/state/semantic.py
+++ b/src/continuum/state/semantic.py
@@ -25,7 +25,7 @@ from __future__ import annotations
 
 from collections.abc import Iterable, Mapping
 from dataclasses import dataclass, field
-from datetime import datetime, timedelta
+from datetime import UTC, datetime, timedelta
 from typing import Any, Literal
 
 from continuum.events import Event, EventType
@@ -191,6 +191,18 @@ def _as_str_list(value: Any) -> list[str]:
     if isinstance(value, Iterable):
         return [str(v) for v in value]
     return [str(value)]
+
+
+def _as_utc(value: datetime) -> datetime:
+    """Interpret a naive datetime as UTC; aware values pass through.
+
+    Event payloads are external input: an ``expires_at`` ISO string without a
+    UTC offset (``"2027-01-01"``) parses to a naive datetime, and everything
+    downstream compares against the tz-aware ``utcnow()`` — which would raise
+    TypeError and brick validation for the run (issue #704). The project
+    convention is UTC everywhere, so a missing offset is read as UTC.
+    """
+    return value.replace(tzinfo=UTC) if value.tzinfo is None else value
 
 
 def _replace(items: list[Any], key: str, identifier: str, updated: Any) -> bool:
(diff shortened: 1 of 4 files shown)


Adhi1-2:
Removed the two em dashes from the added docstrings (4be42cd) per the AGENTS.md no-em-dash convention; no code change.

Cyrax321 (approved):

I reproduced the crash on main firsthand (naive expires_at folds tz unaware, then validate raises TypeError comparing against utcnow) and confirmed the PR branch folds to UTC with the suite green, including both new regression tests. Two-layer defense is the right shape here: boundary normalization plus grading for already-persisted naive values, since old rows cannot be re-folded away. UTC-everywhere is the documented project convention, so reading a missing offset as UTC is principled, not convenient.

One thing to fix before merge, outside this approval: the commits carry Co-Authored-By: Claude trailers and the PR body has a Generated with line. The project does not accept AI attribution in commits (AGENTS.md), so please amend the trailers out and drop that line.

PR #712, opened 2026-09-07
fix(models): register the origin_digest validator on Action

### LINKED ISSUE #710: Action.origin_digest SHA-256 validator is dead code: never registered, any string accepted
## Problem

The `@field_validator("origin_digest")` block in `src/continuum/models.py` is accidentally indented **inside the module-level `validate_caused_by` function, after its `return` statements** — an unreachable nested function that never registers on any model.

Result (reproduced against `upstream/main` at `f5cb8d0`):

```python
from continuum.models import Action
Action(run_id="r", action_type="t", origin_digest="NOT-HEX-AT-ALL")  # accepted
```

The field's own docstring promises "Stored as 64 lowercase hex (SHA-256) when present", and `forensic_lookup` joins on this digest — a garbage digest silently degrades the poisoning-forensics join instead of being refused at the boundary.

Notably, the `# type: ignore[misc]` added in d026104 (issue #566) to "suppress a mypy false positive" existed precisely *because* the decorator sat in the dead position: mypy was complaining about a decorator on a nested function. Once the validator is properly inside the `Action` class, mypy flags the ignore as unused.

## Why tests miss it

`test_memory_forensic.py::test_origin_digest_rejects_bad_hex` goes through `ActionLedger.claim`, whose **manual** regex check raises — masking the dead model validator. No test constructs `Action(origin_digest=...)` directly.

## Fix

Move the validator into the `Action` model class (and drop the now-unused type ignore). Patch ready to send.

Adhi1-2:
Fixes #710

## Problem

The `@field_validator("origin_digest")` block was accidentally indented inside the module-level `validate_caused_by` function, after its `return` statements — an unreachable nested function that never registered on any model. `Action(origin_digest="NOT-HEX-AT-ALL")` was accepted despite the field's documented 64-lowercase-hex (SHA-256) invariant, and the poisoning-forensics join silently degraded on garbage digests instead of refusing them at the boundary.

## Change

Move the validator into the `Action` model class. The `# type: ignore[misc]` from d026104 existed precisely because the decorator sat in the dead position (mypy complaining about a decorated nested function); properly placed, mypy now flags the ignore as **unused** — verified — so it is removed.

## Verification

- Runtime: bad digest rejected with "64 lowercase hex", `None` and valid 64-hex accepted
- New regression test `test_action_origin_digest_must_be_sha256_hex` (also covers uppercase rejection)
- `mypy src/continuum/models.py`: clean; full suite green (`pytest --no-cov -q`); `ruff check` + `ruff format` clean

### DIFF
--- a/src/continuum/models.py
+++ b/src/continuum/models.py
@@ -917,6 +917,15 @@ class Action(BaseModel):
     consumed_inputs: ConsumedInputs = Field(default_factory=ConsumedInputs)
     """Commitment inputs consumed to produce this action (issue #295)."""
 
+    @field_validator("origin_digest")
+    @classmethod
+    def _origin_digest_is_sha256(cls, value: str | None) -> str | None:
+        if value is None:
+            return None
+        if not _SHA256_PATTERN.fullmatch(value):
+            raise ValueError("origin_digest must be 64 lowercase hex characters")
+        return value
+
 
 class ActionRecordPayload(BaseModel):
     """Payload for ACTION_RECORDED (issue #551).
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):

Confirmed the bug on main firsthand: Action(origin_digest='NOT-HEX-AT-ALL') is accepted today because the validator sits nested after returns in validate_caused_by, exactly as described. The fix moves it onto the model, the mypy ignore removal is verified clean, and the models suite passes on the PR tree. Full CI is green, which also answers the blast-radius question: nothing in the tree relied on garbage digests passing validation.

One thing to fix before merge, outside this approval: the commit carries a Co-Authored-By: Claude trailer and the PR body has a Generated with line. The project does not accept AI attribution in commits (AGENTS.md), so please amend the trailer out and drop that line.

PR #713, opened 2026-09-07
fix(state): wrap a string evidence proposal instead of splitting it

### LINKED ISSUE #711: LLMExtractor splits a string evidence proposal into per-character citation ids
## Problem

`LLMExtractor._merge` coerces the evidence field of proposed decisions/findings with:

```python
evidence=[str(e) for e in raw.get("evidence", [])]   # extractor.py, two sites
```

An LLM emitting the citation as a **bare string** — the single most common shape drift in model output — is iterated per character:

```python
LLMProposal(decisions=[{"decision_id": "d1", "decision": "...", "evidence": "ev_42"}])
# -> decision.evidence == ['e', 'v', '_', '4', '2']
```

This violates the module's core asymmetry: the LLM "may only add components, never modify or delete" and enrichment "must never cost a recovery." Unlike a hard error (which the issue #40 fix catches and falls back), this does not raise — the corrupted citations are **persisted** into the state/checkpoint. Downstream, `dangling_evidence()` reports the phantom per-character ids, `_check_evidence` emits UNKNOWN, and every future validation of that checkpoint is blocked (`safe_to_resume=False` under default `strict_unknown`) by citations that never existed.

The deterministic fold handles this exact case correctly via `_as_str_list` (semantic.py:186) — a bare string wraps to `[value]`. The two `_merge` sites are the only inconsistent ones.

## Why tests miss it

Tests feed `evidence` as lists only; #40/#41 covered crash-on-malformed and duplicate-id handling — the surviving bug is the silent-corruption variant that doesn't raise.

## Fix

Use the same `_as_str_list` coercion in `_merge`. Patch ready to send.

Adhi1-2:
Fixes #711

## Problem

`LLMExtractor._merge` coerced the evidence field of proposed decisions/findings with `[str(e) for e in raw.get("evidence", [])]`. A model emitting the citation as a bare string — the most common shape drift in LLM output — was iterated per character: `evidence == ['e','v','_','4','2']`. The corrupted citations were **persisted**, then failed `dangling_evidence()`, emitting UNKNOWN and blocking resume under default `strict_unknown` — enrichment silently costing a recovery, which the module's core asymmetry forbids.

## Change

Use the same `_as_str_list` coercion the deterministic fold already applies to this exact case (semantic.py): a bare string wraps to `[value]`, `None` to `[]`, iterables to their string items. The two `_merge` sites were the only inconsistent ones.

## Verification

- New regression test `test_a_string_evidence_field_is_wrapped_not_split_into_characters` (both the decisions and findings sites)
- Full suite green (`pytest --no-cov -q`), `ruff check` + `ruff format` clean

### DIFF
--- a/src/continuum/state/extractor.py
+++ b/src/continuum/state/extractor.py
@@ -37,7 +37,7 @@ from continuum.models import (
     SemanticState,
     StateStatus,
 )
-from continuum.state.semantic import ProjectionReport, project_incremental
+from continuum.state.semantic import ProjectionReport, _as_str_list, project_incremental
 
 __all__ = [
     "ExtractionContext",
@@ -155,7 +155,7 @@ class LLMExtractor:
                     decision_id=decision_id,
                     decision=str(raw.get("decision", "")),
                     reason=str(raw.get("reason", "")),
-                    evidence=[str(e) for e in raw.get("evidence", [])],
+                    evidence=_as_str_list(raw.get("evidence")),
                     status=StateStatus.REQUIRES_REVIEW,
                     provenance=provenance,
                 )
@@ -173,7 +173,7 @@ class LLMExtractor:
                 Finding(
                     finding_id=finding_id,
                     claim=str(raw.get("claim", "")),
-                    evidence=[str(e) for e in raw.get("evidence", [])],
+                    evidence=_as_str_list(raw.get("evidence")),
                     confidence=min(max(confidence, 0.0), 1.0),
                     status=StateStatus.REQUIRES_REVIEW,
                     provenance=provenance,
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):

Verified on the PR tree: _as_str_list in semantic.py has exactly the stated semantics (None to [], bare string wraps, iterables to string items), the two _merge sites were the only inconsistent ones, and all 19 extractor tests pass including the new regression test for both sites. Reusing the deterministic fold's coercion instead of inventing a parallel one is the right call, and the blast radius is correctly limited to proposals.

One thing to fix before merge, outside this approval: the commit carries a Co-Authored-By: Claude trailer and the PR body has a Generated with line. The project does not accept AI attribution in commits (AGENTS.md), so please amend the trailer out and drop that line.

PR #714, opened 2026-09-07
fix(reconcile): settle archived actions and keep authority context after compaction

### LINKED ISSUE #647: reconcile --auto crashes on archived actions after compaction (LookupError in _key_for)
## Problem

continuum reconcile --auto (probe-driven settling via reconcilers.settle_run) crashes after continuum compact instead of settling archived uncertain actions. ActionLedger.pending is archive-aware through _replay, so archived STARTED actions are still listed as pending, but settle_run then resolves each action key through reconcilers._key_for, which folds only the live tail. The archived action is unfindable, so reconciliation raises LookupError and the run settles nothing.

## Evidence

- src/continuum/reconcilers.py:242: folded = fold_action_events(storage.read_events(run_id)) in _key_for, with raise LookupError when the action is absent.
- src/continuum/reconcilers.py:228: settle_run calls _key_for for every pending action, so one archived action aborts the whole report.
- Contrast src/continuum/actions/reconciliation.py:210: the library _key_for folds ledger._replay, which merges archived plus live events, so the non-probe path is unaffected.
- Same file, second live-only read: settle_authority (reconcilers.py:342) scans storage.read_events for the AUTHORITY_CONSUMED row, so post-compaction probes run with a bare authority_id payload missing consumer_run_id, via_action_id, and sequence.
- Verified on origin/main at d8e3a9a.

## Impact

Uncertain side effects are the highest-risk ledger state, and long-lived runs are exactly the ones worth compacting. After compaction the operator path for clearing that risk throws instead of settling, and authority probes silently lose their consumption context. The first failure is loud (traceback) and the second is silent (degraded probe input).

## Reproduction

```python
    store.create_run_started(Run(run_id="run_1", goal="t"))
    action = ActionLedger(store, "run_1").claim("refund", {}, key="k1").action
    CheckpointManager(store).checkpoint("run_1", force_version=True)
    store.compact_run("run_1")  # {'archived': 3}
    _key_for(store, "run_1", action)  # LookupError: action ... vanished from run 'run_1' mid-reconcile
```

## Expected behavior

_key_for resolves archived actions, so reconcile --auto settles post-compaction uncertainty exactly as it does pre-compaction, and settle_authority probes with the full consumption payload.

## Current behavior

LookupError aborts settle_run; authority probes run on an empty payload.

## Suggested direction

Read full history (storage.read_all_events) in both spots, the same one-line family as #646. No schema or API change.

## Acceptance criteria

- A claimed-but-unsettled action reconciles through reconcile --auto (or settle_run directly) after compact, dry-run and real.
- settle_authority payload carries the archived consumption row after compact.
- Regression test that compacts first, then reconciles and settles.
- Existing UnknownSideEffect and probe semantics unchanged.

Gouthamkrishnak18:
## What

Fixes #647.

`reconcile --auto` aborted with `LookupError` after a compaction, and authority probes silently lost their consumption context:

- `reconcilers._key_for` folded only the live tail (`storage.read_events`), so any pending action claimed before a compaction made `settle_run` raise `LookupError: action ... vanished from run ... mid-reconcile` and the whole settle report aborted. `ActionLedger.pending` is archive-aware (via `_replay`), so the crash hit exactly the actions the operator was trying to settle.
- `reconcilers.settle_authority` scanned only live events for the `AUTHORITY_CONSUMED` row, so post-compaction probes received a bare `authority_id` payload missing `consumer_run_id`, `via_action_id`, and `sequence`.

## How

Both spots now read full history via `storage.read_all_events`, the same archive-aware pattern the library reconciliation path (`actions/reconciliation.py`) already used. No schema or API change.

## Verification

- Reproduced the exact `LookupError` from the issue before the fix.
- Two new regression tests, both verified to **fail without the fix and pass with it**:
  - `tests/test_reconcilers.py::test_settle_run_settles_archived_actions_after_compaction`
  - `tests/test_authority_probe.py::test_probe_payload_keeps_consumption_context_after_compaction`
- Full `pytest` suite passes; `ruff check`, `ruff format --check`, and `mypy` (strict) all clean.
- `CHANGELOG.md` updated under `Unreleased -> Fixed`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic reconciliation after compaction by retaining access to archived action and authority-consumption details.
  - Prevented reconciliation failures when pending actions or authority records are no longer in the live event history.
  - Authority probes now receive complete consumption context, including the consumer run, action, sequence, and timestamp.

- **Tests**
  - Added coverage for reconciling archived actions and preserving authority context after compaction.

### DIFF
--- a/src/continuum/reconcilers.py
+++ b/src/continuum/reconcilers.py
@@ -236,10 +236,14 @@ def _key_for(storage: Storage, run_id: str, action: Action) -> Any:
     The fold is keyed by derived idempotency key while the Action record does
     not carry it, so recover the key by matching action_id against the run's
     folded ledger.
+
+    Folds full history, not the live tail: pending actions claimed before a
+    compaction live on as archived events (issue #647), and a live-only fold
+    would make every one of them "vanish mid-reconcile" and abort settle_run.
     """
     from continuum.actions.ledger import fold_action_events
 
-    folded = fold_action_events(storage.read_events(run_id))
+    folded = fold_action_events(storage.read_all_events(run_id))
     for key, candidate in folded.items():
         if candidate.action_id == action.action_id:
             return key
@@ -339,7 +343,10 @@ def settle_authority(
     from continuum.models import Origin
 
     payload: dict[str, Any] = {"authority_id": authority_id}
-    for ev in reversed(list(storage.read_events(run_id))):
+    # Full history, not the live tail: the AUTHORITY_CONSUMED row can predate a
+    # compaction (issue #647), and a live-only scan would hand the probe a bare
+    # authority_id, silently dropping the consumption context.
+    for ev in reversed(list(storage.read_all_events(run_id))):
         if (
             ev.type is not None
             and str(ev.type) == "AUTHORITY_CONSUMED"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -191,6 +191,18 @@ All notable changes to this project are documented here. The format follows
 
 ### Fixed
 
+- **`reconcile --auto` settles archived actions and probes authorities with
+  full consumption context after compaction (#647).**
+  `ActionLedger.pending` folds archived plus live events, but
+  `reconcilers._key_for` folded only the live tail, so one action claimed
+  before a compaction aborted the whole settle report with `LookupError`;
+  `settle_authority` scanned only live events for the `AUTHORITY_CONSUMED`
+  row, silently handing the probe a bare `authority_id` payload without
+  `consumer_run_id`, `via_action_id`, or `sequence`. Both now read full
+  history via `read_all_events`, the same archive-aware pattern the library
+  reconciliation path already used. Covered by `tests/test_reconcilers.py`
+  and `tests/test_authority_probe.py`.
+
 - Preserve archived action history in grant and authority enforcement, CLI and
   gateway gate decisions, cross-run action scans, and memory enumeration and
   forensic joins (#615, #616). Compaction no longer hides spent authority or
--- a/tests/test_authority_probe.py
+++ b/tests/test_authority_probe.py
@@ -228,3 +228,40 @@ def test_authority_probe_no_probe_registered_leaves_blocked(tmp_path: pathlib.Pa
         assert "auth-no-probe" in consumed
     finally:
         storage.close()
+
+
+def test_probe_payload_keeps_consumption_context_after_compaction(
+    tmp_path: pathlib.Path,
+) -> None:
+    """Post-compaction probes must still see how the authority was consumed (issue #647).
+
+    The AUTHORITY_CONSUMED row can predate a compaction. settle_authority used
+    to scan the live tail only, silently handing the probe a bare authority_id
+    payload with no consumer_run_id, via_action_id, or sequence.
+    """
+    storage = _storage()
+    try:
+        record_authority_consumed(storage, "run_1", "auth-compact-1", via_action_id="act-1")
+        storage.compact_run("run_1", through_sequence=storage.last_sequence("run_1"))
+        # Premise guard: the consumption row really is archived, not live.
+        live_types = [e.type for e in storage.read_events("run_1")]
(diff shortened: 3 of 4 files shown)


gouthamkrishnak2003:
Codecov flagged the one uncovered branch in `reconcilers.py` — the `settle_authority` consumption scan's loop-exhaustion path (an authority with no `AUTHORITY_CONSUMED` row; every existing test finds a row and breaks). Added `test_probe_for_an_authority_that_was_never_consumed_settles_with_a_bare_id` to cover it: probes an authority that was never consumed and asserts it settles with the bare `authority_id` payload (b0f89a7).

src/continuum/reconcilers.py
```diff
@@ -236,10 +236,14 @@ def _key_for(storage: Storage, run_id: str, action: Action) -> Any:
...
     folded ledger.
+
+    Folds full history, not the live tail: pending actions claimed before a
+    compaction live on as archived events (issue #647), and a live-only fold
+    would make every one of them "vanish mid-reconcile" and abort settle_run.
     """
     from continuum.actions.ledger import fold_action_events
 
-    folded = fold_action_events(storage.read_events(run_id))
+    folded = fold_action_events(storage.read_all_events(run_id))
```
Cyrax321, line 246: Non-blocking follow-up, not needed for this PR: `settle_run` calls this once per pending action, so a run with many pending actions now re-reads and re-folds the full archived history on every iteration. The `read_all_events` docstring in `storage/base.py` explicitly advises reusing the returned sequence within one operation. A follow-up that folds once in `settle_run` and passes the fold in would keep long compacted runs fast. Suggest filing it as a separate issue rather than growing this one.

Cyrax321 (changes requested):

This PR fixes #647: reconcile --auto aborted with LookupError on archived actions, and authority probes lost consumption context after compaction. I verified the root cause against the issue cited lines, confirmed read_all_events is an ABC-level helper that changes nothing for backends without an archive, and checked the reversed scan keeps newest-first order. I ran the three new tests on this branch (all pass) and re-ran the two regression tests with the fix reversed (both fail, showing the issue exact two failure modes). The CHANGELOG entry sits correctly under Unreleased/Fixed. No blocking correctness issues found.

Two house-style changes requested before merge:

1. Both commits carry Co-Authored-By trailers and the description ends with a Generated line. This repo keeps commits and code free of tool attribution (see the commit conventions in AGENTS.md), so please reword both commits to drop the trailers (the messages themselves are excellent, keep them verbatim otherwise) and delete the Generated line from the description. A reword plus force-push to your fork branch is the standard route for this.
2. This branch is behind main. Please merge main in before landing.

One non-blocking follow-up is left as an inline comment. Thanks again for the careful verification write-up, especially the fail-without-fix evidence and the premise guard in the probe test. Looking forward to the revision.

gouthamkrishnak2003:
Thanks for the careful review! Both requested items are addressed (force-pushed to the branch):

- Dropped the `Co-Authored-By` trailers from both commits — messages are otherwise verbatim — and removed the "Generated" line from the description.
- `main` is merged in.

On the inline follow-up: agreed that folding once in `settle_run` and passing the fold in is the right shape, especially for long compacted runs with many pending actions. Happy to file that as a separate issue (or put up a small follow-up PR) — just let me know which you'd prefer.

Cyrax321 (approved):
Thank you. Both requested items check out: the trailers are gone from both commits with the messages otherwise verbatim, the Generated line is out of the description, and main is merged in. I confirmed the merge only touched an unrelated file, so the reviewed fix is intact, and required CI (ubuntu matrix, lint, markdown lint, Postgres) is green with the remaining platform jobs still running.

On the follow-up: please file it as a separate issue and claim it there. That keeps this PR shippable now and gives the optimization its own acceptance test.

No blocking issues remain. Marking as ready to merge once the rest of CI finishes.

src/continuum/reconcilers.py
Cyrax321: File it as a separate issue and claim it there, as noted in the approval above. Keeps this one shippable.

PR #715, opened 2026-09-07
fix(compaction): allow a run to be compacted more than once

### LINKED ISSUE #648: second compact on a run fails: anchor checkpoint cannot project without archived RUN_STARTED
## Problem

continuum compact works exactly once per run. Any run that does more work after its first compaction can never be compacted again: the second compact_run fails with ValueError: run 'run_1' could not be anchored: run 'run_1' has no goal: the log never recorded RUN_STARTED.

## Evidence

- compact_run in src/continuum/storage/sqlite.py takes a fresh anchor checkpoint whenever new events arrived since the last version (needs_fresh_anchor), via CheckpointManager(self).checkpoint(run_id, force_version=True).
- src/continuum/checkpoint/manager.py:164: project_current folds self.storage.read_events(run_id), live tail only, even though its docstring says full event history.
- src/continuum/state/semantic.py:628: projection without RUN_STARTED raises ProjectionError, wrapped by compact_run as ValueError: could not be anchored.
- After the first compaction RUN_STARTED lives in events_archive, so the anchor projection concludes the run never started. Verified on origin/main at d8e3a9a.

## Impact

Compaction is presented as repeatable maintenance that bounds replay cost while preserving history. In practice the second run is a hard failure, so long-lived runs (the ones compaction exists for) accumulate an unbounded live tail after their first compact. The error message also misdirects: the log did record RUN_STARTED, it is just archived.

## Reproduction

```python
    store.create_run_started(Run(run_id="run_1", goal="t"))
    first = ActionLedger(store, "run_1").claim("refund", {}, key="k1")
    ActionLedger(store, "run_1").complete(first.key, external_id="e1", result={})
    store.compact_run("run_1")  # {'archived': 3}
    second = ActionLedger(store, "run_1").claim("refund2", {}, key="k2")
    ActionLedger(store, "run_1").complete(second.key, external_id="e2", result={})
    store.compact_run("run_1")  # ValueError: could not be anchored ... never recorded RUN_STARTED
```

## Expected behavior

Repeated compacts succeed: the anchor checkpoint projects over full history (or reuses the stored checkpointed state), and each compact archives only the new prefix.

## Current behavior

The second compact raises ValueError and archives nothing.

## Suggested direction

Project the anchor over read_all_events at checkpoint time for this path, or anchor from the latest stored version plus the live tail. Keep the single-transaction archive marker behavior in compact_run untouched.

## Acceptance criteria

- Two consecutive compacts with work in between both succeed and each reports the new prefix archived.
- verify() passes after both compacts.
- Regression test performing compact, work, compact on SQLite.
- Anchor checkpoint contents unchanged in shape; no schema change.

Gouthamkrishnak18:
## What

Fixes #648.

`continuum compact` worked exactly once per run. Any run that did more work after its first compaction failed its second `compact_run` with:

```
ValueError: run 'run_1' could not be anchored: run 'run_1' has no goal: the log never recorded RUN_STARTED
```

The error also misdirected: the log did record `RUN_STARTED`, it was just archived.

## Root cause

The second compact takes a fresh anchor checkpoint (`needs_fresh_anchor` in `compact_run`) via `CheckpointManager.project_current`, which folded only the live tail (`storage.read_events`) even though its docstring promised "full event history". After the first compaction the live tail begins at the anchor markers with no `RUN_STARTED`, so the anchor projection concluded the run never started.

## How

`project_current` now folds archived plus live events via `storage.read_all_events` — the issue's first suggested direction. Consequences kept deliberate:

- `restore` is untouched: it replays the live tail onto stored checkpoint state, so it keeps reading only live events by design (the archive is already baked into the checkpoint).
- The anchor checkpoint contents are unchanged in shape; no schema change.
- The single-transaction archive marker behavior in `compact_run` is untouched.
- Pre-compaction behavior is identical (`read_all_events` == `read_events` when the archive is empty).

Noted while investigating (out of scope, no change made): `benchmarks/horizon/driver.py` carries a try/except fallback for this exact failure, which the primary path no longer hits.

## Review revision

- **Postgres parity:** `PostgresStorage.compact_run` had the same live-tail-only anchor call and is fixed the same way (`project_current(run_id, full_history=True)` passed as `state=` into the forced checkpoint). A repeated-compact regression test against the Postgres backend (`tests/test_storage_postgres.py::test_pg_a_run_can_be_compacted_repeatedly`) mirrors the SQLite one; it runs in CI's Postgres service container.
- **CHANGELOG punctuation:** the em dashes flagged in review are replaced with commas.
- **Commit hygiene:** the attribution trailers are removed from this PR's commits (messages otherwise verbatim).

## Verification

Acceptance criteria from the issue, all met:

- Two (and three) consecutive compacts with work in between succeed, each reporting the new prefix archived.
- `verify_events` passes after every compact.
- Regression test `tests/test_compaction.py::test_a_run_can_be_compacted_repeatedly` performs compact, work, compact on SQLite — verified to **fail with the exact issue error before the fix and pass after it** — and additionally checks only the new prefix moved and restore still works on the twice-compacted run.
- Postgres suite passes in CI (the new repeated-compact test included; it skips locally without a DSN).
- Full `pytest` suite passes; `ruff check`, `ruff format --check`, and `mypy` (strict) all clean.
- `CHANGELOG.md` updated under `Unreleased -> Fixed`.


gouthamkrishnak2003:
Both findings addressed in 56164cb:

- **Per-turn cost:** `project_current` now reads the live tail by default and takes an opt-in `full_history=True` flag. Only `compact_run`'s forced anchor checkpoint opts in; `evaluate`/`maybe_checkpoint`/`checkpoint(state=None)` and the adapters keep the bounded live-tail read, exactly as on `main`.
- **Stale docstring:** `read_all_events` in `storage/base.py` now spells out which checkpoint path uses full history (the compaction anchor) and which intentionally stay on the live tail (per-turn evaluation and `restore`), and why.

Verified: repeated compacts still succeed with `verify_events` green after each; the default `project_current` on a compacted run still raises `ProjectionError` (unchanged `main` behavior, so per-turn semantics are untouched); full `pytest` suite, `ruff`, and `mypy` (strict) all pass.

Cyrax321 (changes requested):
Thank you for contributing to CONTINUUM, and for the fast turnaround on the earlier bot findings.

This PR fixes #648: the second compact_run failed its anchor checkpoint because project_current folded only the live tail, which carries no RUN_STARTED after the first compaction. I verified the root cause, confirmed every other project_current caller keeps the default live-tail read (identical to main), and checked checkpoint(state=..., force_version=True) accepts the new call shape. I ran the new regression test on this branch (passes) and re-ran it with the anchor opt-in reversed (fails with the exact issue error: could not be anchored, no goal, log never recorded RUN_STARTED). The CHANGELOG entry sits correctly under Unreleased/Fixed. The SQLite path is correct.

Three changes requested before merge:

1. The Postgres anchor path has the same defect and is unfixed. PostgresStorage.compact_run (src/continuum/storage/postgres.py, around line 594) takes its anchor through CheckpointManager(self).checkpoint(run_id, force_version=True), the same live-tail-only call this PR fixes for SQLite. Postgres declares supports_compaction and implements read_archived_events, so a second compact on a Postgres-backed run fails the same way. Please mirror the anchor change there: project with full_history=True and pass state= into checkpoint. The Postgres CI job will exercise it, and a repeated-compact case against that backend would be ideal if practical.
2. CodeRabbit actionable thread on CHANGELOG.md:199 is still open: the entry uses em dashes, which this repo does not allow anywhere. Please accept that suggestion (commas) as posted.
3. Both commits carry Co-Authored-By trailers and the description ends with a Generated line. This repo keeps commits and code free of tool attribution, so please reword both commits to drop the trailers (keep the messages verbatim otherwise) and delete the Generated line from the description. A reword plus force-push to your fork branch is the standard route.

Already verified as addressed: both Gitar findings (per-turn fold cost, stale base docstring) are resolved by the full_history flag and the docstring update, confirmed in the current diff.

Note: most CI checks are still running, so merge waits on those regardless. Thanks again for the careful verification write-up. Looking forward to the revision.

Gouthamkrishnak18:
Thanks for the thorough review, and for catching the Postgres path — you were right that it had the same defect. All three items are addressed and force-pushed as c6c308a (messages reworded, so the branch history was rewritten):

1. **Postgres anchor parity.** `PostgresStorage.compact_run` now takes its anchor through `project_current(run_id, full_history=True)` with the state passed into `checkpoint()`, mirroring the SQLite change exactly. Added `test_pg_a_run_can_be_compacted_repeatedly` to `tests/test_storage_postgres.py`, the SQLite regression's twin on the second engine: compact, work, compact, `verify_events` green after each, archive grows by exactly the new prefix, anchor retained, restore works on the twice-compacted run. It skips locally without a DSN, so the CI Postgres job is what exercises it.

2. **CHANGELOG punctuation.** Both em dashes replaced with commas; a grep confirms no em dashes remain anywhere in the file. The entry also now records that both engines fix the same defect and names both regression tests.

3. **Attribution.** Both PR commits reworded verbatim minus the trailers, and the Generated line removed from this description. The tree is byte-identical to the previously tested state (`git diff --stat` between the old and new heads is empty), so the earlier verification carries over unchanged.

Local checks on the revision: `tests/test_compaction.py` passes in full, the Postgres suite collects cleanly (skips without a DSN as designed), and `ruff`/`mypy` are clean on the touched files.

anya-research:
Thanks for the work here. This branch currently conflicts with main, so a review now would cover a diff that is not what would merge. Please rebase onto latest main and resolve the conflicts, then ping me back here and I will take it from the top.

Cyrax321 (dismissed):
Thank you for contributing to CONTINUUM. Fixing the unbounded live tail for #648 across both engines is the complete version of this fix, and the design respects the cost model: full history only for the rare forced anchor, live tail everywhere per turn, restore untouched. The ValueError to anchored failure message quoted in the changelog matches the reported symptom exactly. Appreciate the thoroughness, including the Postgres twin test.

Verified from the diff: project_current gains full_history defaulting to False so the per-turn path is unchanged, both SQLite and Postgres compact_run take the anchor through the full history projection, storage base docs updated to match, and tests cover SQLite repeat compact plus Postgres. One convergence note: PR #732 fixes the same defect via the anchor checkpoint as fold base with a no archive rescan perf test your approach would not pass. Please look at carrying that test over if the designs combine, or document why the per compact rescan is the accepted cost. CI green everywhere.

Approving. Thanks for covering both engines.

Cyrax321 (dismissed):

Cyrax321 (approved):
Re-approving after the conflict resolution. Both sides fixed #648 independently; the merged version keeps main's full-history default and this branch's flag.

PR #655, opened 2026-09-06
style: remove em dashes from tests, examples, scripts, checkpoint context

Cyrax321:
## Description

First directory batch of #266: 53 em dashes across 19 files in tests/, examples/, scripts/, and the one src string the tests pin (the STALE STATE banner in checkpoint/context.py, which becomes STALE STATE: DO NOT RELY ON with its assertions updated in the same commit). Prose dashes become commas; the single end-of-line dash becomes a full stop.

## Related issues

Refs #266 (remaining: src/, docs, benchmarks, root files; each gets its own PR)

## Types of changes

- Type: style

## Breaking changes

No. Comment, docstring, demo-output, and banner-text wording only. The banner text change is covered by its own asserting tests, which pass.

## How has this been tested?

Python 3.14, editable installation.

- rg confirms zero em dashes remain in the touched directories.
- pytest on all touched test files (recovery context, rewind, action ledger, CLI colour, provenance, recovery engine, storage concurrency): 208 passed.
- python -m py_compile on touched examples, scripts, and src file: clean.
- .venv/bin/ruff check and format --check on the touched paths: passed.
- Full suite plus mypy will run in CI.

## Checklist

No behavior change. CI has not yet run on this PR.

PR #721, opened 2026-09-07
test(concurrency): settle the claim inside the lease hold (#672)

### LINKED ISSUE #672: Flaky: run-lease concurrency test raises UnknownSideEffect on loaded macOS runners
## Problem

tests/test_storage_concurrency.py::test_the_run_lease_restores_exactly_once_under_concurrency fails intermittently on macos-latest Python 3.13 with UnknownSideEffect from a losing claimant, blocking unrelated PRs. Same signature twice today on different diffs:

- PR #655 run (comment-only diff in that file): UnknownSideEffect action charge key b640e167...
- PR #662 run (benchmark-only diff, file untouched): identical failure.

## Mechanism (hypothesis, from reading the test)

The winner claims key k and never completes it. Any thread that acquires the lease after the winner's STARTED record is visible gets UnknownSideEffect instead of fresh/dedup/no-lease, which is correct ledger behavior for a foreign in-flight claim but outside the test's expected outcome set. The test therefore passes only when no loser acquires post-commit, which is timing-dependent and load-sensitive. The file's own commentary already warns timing is a poor witness.

## Expected behavior

Decide the test's contract under contention: either accept UnknownSideEffect as a legal loser outcome, or restructure so losers cannot observe the winner mid-flight (e.g. winner completes before release, or losers assert on lease acquisition only).

## Acceptance criteria

- The test passes deterministically under load, or documents why it cannot and is quarantined/marked accordingly.
- No change to ledger UnknownSideEffect semantics: the ledger is behaving correctly here.

Adhi1-2:
## What

Makes `test_the_run_lease_restores_exactly_once_under_concurrency` deterministic, per #672's acceptance criteria. Ledger `UnknownSideEffect` semantics are untouched: the ledger was behaving correctly, the test's contract was timing-dependent.

## Root cause, reproduced deterministically

The winner claimed key `k` (STARTED) and released the lease **without completing**. Any thread that then acquired the lease saw a foreign in-flight action and correctly raised `UnknownSideEffect`. The test only passed when no loser happened to acquire post-release, which is load-sensitive scheduling, not a contract.

Sequential two-thread repro against current main raises exactly the CI failure from the issue, down to the same key hash:

```
thread A: fresh
UnknownSideEffect reproduced: action 'charge' (key b640e167c5e7...) was interrupted before its outcome was rec
```

That `b640e167...` matches the key quoted in both CI failures linked from the issue (#655, #662).

## Fix

The holder now completes the action before releasing the lease, which is also what real usage does (intercept, perform, complete, all under the run's lease). No later claimant can observe a claim mid-flight, so every interleaving yields: exactly one `fresh`, and every other claimant is either turned away by the lease (`no-lease`) or served the deduplicated result of the completed claim (`dedup`). A new assertion pins that outcome set.

## Verification

- Deterministic repro above fails the assertion path on main; with the fix, no interleaving can produce `UnknownSideEffect`
- 40 consecutive runs of the test: 40 pass, 0 fail
- Full suite green, `ruff check`/`format --check` clean

Closes #672

### DIFF
--- a/tests/test_storage_concurrency.py
+++ b/tests/test_storage_concurrency.py
@@ -261,6 +261,16 @@ def test_the_run_lease_restores_exactly_once_under_concurrency(tmp_path: Path) -
     Wrapping the claim in the run's lease collapses eight simultaneous claimants
     to a single go-ahead, which is what makes docs/multi_agent_isolation.md's
     "one run, one owner at a time" sufficient for exactly-once.
+
+    The holder settles what it starts before letting go of the lease (issue
+    #672). A claim left STARTED at release is a live claim to every later
+    claimant: the ledger correctly raises UnknownSideEffect for a foreign
+    in-flight action, so a test whose winner never completes passes only when
+    no loser happens to acquire the lease after the winner's release, which is
+    timing and not a contract. Completing inside the hold makes the outcome
+    deterministic for every interleaving: one fresh claim, and every other
+    claimant is either turned away by the lease or served the deduplicated
+    result of the completed one.
     """
     from continuum.actions.ledger import ActionLedger
     from continuum.concurrency import SQLiteLeaseCoordinator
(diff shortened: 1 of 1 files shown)


Cyrax321 (approved):

I verified this beyond the green CI: the root cause matches an independent reading of the race (the winner released while STARTED, so any later acquirer correctly raised UnknownSideEffect, and the key hash in the repro matches both CI failures on the issue). Completing inside the hold mirrors real usage and makes every interleaving deterministic, and I ran the concurrency tests 8 consecutive times locally with no failures. Ledger semantics are untouched, which is exactly what the issue asked for.

One thing to fix before merge, outside this approval: the commit carries a Co-Authored-By: Claude trailer and the PR body has a Generated with line. The project does not accept AI attribution in commits (AGENTS.md), so please amend the trailer out and drop that line. Merging will still need that cleanup.

PR #737, opened 2026-09-07
docs(contributing): spell out the three-file ruff rev lockstep (#689)

### LINKED ISSUE #689: Automate or document the three-way ruff rev lockstep that broke #627
## Context

#627 broke CI because the ruff version lives in three places that must move together: the `ruff==` pin in `pyproject.toml`, the `rev:` in `.pre-commit-config.yaml`, and the quoted rev in `CONTRIBUTING.md` (enforced by `tests/test_precommit_config.py`). Dependabot's config (`.github/dependabot.yml`) watches `pip`, `github-actions`, and `uv`, but not the `pre-commit` ecosystem, so automation bumps one pin while the test demands all three.

## Why this matters

Every future ruff bump will fail CI the same way until the lockstep is automated or documented, and the failure looks like broken tests rather than a version skew, which misleads whoever picks the bump PR up.

## Expected outcome

One of: enable the `pre-commit` ecosystem in `dependabot.yml` (checking grouped updates keep the pin test green), or document the three-file lockstep in `CONTRIBUTING.md` next to the existing rev note so the human doing the bump cannot miss it. Propose the direction in a comment first, since it changes automation.

## Step by step for the fix

1. Read `tests/test_precommit_config.py` to see exactly what must stay aligned.
2. Propose the direction in a comment and wait for maintainer agreement.
3. Implement and confirm the pin test passes.

## Files to touch

- `.github/dependabot.yml` or `CONTRIBUTING.md` (per maintainer direction)

Small, ideal for a contributor interested in CI and automation. Follows #627.

Gouthamkrishnak18:
Closes #689.

## What

#627 broke CI because the ruff version lives in three places that must move together, and a dependabot `pip` bump moved only one. Two things made that failure misleading: the bump note in CONTRIBUTING.md named only two of the three pins (it omitted the `rev:` quoted in CONTRIBUTING.md itself), and the pin test failed with a bare assertion that read as a broken test rather than a version skew.

This PR takes the **document** direction proposed on the issue ([comment](https://github.com/Cyrax321/CONTINUUM/issues/689#issuecomment-5574514135)). The dependabot direction was rejected because dependabot groups cannot cross ecosystems: a `pre-commit` update would land as its own PR bumping only `.pre-commit-config.yaml` and would fail the same pin test in the opposite direction.

## Changes

- **CONTRIBUTING.md** - the bump note now lists all three pins, names `tests/test_precommit_config.py` as the enforcer, and says a dependabot PR failing there is drift, not breakage.
- **tests/test_precommit_config.py** - the pin test collects every stale file and fails with `ruff version skew, not a broken test` plus the three places to bump in one PR. New guard test pins dependabot's deliberate lack of a `pre-commit` ecosystem, so a future contributor cannot reintroduce the breakage as a completeness fix.
- **.github/dependabot.yml** - a comment recording the deliberate omission and the reasoning.
- **CHANGELOG.md** - entry under `Unreleased`.

## Verification

- `pytest tests/test_precommit_config.py --no-cov -q` - 3 passed.
- Full suite - green.
- `ruff check` / `ruff format --check` clean.
- Simulated a stale `rev: v0.16.4` in `.pre-commit-config.yaml` and confirmed the new failure message fires with the three-file remedy; simulated re-adding the `pre-commit` ecosystem to dependabot and confirmed the guard test fails with the reasoning. Both simulations reverted.

Direction was proposed on the issue before implementing, per its step 2; the chosen direction changes no automation.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,10 @@
 version: 2
+# The pre-commit ecosystem is deliberately not watched (#689). Dependabot
+# groups cannot cross ecosystems, so a pre-commit update would land as its own
+# PR bumping only .pre-commit-config.yaml, and the three-way ruff rev lockstep
+# (pyproject.toml, .pre-commit-config.yaml, CONTRIBUTING.md, enforced by
+# tests/test_precommit_config.py) would fail CI in that PR. Bump ruff in all
+# three places by hand in one PR.
 updates:
   - package-ecosystem: "pip"
     directory: "/"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ All notable changes to this project are documented here. The format follows
 
 ### Added
 
+- **Documented three-file ruff rev lockstep (#689).** CONTRIBUTING.md now
+  names all three places the ruff version lives (the `ruff==` pin in
+  `pyproject.toml`, `rev` in `.pre-commit-config.yaml`, and the `rev` quoted
+  in CONTRIBUTING.md itself) and the test that enforces them; the pin test's
+  failure message says "version skew, not a broken test" and lists the three
+  files to bump, so a dependabot PR that trips it (#627) is readable as
+  drift. The pre-commit ecosystem's absence from dependabot is recorded in a
+  comment there and pinned by a test: its updates cannot be grouped with pip
+  bumps and would break the lockstep in their own PR.
+
 - **External probe for consumed authorities via reconcilers.json (#557).**
   `reconcile --authority <id>` runs the configured authority probe with the
   recorded consumption payload on stdin; `valid=true` appends
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -148,10 +148,19 @@ To run it manually against all files:
 pre-commit run --all-files
 ```
 
-If you bump the `ruff==` pin in `pyproject.toml`, bump `rev` in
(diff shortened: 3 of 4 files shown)


Cyrax321 (commented):

I verified the three pin tests pass on the PR tree and confirmed the failure mode they guard: the message names the three lockstep files instead of reading as a broken test, and the dependabot omission is now pinned against well-meaning completeness fixes. Direction matches what the issue asked for, and CI is green.

One thing to fix before merge: the commit carries a Co-Authored-By: Claude Code trailer. The project does not accept AI attribution in commits (AGENTS.md), so please amend it out. I will approve once that is done.

Cyrax321 (changes requested):

The document direction matches #689. The three pin coverage, the skew message that names all three files, and the guard test for the deliberate dependabot omission are the correct shape.

Blocking before approval:

1. Remove AI attribution. Both commits still carry Co-Authored-By Claude Code, and the PR body still carries the Generated with footer. Per AGENTS.md the project does not accept AI attribution in commits or PR text. Please amend the trailers out and remove the footer. No code change needed for this step.

2. Green CI on the final tree. Lint and Type-check, the OS matrix, and Postgres were still pending at review time.

Once the trailers and footer are gone and CI is green, this is ready to approve.

Cyrax321 (approved):
Waiving the AI attribution block per maintainer decision. Technical content verified for #689.

Squash message on merge will carry no AI trailers, keeping main history clean. Waiting for remaining Windows checks before merge.

Issue #366, opened 2026-08-25
complete_action launders an UNKNOWN action into COMPLETED with no evidence


Cyrax321:
## What happens

`continuum_complete_action` turns an `UNKNOWN` action into `COMPLETED` with no evidence, no note, and no distinguishable event type. The recovery blocker disappears and the audit trail does not record that it was cleared by assertion.

## Reproduction (MCP only)

```
intercept_action(type="card_charge", key="charge:INV-9001", arguments={"amount": 4200, ...})
  -> proceed=true, action_key=38fb75c0...

fail_action(action_key=38fb75c0..., certain=false,
            error="gateway timeout after the charge request was sent")
  -> status=unknown, side_effect_uncertain=true

complete_action(action_key=38fb75c0...)
  -> status=completed

list_actions -> unresolved: 0
```

A charge whose real-world outcome is genuinely unknown now reads as a clean completion.

## Why

`ActionLedger.complete` has no status guard. It calls `_require(key)` and then unconditionally copies the action to `COMPLETED` with `side_effect_uncertain: False`, whatever the current status. `continuum_complete_action` adds no guard of its own.

The event recorded is `ACTION_RECORDED`, identical to an ordinary first-time completion. Only `reconcile` writes `ACTION_RECONCILED`, so an auditor reading the log cannot tell that an uncertain effect was resolved without a probe. Confirmed in the event log: the legitimate reconcile produced `ACTION_RECONCILED`, the laundering completion produced `ACTION_RECORDED`.

## Why an agent will actually do this

The incentives point at this door. The evidence-gated path is `reconcile_action`, and that path is currently unusable because the identifier it is told to pass is rejected (see the reconcile identifier issue). `complete_action` accepts the key the agent already holds from `intercept_action`, is on the same mutation allowlist, and is the tool the agent is instructed to call routinely. An agent trying to get unstuck reaches for it.

## Suggested fix

Refuse `complete` when the current status is `UNKNOWN`, and say why:

> this action's outcome is unknown, so completing it would assert an effect nobody verified. Check the external system, then call `continuum_reconcile_action(occurred=true)`, which records the evidence.

`STARTED` must keep working, since that is the normal two-phase path. `FAILED` and `COMPENSATED` deserve a decision too: today they silently become `COMPLETED`, which is at least surprising.

If the transition is wanted for a caller that does have evidence, route it through `reconcile` so it lands as `ACTION_RECONCILED` with a note, rather than making the ungated tool the easy one.

[...]

PR #746, opened 2026-09-07
fix(actions): guard fail() against statuses that are not in flight

### LINKED ISSUE #733: ActionLedger.fail() has no status guard, so a late failure report erases a COMPLETED action and the side effect fires twice
## Summary

`ActionLedger.complete()` refuses to settle an action that is not in flight (the status guard added for #366/#374, `src/continuum/actions/ledger.py:1063`). `ActionLedger.fail()` (`src/continuum/actions/ledger.py:1100`) never got the mirror-image guard: it accepts any current status, so a late or duplicated failure report for an action that already COMPLETED flips it to FAILED while the recorded result (the receipt from the real fire) stays on the record.

Once the action reads FAILED, the key is no longer settled, so the next `claim()` for the same operation opens a fresh attempt and the side effect executes a second time. This is the same dedup-defeat class #366 closed for `complete()`, from the other direction.

## Repro

```python
import tempfile, os
from continuum.adapters.generic import GenericAgentAdapter

tmp = tempfile.mkdtemp()
with SQLiteStorage(os.path.join(tmp, "l.db")) as store:
    adapter = GenericAgentAdapter(store)

    fires = []
    def send_invoice():
        fires.append(1)
        return {"receipt": "r-123"}

    # First call: the effect fires and the action is recorded COMPLETED.
    adapter.intercept_action("run_1", "send_invoice", send_invoice, key="inv:42")

    # A late/duplicate failure report arrives for the same key
    # (a retried fail_action RPC, a delayed error handler, a replayed event).
    key = next(k for k, a in ledger.folded().items()
               if a.action_type == "send_invoice")
    action = ledger.fail(key, "late timeout report", certain=True)

    # The next identical call re-fires the side effect.
    adapter.intercept_action("run_1", "send_invoice", send_invoice, key="inv:42")
    print("TOTAL SIDE-EFFECT FIRES:", len(fires))
```

Output: `TOTAL SIDE-EFFECT FIRES: 2` (expected 1: the second call should have been served from the completed record). The invoice is sent twice.

## Expected

`fail()` on a COMPLETED action should refuse, symmetrically to `complete()`: it should raise a `LedgerError` telling the caller the action already completed with a recorded outcome, and pointing at `reconcile(occurred=False)` / `continuum_reconcile_action` if they have evidence the effect did not land.

## Actual

The action silently becomes FAILED (with the stale COMPLETED result still attached), and subsequent identical calls re-execute the side effect.

## Notes

- `fail(certain=False)` (the UNKNOWN path) is the legitimate route for an uncertain report on a STARTED action; the guard needs to keep allowing that.
- `replayguard.protected_call`'s exception path also calls `ledger.fail(...)` (`src/continuum/replayguard.py:139`); it only fires on ALLOW (STARTED), so a guard does not break it, but the fix should confirm that with a test.

Adhi1-2:
## Summary

`complete()` got a status guard in #366 (it refuses to settle anything not in flight); `fail()` never got the mirror-image guard. A late or duplicated failure report for an already-`COMPLETED` action flipped it to `FAILED` while the recorded result stayed on the books. The key stopped reading as settled, the next `claim()` opened a fresh attempt, and the side effect re-fired.

Fixes #733.

## What changed

- `ActionLedger.fail()` now refuses every status that is not in flight, mirroring `complete()`:
  - `STARTED` and `FAILED` settle (re-reporting a `FAILED` action asserts nothing new, the same allowance `complete()` makes for `COMPLETED`)
  - `COMPLETED`, `UNKNOWN`, `COMPENSATED`, `REQUIRES_REVIEW`, `PLANNED` raise a `LedgerError` that points at `reconcile(occurred=False)` / `continuum_reconcile_action`, which records the evidence and the note
- Regression tests pin: the refusal per settled status, that a refused fail leaves the completed outcome settled (dedup cache hit, no pending blocker), the allowed `FAILED` re-report, and that `UNKNOWN` stays a job for `reconcile`.

## Testing

- `tests/test_action_ledger.py`: four new tests / one new parametrization, mirroring the #366 test block

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented completed, compensated, review-required, or unknown actions from being marked as failed.
  * Preserved recorded successful outcomes and deduplication state.
  * Continued to allow already-failed actions to be failed again.
  * Directs uncertain action outcomes to reconciliation.

### DIFF
--- a/src/continuum/actions/ledger.py
+++ b/src/continuum/actions/ledger.py
@@ -1104,8 +1104,27 @@ class ActionLedger:
         — a timeout after the request was sent, for instance. Those become
         ``UNKNOWN`` rather than ``FAILED``, because a timeout is not evidence of
         absence.
+
+        Only in-flight statuses settle here (issue #733). Re-reporting a
+        ``FAILED`` action is allowed, because a caller repeating itself after a
+        dropped response is not asserting anything new. Every other status is
+        refused, mirroring :meth:`complete` (issue #366): a late ``fail`` on a
+        ``COMPLETED`` action used to flip it to ``FAILED`` while its recorded
+        result stayed on the books, which reopened the key and let the next
+        claim re-fire a side effect that had already happened. ``UNKNOWN`` is
+        refused for the same reason ``complete`` refuses it: resolving an
+        uncertain outcome is a correction that needs evidence, which is what
+        :meth:`reconcile` records.
         """
         key, existing = self._require(key)
+        if existing.status not in (ActionStatus.STARTED, ActionStatus.FAILED):
+            raise LedgerError(
+                f"action {existing.action_type!r} is {existing.status.value}, not in flight, so "
+                f"failing it would erase a recorded outcome. If a check confirmed "
+                f"the effect did not happen, call reconcile(occurred=False) "
+                f"(continuum_reconcile_action over MCP), which records the evidence "
+                f"and the note alongside the correction."
+            )
         action = existing.model_copy(
             update={
                 "status": ActionStatus.FAILED if certain else ActionStatus.UNKNOWN,
--- a/tests/test_action_ledger.py
+++ b/tests/test_action_ledger.py
@@ -295,6 +295,81 @@ def test_complete_refuses_every_status_that_is_not_in_flight(
         ledger.complete(outcome.key, external_id="txn-1")
 
 
+@pytest.mark.parametrize(
+    ("settle", "expected_status"),
+    [
+        (lambda led, key: led.complete(key, external_id="txn-1"), "completed"),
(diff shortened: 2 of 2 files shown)


Cyrax321 (changes requested):

This PR guards ActionLedger.fail() against settled statuses, mirroring the complete() guard from #366, and fixes #733. I ran the ledger tests on this branch (green), traced the serve and MCP fail callers through the LedgerError mapping #366 established, and confirmed required CI is green. No blocking correctness issues found.

Cyrax321 (approved):
Verified fix for #733. Source guard plus regression tests, matching scope. Prior changes-requested was attribution-only. Attribution waived per maintainer decision. Approved for merge.

PR #748, opened 2026-09-07
fix(replayguard): build the BLOCK_UNCERTAIN decision with its key

### LINKED ISSUE #735: replayguard BLOCK_UNCERTAIN decision is built without its key, so protected_call raises AssertionError instead of ReplayBlocked
## Summary

`replayguard.evaluate()` builds the `BLOCK_UNCERTAIN` `GuardDecision` without passing `key=` (`src/continuum/replayguard.py:78-82`), unlike every other branch (DENY_UNCLAIMED, ALLOW, SKIP_DUPLICATE, DENY_RECLAIM all pass `key=key`). The public contract says uncertain states raise `ReplayBlocked`; instead `protected_call` hits `assert decision.key is not None` (`src/continuum/replayguard.py:133`) and the caller gets a bare `AssertionError` with no reason, no decision, and no chance to catch `ReplayBlocked`.

Under `python -O` the assert is stripped entirely and `key_to_use` becomes `None`, so the guard proceeds with a null key.

## Repro

```python
import tempfile, os
from continuum.actions.idempotency import idempotency_key
from continuum.replayguard import protected_call, ReplayBlocked

tmp = tempfile.mkdtemp()
with SQLiteStorage(os.path.join(tmp, "rg.db")) as store:

    # Claim, then an uncertain escalation: the public path a retry takes.
    out = ledger.claim("send_invoice", {}, key="i:3")
    ledger.fail(out.key, "timeout", certain=False)  # -> UNKNOWN

with SQLiteStorage(os.path.join(tmp, "rg.db")) as store:
    try:
        protected_call(store, "run_1", action_type="send_invoice",
                       key="i:3", fn=lambda: "SHOULD_NOT_RUN")
    except ReplayBlocked as exc:
        print("expected:", exc.kind)
    except AssertionError as exc:
        print("BUG: AssertionError instead of ReplayBlocked:", exc)
```

Output:

```
BUG: AssertionError instead of ReplayBlocked: GuardDecision(kind=<GuardKind.BLOCK_UNCERTAIN: ...>, message=..., key=None)
```

## Expected

`ReplayBlocked` is raised with the BLOCK_UNCERTAIN decision and its reason ("has an unknown outcome; reconcile first"), exactly as the docstring documents.

## Actual

`AssertionError` (or, under `-O`, a `None` key flowing into the rest of `protected_call`).

## Fix shape

Pass `key=key` when constructing the BLOCK_UNCERTAIN decision. A regression test should call `protected_call` against an UNKNOWN action and assert `ReplayBlocked` (and its kind), not `AssertionError`.

Adhi1-2:
## Summary

`evaluate()` built the `BLOCK_UNCERTAIN` `GuardDecision` without `key=`, unlike every other branch of the decision table. `protected_call` then hit its `assert decision.key is not None` checkpoint and raised a bare `AssertionError` where the documented contract promises `ReplayBlocked`; under `python -O` the stripped assert let a `None` key flow onward instead.

Fixes #735.

## What changed

- The `BLOCK_UNCERTAIN` decision now carries its key, like `DENY_UNCLAIMED` / `ALLOW` / `SKIP_DUPLICATE` / `DENY_RECLAIM` already did

## Testing

- `tests/test_replayguard.py`: `test_block_uncertain_decision_carries_its_key` pins the decision-table invariant; `test_protected_call_raises_replayblocked_for_an_uncertain_action` pins the public contract (ReplayBlocked with kind and key, not AssertionError)

## Summary by CodeRabbit

- **Bug Fixes**
  - Uncertain replay decisions now include the related ledger key.
  - Protected calls with uncertain actions now raise the documented replay-blocking error instead of an assertion failure.
  - Replay now correctly preserves dictionaries containing a `"return"` field, supports non-dictionary results, and remains compatible with previously recorded return data.

- **Tests**
  - Added coverage for ledger-key propagation, replay blocking, and return-value compatibility.

### DIFF
--- a/src/continuum/replayguard.py
+++ b/src/continuum/replayguard.py
@@ -79,6 +79,7 @@ def evaluate(
         return GuardDecision(
             GuardKind.BLOCK_UNCERTAIN,
             f"{action_type!r} {rendered_key!r} has an unknown outcome; reconcile first",
+            key=key,
         )
     return GuardDecision(
         GuardKind.DENY_RECLAIM,
--- a/tests/test_replayguard.py
+++ b/tests/test_replayguard.py
@@ -25,6 +25,7 @@ from continuum.events import EventType  # noqa: E402
 from continuum.models import ActionStatus, Run  # noqa: E402
 from continuum.replayguard import (  # noqa: E402
     GuardKind,
+    ReplayBlocked,
     evaluate,
     langgraph_protected_node,
     protected_call,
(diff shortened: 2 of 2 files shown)


Cyrax321 (changes requested):

This PR passes key= on the BLOCK_UNCERTAIN branch so protected_call raises the documented ReplayBlocked, and fixes #735. One line matching every other evaluate() branch, covered by two new tests. Required CI is green. No blocking correctness issues found.

Cyrax321 (approved):
Verified fix for #735. BLOCK_UNCERTAIN decision built with its key, plus regression tests. Prior changes-requested was attribution-only. Attribution waived per maintainer decision. Approved for merge.

PR #749, opened 2026-09-07
fix(replayguard): envelope journaled results so replay cannot unwrap a caller dict

### LINKED ISSUE #736: protected_call returns a divergent value on replay when the result dict contains the literal key 'return'
## Summary

`replayguard.protected_call` journals dict-shaped results verbatim but unwraps them on replay with a reserved-key convention (`src/continuum/replayguard.py:142-150`):

```python
journal = result if isinstance(result, dict) else {"return": result}
...
value = cached.get("return", cached) if isinstance(cached, dict) else cached
```

A non-dict result is wrapped as `{"return": result}` and correctly unwrapped on replay. But a dict result that itself contains the literal key `"return"` is stored unwrapped, and the replay read then does `cached.get("return", cached)`: it returns `cached["return"]` (just one field) where the first call returned the whole dict. First call and replay return divergent values for the same protected operation, which is exactly what memoisation must never do.

This is the same class as #44, fixed for `intercept_action` in `src/continuum/actions/idempotency.py` with the `RESULT_ENVELOPE_KEY` sentinel; `protected_call` never got the envelope.

## Repro

```python
import tempfile, os
from continuum.replayguard import protected_call

tmp = tempfile.mkdtemp()
with SQLiteStorage(os.path.join(tmp, "rg2.db")) as store:

payload = {"return": "receipt-1", "amount": 100}
k1, r1 = protected_call(store, "run_1", action_type="send_invoice",
                        key="inv:9", fn=lambda: payload)
k2, r2 = protected_call(store, "run_1", action_type="send_invoice",
                        key="inv:9", fn=lambda: payload)
print(r1)  # {'return': 'receipt-1', 'amount': 100}
print(r2)  # 'receipt-1'
print("replay matches first call:", r1 == r2)
```

Output:

```
{'return': 'receipt-1', 'amount': 100}
receipt-1
replay matches first call: False
```

The replay drops `amount` and changes the type of the returned value. A caller branching on the shape of the result (dict vs string) takes a different path on replay than on the original call.

## Expected

The replay of a protected call returns a value equal to the first call's return, whatever keys the dict contains.

## Fix shape

Wrap dict results in an envelope the same way `intercept_action` does (the `RESULT_ENVELOPE_KEY` sentinel from `src/continuum/actions/idempotency.py`), and unwrap only that sentinel on replay, so a caller's literal `"return"` key is never mistaken for the journal marker.

Adhi1-2:
## Summary

`protected_call` journalled dict results verbatim but unwrapped them on replay via the reserved `"return"` key, so a result dict containing that literal key answered differently on its second call than on its first: replay returned one member of the dict instead of the dict. The same class was fixed for `intercept_action` in #44 with an envelope sentinel; this applies it to the guard.

Fixes #736.

## What changed

- `src/continuum/replayguard.py`: results are journalled under the `__return_value__` envelope whenever they need wrapping (non-dicts, and any dict carrying the envelope key or the legacy `"return"` marker); replay unwraps only the envelope, plus the legacy `{"return": ...}` journals of already-stored non-dict results so pre-fix records stay readable. The sentinel is defined locally so the portable core stays importable without the adapter stack.

## Testing

- `tests/test_replayguard.py`: `test_replay_returns_the_callers_own_dict_even_with_a_return_key` (the reported divergence), `test_non_dict_results_round_trip_through_the_envelope`, and `test_a_legacy_return_journal_still_unwraps_on_replay`

### DIFF
--- a/src/continuum/replayguard.py
+++ b/src/continuum/replayguard.py
@@ -28,7 +28,21 @@ from dataclasses import dataclass
 from enum import StrEnum
 from typing import Any
 
-__all__ = ["GuardKind", "GuardDecision", "evaluate", "protected_call", "langgraph_protected_node"]
+__all__ = [
+    "GuardKind",
+    "GuardDecision",
+    "RESULT_ENVELOPE_KEY",
+    "evaluate",
+    "protected_call",
+    "langgraph_protected_node",
+]
+
+# Journal envelope, same literal adapters.generic uses for the same reason
+# (issue #44): a stored result must be distinguishable from the caller's own
+# dict, or the replay path unwraps one member of it instead of the dict
+# (issue #736). Defined locally so the core stays importable without the
+# adapter stack.
+RESULT_ENVELOPE_KEY = "__return_value__"
 
 
 class GuardKind(StrEnum):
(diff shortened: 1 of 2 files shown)


Cyrax321 (changes requested):

This PR envelopes journaled results so a caller dict containing the reserved key replays identically, and fixes #736. I verified the envelope literal is byte-identical to the one in adapters.generic and the legacy return-key path is preserved. Required CI is green. No blocking correctness issues found.

Cyrax321 (approved):
Verified fix for #736. Journaled results enveloped so replay cannot unwrap a caller dict, plus regression tests. Prior changes-requested was attribution-only. Attribution waived per maintainer decision. Approved for merge.

PR #767, opened 2026-09-07
fix(environment): stamp oversized files UNKNOWN_VERSION so a size is never an identity

### LINKED ISSUE #738: FileProvider max_bytes sizing lets a replaced same-size file verify as unchanged (environment check fails open)
## Summary

`FileProvider.capture()` deliberately skips hashing files larger than `max_bytes` and stamps them with a synthetic version `size:<bytes>` (`src/continuum/environment/snapshot.py:127-134`, metadata `{"skipped": "larger than max_bytes"}`). That synthetic string is a size, not an identity, but `diff_environments()` treats it as one: `_unverifiable()` only flags `UNKNOWN_VERSION` (`src/continuum/environment/diff.py:109-110`), so the string comparison `was == now` (`diff.py:185-191`) classifies the resource as verifiably `UNCHANGED`.

The result is silent fail-open of the exact check the subsystem exists to perform. A tracked file replaced with entirely different content of the same byte size passes as "verified unchanged", the validator marks the dependency `VALID / "verified unchanged"` (`src/continuum/state/validator.py:386-388`), and `safe_to_resume` stays `True`. The module's own contract says a resource that could not be inspected must not be treated as intact.

## Repro

```python
import tempfile
from pathlib import Path
from continuum.environment.snapshot import FileProvider, capture
from continuum.environment.diff import diff_environments
from continuum.models import ExternalDependency, Goal, Progress, SemanticState
from continuum.state.validator import validate_state

tmp = Path(tempfile.mkdtemp())
dataset = tmp / "dataset.bin"
dataset.write_bytes(b"A" * 5000)                  # original content
before_env = capture("run_1", FileProvider([dataset], max_bytes=1000))

state = SemanticState(
    run_id="run_1", goal=Goal(description="analyse dataset"), progress=Progress(),
    external_dependencies=[ExternalDependency(resource=str(dataset), kind="file")],
)

dataset.write_bytes(b"B" * 5000)                  # fully replaced, same size
after_env = capture("run_1", FileProvider([dataset], max_bytes=1000))

diff = diff_environments(before_env, after_env)
print("diff delta:", diff.deltas[0].change, "| stable:", diff.stable)
outcome = validate_state(state, current_environment=after_env,
                         checkpoint_environment=before_env)
dep = outcome.report.statuses[0]
print("dependency status:", dep.status, "| detail:", dep.detail)
print("safe_to_resume:", outcome.report.safe_to_resume)
```

Output:

```
diff delta: unchanged | stable: True
dependency status: valid | detail: verified unchanged
safe_to_resume: True
```

Expected: the oversized resource is content-unverified, so the delta must not read as verified-unchanged; a same-size replacement must not yield `VALID` and a safe resume.

## Why tests miss it

`tests/test_environment.py::test_large_files_are_sized_rather_than_hashed` only asserts the `size:5000` version string on a single capture; no test diffs two captures of a same-size, different-content oversized file. The validator tests all use `StaticProvider` with real version strings.

## Fix shape

Either stamp oversized resources with `UNKNOWN_VERSION` (matching `_unverifiable`'s existing treatment) or teach `_unverifiable()` to recognise the synthetic `size:` prefix / `skipped` metadata, so the delta reports the resource as unverifiable and the validator cannot bless it as "verified unchanged". A regression test should diff two same-size different-content captures under a tight `max_bytes`.

Adhi1-2:
## Summary

- Files larger than `max_bytes` are skipped by the hashing providers, but the
  snapshot stamped them with `version=f"size:{stat.st_size}"`. A size is not a
  content identity: swap an oversized file for a different file of identical
  length and the environment diff reports UNCHANGED, so an agent resumes
  believing the environment was verified when nothing about the file's content
  was ever checked.
- Oversized files now stamp `UNKNOWN_VERSION`, the same marker every other
  resource that could not be inspected uses. The file size stays available in
  the snapshot metadata, and the diff's reason chain names `skipped` metadata
  so an UNKNOWN delta on an oversized file says why it could not be verified
  instead of the generic "resource could not be inspected".

Fixes #738

## Testing

- `test_large_files_are_sized_rather_than_hashed` updated to assert the
  UNKNOWN_VERSION stamp plus the preserved `metadata["size"]`.
- `test_a_replaced_same_size_oversized_file_is_not_verified_unchanged`: the
  reproduction from the issue, same-length swap now yields an UNKNOWN delta
  rather than UNCHANGED.
- `test_an_oversized_dependency_does_not_resume_as_verified`: the recovery
  verdict follows the uncertainty instead of resuming.

### DIFF
--- a/src/continuum/environment/diff.py
+++ b/src/continuum/environment/diff.py
@@ -159,6 +159,8 @@ def diff_environments(
             reason = str(
                 current.metadata.get("error")
                 or previous.metadata.get("error")
+                or current.metadata.get("skipped")
+                or previous.metadata.get("skipped")
                 or "resource could not be inspected"
             )
             deltas.append(
--- a/src/continuum/environment/snapshot.py
+++ b/src/continuum/environment/snapshot.py
@@ -125,10 +125,15 @@ class FileProvider(EnvironmentProvider):
             try:
                 stat = path.stat()
                 if self.max_bytes is not None and stat.st_size > self.max_bytes:
+                    # UNKNOWN_VERSION, not a synthetic "size:<n>" stamp: a size
+                    # is not an identity, but diff_environments compared the old
+                    # stamp as one, so a replaced file of the same byte size
+                    # verified as unchanged and the environment check failed
+                    # open (issue #738). The size stays in metadata.
                     captured[key] = EnvResource(
                         name=key,
                         kind="file",
-                        version=f"size:{stat.st_size}",
+                        version=UNKNOWN_VERSION,
                         metadata={"skipped": "larger than max_bytes", "size": stat.st_size},
                     )
                     continue
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -13,7 +13,7 @@ from continuum.environment import (
     diff_environments,
     process_fingerprint,
 )
-from continuum.models import EnvResource
+from continuum.models import EnvResource, StateStatus
 
 # --- capture --------------------------------------------------------------- #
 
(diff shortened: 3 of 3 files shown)


Cyrax321 (changes requested):

This PR stamps oversized files UNKNOWN_VERSION instead of a synthetic size identity, closing the fail-open same-size replacement hole, and fixes #738. I confirmed the constant and the validator UNKNOWN mapping on main. Required CI is green. No blocking correctness issues found.

Cyrax321 (approved):
Verified fix for #738. Oversized files stamped UNKNOWN_VERSION, plus regression tests. Prior changes-requested was attribution-only. Attribution waived per maintainer decision. Approved for merge.

PR #769, opened 2026-09-07
fix(state): diff constraint pins so add/retract is never no semantic change

### LINKED ISSUE #740: diff_states ignores constraint pins, so pin add/retract shows as no semantic change
## Summary

`diff_states()` (`src/continuum/state/diff.py:94-225`) compares goal, progress, decisions, findings, evidence, pending work, plan, approvals, dependencies and model, but never `state.pins`. Constraint pins are first-class `SemanticState` components (#417/#418: active constraints whose hash-tagged markers must reach a resuming agent), but the semantic diff predates them and was never extended; the `Component` enum (`models.py:134-146`) has no `PIN` value either.

Adding or retracting an active constraint pin between two states yields an empty diff, so `continuum diff` (`cli/main.py:825`), recovery contracts and benchmark scoring, all documented consumers of `diff_states`, report "No semantic change" even though the set of constraints the agent must honour changed. A retracted constraint disappearing from the diff is silent data loss in a user-facing correctness surface.

## Repro

```python
from continuum.models import ConstraintPin, Goal, Progress, Provenance, Origin, SemanticState
from continuum.state.diff import diff_states, render_diff

p = Provenance(origin=Origin.DETERMINISTIC, source_sequence=1)
base = dict(run_id="r1", goal=Goal(description="g", provenance=p),
            progress=Progress(provenance=p))
pin = ConstraintPin(constraint_id="c1", sha256="a" * 64, status="active", provenance=p)

added = diff_states(SemanticState(**base), SemanticState(**base, pins={"c1": pin}))
print("pin added   ->", render_diff(added).strip())
retracted = diff_states(SemanticState(**base, pins={"c1": pin}), SemanticState(**base))
print("pin removed ->", render_diff(retracted).strip())
```

Output:

```
pin added   -> No semantic change between v0 and v0.
pin removed -> No semantic change between v0 and v0.
```

Expected: an ADDED / REMOVED entry for the pin (which would also require a `Component.PIN` enum value).

## Why tests miss it

`tests/test_diff.py` has 12 tests; none mentions pins.

## Fix shape

Add `Component.PIN`, extend `diff_states` to diff the `pins` mappings (added / retracted, and changed status or sha256 for a pin present in both), and cover it in `tests/test_diff.py`.

Adhi1-2:
## Summary

- Constraint pins (#417) live on `SemanticState.pins`, but `diff_states`
  compared every other collection and never pins. A pin could be added,
  retracted, or re-pinned against a different digest with the diff reporting
  "No semantic change", so a checkpoint restore or contract review could not
  see that the constraint the run promised to honour had moved.
- Pins are now diffed like every other collection via `_compare_collection`,
  keyed by `constraint_id`, comparing `sha256`, `status` and `pinned_at`,
  under a new `Component.PIN` value so entries are labelled distinctly in the
  rendered diff.

Fixes #740

## Testing

- `test_pin_add_and_retract_are_reported`
- `test_a_pin_whose_digest_changes_is_reported`
- `test_unchanged_pins_produce_no_entries`

----
## Summary by Gitar

- **Bug fixes:**
    - Count archived attempts so the retry budget survives compaction in `src/continuum/...`
    - Guard `fail()` against statuses that are not in flight in `src/continuum/...`

<sub>This will update automatically on new commits.</sub>

### DIFF
--- a/src/continuum/models.py
+++ b/src/continuum/models.py
@@ -144,6 +144,7 @@ class Component(StrEnum):
     MODEL = "model"
     APPROVAL = "approval"
     ENVIRONMENT = "environment"
+    PIN = "pin"
 
 
 class DiffKind(StrEnum):
--- a/src/continuum/state/diff.py
+++ b/src/continuum/state/diff.py
@@ -203,6 +203,18 @@ def diff_states(before: SemanticState, after: SemanticState) -> StateDiff:
         describe=lambda d: f"{d.resource} {d.version or ''}".strip(),
         fields=("version", "checksum", "status", "kind"),
     )
+    # Pins are first-class state (issue #417): the active constraint set a
+    # resuming agent must honour. The semantic diff predates them, so a pin
+    # added or retracted between two states used to yield an empty diff and
+    # `continuum diff` reported "no semantic change" (issue #740).
+    entries += _compare_collection(
+        list(before.pins.values()),
+        list(after.pins.values()),
+        component=Component.PIN,
+        key=lambda p: p.constraint_id,
+        describe=lambda p: f"{p.constraint_id} {p.sha256[:12]}",
+        fields=("sha256", "status", "pinned_at"),
+    )
 
     before_model = before.model.model if before.model else None
     after_model = after.model.model if after.model else None
--- a/tests/test_diff.py
+++ b/tests/test_diff.py
@@ -4,6 +4,7 @@ import pytest
 
 from continuum.models import (
     Component,
+    ConstraintPin,
     Decision,
     DiffKind,
     Evidence,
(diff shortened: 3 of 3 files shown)


Cyrax321 (changes requested):

This PR diffs constraint pins through the standard component machinery under Component.PIN, and fixes #740. I checked the call against the _compare_collection signature. Required CI is green. No blocking correctness issues found.

Cyrax321 (approved):
Verified fix for #740. Diff constraint pins so add/retract is never a no-semantic-change, plus regression tests. Prior changes-requested was attribution-only. Attribution waived per maintainer decision. Approved for merge.

PR #770, opened 2026-09-07
fix(cli): resume exits REQUIRES_HUMAN for a family-blocked parent

### LINKED ISSUE #741: resume exits 0 for a family-blocked parent, so the resume-and-start automation pattern is unsafe
## Summary

`continuum resume` exits 0 for a parent run whose family roll-up is blocked, violating the exit-code safety contract. `cmd_resume` computes `roll_up_children` and downgrades the *presentation*: when the parent's own decision is `RESUME` but any non-terminal child is unsafe, `presented_mode` becomes `"request_human"` and `presented_safe` becomes `False` (`src/continuum/cli/main.py:1262-1267`, the "most cautious signal wins" rule from #243). But the process exit status is still computed from the engine's per-run `decision.mode`, which remains `RESUME` (`cli/main.py:1342`):

```python
return exit_code_for(decision.mode)
```

`exitcodes.py` states the rule as absolute: only a fully verified, safe-to-resume run exits 0. So the documented automation pattern `resume "$PARENT" && ./start-agent.sh` launches an agent onto a family that holds a child with an unreconciled side effect.

The text output compounds the mismatch: it prints `Recovery decision: RESUME` / `Next permitted action: continue` next to the `FAMILY BLOCKED` block, while the JSON payload says `mode: request_human, safe: false`. Three surfaces (text, JSON, exit code) disagree.

## Repro

```python
import io, json, os, tempfile
from continuum.cli import main

tmp = tempfile.mkdtemp(); db = os.path.join(tmp, "fam.db")
    store.create_run(Run(run_id="p1", goal="Send invoices"))          # clean parent
    store.append_event("p1", EventType.RUN_STARTED, {"goal": "Send invoices", "total": 3})
    store.create_run(Run(run_id="c1", goal="risky branch", parent_run_id="p1"))
    store.append_event("c1", EventType.RUN_STARTED, {"goal": "risky branch"})
    ActionLedger(store, "c1").claim("wire_transfer", {}, key="wire:1")  # interrupted side effect

out, err = io.StringIO(), io.StringIO()
code = main(["--db", db, "--json", "resume", "p1"], out=out, err=err)
payload = json.loads(out.getvalue())
print("EXIT CODE:", code, "| mode:", payload["mode"], "| safe:", payload["safe"])
```

Output:

```
EXIT CODE: 0 | mode: request_human | safe: False
```

Text mode also exits 0 while printing `FAMILY BLOCKED: children of this run are not resumable.` Expected: non-zero (`REQUIRES_HUMAN`, 20), matching the presented mode, the way every other non-RESUME verdict exits.

## Why tests miss it

`tests/test_family.py::test_uncertain_child_blocks_the_parent_resume` and `test_tree_limit_cannot_hide_a_child_from_family_safety` assert only `payload["mode"] == "request_human"` / `family_rationale` from the JSON, never the exit code of a family-blocked parent. `tests/test_cli.py::test_only_resume_maps_to_a_zero_exit` tests the `exit_code_for` mapping in isolation and cannot see that `cmd_resume` feeds it the un-presented mode.

## Fix shape

Compute the exit code from the presented mode (or equivalent: `ExitCode.REQUIRES_HUMAN` whenever `presented_safe` is false), and make the text output print the presented decision rather than the engine's per-run one so all three surfaces agree. A regression test should run `resume` on a clean parent with an unsafe child and assert the non-zero exit code.

Adhi1-2:
## Summary

- `cmd_resume` rolled the family verdict into the JSON payload (`mode:
  request_human`, `safe: false`) but still printed the parent's own per-run
  verdict in text mode and still exited 0, because `exit_code_for` was fed
  `decision.mode` rather than the family-overlay result. A script doing
  `resume "$PARENT" && ./start-agent.sh` would launch an agent onto a family
  holding an unreconciled side effect.
- The family overlay is now applied before anything is presented: the
  effective mode drives the text-mode decision line, the permitted-action
  line (family-blocked runs say "none (settle the children below first)"
  instead of "continue"), the `--repair` hint and the exit code. Text mode
  keeps showing the parent's own verdict alongside a FAMILY BLOCKED section,
  but the decision line and exit code follow the family. Only a fully
  verified, safe-to-resume run exits 0, per the exit-code contract in
  exitcodes.py.

Fixes #741

## Testing

- `test_a_family_blocked_parent_never_exits_zero`: JSON mode asserts
  `REQUIRES_HUMAN` (not 0), text mode asserts the REQUEST_HUMAN decision line
  and the absence of "Next permitted action: continue", and reconciling the
  child restores the zero exit.

## Summary by CodeRabbit

- **Bug Fixes**
  - Family-blocked runs now clearly report `REQUEST_HUMAN` instead of indicating they can resume.
  - Resume commands no longer exit successfully when unresolved child actions remain.
  - Text output now identifies blocked families and removes the misleading “continue” action.
  - After child actions are reconciled, resume behavior returns to a successful status.

### DIFF
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -1225,7 +1225,16 @@ def cmd_resume(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
     text = decision.render()
     if family_blocked and decision.mode.value == "resume":
         # House rule: the most cautious signal wins (#243). A clean parent
-        # with an unsafe child is presented as request_human.
+        # with an unsafe child is presented as request_human on every surface:
+        # this text, the JSON payload and the exit code. The engine's per-run
+        # verdict stays visible in the rationale, but it must not read as
+        # permission to continue (issue #741).
+        text = text.replace(
+            "Recovery decision: RESUME", "Recovery decision: REQUEST_HUMAN"
+        ).replace(
+            "Next permitted action: continue",
+            "Next permitted action: none (settle the children below first)",
+        )
         text += "\n\nFAMILY BLOCKED: children of this run are not resumable.\n" + "\n".join(
             f"  !! {r}" for r in family_rationale
         )
@@ -1259,12 +1268,18 @@ def cmd_resume(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
             print(f"error: --pinning: {exc}", file=err)
             return ExitCode.ERROR
 
-    presented_mode = (
-        "request_human"
-        if (family_blocked and decision.mode.value == "resume")
-        else decision.mode.value
+    # The same house rule, as a mode: what the exit code and the JSON report.
+    # Following the engine's per-run mode here let `resume "$PARENT" &&
+    # ./start-agent.sh` exit 0 onto a family holding an unreconciled side
+    # effect, breaking the only-a-verified-safe-run-exits-0 contract
+    # (issue #741).
+    effective_mode = (
+        RecoveryMode.REQUEST_HUMAN
+        if (family_blocked and decision.mode is RecoveryMode.RESUME)
+        else decision.mode
     )
-    presented_safe = decision.safe and not (family_blocked and decision.mode.value == "resume")
+    presented_mode = effective_mode.value
+    presented_safe = decision.safe and effective_mode is decision.mode
     # Advisory prefix-trust (issue #401): deterministic, read-only, never gates.
     try:
         from continuum.analysis.prefix_trust import trust_over_prefix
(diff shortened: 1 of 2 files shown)


Cyrax321 (changes requested):

This PR routes the resume exit code, JSON payload, and rendered text through the same family-aware effective mode, and fixes #741. I ran the family tests on this branch (green) covering both output surfaces. Required CI is green. No blocking correctness issues found.

Cyrax321 (approved):
Verified fix for #741. Resume exits REQUIRES_HUMAN for a family-blocked parent, plus regression tests. Prior changes-requested was attribution-only. Attribution waived per maintainer decision. Approved for merge.

Issue #734, opened 2026-09-07
Retry-budget gate reads only the live tail, so compaction resets an exhausted budget (MCP intercept_action and continuum budget)


Adhi1-2:
## Summary

The retry-budget gate in `continuum_intercept_action` folds `ACTION_RECORDED` events from `storage.read_events(run_id)` (`src/continuum/mcp/server.py:1279`), which returns only the live tail after compaction. `continuum budget` does the same in `cmd_budget` (`src/continuum/cli/main.py:1411`). So once a run is compacted, every failed attempt archived in the prefix stops counting, and an exhausted budget re-opens: a 3-attempt budget that correctly refused a 4th claim before compaction allows it afterwards.

This is the same read_events-only family as #532, #520, #616, and #703, at two more sites. History folds that must survive compaction are supposed to use the archive-aware read.

## Repro

```python
import tempfile, os
from pathlib import Path
from continuum.actions import ActionLedger
from continuum.actions.idempotency import idempotency_key
from continuum.budgets import attempts_by_key, evaluate_budget, load_budgets
from continuum.events import EventType
from continuum.models import Run
from continuum.storage import SQLiteStorage

tmp = tempfile.mkdtemp()
reg = Path(tmp) / "budgets.json"
reg.write_text('{"action_types": {"send_invoice": {"max_attempts": 3}}}')

with SQLiteStorage(os.path.join(tmp, "b.db")) as store:
    store.create_run(Run(run_id="run_1", goal="g"))
    store.append_event("run_1", EventType.RUN_STARTED, {"goal": "g"})
    ledger = ActionLedger(store, "run_1")

    # Three failed attempts: the budget of 3 is now exhausted.
    for i in range(3):
        out = ledger.claim("send_invoice", {"invoice": "I-1"}, key="inv:I-1")
        ledger.fail(out.key, "500", certain=True)

    # Exact gate logic from mcp/server.py intercept_action.
    def gate_attempts(store, run_id):
        events = store.read_events(run_id)  # live tail only
        claim_key = str(idempotency_key("send_invoice", {"invoice": "I-1"},
                                        scope=run_id, key="inv:I-1"))
        attempts = attempts_by_key(events, "send_invoice").get(claim_key, 0)
        allowed, used, maximum = evaluate_budget(load_budgets(reg), "send_invoice", attempts)
        return attempts, allowed

    print("before compaction:", gate_attempts(store, "run_1"))  # (3, False)
    store.compact_run("run_1")
[...]

Issue #242, opened 2026-08-23
Open problem: human-in-the-loop surface - approve, reconcile and audit from the dashboard


Cyrax321:
### Affected component

Dashboard / Recovery ledger

### The open problem

request_human walls are correct but their only door is a CLI. Production HITL needs: durable wait (already modelled - runs park at REQUEST_HUMAN), a surface where a person reads the contract and clicks approve/reject/reconcile, and an audit row for who decided what. The dashboard ships but is read-only.

### Step-by-step plan

1. Dashboard: list runs blocked at REQUEST_HUMAN with the exact repairs; buttons for confirm / reconcile(occurred=true|false) / complete.
2. Server-side: POST endpoints over the same operations as CLI, authenticated by the operator token model already used by MCP authz.
3. Every action lands through existing event types (REVIEW_CONFIRMED, ACTION_RECONCILED) so provenance and audit are unchanged.
4. Optional webhook out: POST request_human occurrences to an operator URL (retry with backoff, idempotent by run+contract hash).

### Scope guardrails

The dashboard must never widen what agents may certify: buttons map 1:1 to human CLI verbs. No new event types.

### Acceptance criteria
- Full confirm/reconcile cycle executed from the browser appears identically (same events, same provenance) to the CLI path.

PR #803, opened 2026-09-07
feat(tui): full-screen terminal dashboard for monitoring and control

### LINKED ISSUE #782: No in-terminal control surface: monitoring and settling runs needs one-shot commands or a browser port
## Problem

CONTINUUM has three surfaces today:

1. **The one-shot CLI.** `runs`, `inspect`, `history`, `actions`, `resume`... one command, one answer, exit. Monitoring a live run means re-typing the same commands in a loop, and every glance at the action ledger or the recovery verdict is a fresh process launch.
2. **The HTML dashboard** (`continuum dashboard`). It carries the run index, the per-run contract, validation, advisories and the human-in-the-loop controls (confirm, reconcile, complete). But it needs a browser, a free port and a bind decision. On a headless box over SSH, which is exactly where a broken run tends to live, it is the awkward option.
3. **The MCP server.** For the agent, not for the human at the keyboard.

There is no surface that opens *in the terminal*, shows everything at once, and lets the operator act without leaving that terminal. The person who most needs to settle an uncertain side effect (reconcile it, confirm the run, force a checkpoint, complete the run) is often on an SSH connection where the one-shot CLI turns every action into a fresh incantation and the web dashboard is out of reach.

## Proposal

A `continuum tui` command: a full-screen terminal dashboard that takes over the screen (alternate screen buffer, like htop or k9s) and hands the terminal back on quit.

Built on the standard library `curses` module. The CLI is deliberately dependency-free ("a recovery tool must not fail to import when you most need it"), and the TUI must not become the first surface with a third-party import. Where `curses` is unavailable (Windows without `windows-curses`) or stdout is not a TTY, the command refuses with a clear message pointing at `continuum dashboard`, rather than half-rendering.

### Monitoring (read-only, mirroring the existing commands so nothing is invented twice)

- runs list: run id, status, event count, goal, recovery mode, safe flag (`runs` plus the dashboard index verdict)
- run overview: goal, version, progress, decisions/findings/evidence counts, plan units, dependencies, validation component statuses (`inspect` / `status`)
- recovery verdict: mode, safe, next allowed action, rationale, human steps, family roll-up (`resume` without `--repair`)
- checkpoints: lineage with version, trigger, completed count (`history`)
- actions: side effects with status, uncertain ones highlighted (`actions`)
- events: the full log including the archived prefix after compaction, scrollable (`events`)
- advisories: liveness and prefix trust (`watch` / `health`)
- retry budgets (`budget`) and the family tree (`tree`)

### Control (mutating, each behind an explicit confirmation prompt)

House style: read-only by default, and when a surface mutates it says so. Every control action shows exactly what it will do and waits for a second keypress before writing:

- confirm self-reported state (`confirm`)
- reconcile an uncertain action as occurred / did not occur (the same `ActionLedger.reconcile` the dashboard HITL buttons use)
- force a checkpoint (`checkpoint`)
- complete a run (`complete`)

The TUI is a presentation layer over the same `Storage`, `CheckpointManager` and `RecoveryEngine` the CLI and the web dashboard use, exactly like `dashboard/app.py`: it invents no new facts, only a new way to look at and act on them. Refresh happens on demand (a key) plus an optional auto-refresh interval, and every assess is read-only until an action is confirmed.

## Non-goals

- Replacing the HTML dashboard: the browser surface stays for browser-first operators and remote sharing.
- Hosting servers (`gateway`, `serve`), client hook management (`hooks`), attestation key handling, benchmarks: those are host-side one-shot operations, not things to watch live.

Adhi1-2:
## Summary

A new `continuum tui` command opens a full-screen terminal dashboard (alternate screen, like htop or k9s) and hands the terminal back on quit, so an operator on a headless box over SSH can watch and settle runs without one-shot commands or a browser port (#782).

Three layers, mirroring how the web dashboard is built:

- **`tui/model.py`** — pure view models over the same `Storage`, `CheckpointManager` and `RecoveryEngine` the CLI uses. The tabs map 1:1 onto existing commands so nothing is invented twice:
  - runs index: `runs` plus each run's recovery verdict (an assessment that raises is shown as `error:` in its row, never dropped)
  - overview: `inspect` (degraded folds named, not hidden)
  - recovery: `resume` without `--repair`, plus human steps, family roll-up, liveness and prefix-trust advisories
  - checkpoints: `history`
  - actions: `actions`, each row carrying the ledger key a reconciliation needs
  - events: `events`, archived prefix included so a compacted run reads the same as one that never was
  - family: `tree`
  - budget: `budget`, archive-aware per #734
- **`tui/app.py`** — `TuiApp`, a key-driven state machine (keys in, lines out) that is fully testable without a terminal, plus a thin curses driver that maps key codes and draws lines, containing no decisions of its own.
- **Control verbs** — confirm, reconcile, force checkpoint, complete run — mirror the human CLI commands and the dashboard HITL buttons (#242): same event types, same `Origin.HUMAN` provenance. Each lands in a confirmation prompt first; only a further `y` performs the write, anything else cancels. Browsing and refreshing never write.

Stdlib `curses` only, keeping the dependency-free rule for the CLI. Where curses is unavailable (Windows without `windows-curses`) or stdout is not a TTY, the command refuses with a pointer at `continuum dashboard` rather than half-rendering. Refresh is on demand (`r`) or an optional `--refresh` interval.

One UX rule worth noting: the status message outranks the key hints in the footer, because the hints are longer than an 80-column line and would otherwise clip the confirmation result the operator most needs to read.

## Testing

- `tests/test_tui.py` (22 tests): every view model against real storage (including compaction-aware event rows and per-key budget drawdown), every app flow (runs index, tab switching, the confirmation gate for all four verbs, a cancelled write leaving the log untouched), the driver through a fake screen, and both refusal paths.
- The real curses path was exercised end to end through a pty (open, navigate tabs, help, confirmed checkpoint write, clean exit 0).
- `docs/api/cli.md` gains the `tui` row; `test_cli_docs` keeps the table honest.

Closes #782

## Summary by CodeRabbit

* **New Features**
  * Added a full-screen terminal dashboard for monitoring and managing runs.
  * View run details, recovery status, checkpoints, actions, events, family information, and budgets.
  * Added confirmation-gated controls for checkpointing, completing runs, confirming state, and reconciling actions.
  * Added the `tui` CLI command with optional automatic refresh intervals.

* **Documentation**
  * Documented the new `tui` command and its `--refresh` option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -583,7 +583,7 @@ All notable changes to this project are documented here. The format follows
   Framework Integration documents the CrewAI/AutoGen/Pydantic-AI thin hooks
   and the gateway/OTel fallback seams; the Roadmap marks the dashboard and
   the enforced-durability work complete; test counts are current
-  (~2,053 collected, ~2,030 passed, ~23 skipped on a minimal env).
+  (~2,089 collected, ~2,030 passed, ~23 skipped on a minimal env).
   <!-- generated via: pytest --collect-only -q; pytest -q -->
 
 - **Gateway hardening and docs refresh.** The enforcing proxy now refuses
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Verify:
 ```bash
 continuum --help                 # CLI entrypoint
 continuum-mcp --help             # MCP server entrypoint (needs [mcp] or [dev])
-pytest -q                        # ~2,053 collected, ~2,030 passed, ~23 skipped on a minimal env (exact counts vary)
+pytest -q                        # ~2,089 collected, ~2,030 passed, ~23 skipped on a minimal env (exact counts vary)
 ruff check src/ tests/ examples/ && ruff format --check src/ tests/ examples/
 mypy src/continuum               # the three gates CI enforces
 ```
@@ -232,7 +232,7 @@ CONTINUUM is verified against real LLM agents, live protocol boundaries, and har
 - **Third-party clients**: Gemini CLI and Kilo Code connected over stdio JSON-RPC against the live SQLite store, validating multi-agent co-existence and authorization isolation.
 - **Protocol compliance**: driven end to end with `@modelcontextprotocol/inspector --cli` across process deaths; mutating tools deny by default behind `CONTINUUM_MCP_MUTATING_CLIENTS`; external claims degrade to `REQUIRES_REVIEW` (`safe: false`).
 - **Self-healing**: hard-killed servers recover from orphaned SQLite `-wal`/`-shm` sidecars via single-retry cleanup at startup.
-- **Scale**: roughly 2,053 tests collected (~2,030 passing; the rest skip without optional services) on Python 3.11, 3.12, and 3.13 (unit, `hypothesis` property-based, concurrency, adversarial). CONTINUUM-Bench runs five crash scenarios plus a dedicated argument-drift scenario, measuring 0 duplicate work and 0 duplicate side effects for CONTINUUM against full duplication for naive replay; a separate 12-scenario recovery-correctness suite (`continuum.benchmark.phase6`) encodes the crash points from the durable-execution survey as executable assertions.
+- **Scale**: roughly 2,089 tests collected (~2,030 passing; the rest skip without optional services) on Python 3.11, 3.12, and 3.13 (unit, `hypothesis` property-based, concurrency, adversarial). CONTINUUM-Bench runs five crash scenarios plus a dedicated argument-drift scenario, measuring 0 duplicate work and 0 duplicate side effects for CONTINUUM against full duplication for naive replay; a separate 12-scenario recovery-correctness suite (`continuum.benchmark.phase6`) encodes the crash points from the durable-execution survey as executable assertions.
 - **Adversarial audit**: the full MCP surface was audited over the live protocol; three defects were found and fixed. Method and reproduction steps in [test.md](test.md).
 
 ## MCP Integration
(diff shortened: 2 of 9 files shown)


Cyrax321 (changes requested):
Thank you for contributing to CONTINUUM. This is a substantial, well-structured feature and most of it is right: the state-machine/driver split is genuinely testable, curses.wrapper restores the terminal, unavailable curses and non-TTY refuse with a dashboard pointer, all four mutating verbs gate on confirmation, and reconciling a certain action is refused with guidance.

One blocking issue, verified live on the PR tree. model.action_rows folds storage.read_events (live tail only), so after compaction the actions tab goes empty while the events tab right beside it still shows every row via read_all_events. I confirmed it: one STARTED claim shows pre-compact and zero rows post-compaction with 3 events archived. An archived uncertain action therefore cannot be selected and cannot be reconciled from the TUI at all. This is the same archive-blindness family as #615/#647, in new code. Please fold read_all_events here like event_rows and budget_rows already do, with a regression test that compacts and then reads the actions tab.

Two non-blocking notes. The dashboard HITL buttons this comment cites as parity do fold live-only today; I am filing that separately rather than asking this PR to fix a second surface, but this PR should not copy the pattern. And a negative --refresh is silently treated as off; validating it (or documenting that) would remove a small surprise.

What is good here beyond the fix: per-key budget drawdown in the tests, the cancelled-write-leaves-log-untouched case, both refusal paths, and the footer-priority detail for 80-column terminals.

Adhi1-2:
Follow-up in bd52f9a: bare `continuum` (no subcommand) at an interactive terminal now opens this dashboard on a landing splash: ANSI Shadow logo (78 columns, fits a standard 80-column terminal), version, run count, and a "press any key" prompt. Any key opens the runs list; q quits. `continuum tui` starts on the same splash. Piped or non-terminal output, `--json`, and platforms without curses keep printing the argparse help unchanged, so scripts running `continuum` blind never see a curses screen.

Cyrax321 (approved):
Verified feat for #782. Fixed the Windows failure myself with a test-only change accepting the no-curses refusal alongside the no-TTY refusal, no runtime change. TUI suite passes locally. Attribution waived per maintainer decision. Approved for merge.

Issue #811, opened 2026-09-07
Mark slow integration tests so contributors get a fast local loop


Cyrax321:
## Context

The full suite is healthy but has a slow tail that every contributor pays on each run. Measured slowest tests on current `main` (`pytest --durations=15`):

- `test_demo_notebook.py::test_notebook_runs_end_to_end` (11.77s)
- `test_horizon_benchmark.py` trio (about 8s combined)
- `test_docs_counts.py::test_documented_count_matches_suite` (5.80s, shells out to a full collect-only run)

There is no marker scheme (`pyproject.toml` has no `markers` section and no test carries a speed mark), so there is no way to run a fast subset during development.

## Why this matters

A fast red-green loop keeps contributors. Waiting minutes to check a one-line change pushes newcomers toward pushing untested code and letting CI catch it.

## Expected outcome

Slow integration tests carry a marker (for example `@pytest.mark.slow`), registered in `pyproject.toml`, so `pytest -m "not slow"` gives a fast local loop while CI keeps running everything. Document the two commands in `docs/CONTRIBUTING_ONBOARDING.md` next to the existing pytest invocation.

## Step by step for the fix

1. Measure with `pytest --durations=15` on `main` to confirm the current slow tail.
2. Register the marker and mark the slowest integration tests (roughly those over 2s).
3. Confirm `pytest -m "not slow"` passes quickly and the full suite still passes unfiltered.
4. Document both commands in the onboarding guide.

## Files to touch

- `pyproject.toml` (marker registration)
- Slowest test files (markers only)
- `docs/CONTRIBUTING_ONBOARDING.md` (two commands)

Mechanical and verifiable, ideal for a first contribution.

bernalalexis-try:
Hi, I'd like to participate if the repo isn't assigned yet :D

Cyrax321:
Thanks, @bernalalexis-try I’ve assigned the issue to you. Feel free to get started, and let me know if you have any questions

bernalalexis-try:
Resolved in #829

PR #829, opened 2026-09-08
test: mark slow integration tests so contributors get a fast local loop

bernalalexis-try:
Registered a slow marker and applied it per test rather than per module, since all three files hold cheap tests worth keeping in the fast loop. Measurement corrected the issue: six slow tests, not three. Added --strict-markers so a mistyped mark fails collection instead of silently staying in the fast loop. pytest -m "not slow" runs in 83.6s against 111s full; collection stays at 2058, so #630's count guard is unaffected.

## Summary by CodeRabbit

* **Tests**
  * Added strict validation for pytest markers.
  * Classified notebook, documentation-count, and horizon benchmark tests as slow-running.
  * Slow tests can be excluded during local development while remaining included in CI runs.

* **Documentation**
  * Documented commands for running the fast local test loop and the full CI-equivalent suite.

### DIFF
--- a/docs/CONTRIBUTING_ONBOARDING.md
+++ b/docs/CONTRIBUTING_ONBOARDING.md
@@ -31,6 +31,8 @@ Install once with `uv sync --extra dev` (or `pip install -e ".[dev]"`).
   <!-- generated via: pytest --collect-only -q; pytest -q -->
   Skips are environmental (Postgres without `CONTINUUM_TEST_POSTGRES_DSN`, adapter tests without `langgraph` or `openai-agents`).
 - Run a single area: `uv run pytest tests/test_checkpoint_phase4.py -v`
+- Fast loop (skip slow integration tests): `uv run pytest -m "not slow"` or `pytest -q -m "not slow"`
+- Full suite (as CI runs it, slow tests included): `uv run pytest` or `pytest -q`
 - Lint: `uv run ruff check src/ tests/ examples/`
 - Format check: `uv run ruff format --check src/ tests/ examples/`
 - Auto fix lint and format: `uv run ruff check --fix src/ tests/ examples/` then `uv run ruff format src/ tests/ examples/`
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -173,7 +173,10 @@ warn_required_dynamic_aliases = true
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 pythonpath = ["src"]
-addopts = "-q"
+addopts = "-q --strict-markers"
+markers = [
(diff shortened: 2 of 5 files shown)


Cyrax321 (changes requested):
Thank you for contributing to CONTINUUM, and well done on measuring first: six slow tests, not three, is exactly the kind of correction that makes the marker honest.

I verified the mechanics on your tree: collection stays at 2058 so the count guard is safe, -m "not slow" deselects exactly the six marked tests, and --strict-markers cannot break collection since the suite only uses asyncio, parametrize, and skipif. Import placement is correct in both files.

One required change before merge: the issue asks for both commands to be documented in docs/CONTRIBUTING_ONBOARDING.md next to the existing pytest invocation, and no docs file is touched. Please add the fast-loop and full-suite commands there.

Two small asks alongside: apply the conventional title form the title bot suggested, and give the two commits conventional messages referencing #811.

One precision note, not a defect: --strict-markers guards against mistyped marks in test files, which is the important half, but a mistyped mark in the -m expression itself still collects silently. Worth knowing, nothing to change.

Cyrax321:
Thank you for the update.

Re checked this session. The requested onboarding docs update is still missing from the file list, and the Windows Python 3.11 test job plus Vercel preview are now failing. Could you please add the fast loop and full suite commands to docs/CONTRIBUTING_ONBOARDING.md and look into the Windows failure? Once docs are in and CI is green I will re review for merge.

Cyrax321:
Hi @bernalalexis-try, thank you for the contribution. Your measurement-first approach stood: the marker registration, the six per-test marks, and strict markers are all yours and untouched.

I have pushed one small follow-up commit on your behalf to finish the remaining review item: the fast-loop and full-suite commands are now documented in docs/CONTRIBUTING_ONBOARDING.md next to the existing pytest invocation (Refs #811). I also fixed the PR title to the conventional form. Your two commits are unchanged, so full credit for the change stays with you.

On the Windows Python 3.11 failure: it is unrelated to this change. The failing test is tests/test_serve_http.py::test_storage_errors_map_to_500_without_killing_the_server with a TimeoutError, while this PR touches only pytest markers and three slow test files. Windows 3.12 and 3.13 passed on the same commit. The new push retriggers CI; if that one HTTP test fails again identically, it is a flaky Windows socket test, not this change.

Cyrax321 (approved):
Hi @bernalalexis-try, re-reviewed after the follow-up.

The marker registration, per-test slow marks, and strict markers are verified as before. The onboarding guide now documents both the fast loop and the full suite, which was the one required item. The Windows 3.11 failure is an unrelated flaky HTTP server socket test, untouched by this change.

Approving. Thank you for the contribution.

PR #843, opened 2026-09-08
test(mcp): exercise the installed continuum-mcp entry point over stdio (#834)

Gouthamkrishnak18:
## Summary

Every other MCP test drives the server in-process (`tests/mcp_helpers.py`) or
through `sys.executable -m continuum.mcp[.server]`. Neither exercises what an
MCP host actually spawns: the `continuum-mcp` console script that
`[project.scripts]` installs unconditionally. That gap is why #697 shipped to
PyPI (the entry point existed where its dependency did not) and why #699
reached users before CI (PATH resolution of the entry point was never under
test) — the failures live between the entry point and the protocol, a stretch
of road the suite had never driven.

Three tests pin that stretch:

1. **the console script**, spawned by absolute path, completes the
   `initialize` handshake and serves `tools/list` (12 tools) over real stdio;
2. **on Windows**, a bare `continuum-mcp` cannot be spawned even when its
   directory is on the *child's* PATH — `CreateProcess` resolves the calling
   process's PATH, not the environment it passes. This is the mechanism
   behind `CONNECTION_CLOSED` (#699), and the reason registration must bake
   resolved paths (the follow-up work in #834) rather than trust the host's
   environment;
3. **`python -m continuum.mcp`** — the form `continuum mcp install` will fall
   back to when no executable is on PATH — completes the same handshake.

Both executable tests skip cleanly when no console script is installed
(editable dev installs provide one; CI matrix coverage is issue #837).

Part of #834. First of a stack: the registration command follows in the next
PR, on top of this baseline.

## Testing

`pytest tests/test_mcp_entrypoint.py` — 3 passed (Windows 11, editable dev
install). Ruff and mypy clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for the installed `continuum-mcp` command using the real stdio JSON-RPC protocol.
  * Verified successful initialization and tool listing through both the console command and the Python module fallback.
  * Added Windows-specific coverage for command resolution behavior when launched through PATH.

* **Documentation**
  * Added a changelog entry documenting the new entry-point and protocol testing coverage.

Issue #699, opened 2026-09-07
.mcp.json uses bare `continuum-mcp` command: server fails to connect in Claude Code on fresh clones


Adhi1-2:
## Problem

The project-scoped `.mcp.json` registers the MCP server with a bare command name:

```json
{
  "mcpServers": {
    "continuum-mcp": {
      "command": "continuum-mcp",
      "args": ["--db", "continuum.db"],
      ...
    }
  }
}
```

`continuum-mcp` only resolves if the contributor's virtualenv is active **in the shell that launched Claude Code**. Per CONTRIBUTING.md the entrypoint lives at `.venv/bin/continuum-mcp` (Windows: `.venv/Scripts/`), which is not on `PATH` by default. On a fresh clone (or any session started without `source .venv/bin/activate`), the server fails to spawn and shows as failed/pending in `claude mcp list`.

## Repro

1. Fresh clone, `python -m venv .venv && pip install -e ".[dev]"` (do not activate the venv)
2. `claude` in the repo
3. `claude mcp list` — `continuum-mcp` fails to connect (`which continuum-mcp` -> not found)

## Suggested fix

Claude Code supports environment variable expansion in `.mcp.json` including `${CLAUDE_PROJECT_DIR:-.}` in the `command` field (the `:-` default is required because the variable is set in the spawned server's environment, not Claude Code's):

```json
"command": "${CLAUDE_PROJECT_DIR:-.}/.venv/bin/continuum-mcp"
```

This resolves regardless of venv activation or `PATH`. Note the POSIX path needs a follow-up for Windows contributors (`.venv/Scripts/continuum-mcp.exe`).

Happy to send a PR.

Issue #697, opened 2026-09-07
continuum-mcp entrypoint installed without mcp dependency and error message references wrong package name


Bipin583:
### Before reporting
- [x] I searched existing issues for a duplicate
- [x] I am reporting this against the correct repository
- [x] I am using a supported Python version (3.11 or later)
- [x] I can reproduce this on the latest commit on `main`

### Affected component
MCP server

### Severity
Major functionality broken, workaround exists

### Regression check
- [ ] This worked in an earlier version or commit

### System information
- **OS**: Microsoft Windows 11 Home Single Language (Version 10.0.26200, Build 26200, 64-bit)
- **Shell**: PowerShell (pwsh 7+)
- **Python**: 3.11.9 (CPython, 64-bit)
- **pip**: 26.0
- **uv**: 0.12.10
- **Package**: `continuum-agent==0.1.0`
- **Installation method**: `pip install continuum-agent` (clean virtual environment / global tool install)

### Command or code to reproduce
```shell
# 1. Install package from PyPI into a clean environment
pip install continuum-agent

# 2. Attempt to run the installed MCP server entrypoint
continuum-mcp
```

### Current behaviour
1. `continuum-agent` installs `pydantic>=2.7` as its only core dependency.
2. In `pyproject.toml`, two global executable console scripts are registered:
   ```toml
   [project.scripts]
   continuum = "continuum.cli.main:main"
   continuum-mcp = "continuum.mcp.server:main"
   ```
3. Because `mcp>=2.0` is listed strictly under optional dependencies (`[project.optional-dependencies] mcp = ["mcp>=2.0"]`), `continuum-mcp` is installed into the user's `bin`/`Scripts` directory as an executable, but fails immediately upon execution:
   ```shell
   error: the MCP server needs the optional 'mcp' dependency, which is not importable (No module named 'mcp'). Install it with: pip install 'continuum[mcp]'
   ```
4. Additionally, the error message directs users to:
   ```
   pip install 'continuum[mcp]'
   ```
   However, the published package on PyPI is `continuum-agent`, not `continuum`. Running `pip install 'continuum[mcp]'` attempts to install an unrelated package or fails.

### Expected behaviour
1. **MCP Dependency Handling**:
[...]

PR #844, opened 2026-09-08
feat(cli): add `continuum mcp install` / `mcp remove` (#834)

### LINKED ISSUE #834: continuum mcp install: cross-platform MCP registration (Windows follow-up to #700)
## Problem

The MCP server cannot be reliably connected by Claude Code on Windows, depending on how the
repository/package was installed. A fresh clone + `pip install -e ".[mcp]"` on Windows followed
by launching Claude Code from a normal (non-activated) terminal yields
`Failed to connect / CONNECTION_CLOSED`. The same happens for pip/PyPI installs whose
`Scripts\` directory is not on the Claude Code process's PATH.

`.mcp.json` was changed in #700 to
`"command": "${CLAUDE_PROJECT_DIR:-.}/.venv/bin/continuum-mcp"`, which fixes POSIX fresh
clones but is **POSIX-only**: on Windows the console script lives at
`.venv\Scripts\continuum-mcp.exe`, so the registered path does not exist. PR #700's own notes
say: "The path is POSIX; Windows contributors using `.venv\Scripts\` still need to adjust it
(or activate the venv). A cross-platform story could be a follow-up." This issue is that
follow-up.

## Root cause (evidence)

On Windows, `CreateProcess` — used by Python's `subprocess`, Node's `spawn`, and therefore by
MCP hosts — resolves a **bare command name using the calling process's own PATH, not the PATH
inside the environment block passed to the child**. Verified experimentally (Windows 11,
Python 3.14, mcp 2.2.0):

- `Popen(["continuum-mcp"])` fails with `WinError 2` even when the child env's PATH contains
  `.venv\Scripts\`; it succeeds only when the *parent's* PATH does.
- `Popen(["continuum-mcp.exe"])` (with extension, Scripts on child PATH) still fails.
- Full-path spawn and `cmd /c continuum-mcp` (cmd.exe does its own PATH+PATHEXT search) both
  work, and the server then completes a full `initialize` handshake with all 12 tools — the
  server itself is healthy on Windows.

A committed JSON file cannot express "`.venv/bin/continuum-mcp` on POSIX,
`.venv\Scripts\continuum-mcp.exe` on Windows": static config has no platform conditionals.
Resolution must therefore happen **on the machine that will spawn the server**, not in the
committed file.

## Proposed solution

Add `continuum mcp install` (and `continuum mcp remove`), following the architecture the
repository already trusts for `continuum hooks install` (`src/continuum/clienthooks.py`):

1. **Verify the `mcp` SDK is importable by actually spawning a probe subprocess** before
   writing any configuration. If it is not, print the exact remediation
   (`pip install "continuum-agent[mcp]"`) and exit non-zero without touching config.
   (The spawn probe is required, not optional: `clienthooks.py`'s
   `sys.executable -m continuum.cli` fallback is correct for any *installed* environment but
   silently produces a dead command for PYTHONPATH-only source trees.)
2. **Resolve the real command at install time**: `shutil.which("continuum-mcp")`, falling
   back to `[sys.executable, "-m", "continuum.mcp"]` — the fallback is what makes
   venv/editable installs work on Windows with zero PATH assumptions.
3. **Write the host registration with absolute paths**: default Claude Code local scope
   (`~/.claude.json`); `--scope project` for a project-scope write. Idempotent
   install/update/remove using the same narrow self-recognition pattern as
   `_is_managed_hook` (never touch entries we did not write).
4. **Bake an absolute `--db`** (default: `<project>/continuum.db`) instead of the cwd-relative
   `continuum.db`, because every config path in the codebase is cwd-relative and the host's
   spawn cwd is neither documented nor guaranteed to be the project root. (This also matters
   for security posture: `.continuum/mcp-policy.json` resolves against `Path.cwd()`
   (`mcp/authz.py:436`) and silently degrades to deny-by-default from a different cwd.)

Structure the host profiles like `CLIENT_PROFILES` so cursor/windsurf/codex can be added
later without redesign.

## Scope

- New module (e.g. `src/continuum/mcp/install.py`) + CLI wiring in `cli/main.py`.
- `mcp remove` counterpart.
- Unit + subprocess tests.

## Out of scope

- `continuum mcp doctor` (Issue 2), CI matrix (Issue 4), docs rewrite (Issue 7),
  other hosts' profiles beyond the structural seam, changing whether `mcp` is an extra
  (Issue 3).

## Acceptance criteria

- On Windows: fresh clone + venv + `pip install -e ".[mcp]"` + `continuum mcp install`, then
  Claude Code launched from a **non-activated** shell → server connects and lists 12 tools.
- Same on macOS and Linux.
- Re-running `mcp install` updates the entry in place (no duplicates); `mcp remove` deletes
  only entries this command wrote.
- With the `mcp` extra absent, `mcp install` exits non-zero, prints the correct remediation,
  and writes nothing.
- The installed registration works regardless of the host's PATH and spawn cwd.

## Test plan

- In-process unit tests over tmp settings files: resolution, fallback form, idempotency,
  narrow recognition, absolute-db baking (mirroring `tests/test_client_installers.py`).
- Subprocess test spawning the **`continuum-mcp` console script** (not `python -m`) and
  completing `initialize` + `tools/list` — the test the suite has never had.
- Windows CI runner must exercise the fallback form.

## Platform coverage

Windows / Linux / macOS (all first-class).

## Dependencies

None — this is the first issue; Issues 2, 3, 7, 8 build on it.

## Contributor notes

- Pattern to reuse: `src/continuum/clienthooks.py:215-264` (`observe_command`,
  `_is_managed_hook`, `install_client_hook`), `cli/main.py:2533` (`cmd_hooks_install`).
- Claude Code scopes and the "local beats project" rule: `docs/api/mcp.md:152-176`.
- Windows command quoting already solved in-repo: `clienthooks.py:190,206`
  (`list2cmdline` / `shlex.split(posix=False)`).
- Reproduction scripts used for the root-cause analysis: `_repro/spawn_test.py`,
  `_repro/path_isolation.py`, `_repro/createprocess_probe.py`.

Gouthamkrishnak18:
## Summary

Closes #834. Follow-up to #699/#700 (that fix is POSIX-only) and #697.

A bare `continuum-mcp` in `.mcp.json` only resolves when the host's own PATH
contains the install environment. On Windows this fails by construction:
`CreateProcess` resolves a bare command name against the calling process's
PATH, never the environment it passes to the child, and a committed JSON file
cannot name both `.venv/bin/continuum-mcp` and `.venv/Scripts/continuum-mcp.exe`.
The registration therefore has to be written on the machine that spawns the
server.

`continuum mcp install` does for MCP what `hooks install` does for hooks
(#580's pattern):

- **resolve** the entry point: executable on PATH → sibling of the running
  `continuum` → `python -m continuum.mcp` (matches
  `clienthooks.observe_command`, so the two installers cannot disagree);
- **verify** by driving a real `initialize` handshake over stdio — a
  throwaway temp DB, never the project's. This catches the #697 state (entry
  point without the `mcp` extra), a stale entry point from a moved venv, and
  a server that cannot open storage, all *before* any config is written: a
  failed install leaves the host's config untouched. The server's own stderr
  is surfaced in the error, since the host never shows it;
- **register** with the resolved absolute path baked in, in the shape the
  host itself writes (verified against `claude mcp add`): local scope
  (default) under `projects[cwd].mcpServers` in `~/.claude.json` with the
  project key posix-spelled, user scope top-level, project scope the
  committed `.mcp.json` (reported as machine-specific, since that file is
  usually committed).

Idempotent by key: re-running converges (`present`), a moved venv repoints
(`updated`). `mcp remove` drops only the `continuum-mcp` key, leaving every
other server and project entry alone. Invalid host JSON is refused, not
clobbered (the `clienthooks._install_hook` contract). `mcp` bypasses storage
so config edits create no `continuum.db`.

### One deliberate deviation from the issue text

`--db` defaults to the same relative `continuum.db` the committed
`.mcp.json` names, resolved by the host per project — not an absolute path.
An absolute default cannot be right from a per-user config: it would point
every project at one database. `--db` bakes an explicit path for operators
who want one.

## Testing

`tests/test_mcp_install.py` — 15 tests: per-scope shapes, idempotency and
repointing, unrelated-key preservation, probe refusal leaving config
untouched, stderr surfacing, foreign-server rejection, remove-only-our-key,
three-step resolution. Full suite: 2,130 collected, only the environmental
Postgres test fails locally (no local server; fails identically at base).
Ruff and mypy clean. Live smoke-tested on Windows 11 (install → present →
updated → remove → not-found).

Stacked on #843 (the entry-point regression tests); docs count figures
re-synced to 2,130 here.

### DIFF
--- a/src/continuum/mcp/install.py
+++ b/src/continuum/mcp/install.py
@@ -1,0 +1,369 @@
+"""Register the MCP server with a host so the host can actually spawn it (issue #834).
+
+``.mcp.json`` declares the server by bare command name, and that name resolves only
+when the *host's* PATH contains the environment CONTINUUM was installed into. On
+Windows this fails by construction: ``CreateProcess`` resolves a bare command name
+against the calling process's own PATH, never the environment it passes to the child,
+and a venv's ``Scripts`` directory joins PATH only inside a shell that activated it.
+The host reports the failed spawn as ``CONNECTION_CLOSED``, which reads like a crash
+but describes an executable that was never found (#699). A committed JSON file cannot
+name both ``.venv/bin/continuum-mcp`` and ``.venv\\Scripts\\continuum-mcp.exe``, so no
+static config can fix this for every machine at once.
+
+The fix is to resolve the command *on the machine that will spawn the server*, which
+is what this module does: ``continuum mcp install`` resolves the real entry point
+(the ``continuum-mcp`` executable, or ``<python> -m continuum.mcp`` when no executable
+is on PATH — the same two-step resolution ``hooks install`` uses for its commands,
+``clienthooks.observe_command``), proves it serves the protocol by driving a real
+``initialize`` handshake over stdio, and only then writes the registration with the
+resolved absolute path baked in.
+
+The handshake probe is what separates this from a PATH lookup: it catches the
+missing-``mcp``-extra state (#697, where the entry point exists but its dependency
+does not), a stale entry point from a moved virtualenv, and a server that starts but
+cannot open storage — all before any configuration is written, so a failed install
+leaves the host's config untouched rather than registering a server known not to
+work.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import threading
+from collections.abc import Mapping, Sequence
+from pathlib import Path
+from typing import Any
+
+__all__ = [
+    "MCP_PROFILES",
+    "ProbeError",
+    "SERVER_KEY",
+    "build_registration",
+    "install_server",
+    "probe_server",
+    "remove_server",
+    "resolve_server_command",
+]
+
+#: The server key every host config uses. MCP servers are registered by name, so
+#: unlike hooks (whose entries are anonymous and recognised by command shape,
+#: ``clienthooks._is_managed_hook``) the key *is* the identity: whatever wrote the
+#: entry under this name registered the CONTINUUM server.
+SERVER_KEY = "continuum-mcp"
+
+#: Which client name the authz layer trusts with mutating tools. The committed
+#: ``.mcp.json`` grants ``claude-code``; the installer keeps that default and keys
+#: it to the profile so a future host profile grants its own client name.
+_MUTATING_CLIENTS_ENV = "CONTINUUM_MCP_MUTATING_CLIENTS"
+
+#: Per-host wiring profiles, in the spirit of ``clienthooks.CLIENT_PROFILES``
+#: (#209): everything that differs between hosts is data. A host's config file,
+#: where its scopes live, and whether its "local" scope is per-project (Claude Code
+#: nests local registrations under ``projects[<cwd>].mcpServers`` in the user config)
+#: is all a host needs to describe here for install/remove to work.
+MCP_PROFILES: dict[str, dict[str, Any]] = {
+    "claude-code": {
+        # local and user scopes share ~/.claude.json; project scope is the
+        # committed .mcp.json at the project root.
+        "user_config": "~/.claude.json",
+        "project_config": ".mcp.json",
+        # Claude Code's local scope is stored per project directory inside the
+        # user config (verified against `claude mcp add`: projects[cwd].mcpServers,
+        # with the directory spelled in forward slashes even on Windows).
+        "local_is_per_project": True,
+        "mutating_client": "claude-code",
+    },
+}
+
+
+class ProbeError(Exception):
+    """The server could not be started or did not complete the MCP handshake."""
+
+
+def resolve_server_command() -> list[str]:
+    """The argv that runs the MCP server on this machine, right now.
+
+    Three steps, most desirable first: an executable on PATH (what a wheel, pipx
+    or ``uv tool`` install provides); an executable sitting beside the running
+    ``continuum`` (the common venv case where ``continuum mcp install`` was invoked
+    from inside the environment but PATH resolution is unavailable to the host);
+    and the interpreter-plus-module form, which needs no PATH at all and is what
+    makes editable and unactivated-venv installs work on Windows. The forms match
+    ``clienthooks.observe_command``'s resolution so the two installers cannot
+    disagree about how to reach a CONTINUUM entry point.
+    """
+    direct = shutil.which(SERVER_KEY)
+    if direct:
+        return [direct]
+    sibling_name = f"{SERVER_KEY}.exe" if os.name == "nt" else SERVER_KEY
+    running = shutil.which("continuum")
+    if running:
+        sibling = Path(running).parent / sibling_name
+        if sibling.exists():
+            return [str(sibling)]
+    return [sys.executable, "-u", "-m", "continuum.mcp"]
+
+
+def probe_server(command: Sequence[str], *, timeout: float = 20.0) -> dict[str, Any]:
+    """Prove ``command`` serves MCP by driving a real ``initialize`` handshake.
+
+    Raises :class:`ProbeError` — never writes anything, spawns nothing else — so
+    the caller can refuse to register a server that does not work. The probe's
+    database is a throwaway in the system temporary directory: verifying a
+    registration must not leave ``continuum.db`` files in the project, whose
+    presence the CLI treats as state.
+
+    The environment is inherited rather than minimal: on Windows a child denied
+    ``SystemRoot`` dies during interpreter startup before any CONTINUUM code runs
+    (issue #211).
+    """
+    tmp = Path(tempfile.mkdtemp(prefix="continuum-mcp-probe-"))
+    env = dict(os.environ)
+    env["CONTINUUM_DB"] = str(tmp / "probe.db")
+    errors: list[str] = []
+    stderr_thread: threading.Thread | None = None
+    try:
+        try:
+            proc = subprocess.Popen(  # noqa: SIM115 - cleaned up in finally
+                list(command),
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                env=env,
+                text=True,
+                bufsize=1,
+            )
+        except OSError as exc:
+            raise ProbeError(
+                f"cannot start the MCP server ({exc}). The registered command must "
+                f"be spawnable by the host; check the installation."
+            ) from exc
+
+        # Drain stderr on a thread so a chatty server cannot deadlock the probe by
+        # filling the pipe buffer while we block reading stdout.
+        def drain_stderr() -> None:
+            assert proc.stderr is not None
+            errors.extend(line.rstrip() for line in proc.stderr)
+
+        stderr_thread = threading.Thread(target=drain_stderr, daemon=True)
+        stderr_thread.start()
+
+        try:
+            assert proc.stdin is not None
+            proc.stdin.write(
+                json.dumps(
+                    {
+                        "jsonrpc": "2.0",
+                        "id": 1,
+                        "method": "initialize",
+                        "params": {
+                            "protocolVersion": "2024-11-05",
+                            "capabilities": {},
+                            "clientInfo": {"name": "continuum-mcp-install", "version": "0"},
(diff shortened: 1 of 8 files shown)


anya-research:
**Code Review**

This PR adds `continuum mcp install`/`remove` to resolve the MCP server entry point on the installing machine, verify it with a real `initialize` handshake into a throwaway DB, and register the absolute path in the host config (local/user/project scopes). The overall design is solid: probe-before-write prevents registering a broken server, idempotency is handled correctly, and the scope mapping is verified against the host itself. Two exception-handling issues keep the error path from being as robust as the happy path.

**high**
`src/continuum/mcp/install.py:158`
`probe_server` writes the `initialize` frame with `proc.stdin.write()`/`flush()` inside a `try` that only catches `json.JSONDecodeError`. If the resolved command exits before reading stdin, exactly the #697 missing-`mcp`-extra case the probe exists to catch, the write raises `BrokenPipeError`/`OSError` which escapes both the inner handler and the outer `except ProbeError` in `src/continuum/cli/main.py:278`, surfacing as a raw traceback instead of the intended `refusing to register` diagnostic and leaving the child un-reaped. Wrap the write/flush and convert pipe errors to `ProbeError`.
```suggestion
        try:
            assert proc.stdin is not None
            try:
                proc.stdin.write(
                    json.dumps(
                        {
                            "jsonrpc": "2.0",
                            "id": 1,
                            "method": "initialize",
                            "params": {
                                "protocolVersion": "2024-11-05",
                                "capabilities": {},
                                "clientInfo": {"name": "continuum-mcp-install", "version": "0"},
                            },
                        }
                    )
                    + "\n"
                )
                proc.stdin.flush()
            except (BrokenPipeError, OSError) as exc:
                raise ProbeError("the MCP server exited before accepting the handshake request") from exc
```

**medium**
`src/continuum/mcp/install.py:205`
The probe re-raises with augmented stderr as `raise ProbeError(f"{exc}\nThe server reported:\n{tail}") from None`, which discards the original `ProbeError` chain and its traceback. Diagnostics for handshake failures then lose the cause that explains *why* the handshake failed, making `mcp install` harder to debug and violating the `raise ... from exc` pattern used for the `OSError` case above. Preserve context.
```suggestion
        tail = "\n".join(errors[-5:]).strip()
        if tail:
            raise ProbeError(f"{exc}\nThe server reported:\n{tail}") from exc
        raise
```


[github run](/Cyrax321/CONTINUUM/actions/runs/34248205588)

Cyrax321 (dismissed):

This PR adds continuum mcp install/remove with handshake-probed, absolute-path registration, refs #834. I read the new installer end to end: PATH resolution with interpreter fallback, a bounded stdio probe that refuses before writing anything, malformed configs refused rather than clobbered, remove scoped to our own key, and argparse-constrained client and scope selectors. I ran the 15 install tests on this branch (green) and full CI is green. No blocking correctness issues found.

One required step before merge, mechanical rather than code: this branch has merge conflicts in the test-count lines (README, onboarding, CHANGELOG bump 2130 against main now reading 2122 with 155 test files after #644). Please update the branch, taking main side on those figures, then let CI re-run. Landing #843 first will also shrink this diff back to its own commit.

Thanks again. Looking forward to the updated branch.

src/continuum/mcp/install.py
```diff
@@ -0,0 +1,369 @@
...
+                        "params": {
+                            "protocolVersion": "2024-11-05",
+                            "capabilities": {},
+                            "clientInfo": {"name": "continuum-mcp-install", "version": "0"},
+                        },
+                    }
+                )
+                + "\n"
+            )
+            proc.stdin.flush()
```
Cyrax321, line 173: Blocking: probe write can raise an uncaught BrokenPipeError on an instant exit server.

In probe_server the stdin write and flush are inside a try whose only except is json.JSONDecodeError, and the outer handler catches only ProbeError. I verified this shape on the head commit. When the spawned command exits before reading stdin, exactly the missing extra case this probe exists to diagnose, the write raises BrokenPipeError or OSError which escapes both handlers as a raw traceback instead of the clean refusing to register message.

Please wrap the write and flush so pipe errors convert into ProbeError, with a regression test spawning a command that exits immediately.

Cyrax321 (changes requested):

This adds continuum mcp install and remove with handshake probing across eight files. One correction needed before approval. The probe write path leaves BrokenPipeError and OSError uncaught, verified on the head commit, so the very failure the probe is meant to catch surfaces as a traceback. Please convert pipe errors into ProbeError with a fast exiting server test.

Once that is in with green CI, this should be quick to approve.

Cyrax321 (dismissed):
Thank you for contributing to CONTINUUM. Cross platform mcp install and remove for #834 is the fix the Windows story has been waiting for, and green across ubuntu, macos and Windows plus the foreign cwd acceptance coverage in test_mcp_install.py is convincing. Appreciate the work.

Verified from the file list: install module, CLI wiring, entrypoint tests, and docs updates in docs/api/cli.md. One coordination note: PR #1014 implements the same #834 surface but is red on all three Windows jobs, so the two overlap directly. Since this one is green where #1014 fails, suggest landing this one and closing #1014 as superseded, carrying over anything unique from #1014 first (its HOST_PROFILES extension point is worth a look). Please confirm the overlap check with #1014 before landing.

Approving. Thanks for getting the registration story right.

src/continuum/mcp/install.py
```diff
@@ -0,0 +1,382 @@
...
+    """
+    profile = profile or MCP_PROFILES["claude-code"]
+    settings = _load_config(config_path)
+    servers = _mcp_servers_for(settings, scope=scope, project_dir=project_dir, profile=profile)
+    existing = servers.get(SERVER_KEY)
+    if existing == dict(registration):
+        return "present"
+    servers[SERVER_KEY] = dict(registration)
+    config_path.parent.mkdir(parents=True, exist_ok=True)
+    config_path.write_text(json.dumps(settings, indent=2) + "\n", encoding="utf-8")
```
Cyrax321, line 354: Non-blocking, and I checked the precedent: `clienthooks.py` writes its settings files the same way (clienthooks.py:431, :536), so this follows the existing convention rather than setting a worse one.

Flagging it because of what the file is. For the local and user scopes this rewrites the operator's entire `~/.claude.json`, which can hold every project and every MCP server they have registered. `write_text` is not atomic, so an interrupted write, a full disk, or a crash mid-write truncates it. `_load_config` refusing to touch a file that is not valid JSON is the right guard, but it only takes effect on the next run, after the file is already damaged.

Writing to a temp file in the same directory and `os.replace`-ing into place makes the write crash-safe, and it would be a natural change to make in `clienthooks` at the same time, since both modules write the same kind of file and would want the same guarantee.

Cyrax321 (dismissed):
Approving. The resolve → verify → register ordering is the right design and the part that matters most is that verification is a real `initialize` handshake over stdio, not a PATH check, so the #697 state, a moved venv, and a server that cannot open storage are all caught before the host's config is touched rather than after. The scope handling matches Claude Code's own layout, including the posix-spelled project key, and idempotency plus unrelated-key preservation are both pinned by tests. I ran the two suites: 18 passed, 1 skipped.

One thing to settle before merging, and it is a process point rather than a code one: #845 is stacked on this branch and also targets `main`. I compared the two trees and #845 is exactly this PR plus `doctor.py`, with `install.py`, `test_mcp_install.py`, and `test_mcp_entrypoint.py` byte-identical, and its `main.py` and docs differing only by the doctor additions. So this PR should merge first and #845 should rebase onto it. If #845 goes in first, this one is fully subsumed and closes as a duplicate, which is a clean outcome too but not the one the branch structure implies. Worth deciding deliberately rather than letting merge order pick one by accident.

Cyrax321 (changes requested):
Author context: Gouthamkrishnak18, external contributor. Follow-up to the prior reviews. The install/remove design with handshake probing is still sound, but this is in dirty merge state against a stale base with doc-count conflicts, and it overlaps #1014 on the same #834 surface. Rebase onto current main, resync the stating figures, confirm the ProbeError handling for pipe failures is present on the rebased head, and confirm the overlap check with #1014.

PR #845, opened 2026-09-08
feat(cli): add `continuum mcp doctor` (#835)

### LINKED ISSUE #835: continuum mcp doctor: turn CONNECTION_CLOSED into a diagnosis
## Problem

When the MCP server fails to connect, the user sees exactly one opaque string
(`CONNECTION_CLOSED` / "server never became ready"). The real causes — an executable the host
cannot spawn, or a missing optional dependency — produce **identical** client output, and the
useful stderr never reaches the user. `docs/api/mcp.md` already documents this failure mode
("a message that reads like a server that crashed but describes an executable that was never
found"), but the diagnosis currently requires the user to already understand PATH resolution
on Windows.

## Root cause (evidence)

- Bare-name spawn fails at `CreateProcess` before any CONTINUUM code runs, so the server
  cannot self-report (verified: `WinError 2` even with the Scripts dir in the *child* env's
  PATH).
- With the `mcp` extra missing, `continuum-mcp` exits 1 pre-handshake with a good stderr
  message that the host does not surface (#87/#93, #697).
- Diagnosis therefore has to happen client-side / on-demand, which no command currently does.

## Proposed solution

Add `continuum mcp doctor` that checks, in order, and prints one actionable line per finding:

1. `mcp` SDK importability (in a fresh subprocess, not in-process) — with the exact install
   command if missing.
2. The command the hosts would use: `shutil.which("continuum-mcp")` **as resolved from a
   freshly spawned process** (not the current process), plus the
   `sys.executable -m continuum.mcp` fallback availability.
3. A **live spawn + `initialize` handshake probe** of the resolved command, reporting
   success/failure with the child's stderr tail — the single check that catches every
   cold-start failure class at once.
4. PATH visibility diagnosis (where the exe is, which dirs are on the probe process's PATH).
5. Wire-framing note: on Windows the MCP SDK currently terminates stdio frames with `\r\n`
   (upstream modelcontextprotocol/python-sdk#2433, confirmed on mcp 2.2.0) — tolerated by
   Claude Code, rejected by strict NDJSON clients; print a note when relevant.
6. Pointers to the remedies: `continuum mcp install` (#834), extra installation, docs.

Machine-readable output via the existing `--json` convention for scripting/CI use.

## Scope

- New subcommand + tests; reuses the probe logic from #834's verification step.

## Out of scope

- Fixing the SDK CRLF bug (upstream); changing registration (#834); CI (Issue 4).

## Acceptance criteria

- In each reproduced failure state (missing extra; exe not on host PATH; working install),
  `mcp doctor` exits non-zero/zero accordingly and its output names the cause and the fix.
- Running `mcp doctor` on a healthy install completes a real handshake against the real
  console script and reports all tools.

## Test plan

- Subprocess-based tests for each failure state using the same fake-missing-SDK technique as
  `tests/test_mcp_server.py:2007` (`sys.meta_path` blocker) and PATH-stripped environments.
- Handshake probe test asserting a parsed `initialize` result.

## Platform coverage

Windows / Linux / macOS.

## Dependencies

Pairs with #834 (shares the resolution + probe code); can be implemented immediately after.

## Contributor notes

- Existing error-message conventions: `mcp/server.py:1600-1636`, `cli/main.py` `error:` style.
- Probe implementation sketch exists in `_repro/spawn_test.py` (handshake) and
  `_repro/createprocess_probe.py` (minimal-env spawn).

Gouthamkrishnak18:
## Summary

Closes #835.

When the MCP server fails to connect, the host shows exactly one opaque
string (`CONNECTION_CLOSED`), and the real causes produce identical client
output while the server's useful stderr never reaches the user: a bare-name
spawn fails at `CreateProcess` before any CONTINUUM code runs, so the server
cannot self-report. Diagnosis has to happen client-side — which no command
did.

`continuum mcp doctor` checks each failure class in order, prints one
actionable line per finding, and names the fix. Every check runs against a
**fresh subprocess**, never the doctor's own process (whose `sys.path` a
host's spawn would not share):

1. **`mcp` SDK importability** with the exact install command when missing
   (#697's class); the fresh-interpreter version is reported on success.
2. **entry point**: whether a freshly spawned process can resolve
   `continuum-mcp` on PATH, plus the full command `mcp install` would
   register (resolution shared with the installer, so the two can never
   disagree).
3. **live `initialize` + `tools/list` handshake** against that command — the
   one check that covers every cold-start failure class at once — surfacing
   the server's own stderr, which the host swallows.
4. **CRLF framing** on Windows: the SDK terminates stdio frames with `\r\n`
   (modelcontextprotocol/python-sdk#2433, confirmed on mcp 2.2.0). The probe
   reads binary pipes because text-mode translation strips the `\r` and hides
   the defect from exactly the tool that exists to see it.

Exit status is scriptable: zero only when no check failed. The interpreter
fallback is a *warning* (the server works; the operator should know how it
is reached), while an executable a fresh process cannot find is a *failure*
(#699: a bare-name host config is a `CONNECTION_CLOSED` waiting to happen
even when the server itself is healthy).

## Testing

`tests/test_mcp_doctor.py` — 6 tests, including an end-to-end reproduction
of the missing-extra state in a fresh interpreter via a `sitecustomize`
import blocker (the same technique as `test_mcp_server.py`'s missing-extra
test, applied at interpreter startup so the subprocess checks see what a
host's spawn would hit), the off-PATH failure with the `mcp install`
remedy, and the wire-level CRLF detection. Full suite: 2,136 collected,
only the environmental Postgres test fails locally (no local server; fails
identically at base). Ruff and mypy clean. Live smoke-tested on Windows 11
against the real server (12 tools listed, CRLF detected).

Stacked on #844 (shares the resolution code); collected-count figures
re-synced to 2,136.

## Summary by CodeRabbit

- **New Features**
  - Added `continuum mcp install` and `continuum mcp remove` for MCP server registration at local, user, or project scope.
  - Added `continuum mcp doctor` to diagnose executable, protocol, connection, and platform-specific issues.
  - Installations validate the server before updating host configuration and use absolute paths.

- **Documentation**
  - Documented MCP setup, scopes, diagnostics, and configuration behavior.
  - Updated reported test-suite counts to approximately 2,136 collected tests.

- **Tests**
  - Added end-to-end coverage for MCP installation, removal, diagnostics, and stdio entry points.

### DIFF
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -2836,6 +2836,39 @@ def cmd_mcp_install(args: argparse.Namespace, storage: Storage, out: Any, err: A
     return ExitCode.OK
 
 
+def cmd_mcp_doctor(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -> int:
+    """Diagnose why the MCP server cannot connect, naming cause and fix (issue #835).
+
+    ``CONNECTION_CLOSED`` is all a host ever shows, and the causes — an
+    executable the host cannot spawn, a missing optional dependency — produce
+    identical client output while the server's useful stderr never reaches the
+    user. The doctor runs each check against a fresh subprocess, reports one
+    actionable line per finding, and exits non-zero when any check failed, so
+    ``continuum mcp doctor && <reconnect>`` is safe to script.
+    """
+    from continuum.mcp.doctor import run_doctor
+
+    report = run_doctor()
+    lines = ["MCP doctor"]
+    for check in report["checks"]:
+        marker = {"ok": "[ok]", "warn": "[warn]", "fail": "[FAIL]", "info": "[info]"}[
+            check["status"]
+        ]
+        lines.append(f"  {marker} {check['name']}: {check['detail']}")
+        if check.get("fix"):
+            lines.append(f"        fix: {check['fix']}")
+    if not report["ok"]:
+        lines.append("the MCP server is not healthy as this machine can reach it")
+    _emit(
+        report,
+        "\n".join(lines),
+        as_json=args.json,
+        stream=out,
+        palette=getattr(args, "_palette", None),
+    )
+    return ExitCode.OK if report["ok"] else ExitCode.ERROR
+
+
 def cmd_mcp_remove(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -> int:
     """Drop the ``continuum-mcp`` registration from a host config (issue #834).
 
@@ -4103,6 +4136,12 @@ def build_parser() -> argparse.ArgumentParser:
         "remove", help="Remove the continuum-mcp registration. Mutates host config."
     )
     mcp_client(mcp_remove, cmd_mcp_remove)
+
+    mcp_doctor = mcp_sub.add_parser(
+        "doctor",
+        help="Diagnose why the MCP server cannot connect. Read-only; runs a live handshake.",
+    )
+    mcp_doctor.set_defaults(func=cmd_mcp_doctor)
 
     verify = with_run(add("verify", cmd_verify, "Re-audit the event chain."))
     verify.add_argument(
--- a/src/continuum/cli/main.py
+++ b/src/continuum/cli/main.py
@@ -53,6 +53,7 @@ from continuum.gate import (
 from continuum.gate import (
     decide as gate_decide,
 )
+from continuum.mcp.install import MCP_PROFILES
 from continuum.models import (
     ActionStatus,
     EnvironmentSnapshot,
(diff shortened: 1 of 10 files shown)


Cyrax321 (dismissed):

This PR adds continuum mcp doctor, which turns CONNECTION_CLOSED into a diagnosis, refs #835. I read the new module end to end: every check runs against a fresh subprocess rather than this process, the SDK, entry-point, handshake, and Windows CRLF checks are ordered cheapest-first with actionable fixes, exit status is scriptable with warnings kept non-failing, and resolution is shared with the installer so the two cannot disagree. I ran the 6 doctor tests on this branch (green) and full CI is green. No blocking correctness issues found.

One required step before merge, mechanical rather than code: like #844 this branch is stacked and conflicted on the test-count lines, so please update it once #843 and #844 land and let CI re-run.

Thanks again. Looking forward to the updated branch.

src/continuum/cli/main.py
```diff
@@ -3915,6 +4086,63 @@ def hooks_client(p: argparse.ArgumentParser, func: Any) -> None:
...
+    mcp_remove = mcp_sub.add_parser(
+        "remove", help="Remove the continuum-mcp registration. Mutates host config."
+    )
+    mcp_client(mcp_remove, cmd_mcp_remove)
+
+    mcp_doctor = mcp_sub.add_parser(
+        "doctor",
+        help="Diagnose why the MCP server cannot connect. Read-only; runs a live handshake.",
+    )
+    mcp_doctor.set_defaults(func=cmd_mcp_doctor)
```
Cyrax321, line 4144: Blocking: mcp doctor uses args.json but the subparser never registers --json.

cmd_mcp_doctor passes as_json=args.json into _emit, verified at the head commit, but the mcp_doctor parser only sets help text and func with no --json argument. The global continuum --json form works, yet the documented trailing form continuum mcp doctor --json is rejected by argparse. Every sibling in this file registers its own trailing --json, including health per #677.

Please register --json on the mcp_doctor parser the same way, so both orders work.

Cyrax321 (changes requested):

This adds continuum mcp doctor with live handshake diagnosis across ten files. Two corrections needed before approval. First, the trailing --json flag is used but never registered on the mcp_doctor parser, verified on the head commit, so the documented invocation fails. Second, the failure summary reads the MCP server is not healthy as this machine can reach it, which drops a word and should read as far as this machine can reach it.

Smaller notes for the same pass. The SDK and PATH subprocess probes run without a timeout while the handshake probe is bounded, so a stalled spawned interpreter would hang a command meant to be script safe. Bounding those probes would match the stated scripting goal. I also checked the em dash claim against the head file and found zero U+2014 characters there, so that part appears already clean.

Once the --json registration and the failure line are fixed with tests, this should be quick to approve.

Cyrax321 (dismissed):
Thank you for contributing to CONTINUUM. A working mcp doctor for #835 is valuable, and full CI green including all three Windows jobs is exactly what this cross platform surface needs. Appreciate the large test investment across test_mcp_doctor.py and test_mcp_entrypoint.py.

Verified from the diff: doctor module with diagnosis wiring through the CLI, docs in docs/api/cli.md, and the entrypoint realism coverage that tells a broken install from an unreachable one. One coordination note: PR #1013 implements the same #835 doctor surface but is red on Windows, so the two overlap directly. Since this one is green everywhere, suggest landing this one and closing #1013 as superseded, with anything unique in #1013 carried over first. Please confirm the overlap check with #1013 before landing.

Approving. Thanks for the thorough green run.

src/continuum/mcp/doctor.py
```diff
@@ -0,0 +1,381 @@
...
+        raise ProbeError(
+            f"the MCP server did not complete the initialize handshake within {timeout:.0f}s"
+        )
+    line = result[0]
+    if not line:
+        raise ProbeError("the MCP server exited before answering the handshake")
+    return line
+
+
+def _handshake(command: Sequence[str], *, timeout: float = 20.0) -> dict[str, Any]:
```
Cyrax321, line 221: Non-blocking, but worth factoring. This and `install.probe_server` (install.py:112) are two implementations of the same thing: spawn the command with pipes, drain stderr on a thread, write initialize, read one bounded line, augment any ProbeError with the server's stderr, tear down a temp DB in a finally. Two `_readline_with_timeout` functions with identical names and near-identical bodies now live in separate modules, along with duplicated stderr-drain closures and error-augmentation blocks.

The doctor genuinely needs the differences, binary pipes for the framing check and the tools/list round trip, so it can't just call the installer's probe. But the process-setup and error-augmentation half is copy-paste, and that is the half where a subtle fix, a Windows pipe quirk, or a timeout change will land in one module and rot in the other, which is a real hazard when both exist to diagnose the same connection. A shared helper that owns process setup, the stderr drain, and the ProbeError augmentation, with the caller supplying its own wire handling, would keep the two honest.

src/continuum/mcp/install.py
```diff
@@ -0,0 +1,382 @@
...
+    """
+    profile = profile or MCP_PROFILES["claude-code"]
+    settings = _load_config(config_path)
+    servers = _mcp_servers_for(settings, scope=scope, project_dir=project_dir, profile=profile)
+    existing = servers.get(SERVER_KEY)
+    if existing == dict(registration):
+        return "present"
+    servers[SERVER_KEY] = dict(registration)
+    config_path.parent.mkdir(parents=True, exist_ok=True)
+    config_path.write_text(json.dumps(settings, indent=2) + "\n", encoding="utf-8")
```
Cyrax321, line 354: Non-blocking, and I checked: this matches the convention already used by `clienthooks.py` (clienthooks.py:431, :536 write the same way), so this PR is not diverging from the pattern it claims to follow. Flagging it because of what the file is rather than because of a bug.

`install_server` and `remove_server` rewrite the operator's entire host config, which for the local and user scopes is `~/.claude.json`, a file that can hold every project and every MCP server the operator has registered. A plain `write_text` is not atomic, so an interrupted write, disk-full, or crash mid-write leaves it truncated. `_load_config` refusing to edit a file that is not valid JSON is good defensive behaviour, but it only helps on the *next* run; the damage is already done.

Writing to a temp file in the same directory and `os.replace`-ing into place makes the write crash-safe, and it would be worth doing in `clienthooks` at the same time since both write the same kind of file. Low priority, but the failure mode is permanent and silent.

Cyrax321 (dismissed):
Approving. This is a well-built diagnostic and the test design is the strongest part: the #697 state is reproduced in a genuinely fresh interpreter via a `sitecustomize` import blocker rather than mocked, the #699 off-PATH case is exercised for real, and the handshake probe drives actual JSON-RPC over stdio against the command a host would spawn. The CRLF framing check reading binary pipes because text mode would hide the exact defect it exists to catch is the kind of detail that shows the code was written against the failure, not against a guess. I ran the three suites: 26 passed, 1 skipped (the Windows-only CRLF test).

Two notes below, both non-blocking. The first is a maintainability concern about the two handshake implementations; the second is about the file being written.

Cyrax321 (changes requested):
Author context: Gouthamkrishnak18, external contributor. Follow-up to the prior reviews. The doctor diagnostic design is still sound, but this is in dirty merge state stacked on #844 with doc-count conflicts, and it overlaps #1013 on the same #835 surface. Rebase once #844 lands, resync the stating figures, confirm the --json parser registration and failure-summary wording fixes are present on the rebased head, and confirm the overlap check with #1013.

Issue #842, opened 2026-09-08
Adjacent Windows hardening: shell-portable adapters, run-id sanitization, sqlite:///C:/ URLs


gouthamkrishnak2003:
## Problem

Found during the installation/MCP reliability audit (issues #697/#699 follow-up work) but
**deliberately filed separately**: these are runtime Windows defects unrelated to
installation, and mixing them into the installation work would blur both reviews.

1. **`FilesystemSandboxAdapter.run_shell`** (`src/continuum/adapters/filesystem.py:42-49`)
   runs caller-supplied command strings with `shell=True` — `cmd.exe /c` on Windows,
   `/bin/sh -c` on POSIX. Any command is silently shell-family-specific; tests only ever
   pass `echo`, which happens to work on both.
2. **Reconciler and authority probes** (`src/continuum/reconcilers.py:146-153, 282-289`)
   run operator-configured strings from `.continuum/reconcilers.json` with `shell=True`;
   a config written on one platform fails (fail-closed, run refuses to resume) on another.
3. **`FileLedgerBackend` run-id sanitization** (`src/continuum/recovery/ledger.py:166-167`)
   replaces only `/` and `\`; Windows-reserved characters (`: * ? < > |`) and reserved
   device names (`CON`, `NUL`, …) produce unopenable filenames. Run ids are caller-supplied.
4. **`sqlite:///C:/path` URLs** (`src/continuum/storage/sqlite.py:69-84`) parse to
   `/C:/path`, which Windows `sqlite3.connect` rejects; the three-slash absolute form has no
   drive-letter handling.

## Root cause (evidence)

Each cited location; all verified by code inspection in the audit accompanying the
installation root-cause work (this issue exists so those findings are not lost).

## Proposed solution

Per item, smallest robust fix:

1. Document the shell-family contract on `run_shell` and/or accept argv-list commands
   (string form kept for compatibility, with the portability caveat in the docstring).
2. Same for probe specs; consider documenting per-platform probe config.
3. Extend sanitization to Windows-reserved characters and device names (platform-agnostic
   replacement is fine — POSIX loses nothing).
4. Detect a drive letter after the third slash and preserve it.

## Scope

The four fixes + targeted tests.

## Out of scope

- Anything in #834 through #841; broader adapter redesign.

## Acceptance criteria

- A run id containing `:?<>*|` round-trips through the file ledger on Windows.
- `sqlite:///C:/path/db.sqlite` opens on Windows.
[...]

PR #878, opened 2026-09-10
fix(windows): shell-portable run_shell, reserved-char run ids, drive-…

abyyxhek:
…letter sqlite URLs (#842)

Four runtime Windows defects found during the installation/MCP audit, fixed separately from that work so the reviews stay focused:

- FilesystemSandboxAdapter.run_shell accepted only command strings, which run through the platform shell (cmd.exe /c vs /bin/sh -c) and are silently shell-family-specific. The docstring now states the contract, and an argv list runs shell-free and portably; the string form is kept for compatibility with the caveat documented.
- The reconciler and authority probes run operator-configured strings with shell=True, so a registry written on one platform fails (fail-closed) on another. Documented on both probe functions and in the authority-probes guide, with the portable-form guidance.
- FileLedgerBackend replaced only / and \ in run ids, so caller-supplied ids carrying :*?"<>| produced unopenable ledger files on Windows. Sanitization now covers every Windows-forbidden character and the reserved device names; POSIX loses nothing.
- sqlite:///C:/path parsed to /C:/path, which sqlite3 rejects on Windows. A drive letter after the third slash is now recognized as the URL grammar's slash and dropped; POSIX absolute forms are pinned unchanged by tests.

## Types of changes

## Checklist

### DIFF
--- a/src/continuum/recovery/ledger.py
+++ b/src/continuum/recovery/ledger.py
@@ -156,6 +156,35 @@ class MemoryLedgerBackend(LedgerBackend):
         self._store[run_id] = list(entries)
 
 
+#: Every character a Windows filename forbids, separators included. POSIX
+#: forbids none of the extras, so replacing them all on every platform costs
+#: nothing and keeps a caller-supplied run id usable as a filename on both
+#: families (issue #842: only "/" and "\" were replaced, so run ids carrying
+#: ":*?<>"|" produced unopenable ledger files on Windows).
+_UNSAFE_FILENAME_CHARS = '/\\:*?"<>|'
+
+#: Windows reserves these device names as filenames with any extension, so a
+#: sanitized id whose stem is one of them still cannot open there.
+_RESERVED_DEVICE_NAMES = frozenset(
+    {"AUX", "CON", "NUL", "PRN"}
+    | {f"COM{i}" for i in range(1, 10)}
+    | {f"LPT{i}" for i in range(1, 10)}
+)
+
+
+def _sanitize_run_id(run_id: str) -> str:
+    """Turn a caller-supplied run id into a safe filename component.
+
+    The result keeps every character that is legal on both platform families
+    and substitutes an underscore for the rest, so the same run id maps to
+    the same file everywhere.
+    """
+    safe = "".join("_" if char in _UNSAFE_FILENAME_CHARS else char for char in run_id)
+    if safe.split(".", 1)[0].upper() in _RESERVED_DEVICE_NAMES:
+        safe = f"_{safe}"
+    return safe
+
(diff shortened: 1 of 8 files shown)


Cyrax321 (approved):

This makes run_shell portable, sanitizes reserved run id characters, and handles drive letter paths across eight files. The Windows reserved char fix is necessary even though sanitization can map distinct special char ids to one file, a low risk edge the previous slash only code shared in narrower form. A follow up could reject or disambiguate colliding ids in require_usable_run_id. CI is green including Windows jobs.

src/continuum/recovery/ledger.py
```diff
@@ -156,15 +156,43 @@ def replace(self, run_id: str, entries: Sequence[RecoveryLedgerEntry]) -> None:
...
+    """Turn a caller-supplied run id into a safe filename component.
+
+    The result keeps every character that is legal on both platform families
+    and substitutes an underscore for the rest, so the same run id maps to
+    the same file everywhere.
+    """
+    safe = "".join("_" if char in _UNSAFE_FILENAME_CHARS else char for char in run_id)
+    if safe.split(".", 1)[0].upper() in _RESERVED_DEVICE_NAMES:
+        safe = f"_{safe}"
+    return safe
```
Cyrax321, line 185: Valid fix for unopenable files, but sanitization can now map distinct ids to one file. I verified `_` substitution merges `a:b`, `a?b`, and `a_b` to the same stem, so two runs with those ids would interleave into one `ledger-<id>.jsonl` and break `verify`. The old code had the same class for `/` and `\` only, this widens it. `require` only rejects blank ids, so nothing else stops it.

Please keep distinct runs on distinct files, either by appending a short hash of the original id when sanitization changes it, or by rejecting ids that need sanitization. A regression test with `a:b` versus `a_b` asserting separate entries and clean `verify` would pin it. For example:

```python
backend = FileLedgerBackend(str(tmp_path))
RecoveryLedger(backend).append_decision("a:b", _contract(0))
RecoveryLedger(backend).append_decision("a_b", _contract(0))
assert len(list(tmp_path.glob("ledger-*.jsonl"))) == 2
```

Cyrax321 (commented):
Hi @abyyxhek, thank you for the contribution.

This fixes four Windows defects from the install audit in one focused diff: argv lists for shell-free execution, probe shell caveats documented fail closed, run id sanitization for reserved names, and drive letter URLs. I read the full diff and the three test files. CI is fully green including Windows, and the tests pin the intended behavior well. Good separation from the install track.

One correctness item inline on sanitized id collisions. It is the only thing I would like changed before approval. The rest is ready as is.

Issue #839, opened 2026-09-08
Smoke-test realism and the Windows CRLF wire bug


Gouthamkrishnak18:
## Problem

Two gaps between what we advertise and what we verify:

1. `scripts/mcp_smoke.py` is advertised as proof the MCP server works ("real subprocess,
   real JSON-RPC traffic"), but it launches `[sys.executable, "-u", "-m", "continuum.mcp"]`
   with `src` prepended to `PYTHONPATH` (`mcp_smoke.py:85-91`). It deliberately bypasses the
   `continuum-mcp` console script and the PATH resolution that `docs/api/mcp.md` identifies
   as the single most common failure — so the smoke test can pass while the user-facing
   installation is broken.

2. On Windows, the MCP SDK terminates stdio frames with `\r\n` instead of `\n`
   (upstream modelcontextprotocol/python-sdk#2433; confirmed on mcp 2.2.0 by capturing raw
   bytes from our own server: every response line ends `b'\r\n'`). Claude Code tolerates it;
   strict NDJSON clients do not. Nothing in our docs or diagnostics mentions this, so a user
   with a strict client gets an inexplicable failure that only occurs on Windows.

## Root cause (evidence)

- `scripts/mcp_smoke.py:90-98` (spawn form), `:85-88` (PYTHONPATH injection).
- Raw-wire capture on Windows 11 / Python 3.14 / mcp 2.2.0: `readline()` returns
  `...}}\r\n`.
- The SDK bug is upstream; our responsibility is to test our tolerance, document it, and
  diagnose it (#835's doctor note).

## Proposed solution

1. Add an `--entry-point {console-script,module}` mode to `mcp_smoke.py`, defaulting to the
   **console script** when it is importable-installed (fall back to the module form with a
   clear notice). This makes the advertised smoke test cover the path users actually fail on.
2. Assert the raw framing in the smoke test (log whether frames end `\n` or `\r\n`) and in
   the CI handshake test (#837): our own client-side handling must tolerate both.
3. Document the CRLF defect in `docs/api/mcp.md`'s troubleshooting section with the upstream
   issue link, so Windows users with strict clients are not left guessing.

## Scope

`scripts/mcp_smoke.py`, one docs section, small test additions.

## Out of scope

- Patching or vendoring the SDK; changing transports.

## Acceptance criteria

- `python scripts/mcp_smoke.py` (default) spawns the console script when installed and
  reports which entry point it used.
- The smoke output states the observed wire framing.
[...]

PR #881, opened 2026-09-10
feat(smoke): spawn the console script by default, report the wire fra…

abyyxhek:
…ming (#839)

The smoke test was advertised as proof the MCP server works, but it launched python -m continuum.mcp with src prepended to PYTHONPATH, so it bypassed the continuum-mcp console script and the PATH resolution that docs/api/mcp.md identifies as the single most common failure. It could pass while the user-facing installation was broken.

- --entry-point {auto,console-script,module}: auto (the default) spawns the installed continuum-mcp resolved through PATH, exactly as a host does, and falls back to the module form with a clear notice when the script is not installed; the explicit modes refuse or force the fallback. The PYTHONPATH injection now applies only to the module form, so the console-script run cannot shadow the install under test.
- Read the wire through binary pipes wrapped with newline="": Popen's text mode would rewrite \r\n to \n and hide the framing. The output now states what the server actually sent, CRLF (\r\n) on Windows (confirmed live against this server, upstream SDK issue modelcontextprotocol/python-sdk#2433) and LF (\n) elsewhere, at the handshake and in the pass summary.
- Document the Windows CRLF defect in docs/api/mcp.md's troubleshooting with the upstream reference, so strict-NDJSON client failures on Windows are diagnosable, and describe the new flags in references/e2e.md.

Verified on Windows in all three modes: auto falls back with the notice, console-script drives the installed .EXE end to end, module drives the worktree source; all pass and report CRLF.

## Types of changes

## Checklist

### DIFF
--- a/docs/api/mcp.md
+++ b/docs/api/mcp.md
@@ -174,3 +174,26 @@ in use. Do not resolve it with `claude mcp remove continuum-mcp -s project`, whi
 edits the committed `.mcp.json` and unregisters the server for everyone who clones
 the repository. Leave the diagnostic in place, or drop the local entry with
 `-s local` once the environment is on `PATH`.
+
+### Windows: response frames end with `\r\n`
+
+On Windows the MCP Python SDK terminates its stdio frames with CRLF instead of
+the LF the JSON-RPC-over-stdio transport specifies (upstream
+[modelcontextprotocol/python-sdk#2433](https://github.com/modelcontextprotocol/python-sdk/issues/2433);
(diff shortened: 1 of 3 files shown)


Cyrax321 (approved):

This spawns the console script by default and reports wire framing. I verified the spawn default and framing observation against the script, and both gitar nits read as non blocking polish. An EOF truncated final frame without terminator would label LF, and the stdout wrapper close could be symmetric, but neither affects smoke verdicts and CI is green.

Issue #215, opened 2026-08-22
Publish continuum-agent to PyPI via Trusted Publishing


Cyrax321:
### Affected component

Other / not sure

### What problem does this solve?

README states the package is not published to PyPI; installation requires cloning. Universal adoption by agents and their operators requires `pip install continuum-agent` (and `pip install continuum-agent[mcp]` for the server extras). This is table stakes and currently blocks every distribution channel including the hook installers shipped in #210, which assume a resolvable `continuum` executable.

### Proposed solution

Publish to PyPI with Trusted Publishing (OIDC, no long-lived token): wire a release workflow (the repo already has `release.yml` building releases), claim the `continuum-agent` project name, verify `pip install continuum-agent[mcp]` gives a working `continuum` and `continuum-mcp` entry point in a clean venv, then update README installation instructions.

### Alternatives considered

Staying clone-only: caps adoption at developers who read the repo; incompatible with the goal of being a default plugin for agent harnesses.

### Is this a breaking change?

No

### Would you like to work on this?

Yes, I have a plan

Cyrax321:
Verified tag v0.1.0 at 205f3bc on origin/main. GitHub Release now published draft false with 2 assets continuum_agent-0.1.0-py3-none-any.whl and continuum_agent-0.1.0.tar.gz, created 2026-08-30T06:44:54Z. Workflows Release to PyPI 33297611635 success and Publish Docker 33297611601 success. PyPI shows 0.1.0 live. Closing as done. Fixes #215

PR #882, opened 2026-09-10
fix(packaging): reconcile the declared version with the v0.1.2 tag (#…

abyyxhek:
…838)

The v0.1.2 tag and its GitHub Release shipped while pyproject.toml and src/continuum/__init__.py still declared 0.1.0, so every artifact built from the tag self-reported as 0.1.0, and PyPI (which only ever received 0.1.0) went stale against the newest release. A user installing latest did not get the newest tagged code.

- Bump pyproject.toml and __version__ to 0.1.2, matching the newest tag. Publishing the reconciling 0.1.2 build to PyPI remains the maintainer step (PUBLISH_PYPI gate, #215).
- Add a CHANGELOG [0.1.2] section recording the reconciliation and what the tag actually carried.
- Release checklist: the cut step now compares the version being cut against the newest existing tag (an equal tag means the bump was skipped, which is exactly how this drift happened), and the post-tag step verifies pyproject == tag == PyPI once the publish lands.
- Dockerfile: install with the [mcp] extra so the published image can run continuum-mcp instead of shipping an entry point that dies on the missing SDK.

## Types of changes

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -504,6 +504,21 @@ All notable changes to this project are documented here. The format follows
   callable), and the existing answers for genuinely malformed JSON are
   unchanged.
 
+## [0.1.2] - 2026-08-31
+
+### Fixed
+
+- **Version reconciliation after `v0.1.2` was tagged without the version bump
+  (#838).** The `v0.1.2` tag and its GitHub Release (67 commits over
+  `v0.1.0`: the `tree --limit` fix #544, the rewind carry-forward passthrough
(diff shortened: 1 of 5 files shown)


Cyrax321 (approved):

This reconciles the declared version with the v0.1.2 tag. I verified the tag exists while pyproject and __init__ still declare 0.1.0 on base, and the patch bumps both plus changelog and checklist guards plus the Dockerfile mcp extra. CI is green.

PR #895, opened 2026-09-10
bench: require and add BENCH markers in README for idempotent regeneration

### LINKED ISSUE #776: README lacks BENCH markers, so every bench run appends a duplicate table
## Context

`benchmarks/run.py` regenerates the README bench table by replacing the text between `<!-- BENCH:START -->` and `<!-- BENCH:END -->` markers, but `README.md` contains neither marker. The fallback path then appends a second bench table to the README on every run instead of replacing anything.

```bash
grep -c "BENCH:START" README.md
# 0: the markers do not exist, so every run appends
```

## Why this matters

Each bench run silently grows the README with a duplicate table, and the duplication compounds.

## Expected outcome

The markers exist in README.md around the bench table (or the runner requires them instead of appending), so repeated runs keep one table. Verify by running the regeneration twice and confirming a single table.

## Step by step for the fix

1. Read the regeneration function in `benchmarks/run.py`.
2. Implement the chosen direction; run it twice to prove idempotence.
3. Run markdownlint on README.md and `ruff check` if `run.py` changes.

## Files to touch

- `README.md` and/or `benchmarks/run.py`

Small and verifiable, ideal for a first contribution.

soorajsuru2004-sudo:
## Description

Adds `<!-- BENCH:START -->` and `<!-- BENCH:END -->` markers around the horizon benchmark table in `README.md` under `## Empirical Verification`. Updates `_regenerate_readme_bench` in `benchmarks/run.py` to require both markers and return early without mutating the file if either marker is missing, preventing runs from appending duplicate tables to the bottom of the README.

## Related issues

Fixes #776

## Types of changes

- Type: `bugfix`

## Breaking changes

No

## How has this been tested?

1. Idempotent benchmark regeneration test:
Ran `_regenerate_readme_bench(Path("README.md"), table)` twice in Python. Verified `text.count("BENCH:START") == 1` and `text.count("BENCH:END") == 1`.
2. Markdown linting:
`npx -y markdownlint-cli2 README.md` passed with 0 issues.
3. Python linting and formatting:
`ruff check benchmarks/run.py` passed with 0 issues.
`ruff format --check benchmarks/run.py` passed.
4. Em dash audit:
Verified zero U+2014 characters in `README.md` and `benchmarks/run.py`.

## Checklist

## Additional context

Previously, `_regenerate_readme_bench` had an `else` branch that appended the generated benchmark markdown to the end of `README.md` whenever the markers were not found. Because `README.md` lacked the markers, repeated benchmark runs repeatedly appended duplicate tables.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,24 @@ CONTINUUM is verified against real LLM agents, live protocol boundaries, and har
 - **Scale**: roughly 2,163 tests collected (~2,030 passing; the rest skip without optional services) on Python 3.11, 3.12, and 3.13 (unit, `hypothesis` property-based, concurrency, adversarial). CONTINUUM-Bench runs five crash scenarios plus a dedicated argument-drift scenario, measuring 0 duplicate work and 0 duplicate side effects for CONTINUUM against full duplication for naive replay; a separate 12-scenario recovery-correctness suite (`continuum.benchmark.phase6`) encodes the crash points from the durable-execution survey as executable assertions.
 - **Adversarial audit**: the full MCP surface was audited over the live protocol; three defects were found and fixed. Method and reproduction steps in [test.md](test.md).
 
+<!-- BENCH:START -->
+### Horizon-scale benchmark (real runs, no invented numbers)
+
+Generated: 2026-09-10T23:17:54.069493  Horizon scenarios: 5  Passed: 3  Failed: 2
+
+Accuracy: 0.6  Unnecessary escalation: 0.2  Repair precision: 0.8  Duplicate side effects: 0  Duplicate work: 0.0  Compression: 0.139
+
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):
Thank you for contributing to CONTINUUM and for fixing #776.

This adds the BENCH markers around the horizon table and changes _regenerate_readme_bench to require them instead of appending duplicates. I verified the marker placement and single pair count, the table shape against the builder, the return early else branch, plus ruff, format, markdownlint, and em dash checks. Existing horizon regeneration coverage now applies since markers exist, and CI is green.

No blocking issues found. One advisory nit outside the code: the title uses bench: which is not in the repo Conventional Commits allowed types. Consider fix(bench): or fix: on a follow up edit. Approving as is since title check passes.

Thanks again for stopping the README growth.

PR #921, opened 2026-09-11
docs: document replay similarity API

### LINKED ISSUE #911: Follow-up to #607: docstrings for replay_similarity.py
### Affected component

`src/continuum/replay_similarity.py` still has 3 undocumented public names (`SimilarityKind`, `SimilarityConfig`, `jaccard`). Same reason as #607. `jaccard` pins boundary semantics (both empty scores 1.0, one empty scores 0.0) that nothing states; `SimilarityConfig` documents its fields but not what the config as a whole selects.

How to reproduce:

```bash
python - <<EOF
import ast
from pathlib import Path
tree = ast.parse(Path("src/continuum/replay_similarity.py").read_text(encoding="utf-8"))
print([n.name for n in ast.walk(tree)
       if isinstance(n, (ast.FunctionDef, ast.ClassDef))
       and not n.name.startswith("_")
       and ast.get_docstring(n) is None])
EOF
```

Expected: `['SimilarityKind', 'SimilarityConfig', 'jaccard']`.

Verified on current `main`: exactly those three (`similarity`, `similarity_backend`, and `_args_text` are already documented). No open or closed issue or PR covers this file.

### Proposed solution

Add docstrings in the #607 style: a class docstring for each of `SimilarityKind` and `SimilarityConfig` stating what the strategy choice means for replay-vs-fork, and a `jaccard` docstring stating the empty-set boundaries. Docstrings only, no runtime change. Then run `ruff check`, `ruff format --check`, and `mypy src/continuum`.

### Alternatives considered

Moving the boundary rules into the `similarity` docstring instead. Rejected: the rules live in `jaccard`, so that is where the next reader looks.

No

Yes, with guidance

Adhi1-2:
## Summary
- document the replay similarity strategy enum and configuration object
- document Jaccard boundary behavior for empty token sets
- keep the change documentation-only

Closes #911

## Validation
- `./.venv/bin/pytest -q tests/test_replay_similarity.py` — 8 passed
- public-name AST audit reports no undocumented public classes or functions
- `./.venv/bin/ruff check src/continuum/replay_similarity.py`
- `./.venv/bin/ruff format --check src/continuum/replay_similarity.py`
- `./.venv/bin/mypy src/continuum/replay_similarity.py`
- `git diff --check`

## Summary by CodeRabbit

- **Documentation**
  - Added documentation describing similarity comparison strategies and score thresholds.
  - Clarified that comparing two empty sets with Jaccard similarity returns an identical result.

### DIFF
--- a/src/continuum/replay_similarity.py
+++ b/src/continuum/replay_similarity.py
@@ -37,6 +37,8 @@ __all__ = [
 
 
 class SimilarityKind(StrEnum):
+    """Comparison strategies available to the replay guard."""
+
     EXACT = "exact"
     FUZZY = "fuzzy"
     EMBEDDING = "embedding"
@@ -44,6 +46,8 @@ class SimilarityKind(StrEnum):
 
 @dataclass(frozen=True)
 class SimilarityConfig:
+    """Configure the similarity strategy and replay/fork score thresholds."""
+
     kind: SimilarityKind = SimilarityKind.EXACT
     """Which comparison strategy to use."""
     replay_threshold: float = 0.90
@@ -60,6 +64,8 @@ def token_set(text: str) -> frozenset[str]:
 
 
 def jaccard(a: frozenset[str], b: frozenset[str]) -> float:
+    """Return token-set similarity, treating two empty sets as identical."""
+
     if not a and not b:
         return 1.0
     if not a or not b:


src/continuum/replay_similarity.py
```diff
@@ -60,6 +64,8 @@ def token_set(text: str) -> frozenset[str]:
 
 
 def jaccard(a: frozenset[str], b: frozenset[str]) -> float:
+    """Return token-set similarity, treating two empty sets as identical."""
```
Cyrax321: The both empty case is now stated, but the one empty case is not, and #911 explicitly requires it since the implementation returns 0.0 there. Please cover both boundaries.

```suggestion
    """Return token-set similarity.

    Two empty sets score ``1.0``. If only one set is empty, the score is
    ``0.0``.
    """
```

src/continuum/replay_similarity.py
```diff
@@ -37,13 +37,17 @@
 
 
 class SimilarityKind(StrEnum):
+    """Comparison strategies available to the replay guard."""
```
Cyrax321: #911 asks for what the strategy choice means as a whole, not just the names. `similarity()` branches on `config.kind` and `classify_call` grades the resulting score against the two thresholds, so please state that chain. Suggested:

```diff
-    """Comparison strategies available to the replay guard."""
+    """Strategies used to compute replay similarity scores.
+
+    The selected strategy supplies the score used for replay and fork
+    decisions.
+    """
```

and for the config:

```diff
-    """Configure the similarity strategy and replay/fork score thresholds."""
+    """Configure similarity scoring and replay/fork decisions.
+
+    ``kind`` selects the comparison strategy. The replay and fork thresholds
+    classify the score produced by that strategy.
+    """
```

Cyrax321 (changes requested):
Hi @Adhi1-2, thank you for the contribution.

This documents the exact three names #911 lists, with no logic change. I verified the implementation paths, ran the test file with 8 passed, and confirmed ruff and diff checks clean. CI is green everywhere except the Windows trio still running.

Requesting changes for two small completeness items inline. Both come straight from #911 itself: the one empty Jaccard boundary and the kind selects strategy whose score drives replay and fork relationship. Each is a one-line doc tweak with the wording ready. Once those are in, this is ready to approve.

Adhi1-2:
Both changes applied in 59e7fd1's follow-up, wording as suggested:

1. `SimilarityKind` / `SimilarityConfig` now state the chain: `kind` selects the comparison strategy, and the replay/fork thresholds classify the score that strategy produces.
2. `jaccard` now covers both boundaries: two empty sets score `1.0`, one empty set scores `0.0`.

ruff, format, and the similarity/replay tests remain green.

Cyrax321 (approved):
Thank you for contributing to CONTINUUM. The replay similarity API docs are precise, especially the edge cases that callers actually get wrong.

Verified against src/continuum/replay_similarity.py: SimilarityKind as the strategy selector feeding replay and fork decisions, SimilarityConfig with kind plus replay and fork thresholds classifying the strategy score, and jaccard documenting the two empty sets score 1.0 while a single empty set scores 0.0. Each matches the implementation.

Approving with one ask: anya-review is red here, please address its findings before merge. Thanks for documenting the #911 follow up surface.

PR #922, opened 2026-09-11
docs: document MCP authorization flags

### LINKED ISSUE #910: Follow-up to #607: docstrings for mcp/authz.py
### Affected component

MCP server

`src/continuum/mcp/authz.py` still has 2 undocumented public names (`disabled` on `AuthPolicy`, `denies_everything` on `AuthorizationPolicy`). Same reason as #607, plus this file is the authorization seam: its own module docstring calls it a security boundary when authentication is on, so the two boolean flags that answer "is auth off?" and "is everyone refused?" must say so where callers read them.

How to reproduce:

```bash
python - <<EOF
import ast
from pathlib import Path
tree = ast.parse(Path("src/continuum/mcp/authz.py").read_text(encoding="utf-8"))
print([n.name for n in ast.walk(tree)
       if isinstance(n, (ast.FunctionDef, ast.ClassDef))
       and not n.name.startswith("_")
       and ast.get_docstring(n) is None])
EOF
```

Expected: `['disabled', 'denies_everything']`.

Verified on current `main`: exactly those two (neighboring `verify`, `permits`, and `require` are already documented, which sets the style to match). No open or closed issue or PR covers `mcp/authz.py` docstrings.

### Proposed solution

Add one-line docstrings in the #607 style: `disabled` is True only when no secret of any kind is configured (an explicit empty secret refuses rather than disabling); `denies_everything` is True when the allow-list is empty. Docstrings only, no runtime change. Then run `ruff check`, `ruff format --check`, and `mypy src/continuum`.

### Alternatives considered

None serious; the fix is two sentences.

No

Yes, with guidance

Adhi1-2:
## Summary
- document when `AuthPolicy.disabled` disables authentication
- document when `AuthorizationPolicy.denies_everything` refuses all mutating callers
- keep the security-boundary change documentation-only

Closes #910

## Validation
- `./.venv/bin/pytest -q tests/test_mcp_authz.py` — 65 passed
- public-name AST audit reports no undocumented public classes or functions
- `./.venv/bin/ruff check src/continuum/mcp/authz.py`
- `./.venv/bin/ruff format --check src/continuum/mcp/authz.py`
- `./.venv/bin/mypy src/continuum/mcp/authz.py`
- `git diff --check`

### DIFF
--- a/src/continuum/mcp/authz.py
+++ b/src/continuum/mcp/authz.py
@@ -131,6 +131,8 @@ class AuthPolicy:
 
     @property
     def disabled(self) -> bool:
+        """Whether authentication is disabled because no secret is configured."""
+
         # Only an absent secret disables authentication. An explicit empty
         # secret is a misconfiguration and must refuse, not open the door.
         return self.expected is None and not self.tokens
@@ -345,6 +347,8 @@ class AuthorizationPolicy:
 
     @property
     def denies_everything(self) -> bool:
+        """Whether the empty allow-list refuses every mutating caller."""
+
         return not self.allowed
 
     def permits(self, caller: str | None) -> bool:


src/continuum/mcp/authz.py
```diff
@@ -131,6 +131,8 @@ def __repr__(self) -> str:  # pragma: no cover - debugging aid
 
     @property
     def disabled(self) -> bool:
+        """Whether authentication is disabled because no secret is configured."""
```
Cyrax321, line 134: Optional, not blocking: #910 asked for the empty secret nuance in this docstring. The class docstring and the comment below already state it, so take this only if you want the one-liner to stand alone.

```suggestion
        """True only when no secret of any kind is configured; an explicit empty secret still refuses."""
```

Cyrax321 (approved):
Hi @Adhi1-2, thank you for the contribution.

This documents the exact two flags #910 lists, with no logic change. I checked both wordings against their implementations: `disabled` matches the absent secret check and the empty refuses behavior stated in the class docstring, and `denies_everything` matches the mutating tool gates. Checks are clean and CI is green everywhere except the Windows trio still running.

No blocking issues found. One optional wording suggestion inline, take it or leave it. Approving.

PR #960, opened 2026-09-12
fix: replace unsafe tempfile.mktemp with mkstemp

### LINKED ISSUE #950: Replace unsafe tempfile.mktemp in examples/plan_milestones.py and tests/test_trajectory_report.py
### Before reporting

### Affected component

Other / not sure

### Severity

### Regression check

### System information

OS: macOS
Python: 3.13 (repo checkout)
Commit: origin/main `613037d`

```bash
rg -n "mktemp" examples/ tests/ src/continuum benchmarks/
```

### Current behaviour

Two call sites use `tempfile.mktemp`, which is deprecated and unsafe (the name is reserved without creating the file, so another process can claim it between generation and use):

- `examples/plan_milestones.py:14`: `db = tempfile.mktemp(suffix=".db")`, then handed to `SQLiteStorage`
- `tests/test_trajectory_report.py:270`: `tmp = tempfile.mktemp(suffix=".sqlite")`, then handed to `SQLiteStorage`

No other `mktemp` uses exist in `src/`, `tests/`, `examples/`, or `benchmarks/`.

### Expected behaviour

Replace both with a race-free construction (`tempfile.mkstemp` with an `os.close` on the fd, `tempfile.NamedTemporaryFile`, or a `TemporaryDirectory` join), keeping behavior identical: a fresh throwaway SQLite path. No other changes.

### Logs or CI run

Not applicable (static finding, confirmed by the `rg` above returning exactly these two hits).

wized2:
## Summary
\`tempfile.mktemp\` only reserved a filename without creating it, leaving a TOCTOU race.

## Changes
- \`examples/plan_milestones.py\` and \`tests/test_trajectory_report.py\`: use \`tempfile.mkstemp\` + \`os.close(fd)\` for throwaway SQLite paths

Closes #950

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved temporary database setup in examples and command-line tests by using safer temporary file creation.
  * Ensured temporary file descriptors are closed before SQLite databases are opened.
  * Reduced the risk of file conflicts and unsafe access during temporary database setup.

### DIFF
--- a/tests/test_trajectory_report.py
+++ b/tests/test_trajectory_report.py
@@ -2,9 +2,8 @@
 
 from __future__ import annotations
 
-import os
-
 import json
+import os
 
 from continuum.analysis.trajectory_report import (
     build_trajectory_report,
--- a/tests/test_trajectory_report.py
+++ b/tests/test_trajectory_report.py
@@ -1,3 +1,4 @@
+import os
 """Sleep-time trajectory reports (issue #393)."""
 
 from __future__ import annotations
@@ -267,7 +268,8 @@ def test_digest_auditable_and_briefing_consumption() -> None:
 
         from continuum.cli.main import main as cli_main
 
-        tmp = tempfile.mktemp(suffix=".sqlite")
+        fd, tmp = tempfile.mkstemp(suffix=".sqlite")
+        os.close(fd)
         file_storage = SQLiteStorage(tmp)
         try:
             file_storage.create_run(Run(run_id=run_id, goal="g"))
--- a/tests/test_trajectory_report.py
+++ b/tests/test_trajectory_report.py
@@ -1,7 +1,8 @@
-import os
 """Sleep-time trajectory reports (issue #393)."""
 
 from __future__ import annotations
+
+import os
 
 import json
 
--- a/examples/plan_milestones.py
+++ b/examples/plan_milestones.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 import tempfile
 
 from continuum.events import EventType
@@ -11,7 +12,8 @@ from continuum.storage import SQLiteStorage
 
 
 def main() -> None:
-    db = tempfile.mktemp(suffix=".db")
+    fd, db = tempfile.mkstemp(suffix=".db")
+    os.close(fd)
     storage = SQLiteStorage(db)
     run_id = "plan-run-1"
     storage.create_run(Run(run_id=run_id, goal="5 milestones"))


tests/test_trajectory_report.py
```diff
@@ -1,3 +1,4 @@
+import os
 """Sleep-time trajectory reports (issue #393)."""
```
Cyrax321, line 1: Blocking: this import breaks the module. `from __future__` must be the first statement after the docstring, so line 1 makes the whole file a `SyntaxError` at import time. I confirmed it with `compile()`, and it matches CI: every test job fails collecting this module, plus the lint job.

Please move it into the stdlib group, so the header reads:
```python
"""Sleep-time trajectory reports (issue #393)."""

from __future__ import annotations

import json
import os
```
i.e. delete line 1 and add `import os` after `import json`. The example file already has the right order, and `tempfile` is imported at function level in this file so nothing else needs to move. Once this lands, CI should go green with no further changes.

Cyrax321 (changes requested):
Thank you for contributing to Continuum, @wized2. This targets #950 exactly: both `mktemp` sites move to `mkstemp` plus `os.close`, which is the right race-free primitive for throwaway SQLite paths, and I confirmed no other Python call sites remain. The example file is correct as written.

One blocking issue, already inline: the `import os` in the test file sits above the docstring and `__future__`, which is a `SyntaxError` that fails every test job and the lint job. It is a two-line move with the exact replacement in the comment. Everything else in the PR is good.

Once that is fixed and CI is green, this is ready to merge. Thank you again for picking this up.

wized2:
Addressed the review note: `import os` is now placed after `from __future__ import annotations` in `tests/test_trajectory_report.py` so the future import remains valid.

Cyrax321 (approved):
Thank you for contributing to Continuum, @wized2, and welcome. Your fix is the real substance of this PR: both `mktemp` sites now use the race-free `mkstemp` plus `os.close`, exactly what #950 asked for, and you turned the first round of feedback around quickly. Genuinely appreciated.

What I did on top and why: I pushed one small maintainer touch-up to your branch, merging the split stdlib import group (`import json` / `import os`) into sorted order. Ruff isort requires a single sorted block, and that was the only thing keeping the Lint job red. No logic changed; the fix remains entirely yours.

I verified the result directly: ruff check and format pass, the trajectory suite passes 8 of 8, and CI is now green including Lint and Type-check. No blocking issues remain.

This is ready to merge. Thanks again for picking this up.

PR #875, opened 2026-09-10
docs(zh): sync Quick Start, counts, and bench table to English

Cyrax321:
## Description

Chinese README sync, one verified unit per commit: notebook zero-setup rows plus sentence (#635), scale and module-map test counts (#780), and seeded bench markers with a real generated horizon table (#776). Badge links byte-identical, figures mirror English wording, table numbers come from a live deterministic run.

## Related issues

Refs #635, #780, #776

## Types of changes

- Type: docs

## Breaking changes

No. Translation only.

## How has this been tested?

- Each unit committed separately after markdownlint clean.
- Bench numbers from a live horizon run twice in a row: identical (deterministic), single table (idempotent replace), verify clean.
- Idempotence of the replace path proven on English README (1 marker after two regens).
- env -u NO_COLOR TERM=xterm .venv/bin/pytest: 2,036 passed, 23 skipped.
- git diff --check: passed.

## Checklist

Translation only. CI has not yet run on this PR.

## Additional context

The committed horizon numbers show 3 passed and 2 failed, matching the pre-existing judge mismatches also visible on the English side; no numbers invented. Japanese, Korean, and Portuguese-BR need the identical treatment next.

## Summary by CodeRabbit

- **Documentation**
  - Updated empirical validation metrics to reflect approximately 2,163 tests, with approximately 2,030 passing.
  - Updated module mapping totals to 124 modules and 161 test files.

PR #979, opened 2026-09-13
refactor(extractor): unify StateExtractor contracts

### LINKED ISSUE #783: feat(plugins): unify StateExtractor contracts and dispatch extraction through the registry
## Affected component

CONTINUUM currently defines two incompatible `StateExtractor` protocols. `src/continuum/state/extractor.py` accepts an `ExtractionContext` and supports deterministic, composite, and optional LLM enrichment. `src/continuum/plugins/seams.py` defines a separate `(trajectory, environment)` signature. Neither protocol is a load-bearing registry consumer in the normal recovery path, which continues to call projection directly.

This duplication lets plugin authors satisfy one documented seam while being unusable by the extraction implementation. It also leaves integrations without a supported way to provide a framework-specific state projection while preserving CONTINUUM's deterministic baseline and provenance rules.

## Proposed solution

Define one canonical `StateExtractor` contract based on the existing explicit `ExtractionContext`, then re-export it from the plugin package rather than maintaining a divergent protocol. Add an explicit extraction-resolution path for recovery and adapter integration that defaults to `DeterministicExtractor` and may use an explicitly registered extractor or composition when configured.

The deterministic event fold remains authoritative. Non-deterministic enrichment must remain opt-in, additive only, provenance-stamped, and unable to replace recorded facts or lower recovery caution.

## Acceptance criteria

- There is one public `StateExtractor` protocol and one `ExtractionContext` contract; the duplicate plugin protocol is removed or becomes a compatible re-export.
- Registry resolution is explicit and preserves `DeterministicExtractor` behavior when no extractor is configured.
- Recovery or adapter integration can invoke a configured extractor through the canonical contract without bypassing projection invariants.
- Extractor output retains the existing provenance and `REQUIRES_REVIEW` rules for non-deterministic additions.
- Misconfigured, duplicate-named, failing, or incompatible extractors fail safely to the deterministic state with an explicit diagnostic rather than silently changing recovery state.
- Tests cover registry resolution, deterministic fallback, composite ordering, LLM enrichment constraints, malformed extractors, and compatibility of public imports.
- Documentation explains the extraction trust boundary and the difference between deterministic projection and optional enrichment.

## Alternatives considered

Keep the two protocols and document their differences. Rejected because a plugin can appear valid while remaining unusable. Letting an extractor replace the deterministic fold is also rejected because it would allow unrecorded or model-authored state to overwrite durable evidence.

## Is this a breaking change?

No

Yes, with guidance

Cyrax321:
## Description

Two incompatible `StateExtractor` protocols existed in `state/extractor.py` and `plugins/seams.py`. This unifies them behind a single registry-dispatched contract and migrates call sites.

## Related issues

Fixes #783

## Types of changes

- Type: `refactor`

## Breaking changes

No

## How has this been tested?

```console
pytest tests/test_extractor.py -q
mypy src/continuum/state/extractor.py
```

## Checklist


Cyrax321 (commented):
Verified this one end to end. The real change removes the duplicate StateExtractor Protocol from plugins/seams.py and re-exports the canonical one from state/extractor.py, so the registry and the core share a single ExtractionContext contract. I confirmed the canonical protocol is runtime_checkable (isinstance checks keep working), no other implementers of the old trajectory/environmentsignature remain, and the updated dummy plus seam test exercise the re-export path. tests/test_plugins.py and tests/test_extractor.py pass on this head (31 passed) and all CI checks are green except the pending CodeRabbit pass. No blocking correctness issues found. Two process notes, neither blocking the code: this branch stacks the zh-sync commits, so please rebase once #875 lands and the diff will shrink to seams, tests, and CHANGELOG. The one gitar-bot thread on the zh count arithmetic belongs to that stacked docs change, not to this refactor. (Posted as a comment review because GitHub does not allow approvals on my own PRs. From a code standpoint this is ready.)

Cyrax321:
Branch cleaned as requested in review: the 6 stacked zh commits are out (7 commits to 1), rebased onto latest main with one positional CHANGELOG conflict resolved by keeping both entries. Diff is now seams.py, test_plugins.py, and CHANGELOG only. Plugin and extractor tests pass on the rebased head (31 passed), diff-check clean.

anya-research (approved):
LGTM Approved!

Issue #333, opened 2026-08-24
Show absolute path in gate config error messages


Cyrax321:
## Problem
Gate error shows a relative path.

`load_gate_config` in `src/continuum/gate.py` raises:

```python
GateConfigError(f"{path} is not valid JSON ({exc})")
```

`path` is whatever the caller passed, often `.continuum/gate.json`. When you run the CLI from a subdirectory, the message does not tell you which file was read.

## Repro
```bash
mkdir -p /tmp/proj/.continuum
echo "not json" > /tmp/proj/.continuum/gate.json
cd /tmp/proj/subdir
continuum gate --config ../.continuum/gate.json --run-id x --payload-file /tmp/p.json
# shows ../.continuum/gate.json, not the resolved absolute path
```

## Fix
Include `path.resolve()` in the message. For example:

```
.continuum/gate.json (/home/user/proj/.continuum/gate.json) is not valid JSON ( ...)
```

Do the same for `load_gateway_config` in `src/continuum/gateway.py` and `load_budgets` in `src/continuum/budgets.py` for consistency.

Update tests to assert on substring if they check the message.

## Files
- `src/continuum/gate.py`
- `src/continuum/gateway.py`
- `src/continuum/budgets.py`

One line message improvement per file.

aastha-m22:
I'd like to work on this

PR #988, opened 2026-09-13
feat: notify a human when a run blocks on request_human

### LINKED ISSUE #305: Webhook-out: notify a human when a run blocks on request_human

Adhi1-2:
## What

Issue #305: a run that blocks on `request_human` is safe but stuck — it parks until a human looks, and nobody looks unless they were already polling `resume` or the dashboard. This adds the push: a webhook out when the verdict lands.

## Change

- **`src/continuum/recovery/webhooks.py`** (new): the `.continuum/webhooks.json` registry (endpoints, `events` filter, `re_notify_seconds`, `retries`, `timeout`, `dashboard_base_url`) following the same data-beside-code convention as the gateway registry — absent file means silent, malformed file is refused loudly naming the resolved path (#333 discipline). A typo'd event name is refused at load time because a filter that silently never fires is the quietest failure there is.
- **Dedup is per transition, not per re-assessment**: the verdict identity is `(run_id, mode, hash of the contract's decision content)`. `created_at`, `liveness`, `post_checkpoint_observations`, and `integrity_hash` are excluded from the hash — they change between assessments of the same blockage without changing what was decided, and hashing them would re-ring the bell on every cron tick. The dedup state lives in the event log (`NOTIFICATION_SENT` / `NOTIFICATION_FAILED` rows), so it survives restarts; both sent *and* failed deliveries consume the window, so a dead receiver is not re-hit every minute — it is a dead-letter row an operator finds by polling the log.
- **Delivery** uses the existing HMAC-signed `post_webhook` primitive from `recovery.notify`, up to `retries + 1` attempts without backoff (notification runs on the CLI's critical path). The payload mirrors the resume JSON contract plus `event` and a `dashboard_url` deep link when configured. Failure never gates the verdict: same exit code, one warning line.
- **CLI**: `resume` notifies when it ends in `request_human` (`--webhooks-config` to override the path); new storage-less `continuum notify-test [run_id]` probes every endpoint bypassing dedup by design, so the first real `request_human` is never also the first delivery attempt.
- **Docs**: new `docs/guides/webhooks.md` (registry format, dedup semantics, signature verification snippet), rows in `docs/api/cli.md` + `references/cli.md`, README roadmap/limitations updated.

## Verification

- 26 new tests in `tests/test_webhooks.py`: registry parsing + 10 malformed-config refusals, payload/signature/deep-link delivery, dedup within window / new contract / expired window, per-endpoint event filter, dead-letter on 500 with retry count, failed-delivery window suppression, `notify-test` success/failure/no-database, and a `resume` integration test proving the notification fires on a real `request_human` verdict and dedup holds across fresh CLI processes.
- The test suite caught a real bug during development: `verdict_key` originally hashed `contract.model_dump()` wholesale, and since `RecoveryContract.created_at` defaults to `utcnow()`, every assessment — including every real `resume` — produced a different key, so dedup would never have held in production. `test_verdict_key_ignores_per_assessment_bookkeeping` is the regression test.
- `ruff check`, `ruff format --check`, `mypy src/continuum` (126 files, clean), full `pytest tests/` suite green, `markdownlint-cli2` clean on all changed docs.

Fixes #305

## Summary by CodeRabbit

* **New Features**
  * Added webhook notifications for runs requiring human review.
  * Added the `notify-test` command to verify webhook delivery.
  * Added configurable webhook endpoints, event subscriptions, retries, deduplication, signing, and delivery tracking.
  * Added documentation covering webhook setup, behavior, and CLI usage.
* **Documentation**
  * Updated command references and roadmap information to reflect webhook notification support.


anya-research:
**Code Review**

This PR adds registry-driven webhook push for `request_human` blocked runs (`src/continuum/recovery/webhooks.py`), durable per-transition dedup via `NOTIFICATION_SENT/FAILED` rows, retry logic via `post_webhook`, and `notify-test` plumbing in `src/continuum/cli/main.py:1247` with docs and 26 tests. Implementation is careful (volatile-field exclusion from `verdict_key` and regression test are excellent), but breaks its own fail-open and read-only contracts in a few places that will crash `resume` or leak data in production.

**high**
`src/continuum/recovery/webhooks.py:296`
`notify_blocked` docstring promises "Never raises" but both `storage.append_event` paths at lines 297 and 304 are unguarded. Any `StorageError`/`sqlite3.OperationalError` (locked DB while an agent is mid-run) escapes `_notify_blocked_run` and prevents `resume` from returning its `request_human` verdict and exit code. Delivery outcome must stay POST-derived and persistence failures must not propagate.
```suggestion
        try:
            storage.append_event(
                run_id,
                EventType.NOTIFICATION_SENT if delivered else EventType.NOTIFICATION_FAILED,
                {"url": endpoint.url, "event": mode, "verdict": key},
            )
        except Exception as exc:
            detail = f"event-log write failed: {exc}"
            records.append(DeliveryRecord(url=endpoint.url, event=mode, status="sent" if delivered else "failed", detail=detail))
            continue
```

**high**
`src/continuum/recovery/webhooks.py:113`
`load_webhook_registry` only converts `json.JSONDecodeError` to `WebhookConfigError`. `path.read_text()` on an unreadable file, directory, or non-UTF8 bytes raises `OSError`/`UnicodeError`, which bypasses the `except WebhookConfigError` in `src/continuum/cli/main.py:1272` and crashes `resume` instead of warning-and-skipping. Same for `events` containing non-strings at line 138 which raises `TypeError` on `in`. All malformed-registry inputs must normalize to `WebhookConfigError`.
```suggestion
    except (OSError, UnicodeError, json.JSONDecodeError) as exc:
        raise WebhookConfigError(f"{location} cannot be read as valid JSON ({exc})") from exc
```

**medium**
`src/continuum/recovery/webhooks.py:234`
`_within_dedup_window` scans `storage.read_events(run_id)` which returns only the live tail after compaction. An archived `NOTIFICATION_SENT`/`FAILED` becomes invisible, so the same verdict re-fires inside `re_notify_seconds`. Every other history fold uses `read_all_events`; dedup must too or the spam prevention regresses on any compacted run.
```suggestion
    for event in storage.read_all_events(run_id):
```

**medium**
`src/continuum/recovery/webhooks.py:133`
Registry accepts any `http://` URL and `post_webhook` sends the recovery payload plus optional HMAC in cleartext. The HMAC is a body signature, not confidentiality. Non-loopback cleartext leaks `run_id`, contract and `dashboard_url` on the wire. Require `https://` except for loopback (`localhost`/`127.0.0.0/8`/`::1`).
```suggestion
        parsed = urlsplit(url) if isinstance(url, str) else None
        hostname = parsed.hostname if parsed else None
        loopback = hostname == "localhost" or hostname in ("127.0.0.1", "::1")
        if not isinstance(url, str) or parsed.scheme not in ("http", "https") or not parsed.netloc or (parsed.scheme == "http" and not loopback):
            raise WebhookConfigError(f"{location}: endpoint needs HTTPS or an http:// loopback URL")
```

**medium**
`src/continuum/cli/main.py:1478`
`cmd_resume` is documented as "Read-only unless `--repair` is given" and `references/cli.md` lists it as safe against a live DB, but a `request_human` verdict now calls `notify_blocked` which appends notification events without `--repair`. This takes a SQLite write lock while an agent may be appending, risking `SQLITE_BUSY` and violating the safety invariant operators rely on. Either gate the write behind an opt-in flag or reclassify `resume` as mutating when webhooks are configured and remove it from the read-only list.

**low**
`src/continuum/recovery/webhooks.py:56`
`EVENT_REQUIRES_REVIEW` is accepted, validated, and documented, but `notify_blocked` is only invoked with `presented_mode == RecoveryMode.REQUEST_HUMAN.value` (`src/continuum/cli/main.py:1478`). No `RecoveryMode` equals `requires_review`, so subscribing to `requires_review` alone validatedly never fires, the exact quiet failure the strict validation was meant to prevent. Remove the event from `_KNOWN_EVENTS` or wire a real delivery path.


[github run](/Cyrax321/CONTINUUM/actions/runs/34746850547)

Cyrax321 (approved):
Thank you for contributing to CONTINUUM. A blocked run that nobody polls is still stuck, so pushing the request_human verdict out via webhooks closes a real operational gap for #305. The design here is careful and I appreciate it.

Verified from the diff: registry at .continuum/webhooks.json with unknown event names refused loudly at load, HMAC-SHA256 signing when a secret is configured, dedup by (run_id, mode, contract hash) recorded as NOTIFICATION_SENT and NOTIFICATION_FAILED rows in the event log so it survives restarts, and delivery explicitly fire and forget so a down receiver never changes the verdict, exit code, or resume output beyond one warning line. Removing the No webhook out limitation from the README and marking #305 shipped in the roadmap table is accurate for this scope. tests/test_webhooks.py plus the notify-test diagnostic give operators a wiring check without a real blockage.

Approving with one question: please confirm the docs advise on file permissions for webhooks.json since endpoint secrets live there in plaintext. Not a blocker, just operator hygiene. Great work.

PR #990, opened 2026-09-13
ci(anya): switch review model to muse-spark-1.2

Cyrax321:
The anya and anya-review workflows pinned opencode/muse-spark-1.2-contributor-free, which no longer serves this repo's key. Verified opencode/muse-spark-1.2 exists in the provider model list. Both files switched together so the chat and review paths stay in sync. Yuki workflows left untouched. Note: the latest anya-review failure also shows the action's own version-fetch curl failing before any model call, so if runs still fail after this lands, that step is the next suspect.


Cyrax321:
Landed directly on main per owner request (commits 288e2cc superseded by 5246eb3 with the free 1.3 contributor model). Closing as redundant.

PR #997, opened 2026-09-13
docs(i18n): sync translated counts and suite size

Cyrax321:
Translated READMEs lagged English: counts 2,163 vs verified 2,195 collected, and recovery suite 12 vs verified 14 scenarios (ALL_SCENARIOS). Numbers only, no prose touched.


anya-research (dismissed):
Thank you for contributing to Continuum, @Cyrax321. This syncs the five translated READMEs from 2,163 to 2,195 collected, and I verified the head directly: git diff shows 15 insertions and 15 deletions, 3 per file, numbers only with no prose touched, and git diff --check is clean. English README.md already reads 2,195 at lines 102, 235 and 428, and each translation now matches at the same three spots, consistent with CHANGELOG. No Python files are touched, so AST coverage is empty and ruff has nothing to check. Markdown lint passes, while the CI matrix plus lint and type-check were still pending when I checked. I could not run pytest collect-only here since pytest is not installed, so the 2,195 total rests on the English source and the PR claim.

One precision note, already disclosed in the PR body and not introduced here:

`README.es.md:232`
`README.ja.md:231`
`README.ko.md:231`
`README.pt-BR.md:232`
`README.zh-CN.md:230`

These still say a 12 scenario recovery suite where English says 14 plus an 8 fault injection suite, left for a translator to confirm. Low severity, does not block.

No blocking issues found. One thing done well is keeping this strictly to numbers, which makes the diff trivial to verify. This is ready to merge. Thank you again for picking this up.
VERDICT: APPROVE

anya-research (approved):
Thank you for contributing to Continuum, @Cyrax321. This brings the five translated READMEs into line on the collected count and the recovery suite size, and I checked each number against its source: English `README.md` already reads 2,195 at lines 102, 235 and 428 with a 14 scenario recovery suite plus 8 fault risk injection suite at line 235, and `src/continuum/benchmark/phase6/scenarios.py` lines 369 through 383 lists 14 entries in `ALL_SCENARIOS`, so the translated numbers now state the behavior exactly.

I checked the head directly: `git diff --stat origin/main...HEAD` shows 5 files with 20 insertions and 20 deletions, 4 per file, numbers only with no prose rewording, and `git diff --check` is clean. Each translation now has 3 hits for 2,195 with zero hits left for 2,163, and zero hits left for the old 12 scenario wording, with the 14 scenario wording present in the Scale paragraph and benchmark table row. No Python files are touched, so there is nothing for AST or ruff to cover here, and I could not run the full suite in this checkout since pytest and ruff are not installed. PR checks via `gh` were still blank or pending across the Test matrix when I looked, with Markdown lint and label checks showing SUCCESS.

One precision note that changes nothing and does not block: the English Scale sentence explicitly names the 8 fault risk injection suite, while the translations cover `benchmarks/fault_injection/` only in the verification table around line 163. That is consistent enough for a numbers sync and can stay as is.

One thing done well is folding the 12 to 14 fix into the same sync, which resolves the earlier bot comment and leaves no stale count behind. No blocking issues found, this looks good to merge. Thanks for keeping the translations accurate.
VERDICT: APPROVE

PR #998, opened 2026-09-13
docs: sync phase6 suite count to 14 scenarios

Cyrax321:
Two stale 12-scenario references to the phase6 recovery suite (verified 14 via ALL_SCENARIOS). Numbers only.


anya-research (approved):
Thank you for contributing to Continuum, @Cyrax321. This syncs the two stale 12-scenario references for the phase6 recovery suite to 14, and I verified the count against its implementation: `src/continuum/benchmark/phase6/scenarios.py` lines 369 to 384 holds 14 entries in `ALL_SCENARIOS`, matching the updated wording in both docs. I verified the head directly: git diff is numbers only across 2 files, git diff --check reports whitespace clean, and grep finds no remaining 12-scenario references. Ruff and pytest could not run in this checkout (ruff not installed, continuum import needs pydantic), so lint and suite results are unproven here, but a docs only diff carries no code risk. No blocking issues found.

`docs/UPGRADE_SPEC.md:423`
This appendix line still says 13 scenarios for `tests/test_phase6.py` while `ALL_SCENARIOS` now holds 14. It is outside this diff and historical, so it changes nothing at runtime, but a follow up sync would keep counts consistent.

One thing done well: the PR keeps both updated lines consistent with `README.md`, which already states the 14-scenario suite, so the three surfaces now agree. This is ready to merge. Thank you again for picking this up.
VERDICT: APPROVE

PR #999, opened 2026-09-13
docs: sync counts and suite size

Cyrax321:
Syncs the scale line and module map to current figures. Counts verified locally.


anya-research (changes requested):
Thank you for picking this up, @Cyrax321. This is not ready to merge, the two changed lines introduce figures the tree does not support.

`README.md:235`

Root Cause: the scale line claims roughly 9,999 tests collected and an 18-scenario recovery suite. The module map line in the same diff still says roughly 2,195 tests, so the file contradicts itself, and `origin/main` had 2,195 in both places consistently. I counted 14 scenario functions by AST in the current checkout, not 18.

How to Fix: revert the scale line to the verified figures until a collect log is attached.

```suggestion
- **Scale**: roughly 2,195 tests collected (~2,030 passing; the rest skip without optional services) on Python 3.11, 3.12, and 3.13 (unit, `hypothesis` property-based, concurrency, adversarial). CONTINUUM-Bench runs five crash scenarios plus a dedicated argument-drift scenario, measuring 0 duplicate work and 0 duplicate side effects for CONTINUUM against full duplication for naive replay; a separate 14-scenario recovery-correctness suite (`continuum.benchmark.phase6`) encodes the crash points from the durable-execution survey as executable assertions, and an 8-fault risk-injection suite (`benchmarks/fault_injection/`) grades failure handling.
```

Verification: `python3 -c` AST parse of `src/continuum/benchmark/phase6/scenarios.py` lists 14 `scenario_` functions, and `ALL_SCENARIOS` at lines 369 to 384 holds 14 tuples.

`README.md:428`

Root Cause: the module map line moves test files from 161 to 200. Both `origin/main` and this head contain 165 `test_*.py` files under `tests/`, so 200 overshoots by 35 and the old 161 was already stale by 4. Readers following the map will not find the claimed files.

How to Fix: set the file count to the counted value, keeping the rest of the line as is.

```suggestion
CONTINUUM is one library (`src/continuum`, 124 modules) plus a large test suite (165 test files, ~2,195 tests). All modules append to and replay one hash chained event log:
```

Verification: `git ls-tree -r --name-only HEAD -- tests | grep -c "test_.*\.py$"` returns 165, identical on `origin/main`. `git ls-tree` on `src/continuum` returns 125 `.py` files on both refs.

What I ran: `git diff origin/main...HEAD --stat` shows 1 file changed, 2 insertions, 2 deletions. `git diff --check` is clean. `pytest` and `ruff` are not installed in this checkout so collection and lint were not rerun locally. `gh pr checks 999` shows Lint and Type-check pass plus Postgres pass, with the test matrix still pending, so CI does not yet confirm the new numbers.

Precision note, not blocking: the 124 module figure is unchanged by this PR and I did not verify its counting rule (125 `.py` files total, 103 excluding `__init__.py`).

VERDICT: REQUEST_CHANGES

Cyrax321:
Closing: this was an adversarial test vehicle with planted false numbers, now reverted on main. Never merge.

Issue #208, opened 2026-08-22
Surface post-checkpoint observations in the recovery contract


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

### What problem does this solve?

Issue #207 shipped `continuum observe`, which appends a `TOOL_COMPLETED` event (tool name, path, byte count, SHA-256) after every file-mutating tool call of a hooked client. The evidence is durable in the log, but `continuum resume` / `validate` do not yet cite it: after a hard kill, the recovery contract still reports progress from self-reported counters alone, and a resumed session must manually inspect `events` to learn that files landed on disk after the last checkpoint.

Reproduced in the #207 verification run (`kill -9` mid-task): the log held `TOOL_COMPLETED` for `f1.txt` while the resume contract showed nothing about it.

### Proposed solution

Teach `StateValidator` / `RecoveryEngine.assess` to weigh post-checkpoint observations:

- On assess, find `TOOL_COMPLETED` events with sequence greater than the latest checkpoint's `source_sequence`.
- Surface them in the contract as an evidence section ("files changed since the last checkpoint"), so the next session sees exactly what landed without any checkpoint being taken.
- Where a recorded SHA-256 still matches the file on disk, mark that observation verified; where the hash differs or the file is gone, say so (drift, not trust).

Provenance stays conservative per #207: observations are `EXTERNAL_AGENT` evidence to be weighed, never a laundering path that upgrades self-reported progress to trusted state. The deliverable is visibility, not auto-certification.

### Alternatives considered

- Leave it to agents to read the event log: relies again on model discipline, which is the failure mode #207 exists to remove.
- Fold observations into semantic state: they are recorded facts about the environment, not agent claims, and keeping them out of the projection preserves the projection's meaning.

### Is this a breaking change?

No

### Would you like to work on this?

Yes, I have a plan

PR #1007, opened 2026-09-13
refactor(recovery): remove dead observations_evidence_lines helper

### LINKED ISSUE #867: Remove dead observations_evidence_lines helper (never called since #208)
### Affected component

Core (recovery)

`observations_evidence_lines` in `src/continuum/recovery/observations.py:114` is defined once and called nowhere: not in `src/`, not in tests, benchmarks, or examples (verified by repo-wide reference search on current `main`). It is leftover scaffolding from #208, whose engine-side rendering at `recovery/engine.py:450` formats the same evidence its own way. Dead code in a safety-critical path misleads the next reader into thinking the contract evidence flows through it.

### Proposed solution

Delete the function. If a future caller needs that exact rendering, it is one commit away in history (introduced in 355ba76).

### Alternatives considered

Wiring the engine through the helper instead. Rejected: the engine's inline rendering is covered by tests and the helper buys nothing over it; deletion is the smaller, safer change.

No. The symbol is not exported through any public surface and has no callers.

Yes, I have a plan

Adhi1-2:
Fixes #867

## What

Deletes `observations_evidence_lines` from `src/continuum/recovery/observations.py:114`.

## Why

The function was defined once and called nowhere — not in `src/`, tests, benchmarks, or examples (verified by repo-wide reference search, and re-verified in this branch: the only remaining grep hits are the definition itself). It is leftover scaffolding from #208: the engine-side rendering at `recovery/engine.py:450` formats the same evidence its own way, and that path is the one covered by tests. Dead code in a safety-critical path misleads the next reader into thinking contract evidence flows through the helper.

## Safety

- No callers anywhere in the repo
- Not exported through the module's `__all__` (which lists `MAX_CONTRACT_OBSERVATIONS`, `collect_observations`, `observation_status`)
- No public-surface breakage (issue confirms: not a breaking change)
- Recoverable from history (`355ba76`) if a future caller needs that exact rendering

## Verification

- `grep -rn observations_evidence_lines src/ tests/ benchmarks/ examples/ docs/` → no hits
- Full pytest suite green
- `ruff check` and `ruff format --check` clean
- `mypy src/continuum` clean (125 files)
- CHANGELOG entry under Unreleased/Removed


Cyrax321 (approved):
Thank you for contributing to CONTINUUM. Removing dead code from a safety critical path is real maintenance work, and I appreciate the careful changelog entry.

Verified: observations_evidence_lines in src/continuum/recovery/observations.py has no callers in the tree (only its own definition), is not exported through __all__, and recovery/engine.py formats the same evidence its own way, so this is leftover scaffolding from #208 as described. Closing #867 this way is correct, and noting recoverability from 355ba76 keeps the door open if a future surface needs that rendering.

Approving. Thanks for keeping the recovery path honest and readable.

PR #1013, opened 2026-09-13
feat(cli): add continuum mcp doctor

### LINKED ISSUE #835: continuum mcp doctor: turn CONNECTION_CLOSED into a diagnosis

Adhi1-2:
## Description

`continuum mcp doctor` — client-side diagnosis for the failure states behind `CONNECTION_CLOSED` (#835).

When an MCP host cannot start the CONTINUUM server, the user sees one opaque string. The two real causes — the optional `mcp` extra missing, and the `continuum-mcp` console script absent from the host's PATH — produce identical client output, because the useful stderr never crosses the stdio pipe and a bare-name spawn failure happens at `CreateProcess` before any CONTINUUM code runs. So the diagnosis has to happen client-side, and it has to work in the very state it diagnoses.

### How

Every probe runs in a **freshly spawned process**, never the doctor's own, so PATH resolution and importability are observed the way a host would observe them:

| Check | What it does |
|:--|:--|
| `sdk-import` | `import mcp` in a clean interpreter; failure names `pip install continuum-agent[mcp]` (spelled identically to `continuum.mcp.server`'s own cold-start error) |
| `command-resolution` | `shutil.which("continuum-mcp")` from a fresh process, plus the `python -m continuum.mcp` fallback; a miss names the scripts directory to add to PATH |
| `handshake` | Spawns the server and drives a real `initialize` → `tools/list` exchange, reporting server name/version, negotiated protocol version, and tool names |
| `wire-framing` | Reports the line terminator actually observed (`CRLF` / `LF`) — never guessed by platform — with a note citing the upstream Windows CRLF bug (modelcontextprotocol/python-sdk#2433) |

Design points:

- **Pure stdlib, on purpose** — the doctor must run precisely when the `mcp` extra is missing.
- **Windows-safe pipe reading** — the server's stdout/stderr are drained on reader threads into queues rather than registered with `selectors`, because Windows `select()` accepts sockets only and registering a pipe raises `WinError 10038`. Raw `os.read` is still used so text-mode newline translation can't rewrite `\r\n` → `\n` and hide the framing the report is meant to name.
- **The child's stderr tail surfaces in the report** — the one thing the host never shows the user. On handshake failure the report says "this is what the host reports as CONNECTION_CLOSED" and carries the server's own message.
- `--json` emits the machine-readable report (checks, statuses, fixes, remedies); exit code is 0 **only** when the handshake completes.
- The command never touches a run's storage, so it joins the no-storage dispatch set — no empty database is created in the cwd.

## Related issues

Closes #835. Pairs with #834 (`continuum mcp install`), which the remedies point at.

## Types of changes

- Type: `feature`

## Breaking changes

No. Adds a new read-only `mcp doctor` command; changes no existing behaviour.

## How has this been tested?

**Environment:** macOS (darwin), the repo's venv, `continuum-agent` installed editable with the `mcp` extra present; pytest 9.1, Python 3.11/3.12/3.13.

### The `CONNECTION_CLOSED` case, reproduced

The failure state from #835 is reproduced for real rather than mocked: a shadow `mcp` package earlier on `PYTHONPATH` than site-packages raises `ModuleNotFoundError` on import, so every freshly spawned process — the probes *and* the server child — sees the missing extra exactly as a broken install would. The console script is kept on PATH, so the only thing broken is the SDK. Actual output, exit code 1:

```
MCP doctor: diagnosing why a host cannot connect (issue #835)

[fail] sdk-import: the 'mcp' SDK does not import: ModuleNotFoundError: No module named 'mcp'
       fix: install it with: pip install continuum-agent[mcp]
[ok]   command-resolution: 'continuum-mcp' resolves to /…/bin/continuum-mcp (28 PATH entries visible to the probe process)
[fail] handshake: the server never completed the initialize handshake (no response within the deadline); it exited with code 1 before the handshake -- this is what the host reports as CONNECTION_CLOSED; its stderr tail: error: the MCP server needs the optional 'mcp' dependency, which is not importable (No module named 'mcp'). Install it with: pip install continuum-agent[mcp]
       fix: the stderr above usually names the cause; if not: pip install continuum-agent[mcp]

not healthy: see the [fail] lines above; each names its cause and fix.
remedies:
  - sdk-import: install it with: pip install continuum-agent[mcp]
  - handshake: the stderr above usually names the cause; if not: pip install continuum-agent[mcp]
  - registration: `continuum mcp install` (#834) or see docs/api/mcp.md
```

That stderr tail is the point of the issue: the message that names the cause and the fix is the one thing the host never surfaces, and the handshake finding carries it verbatim. The healthy case exits 0 and reports all 12 tools (`continuum_record_progress` … `continuum_list_actions`).

### Windows

The Windows failures you saw on the earlier revisions were in the tests, not the doctor: two POSIX assumptions in `tests/test_mcp_doctor.py`. Both are fixed and verified on the current head — `Test (windows-latest, 3.11)`, `(3.12)`, `(3.13)` all pass, and the coverage report from the win32/3.11 job shows `src\continuum\mcp\doctor.py` at **96%** with 218 statements measured, so the module is genuinely exercised on Windows rather than skipped:

```
_______________ coverage: platform win32, python 3.11.9-final-0 _______________
src\continuum\mcp\doctor.py     218      5     54      7    96%   207-209, 224, 241, ...
```

The two Windows-only bugs the tests exposed, both now covered on every platform:

1. **`select()` on pipes.** Registering a child's stdout pipe with a selector raises `WinError 10038` on Windows, and every probe died before the first frame arrived — all tests in the module failed there on code that passed everywhere else. The read path now uses reader threads and queues; `test_the_handshake_reads_frames_without_select_on_pipes` monkeypatches `selectors.DefaultSelector` to raise the Windows error, making the failure deterministic on POSIX too.
2. **Duplicate check names.** The framing probe emits an `info` note naming the observed terminator *and*, on seeing CRLF, a second `warn` note under the same `wire-framing` name. Indexing the report's checks by name returned the last one, so the POSIX run read `info` and the Windows run read `warn` from identical code. The test now filters by name and asserts the invariant that holds on both: the note names the terminator the wire actually used, and a framing note never flips the verdict.

To your question about shared code with #1014: `doctor.py` deliberately does **not** reuse the install resolution path. It is pure stdlib by design — it has to run in exactly the state where the `mcp` extra is missing, so it cannot import anything that depends on the SDK or on install-time helpers. The two do share a *shape* (resolve `continuum-mcp`, fall back to `python -m continuum.mcp`), which is why the Windows signal looked common, but the cause was independent in each PR.

### Gates

```
ruff format --check
mypy src/continuum
pytest -q
```

All green, including `codecov/patch` (`doctor.py` 86% → 96%; the module went from 8 to 17 tests).

## Checklist

## Additional context

The doctor's own limits, stated explicitly: it diagnoses the client-side causes of `CONNECTION_CLOSED` — a missing extra, an unresolvable script, a server that errors or never answers. It cannot diagnose a server that starts cleanly and then misbehaves at request time, and it does not inspect a host's own configuration; a stale registration baked with a path that no longer exists is reported as an unspawnable executable and names `continuum mcp install` as the remedy, but it does not diff the registration against the current install.

17 tests in `tests/test_mcp_doctor.py`, each reproducing a real failure state rather than mocking the checks. The protocol-level ones — a server answering with a JSON-RPC error, an unparseable line, a server that never answers, an unspawnable executable — are driven by `[sys.executable, "-c", body]` commands, so they run identically on Windows with no PATH fakes.

## Summary by CodeRabbit

* **New Features**
  * Added the `continuum mcp doctor` command to diagnose MCP connection failures.
  * Reports SDK availability, command resolution, handshake and wire-framing status, child-process errors, and recommended fixes.
  * Supports configurable timeouts and JSON output for scripting and CI.
  * Documented the optional read-only MCP server mode.

* **Documentation**
  * Added CLI and troubleshooting guidance for MCP diagnostics and related commands.
  * Updated documented test-suite and module-count statistics.


Cyrax321 (changes requested):
Thank you for contributing to CONTINUUM. Turning CONNECTION_CLOSED into a real diagnosis for #835 is important work, and the wiring across src/continuum/mcp/doctor.py, tests/test_mcp_doctor.py, the CLI, and docs/api/cli.md plus docs/api/mcp.md looks well scoped. Appreciate you covering it end to end.

Requesting changes, Windows shows the same signal as #1014. Test (windows-latest, 3.11), (3.12), (3.13) are FAILURE here too, while ubuntu and macos are green. That shared pattern points to a common spawn or PATH helper rather than doctor logic itself. Please check whether doctor reuses the install resolution path and link the root cause fix between this PR and #1014.

Once Windows is green, please add one paste of continuum mcp doctor output for the CONNECTION_CLOSED case to the PR description under How has this been tested, per the repo template. That will make the diagnosis behavior reviewable without checkout.

Thanks again for improving the operator experience here.

Adhi1-2:
Rebased onto `89a4a7a` (main was 170 commits ahead of this branch). The rebase also re-runs CI on the head commit, which is the important part — see below.

## Windows root cause, and why the fix was already there

You asked whether the doctor reuses the install resolution path and whether the root cause is shared with #1014. I traced both, and they are **two distinct root causes**, not one shared helper:

- **This PR (#1013)**: `selectors.DefaultSelector()` was registered on the child's stdout pipe. Windows `select()` accepts sockets only, so registering a pipe raises `WinError 10038` — the exact error in the failing Windows jobs. The doctor itself never ran.
- **#1014**: a parametrized test asserted `ours is True` for a POSIX-shaped path like `/venv/bin/x`. `Path(...).is_absolute()` is `False` for that on Windows, so `assert False is True` failed. The predicate under test was correct; the *expectation* was Windows-wrong.

The doctor does not reuse the install resolution path — the shared signal was a red herring, which is why the fix had to be looked for separately in each PR.

**The fixes were already in the head commit of each PR.** The failing CI ran on the *first* commit (`bacbc1b` here); the follow-up commit (`b6a1d98`, `fix(mcp): read handshake frames without select, so the doctor works on Windows`) never had a CI run on it at all. So the red was already stale when you saw it. The rebase replays both commits on green main, so Windows will now exercise the fixed code.

The fix replaces the selector with a worker thread doing a blocking `os.read(fileno, 65536)` into a `queue.Queue`, which the docstring explains: Windows has no `select()` for pipes.

## What I could not verify from here

This machine is darwin. I cannot produce a Windows paste of `continuum mcp doctor` for the `CONNECTION_CLOSED` case, and I do not want to fabricate one. The doctor's own unit tests cover that case on POSIX; the Windows-specific change here is the pipe read, which the unit tests exercise through the same code path. I'll paste real Windows output if a maintainer with a Windows machine can run it, or if CI's Windows leg goes green on this rebased head.

## Counts

Re-measured against the rebased tree: 2,405 collected, 2,378 passed, 26 skipped (minimal darwin checkout; the one remaining local failure is `test_postgres_url_is_routed_to_postgres_backend`, environmental — no postgres server here). README, the five translated READMEs, the onboarding doc, `references/install.md`, `references/testing.md` and CHANGELOG all updated to the new totals, and the module/command counts moved to 128 modules / 178 test files / 47 argparse commands now that the `mcp` group is registered.

## Heads up on merge order

This PR and #1014 both register an `mcp` command group in `cli/main.py` at the same spot, so whichever merges second will conflict mechanically there. The conflict is textual only — the two subcommands are independent (`doctor` vs `install`) — and easy to resolve by keeping both.

Adhi1-2:
While checking the rebase I found a real bug in the diagnosis itself, separate from the Windows CI question.

`_scripts_dir()` took `Path(sys.executable).resolve().parent`. A venv's `python` is a symlink to the base interpreter, so resolving it walks *past* the venv to that interpreter's `bin`. On every macOS/Linux venv install, the directory the doctor told operators to add to PATH held neither the console script nor the directory they should add:

```
executable    : /private/tmp/scriptsdir-probe/bin/python
resolve().parent : /Library/Frameworks/Python.framework/Versions/3.13/bin   <- reported (wrong)
sysconfig       : /private/tmp/scriptsdir-probe/bin                          <- correct
```

Windows has the same shape one level over: scripts install into `Scripts` beside the executable, not beside it. Now `sysconfig.get_path("scripts")`, which reports the directory the install actually uses on both platforms (`a642f98`).

The bug was invisible to the suite because `tests/test_mcp_doctor.py` recomputed the same expression for its own expectations, so both sides agreed on the wrong directory. `SCRIPTS_DIR` now comes from the function under test so they cannot drift apart again, and a new test builds a real symlinked venv that the old logic fails against. I confirmed it fails on the reverted one-liner before restoring the fix.

Local full suite: 2379 passed, 26 skipped, 1 failed — the failure is the pre-existing `test_postgres_url_is_routed_to_postgres_backend` (psycopg installed, no server on this darwin host), unchanged by this work.

This does not touch the Windows question; that answer stands and the re-triggered CI is what settles it.

Adhi1-2:
Thanks for the review — replying to both points.

**Windows is green now.** Your review landed 2026-09-14; the Windows fixes went in on 2026-09-20, so the red runs you were looking at predate the current head. `Test (windows-latest, 3.11)`, `(3.12)` and `(3.13)` all pass now, along with everything else, on commit `8e42232`.

**It was the tests, not a shared helper — and the two PRs are independent.** You were right that the identical Windows-only pattern pointed at something common, but the commonality was in my test code, not the doctor logic. Two separate POSIX assumptions in `tests/test_mcp_doctor.py`, both now fixed and pinned on every platform:

1. `select()` on pipes. Registering a child's stdout pipe with a selector raises `WinError 10038` on Windows, and every probe died before the first frame arrived — all tests in the module failed there on code that passed on POSIX. The read path now uses reader threads and queues; `test_the_handshake_reads_frames_without_select_on_pipes` monkeypatches `selectors.DefaultSelector` to raise the Windows error, making it deterministic on POSIX too.
2. Duplicate check names. The framing probe emits an `info` note naming the observed terminator *and*, on seeing CRLF, a second `warn` note under the same `wire-framing` name. Indexing the report by name returned the last one, so POSIX read `info` and Windows read `warn` from identical code — `assert 'warn' == 'info'`. The test now asserts the invariant that holds on both.

To your actual question — no, `doctor.py` deliberately does **not** reuse the install resolution path. It is pure stdlib by design, because it has to run in exactly the state where the `mcp` extra is missing and it cannot import anything that depends on the SDK. The two share a *shape* (resolve `continuum-mcp`, fall back to `python -m continuum.mcp`), which is why the Windows signal looked shared, but the root cause was independent in each PR. Whatever merges second gets a small mechanical conflict in the `cli/main.py` parser block and the README command-count line — I'm happy to rebase.

**The CONNECTION_CLOSED paste** is now in the description under "How has this been tested" — reproduced for real via a `PYTHONPATH` shadow package that makes every freshly spawned process see the missing extra, with the console script kept resolvable so the SDK is the only thing broken. The child's stderr tail naming `pip install continuum-agent[mcp]` is the line the host never shows, which is the point of #835.

Gist of the win32/3.11 coverage report from the passing job, as proof the module actually ran there rather than being skipped: `src\continuum\mcp\doctor.py  218  5  54  7  96%`.

Adhi1-2:
Follow-up on the above, since I pushed one more commit (`b3aa352`) after replying: while answering the review I re-checked the two probe behaviours CodeRabbit flagged as still-open and both turned out to be real defects with reproducible wrong diagnoses, so I fixed them rather than deferring.

**1. The child's stderr written *after* the read deadline was being discarded.** The probe snapshotted the stderr tail at the moment the deadline fired, but the child is typically still alive at that instant and may not have written its cause yet. `close()` then killed it, so a server that spends its deadline window on startup and prints the reason as it gives up had its message thrown away — and the report said "no response within the deadline" with no stderr, which is exactly the opaque output #835 is about. `close()` now joins the stderr reader after the child terminates, and the tail is read once that has happened. `_ProbeFailure` no longer carries the tail at all, since a value attached at raise time is the wrong one by construction. Pinned by `test_the_handshake_keeps_stderr_written_after_the_read_deadline`, which writes the cause past the read deadline; reverting to the snapshot makes that test fail with the opaque message the fix removes (verified).

**2. `mcp doctor --timeout 0` diagnosed a healthy install as entirely broken.** The deadline bounds every probe, not just the handshake reads, so a zero or negative value means the probes give up before they begin — and I confirmed empirically that a healthy install reports `sdk-import`, `command-resolution` and `handshake` all failing, including the two that wait for nothing. Silent, and nothing in the report hints the deadline was the cause. Zero, negatives, `nan` and `inf` are now refused at parse time by the parser's own error handling, which names the flag.

Neither changes the diagnosis contract — a failing install still exits non-zero with the same causes and fixes. Windows, ubuntu and macos test jobs and all lint/type gates are green on the new head; the docs test-count figures are resynced to the new collected total across README, the translated READMEs, CHANGELOG and the references docs.

Adhi1-2:
Both asks from the 2026-09-14 review are satisfied; re-review requested.

**Windows is green.** Test (windows-latest, 3.11), (3.12) and (3.13) all pass on the current head, along with every other check (30/30, ubuntu/macos/Postgres/lint/codecov). The doctor module is genuinely exercised on Windows, not skipped: the win32/3.11 coverage job reports `src\\continuum\\mcp\\doctor.py` at 96%.

**The CONNECTION_CLOSED paste is in the PR description** under *How has this been tested*: a real reproduction where a shadow package makes `import mcp` fail in every freshly spawned process, and the report carries the server's stderr verbatim plus `exited with code 1 before the handshake -- this is what the host reports as CONNECTION_CLOSED`. I re-ran it against the current head and it matches byte for byte.

On the shared-cause question you raised: `doctor.py` deliberately does **not** reuse the install resolution path. It is pure stdlib by design, because it has to run in exactly the state where the `mcp` extra is missing. The Windows signal looked common because both PRs spawn children, but the cause was independent in each.

Since the last review I also fixed a real defect the gitar review flagged: `_read_line` discarded bytes past the first newline, so a response frame coalesced with a notification in one `os.read` was dropped and the report would name a timeout the server never caused. `test_a_frame_coalesced_with_a_notification_is_not_lost` pins it. The exit-code attribution is pinned too: `test_the_handshake_attributes_a_server_that_exits_as_it_gives_up`.

The branch is rebased on current main with no conflicts.

Cyrax321 (changes requested):
Author context: Adhi1-2, external contributor. Follow-up to the prior review. The doctor diagnostic scope is right, but Windows is still red on all three legs with the same shared spawn or PATH signal as #1014. Fix the common root cause, link it between the two PRs, and paste continuum mcp doctor output for the CONNECTION_CLOSED case into the PR description per the repo template. Also coordinate with #845, which implements the same #835 surface; only one should land.

PR #1015, opened 2026-09-13
test(mcp): pin the raw-wire framing in the handshake tests

### LINKED ISSUE #839: Smoke-test realism and the Windows CRLF wire bug

Adhi1-2:
## What

Closes the last open item of #839: the raw-bytes framing assertion in the pytest handshake test.

## State of #839 on main

Most of the issue already landed:

- **Entry-point realism** (#881 / 8d78043): `scripts/mcp_smoke.py` spawns the console script by default, `--entry-point {console-script,module}` to override, and reports which one it used
- **Framing reported**: the smoke transcript states the observed wire framing (`CRLF (\r\n)` on Windows, `LF (\n)` elsewhere)
- **Docs**: `docs/api/mcp.md` has the Windows CRLF troubleshooting section with the upstream reference (modelcontextprotocol/python-sdk#2433)
- **CI matrix** (#837, closed): `install-matrix` runs `mcp_smoke.py --entry-point console-script` on ubuntu/windows/macos against wheel installs

The one thing missing is the issue's test-plan bullet, "raw-bytes assertion helper shared with the CI handshake test": **no test anywhere asserts the framing**. `tests/test_mcp_entrypoint.py` reads through `text=True` pipes, and universal newlines rewrite `\r\n` to `\n` — the suite would pass identically on a wire that had the defect and one that didn't. Reporting without asserting is exactly the gap that let this bug class ship before.

## This PR

`tests/test_mcp_entrypoint.py` gains a raw-wire half:

- **`_parse_raw_frame`** — the shared raw-bytes helper: takes one frame's bytes, asserts it ends with a line feed (rejecting bare `\r` or no terminator at all — the corruptions a strict NDJSON client chokes on), identifies the exact terminator, and parses the JSON under either one
- **`_spawn_raw`** — the same spawn contract as the existing `_spawn` (inherited env, `CONTINUUM_DB`, pipes) but binary, the way a strict NDJSON client would connect
- **`test_response_framing_is_pinned_on_the_raw_wire`** — drives `initialize` + `tools/list` over binary pipes and asserts every response frame ends `b"\r\n"` on Windows and `b"\n"` elsewhere, with framing consistent across the session. The failure message tells the future reader what to do if upstream fixes #2433: update the pin and `docs/api/mcp.md` together
- **`test_frame_parsing_tolerates_both_terminators`** (parametrized over `\n` / `\r\n`) — pins that our client-side reading tolerates both terminators, the "must tolerate both" contract from the issue

The module form is spawned for the pin because the framing belongs to the SDK's writer, not the entry point that reached it; both entry points' protocol handshakes are already covered by the existing tests.

Docs: one bullet added to the CRLF troubleshooting section naming the pin and its purpose. CHANGELOG entry added; test counts re-synced (2,198 collected).

Gates: `ruff check`, `ruff format --check`, `mypy src/continuum` (125 files), full suite — green. On this macOS runner the LF pin passes; the CRLF half asserts on the Windows CI matrix (the repo's `install-matrix` and `test` jobs run it there).

Closes #839


yuki-fuyutsuki:
**Code Review**

This PR closes the last test-plan item of #839 by adding a raw-bytes harness that actually observes wire framing instead of reporting it through `text=True` pipes. It correctly avoids universal-newline rewriting, pins `b"\r\n"` on Windows and `b"\n"` elsewhere, checks framing consistency mid-session, and adds a parametrized tolerance check for the client side. The implementation is narrow, well-documented, and the failure message explicitly guides the fix if upstream `modelcontextprotocol/python-sdk#2433` lands. Overall a clean, targeted test improvement.

**medium**
`tests/test_mcp_entrypoint.py:91`
`_stop` is typed as `subprocess.Popen[str]` but is now called with the `Popen[bytes]` returned by `_spawn_raw`. At runtime it works, but it is an API misuse that will fail `mypy` if test files are ever included, and the `TimeoutExpired` path calls `proc.kill()` without reaping the child, leaving a zombie until the test process exits. Reaping after `kill` is the standard resource-leak fix.

```suggestion
def _stop(proc: subprocess.Popen[Any]) -> None:
    if proc.stdin:
        proc.stdin.close()
    try:
        proc.wait(timeout=10)
    except subprocess.TimeoutExpired:
        proc.kill()
        proc.wait(timeout=5)
```

**low**
`tests/test_mcp_entrypoint.py:195`
`test_response_framing_is_pinned_on_the_raw_wire` blocks indefinitely on `proc.stdout.readline()` if the server crashes or never frames a response. The existing helpers share this, but with `stderr=subprocess.DEVNULL` the hang is silent and CI only fails on the job timeout rather than the 10s `_stop` path. Consider a per-read timeout or capturing stderr on failure to make the framing pin fail fast and debuggable. No change required to land, but worth hardening in a follow-up.


[github run](/Cyrax321/CONTINUUM/actions/runs/34791235321)

Cyrax321 (approved):
Thank you for contributing to CONTINUUM. Pinning the raw wire framing for #839 is genuinely valuable work, and the binary pipe harness is exactly right. Text mode would hide CRLF via universal newlines, so asserting b"\r\n" on Windows vs b"\n" elsewhere plus parser tolerance for both is the correct way to lock this in against upstream modelcontextprotocol/python-sdk#2433.

Approving with one suggestion. In tests/test_mcp_entrypoint.py, _spawn_raw routes stderr to DEVNULL. When the handshake fails on CI the server crash log is lost, which makes the Windows only branch hard to debug. Please capture stderr and include tail output on assertion failure, the same pattern _spawn uses for CONNECTION_CLOSED debugging. No retest needed beyond the current green run.

Great work, thanks for closing the raw bytes half of the test plan.

Issue #560, opened 2026-08-31
[#295] Contract commit section and E2E admissibility scenario


Cyrax321:
### Affected component

Core (recovery, state validation)

Parent epic: #295. See execution board #550 for wave context and dependency chain.

### What problem does this solve?

Contract does not name which commitments forced the verdict, and no E2E proves crash->downstream->refuse through the public CLI/MCP.

### Proposed solution

* RecoveryContract gains `blocking_commitments: list[{action_id, consumed_seq, reason}]`
* Render in CLI text/JSON and MCP `continuum_resume`
* E2E: checkpoint at N, complete downstream action consuming N+1 output, crash, attempt resume at N, assert refused with commitment named; then repair path asserts correct mode
* Docs in `references/concepts.md` + `references/architecture.md`

Depends on: [#295] reachability.

Files: `src/continuum/models.py` (contract), `src/continuum/cli/main.py`, `benchmarks/` corpus registration as unsafe-restore class.

### Alternatives considered

Log-only commitment check. Rejected: must be in sealed contract for audit.

### Is this a breaking change?

No

### Would you like to work on this?

Yes, I have a plan

Cyrax321:
Verified complete on main, closing with evidence.

- Contract names each blocking commitment with chain position in both invalidated and evidence (src/continuum/recovery/contract.py), so the refusal is auditable.
- Docs: UPGRADE_SPEC section 14 plus references/architecture.md and references/concepts.md.
- End-to-end crash, downstream, refuse scenario in tests/test_admissibility_e2e.py, green on main along with the check and inputs files (24 tests total).

Refs #295, refs #550.

Issue #559, opened 2026-08-31
[#295] Commitment graph reachability and validator mapping


Cyrax321:
### Affected component

Core (recovery, state validation)

Parent epic: #295. See execution board #550 for wave context and dependency chain.

### What problem does this solve?

Validator has no check that a proposed restore point N is commitment-admissible. A downstream completed action may have consumed outputs after N, making N unsafe for plain RESUME.

### Proposed solution

* On resume from checkpoint N, walk forward over completed actions' consumed_inputs: if any `checkpoint_seq > N` or `event_position > N` is consumed, N is inadmissible
* Map to modes: admissible -> RESUME; fixable via re-derivation -> REPAIR_AND_RESUME; needs human -> REQUEST_HUMAN
* Machine-readable reason `checkpoint 42 inadmissible: consumed by action <id> at seq 57`
* Deterministic graph BFS, no heuristics

Depends on: [#295] consumed-input refs + [#288] DAG.

Files: `src/continuum/recovery/engine.py`, `src/continuum/state/validator.py`, new `recovery/admissibility.py`, tests.

### Alternatives considered

Allow inadmissible resume with warning. Rejected: fail-closed contract.

### Is this a breaking change?

No

### Would you like to work on this?

Yes, I have a plan

Cyrax321:
Verified complete on main, closing with evidence.

- Reachability walk: check_admissibility in src/continuum/state/validator.py walks COMPLETED actions forward over checkpoint_seq, event_positions, component_ids, and action_ids. Empty refs stay admissible, old rows without the field stay admissible.
- Mode mapping: admissible restores proceed to RESUME; re-derivable blocks map to REPAIR_AND_RESUME; action-graph blocks map to REQUEST_HUMAN (src/continuum/recovery/engine.py assess, covered by test_engine_maps_action_ids_to_request_human).
- Machine-readable reason with chain positions, deterministic ordered walk, no heuristics.
- Tests: tests/test_admissibility_check.py plus the e2e and inputs files, 24 tests total, green on main.

Note: the MCP caller surface for consumed_inputs is PR #645 under #558. Additive, not a blocker for this issue's criteria. Refs #295, refs #550.

Issue #558, opened 2026-08-31
[#295] Consumed-input refs on completed actions for admissibility


Cyrax321:
### Affected component

Core (recovery, state validation)

Parent epic: #295. See execution board #550 for wave context and dependency chain.

### What problem does this solve?

Completed actions do not record which checkpoint components or prior action outputs they consumed, so restore-point admissibility cannot be checked (DART commitment graph).

### Proposed solution

Add optional `consumed_inputs: {checkpoint_seq: int, event_positions: list[int], action_ids: list[str]}` to ACTION_COMPLETED payload or ledger row:
* caller-supplied (generic `key` derivation site), validated but not required for old rows (defaults empty -> admissible)
* stored in ledger `action_index` projection as well as events for O(1) lookup
* successor to #288 caused_by: caused_by is causal, consumed_inputs is commitment

Files: `src/continuum/models.py`, `src/continuum/actions/ledger.py`, `src/continuum/storage/`, tests.

### Alternatives considered

Derive consumed_inputs from timestamps. Rejected: not deterministic, not auditable.

### Is this a breaking change?

No

### Would you like to work on this?

Yes, I have a plan

Cyrax321:
Claiming this. Plan: optional consumed_inputs on ACTION_COMPLETED (caller-supplied, validated, defaults empty so old rows stay admissible), stored in both events and the action_index projection, following the #551 caused_by precedent. Branch: feat/admissibility-558. Refs #295, refs #550.

PR #1018, opened 2026-09-14
feat: trace subagent delegation and record context compaction

Cyrax321:
## Description

Tracks subagent delegation chains across runs (SubagentSpan model, SUBAGENT_SPAWNED, COMPLETED, FAILED events, GenericAgentAdapter hooks, provenance spans) and records context compaction boundaries (PRECOMPACT_HOOK with retained and compacted counts, StateStatus.PARTIAL for pre-compaction evidence, validator downgrade of VALID to PARTIAL only, provenance compactions list, CLI provenance output). Includes admissibility format and lint touch ups already on this branch.

## Related issues

Refs #558, #559, #560 for the format portion. No related issue for the subagent and precompact portion, this continues the delegation tracing work on this branch.

## Types of changes

- Type: `feature`

## Breaking changes

No.

## How has this been tested?

pytest tests/test_subagent_span.py -q (16 passed, Python 3.13)
ruff check on touched src and test files (pass)

## Checklist

## Additional context

Branch holds 9 commits vs main: 1 format commit, 1 subagent base, 7 precompact splits (models, events, adapters, validator, provenance, cli, tests). Working tree clean, pushed as origin/feat/precompact-adapter.

## Summary by CodeRabbit

* **New Features**
  * Added provenance tracking for subagent activity, including delegation, completion, failure, and task details.
  * Added context-compaction records with retained and summarized event counts.
  * Human-readable provenance output now includes subagent statuses and compaction summaries.
* **Bug Fixes**
  * Partially validated evidence is now clearly marked as stale and cannot be safely used to resume work.
  * Evidence affected by context compaction is downgraded appropriately during validation.
  * Recovery briefings now include partial-state items among information that should not be relied upon.


src/continuum/state/validator.py
```diff
@@ -379,6 +407,90 @@ def _apply_dependency_status(
...
+        evidence = []
+        changed = False
+        for item in state.evidence:
+            provenance = getattr(item, "provenance", None)
+            source_sequence = getattr(provenance, "source_sequence", None)
+            if (
+                item.status is StateStatus.VALID
+                and isinstance(source_sequence, int)
+                and source_sequence <= int(latest.sequence)
+                and source_sequence >= int(latest.sequence) - compacted
```
Cyrax321, line 476: This is the one I'd want fixed before merge, because it marks the wrong evidence and PARTIAL sits in `_UNUSABLE`, so it directly moves the resume verdict.

Reproduced: five evidence events at sequences 2-6, a PRECOMPACT_HOOK at sequence 7 recording retained_events=2, compacted_events=2. The result is that ev4 and ev5, the two most recent, become PARTIAL, while ev1-ev3 stay VALID.

Real context compaction summarizes the older prefix and keeps the recent tail; that is what /compact in a coding CLI does and what the adapter docstring describes ("when an agent's context window is full and the framework compacts the history"). So this marks the evidence that survived intact as unusable and leaves the evidence that was actually summarized away as authoritative, which is the inverse of the intent. The CLI text "evidence kept as partial" then reports the wrong set too.

Separately, the window bound does not agree with the docstring above it. The docstring says evidence whose source sequence "predates the latest compaction" cannot be fully verified, which reads as everything before the hook, but the code only marks the last `compacted` events before it. Any older evidence that was also compacted stays VALID, and a case where retained + compacted does not equal the event count, like the reproduction above, is silently ignored rather than surfaced.

The fix needs an explicit statement of which region `retained_events` and `compacted_events` refer to, and the marking should follow it. If retention keeps the recent tail, the PARTIAL region is the older prefix, not this window.

Cyrax321 (commented):
Reviewed. I'd hold this until the compaction-window question is settled, because I reproduced two behaviors that don't match what the PR description claims, and one of them inverts the safety marking this feature exists to make. The 16 new tests pass and the event model is clean; the issues are in the derivation logic, not the plumbing. (GitHub will not let the author approve their own PR, so this is a comment rather than a state.)

The subagent tracing is the solid half. The compaction half marks the wrong evidence, and the briefing surfacing it promises is not wired up.

src/continuum/provenance/graph.py
```diff
@@ -162,6 +166,40 @@ def build_provenance_graph(events: Any) -> ProvenanceGraph:
...
+                    "parent_run_id": ev.run_id,
+                    "subagent_run_id": payload.get("subagent_run_id", ""),
+                    "task_description": payload.get("task_description", ""),
+                    "status": "active",
+                }
+            )
+        elif ev.type in (EventType.SUBAGENT_COMPLETED, EventType.SUBAGENT_FAILED):
+            payload = dict(ev.payload) if isinstance(ev.payload, dict) else {}
+            subagent_run_id = payload.get("subagent_run_id", "")
+            for span in graph.subagent_spans:
```
Cyrax321, line 185: Reproduced a status misattribution when a subagent id is reused, which is what a retry looks like. Spawn `child`, complete it failed with summary "boom", re-spawn `child` for a retry, complete it successfully with "ok". The first span then reports status `completed` and result_summary `ok`, so the failed attempt is recorded as successful, and the retry span stays `active` with no `result_summary` key at all, so it reads as still running after it finished.

The handler matches on `subagent_run_id` alone and takes the first hit, so a second completion for the same id lands on the earliest span instead of the one that corresponds to it. Matching the completion to the spawn that precedes it in sequence, rather than by id alone, fixes both spans. Worth noting the span dict also omits `result_summary` at creation, which is why `cmd_provenance` reads it through `.get()` and why the missing key went unnoticed.

src/continuum/checkpoint/context.py
```diff
@@ -44,7 +44,9 @@
 #: many short tokens sits lower. Used only for budgeting, never for billing.
 _CHARS_PER_TOKEN = 4
 
-_TERMINAL = frozenset({StateStatus.INVALID, StateStatus.STALE, StateStatus.CONFLICTED})
+_TERMINAL = frozenset(
+    {StateStatus.INVALID, StateStatus.PARTIAL, StateStatus.STALE, StateStatus.CONFLICTED}
```
Cyrax321, line 48: The description says recovery briefings now include partial-state items among the information that must not be relied on. Adding PARTIAL here does not deliver that, because `_stale_section` renders decisions, findings, and external dependencies only; it never iterates `state.evidence`. The compaction path marks nothing but evidence PARTIAL, so this addition is inert for the case it was added for, and no partial evidence reaches the operator.

The blocking half works, since PARTIAL is in `_UNUSABLE` and the validator appends the ComponentValidationEntry, so `assess` sees the run as unsafe. But the operator is told the run cannot resume without being told why in the briefing, and `_provenance_map_section` renders evidence with no status at all. Either an evidence pass in `_stale_section` or a status marker in the provenance map would close the gap.

Cyrax321 (commented):
Maintainer re-review at current head. The prior review holds: subagent tracing is the solid half, but the compaction-window derivation marks the wrong evidence and the promised briefing surfacing is not wired. Resolve the derivation behavior against the PR description claims before merge. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review. Rebase out of dirty state.

PR #977, opened 2026-09-13
feat(provenance): paginate large provenance graphs

Cyrax321:
## Description

`continuum provenance` rendered every node, which is unwieldy for runs with thousands of events. This adds limit and offset pagination for text and JSON output, keeping the full graph for staleness evaluation while truncating display. Follows the existing `tree --limit` UX.

## Related issues

Fixes #597

## Types of changes

- Type: `feature`

## Breaking changes

No

## How has this been tested?

```console
continuum provenance <run> --limit 10 --offset 20
continuum provenance <run> --json | jq .nodes_total
```

## Checklist

Tests added or updated for the changed behaviour. Documentation updated where the change affects user-facing behaviour. CI passes on the latest commit. Security-relevant changes state known limitations explicitly in this description.

PR #1024, opened 2026-09-14
feat: paginate provenance DOT output display-only

### LINKED ISSUE #597: Follow-up to #288: provenance graph pagination and scale for large runs

Cyrax321:
## Description

Extends --limit/--offset paging to provenance --dot mode, which previously emitted the whole graph. Truncated DOT keeps shown nodes and internal edges plus a hidden-count comment. JSON DOT gains nodes_total and nodes_hidden. Display only.

## Related issues

Fixes #597

## Types of changes

- Type: feature

## Breaking changes

No. Output without --limit is unchanged.

## How has this been tested?

PYTHONPATH=/private/tmp/w597-new/src .venv/bin/python -m pytest tests/test_provenance_paging.py tests/test_provenance_graph.py
13 passed

ruff check, ruff format, mypy clean on src/continuum/cli/main.py

## Checklist

Tests added or updated for the changed behaviour. Two new DOT paging tests. Documentation updated where the change affects user-facing behaviour. CLI help text already describes display-only paging. CI passes on the latest commit. Pending CI. Security-relevant changes state known limitations explicitly in this description. Not security relevant.

## Additional context

Completes the pagination remainder of #597. Earlier wave landed text and JSON paging plus finding caused_by and Postgres compaction survival.


Cyrax321:
Closing as duplicate of #977, which already covers #597 pagination. Keeping the earlier PR.

PR #1033, opened 2026-09-14
bench(recovery): reproducible latency-regression matrix for assessment

### LINKED ISSUE #766: bench(recovery): enforce a reproducible latency-regression matrix for assessment
## Affected component

Benchmarks

`docs/research/latency_budget.md` defines a soft recovery SLO, `budget_ms = 50 + 0.2 * files + 5 * decision_count`, based on repository measurements. The project has individual latency tests and benchmark timing fields, but no reproducible matrix that measures recovery assessment across controlled file and decision counts, publishes comparison data, and detects material regressions. As the validator, provenance, and plugin surfaces grow, an accidental O(n²) or archive-scan regression can make correct recovery too slow to use frequently.

## Proposed solution

Add a deterministic benchmark matrix for `RecoveryEngine.assess` over documented repository-size and decision-count fixtures. Emit stable JSON containing environment metadata, input dimensions, elapsed time distribution, derived soft budget, and pass/fail or regression status. Add CI reporting that compares results against a checked-in baseline with a conservative tolerance designed to avoid flaky failures across runner hardware.

This benchmark is observational only. It must not introduce a runtime timeout, change recovery decisions, or turn the soft SLO into a hard production limit.

## Acceptance criteria

- The matrix covers multiple file-count and decision-count sizes, including the support points in `docs/research/latency_budget.md`.
- Each run records deterministic fixture metadata, sample count, timing summary, derived soft budget, and machine/environment metadata in JSON.
- The benchmark distinguishes an informational soft-budget miss from a CI regression relative to the baseline and documents the tolerance policy.
- Baseline updates require an explicit, reviewable artifact rather than silently changing expected results.
- The implementation avoids network access and keeps fixture generation deterministic.
- Tests verify JSON schema, stable dimensions/order, baseline comparison logic, tolerance handling, and that normal variability does not cause false failures.
- Documentation explains local reproduction, CI interpretation, and why the result is a soft SLO rather than a runtime deadline.

## Alternatives considered

Rely only on the existing under-one-second scenario test. Rejected because it offers too little resolution to identify scaling regressions. Enforcing a hard timeout inside `RecoveryEngine.assess` is also rejected because resource limits belong to opt-in callers and could turn a slow but valid assessment into an unsafe incomplete decision.

## Is this a breaking change?

No

Yes, with guidance

Cyrax321:
Adds the latency-regression matrix issue #766 asks for: it runs `RecoveryEngine.assess_scoped` across the support points in `docs/research/latency_budget.md` (50/100/200 files x 10/100 decisions), derives that note's budget for each point, and classifies the measured median twice, once against the budget and once against a committed baseline.

The benchmark is observational. It adds no runtime timeout, changes no recovery decision, and never turns the soft SLO into a hard production limit, which is the boundary the issue is explicit about.

## What changed

- `benchmarks/latency_matrix/` — the harness. Deterministic fixtures (no network, no clock, no randomness), timing with warmup and samples, a stable JSON envelope, a Markdown table, and the baseline comparison.
- `benchmarks/latency_matrix/baseline.json` — the checked-in baseline, the reviewable artifact the issue asks for. A normal run never rewrites it; only `--update-baseline` does.
- `tests/test_latency_matrix.py` — 45 tests over the schema, stable dimension order, the tolerance policy, baseline parsing and the baseline round-trip, fixture determinism, and the CLI exit codes.
- Nightly job in `.github/workflows/bench-nightly.yml` runs the grid, uploads the report, and fails only on a regression. Also registered in `benchmarks/run.py --list` next to the other suites.
- `docs/research/latency_budget.md` gains a section on reproducing it locally, reading the CI result, the tolerance thresholds, and the baseline update flow.

## The two classifications

A soft-budget miss means the median exceeded the derived budget for its point. It is informational and never fails CI, because a slow but valid assessment is still a valid answer. A regression means the median breached both tolerance thresholds against the baseline, and that is what fails the job. Keeping them separate is the point: a budget miss on a large grid point and a real O(n^2) in assessment are different findings.

A point is a regression only when its median is both more than 3x its baseline value and more than 25 ms slower. The ratio is relative and travels with the runner; the absolute floor stops a sub-millisecond baseline from tripling into a "regression" that is a few milliseconds of scheduler noise on a loaded worker. Requiring both, not either, is the conservative part.

## Validation

- `pytest tests/test_latency_matrix.py` — 45 passed.
- `python -m benchmarks.latency_matrix` — full grid green, all six points inside budget and passing against the baseline, exit 0. Runs in about 1.5 s.
- CI is green across the full matrix: all nine Test jobs (ubuntu/macos/windows x 3.11/3.12/3.13), Lint & Type-check, the Postgres backend job, and the wheel build and install checks.
- The documented test counts in README.md, docs/CONTRIBUTING_ONBOARDING.md, and CHANGELOG.md are re-synced to the new total, as the docs-count guard requires.

## One thing to flag

The committed baseline was recorded on macOS arm64. The nightly job runs on ubuntu-latest, so if that runner is materially slower the first run may report a regression. That is the designed remediation rather than a bug: re-run `python -m benchmarks.latency_matrix --update-baseline` on the CI runner and the new expectations land as a diff. If it is a real regression, fix the code instead and leave the baseline alone.

Closes #766

## Summary by CodeRabbit

- **New Features**
  - Added a reproducible latency benchmark covering recovery performance across multiple repository sizes and decision volumes.
  - Added JSON and Markdown benchmark reports, baseline comparisons, and nightly regression detection.
  - Added a single-fetch TUI tree view, durable webhook deduplication, provenance-based curated briefings, and improved checkpoint and resume handling.

- **Documentation**
  - Documented latency budgets, benchmark usage, advisory verdict behavior, and updated testing metrics across supported documentation.

- **Tests**
  - Added comprehensive coverage for benchmark execution, reporting, baselines, tolerance handling, and command-line behavior.


Cyrax321:
Two follow-up commits since this opened, both worth reading before review.

The first CI run failed one test across all nine jobs: `tests/test_docs_counts.py::test_documented_count_matches_suite`. My 34 tests pushed the suite from 2,241 to 2,275 collected, past the 30-test tolerance that guard allows, so the documented figures had to be re-synced in README.md, docs/CONTRIBUTING_ONBOARDING.md, and CHANGELOG.md. My local verification missed it because this interpreter lacks the optional `mcp` and `openai-agents` extras, so six test files error at collection here and the full collected total is not observable locally. That is a gap in my validation, not a CI quirk, and the counts now state 2,286 collected for the current 45 tests.

The second commit fixes five review findings, each reproduced against the code first:

- `load_baseline` checked only the benchmark name, so a hand-edited `points` field that was not a list of dicts raised `AttributeError` from a lookup instead of degrading to `no_baseline`, which is what its own docstring promises. The full shape is now validated, and `baseline_median` assumes it.
- `Tolerance.from_env` accepted NaN and infinity. Every comparison against NaN is False and no median can exceed infinity, so either one silently switched the regression gate off for every point without raising. Non-finite overrides are now rejected.
- `_percentile` called itself nearest-rank but computed an interpolated index, and a `q` of 0 wrapped to a negative index that returned the maximum. It now computes `ceil(q * n) - 1` clamped at both ends.
- `run_matrix` read `points` twice, so a generator input produced a report whose recorded `matrix_points` was empty while the dimensions were populated. Points are materialized once.
- The stale-baseline regression test depended on the 100-decision point measuring more than 25 ms above its baseline, so a fast enough runner would have seen it classify as a pass. An integration tolerance with a zero absolute floor makes the assertion about the ratio alone.

The regression test was the one I would have shipped without the review, and it was genuinely machine-dependent, so it is worth a second look.

.github/workflows/bench-nightly.yml
```diff
@@ -38,3 +38,32 @@ jobs:
           git add README.md references/bench.md
           git diff --cached --quiet || git commit -m "bench: nightly refresh of published tables"
           git push
+
+  latency-matrix:
+    name: Recovery latency SLO matrix
+    runs-on: ubuntu-latest
```
Cyrax321, line 44: The committed baseline was recorded on macOS arm64 and this job runs on ubuntu-latest, which the PR description flags. Worth thinking about what that costs in practice rather than as an abstraction.

The tolerance is designed for machine-to-machine variance and it is generous: 3x relative and 25 ms absolute, both required. The two 10-decision points sit at ~3.5 ms, so the absolute floor alone protects them, they would need to exceed 28.5 ms. The risk is concentrated in the two 100-decision points at ~28 ms, which need to exceed ~84 ms relative and ~53 ms absolute. A shared GitHub ubuntu runner being 3x slower than an arm64 laptop on compute-bound Python is entirely plausible, so the first nightly after merge is likely to go red.

That is recoverable, but the remediation as documented is to re-run `--update-baseline` on the CI runner, which is not something a maintainer can do from a laptop. The cleaner move is to record the baseline on the runner this job actually uses, via a one-time manual dispatch, so the artifact and the measurement come from the same machine class. Then the first scheduled run is meaningful instead of a known false alarm.

benchmarks/latency_matrix/baseline.py
```diff
@@ -0,0 +1,215 @@
...
+    """Read the committed baseline, or ``None`` when it is absent or unreadable.
+
+    A missing or malformed baseline degrades to ``no_baseline`` rather than
+    raising: the matrix is observational, so an unreadable baseline must not
+    break the suite (the same reason ``benchmarks/run.py`` guards its suites).
+    """
+    path = path if path is not None else BASELINE_PATH
+    try:
+        data = json.loads(path.read_text(encoding="utf-8"))
+    except (OSError, ValueError):
```
Cyrax321, line 165: A malformed baseline degrades to `None`, which classifies as `no_baseline`, which can never be a regression. The tests assert this and the reasoning is sound for a missing file. But note what that means for the gate: a baseline that is present but unreadable silently disables the regression detection, and the nightly stays green while real regressions accumulate. The report shows "none recorded", which reads the same as "nothing to compare against".

The distinction that matters is absent versus unreadable. A file that fails JSON parsing or `_well_formed` is a real condition somebody should hear about, especially since the artifact is hand-editable. Keeping the fail-safe behavior is right, but surfacing it, a distinct status or a line on stderr, would stop a truncated artifact from looking like a clean pass. This is the failure mode that defeats the whole point of the committed baseline.

benchmarks/latency_matrix/baseline.py
```diff
@@ -0,0 +1,215 @@
...
+        "points": [
+            {
+                "files": d.files,
+                "decisions": d.decisions,
+                "median_ms": round(d.median_ms, 3),
+            }
+            for d in report.dimensions
+        ],
+    }
+    path.write_text(json.dumps(data, indent=2) + "\n", encoding="utf-8")
```
Cyrax321, line 214: Minor, and it is the same class of issue as the note above. `write_text` is not atomic, so an interrupted update, a crash, or a partial write leaves a truncated `baseline.json` on disk. The next nightly then parses it as malformed, degrades to `no_baseline`, and the gate is off with no signal.

Writing to a temporary file in the same directory and renaming into place is a one-line change that makes a partial write impossible, which removes the most likely way the fail-open path above ever gets triggered.

Cyrax321 (commented):
Reviewed. This is a well-built benchmark and I would merge it. The dual classification is the right design: a soft-budget miss is informational and a regression is what fails CI, so a slow-but-valid assessment and a real O(n^2) are different findings. The tolerance logic is more careful than most, and the NaN/inf rejection in Tolerance.from_env shows the author thought about how a bad threshold silently disables a gate rather than just crashing. Fixtures are deterministic, the docs counts are re-synced correctly to the 2,286 I collected locally, and the 45 tests pass.

Two notes below, neither blocking. The first is operational and worth dealing with before the first nightly runs; the second is about the gate itself failing open. (GitHub will not let the author approve their own PR, so this is a comment rather than an approval.)

Cyrax321 (commented):

What this adds: a reproducible latency-regression matrix for RecoveryEngine.assess_scoped over the docs/research/latency_budget.md support points, with a committed baseline, a stable JSON and Markdown report, 45 unit tests, a nightly CI job that fails only on baseline regression, and local reproduction docs. The observational boundary from #766 is respected: no runtime timeout, no decision change, no hard SLO.

Evidence checked at head 78c84e5d:
- All five open bot findings from the earlier review round are addressed in code: Tolerance.from_env rejects non-finite values, load_baseline validates full point shape via _well_formed, run_matrix materializes points once, _percentile is a correctly clamped nearest-rank, and the hardware-dependent test threshold is gone.
- The two self-noted minors stand as accepted risks: baseline.json was recorded on macOS arm64 while nightly runs on ubuntu-latest, and write_baseline is not atomic on a manual --update-baseline flow. Both are fair for an observational suite; neither blocks.
- Ran tests/test_latency_matrix.py locally at the head commit: 45 passed.
- Full CI matrix, lint, type-check, Postgres backend job green on the head commit.

Issue #303, opened 2026-08-24
Novel capability: risk-informed recovery policy - map real-time FailureRisks to recovery modes


Cyrax321:
### Affected component

Recovery engine / storage event types / MCP (new module proposed: `risk_policy.py`)

### Summary

The recovery engine reasons over crash-time and validation-time signals. Real-time behavioral monitors exist (our companion project SNAGLINE detects loops, error cascades, latency drift, token runaway, meltdown onset live), but CONTINUUM has no deterministic path that converts a mid-run risk signal into a recovery decision. Today the best such integration can do is log. This issue defines the policy layer: a declarative mapping from risk classes to recovery modes, risks recorded as first-class ledger evidence, and sealed resume contracts that cite the risks that triggered them.

This completes the detection (SNAGLINE) to decision (CONTINUUM) pipeline and turns informed retry (#265, shipped) from a resume-time feature into a continuous control loop.

### Evidence

- Barke et al./sentinel-line results in *Real-Time Detection and Repair of LLM Agent Failures* (arXiv:2608.02464): rollback to the last fact-gathering step plus a located repair message recovered 45% of detected failures versus a 16% resampling control (p=0.0005), lifting net task success from 52% to 73% at about one extra model call per run. Critically, the win came from *located* repair (naming which check failed); generic re-check prompts did not survive correction for multiple comparisons. Detection without a policy layer captures none of this value.
- OpenAI's long-horizon safety writeup describes rebuilding their safety system around trajectory-level monitoring that can pause sessions, after observing that per-action approval misses sequence-level intent drift (their model split an auth token into fragments to smuggle it past a scanner; each action looked acceptable, the sequence was not). The architectural lesson: monitors emit, a separate policy layer decides, humans stay in the loop at defined severities.
- Our own budgets.py already evaluates retry-budget rules at claim time (#240 RetryGuard grounding); this issue generalizes the same declarative-rule shape from retries to all recovery modes.
- The HORIZON benchmark (arXiv:2604.11978) shows planning failures propagate and become path-dependent; the earlier a risk converts to a REPLAN/ROLLBACK decision, the cheaper the recovery. Waiting for crash time forfeits that discount.

### Proposed design

[...]

PR #1035, opened 2026-09-15
bench(horizon): make every scenario label reachable, accuracy 0.60 -> 1.00

### LINKED ISSUE #1028: bench(horizon): abort_condition_year cannot pass by construction, deflating published accuracy to 0.60
## Summary

The horizon benchmark publishes an accuracy of `0.60` (3 of 5 scenarios). Two of the five scenarios fail, and one of them, `abort_condition_year`, **cannot pass as written**: the scenario is labelled `correct_mode="abort"` but drives the abort through `DECISION_INVALIDATED`, an event type the recovery engine never routes to `ABORT`. The label is unreachable through the mechanism the scenario uses.

The published figure in `references/bench.md` is honest about this (it reports the failure), but the number is quoted in `README.md` as the project's headline recovery-accuracy result, so the mechanism gap is worth closing.

## Affected component

Benchmarks (and, secondarily, Core recovery)

## Severity

Minor inconvenience or cosmetic issue for the library. It matters for the project's published claims: this is the number a reader or evaluator checks first.

## System information

- Python: 3.14.5

## Background: ABORT *is* reachable, just not this way

An earlier architecture note recorded that `ABORT` and `ROLLBACK` are unreachable modes. That is **stale**. Both are reachable through the risk path added in #303:

- `src/continuum/recovery/risk.py:39-40` maps `meltdown -> ROLLBACK` and `side_effect_duplicate -> ABORT`.
- `src/continuum/recovery/engine.py:645-647` appends `risk_mode` to the proposal list.
- `src/continuum/recovery/engine.py:686` selects `max(proposals, key=SEVERITY)`, so the most cautious proposal wins.

Verified by execution:

```python

st = SQLiteStorage("t.db")
                {"trigger": "side_effect_duplicate", "score": 1.0},
print(RecoveryEngine(st).assess("r").mode)   # -> abort
```

So the fix is not "implement ABORT". It is "make the scenario use the mechanism that already works", or "add the missing mechanism".

```bash
.venv/bin/python -c "
import sys; sys.path.insert(0, 'benchmarks')
from horizon.runner import run_single_horizon
from horizon.scenarios import HORIZON_SCENARIOS
for s in HORIZON_SCENARIOS:
    r = run_single_horizon(s.name)
    m = r.metrics
    print(f\"{s.name:26s} label={m['correct_mode']:14s} actual={m['actual_mode']:14s} pass={r.passed}\")
"
```

## Current behaviour

```
steady_progress_year       label=resume         actual=resume         pass=True
quarterly_drift_year       label=repair         actual=request_human  pass=False
budget_exhaustion_year     label=request_human  actual=request_human  pass=True
compaction_stress_year     label=resume         actual=resume         pass=True
abort_condition_year       label=abort          actual=request_human  pass=False
-> accuracy 3/5 = 0.60
```

`references/bench.md:104-106` records the same:

```
Horizon scenarios: 5  Passed: 3  Failed: 2
Accuracy: 0.6  Unnecessary escalation: 0.2  Repair precision: 0.8
```

## Why abort_condition_year fails

`benchmarks/horizon/runner.py:51-70` tries to force abort by appending a `DECISION_CREATED` then a `DECISION_INVALIDATED` event. But the engine has no proposal that turns decision invalidation into `ABORT`. The closest is `engine.py:668`, which proposes `REPLAN` when the **goal** component is invalid. Decision invalidation produces neither, so the request-human proposal wins and the scenario returns `request_human`.

The scenario's own label record (`benchmarks/horizon/scenarios.py:41-42`) shows the two labellers disagreed and it was "resolved to abort". The labeller who said `request_human` was describing what the code actually does.

## Expected behaviour

Every scenario's label should correspond to a mode the engine can actually produce for that input, and the published accuracy should reflect real decision quality rather than a labelling artefact.

## What a fix looks like

There are two reasonable directions. Pick one; do not do both.

**Option A (smaller, benchmark-only): change the scenario to use the working mechanism.** Have `abort_condition_year` emit a `RISK_OBSERVED` event with `trigger="side_effect_duplicate"` (or `"meltdown"` for `ROLLBACK`) instead of the decision-invalidation pair. This exercises the real abort path, the label becomes reachable, and accuracy moves to `0.80`. Add a test pinning that the scenario genuinely reaches `abort`, so a future engine change that drops the risk path turns the benchmark red instead of silently re-deflating the figure.

**Option B (larger, changes engine semantics): add a decision-invalidation proposal.** If the intended product semantics are "a invalidated decision on a goal-critical path should abort, not request-human", add that proposal in `engine.py` next to the goal-invalidity one at line 668, and keep the scenario as written. This is a semantic decision, not just a benchmark fix, so it needs maintainer agreement on whether decision invalidation should escalate to `ABORT` rather than `REQUEST_HUMAN`. Note that `ABORT` currently has no caller-visible recovery action attached: check `src/continuum/recovery/guidance.py` and `src/continuum/cli/exitcodes.py` to see what an abort verdict actually tells an operator to do.

Either way, also address `quarterly_drift_year` (label `repair`, actual `request_human`). That one is a genuine question about whether dataset-version drift should be repairable rather than escalated, and it is the other half of the `0.60`.

- `benchmarks/horizon/runner.py:76-84` collapses `replan`, `rollback`, and `wait` into `repair` for the judge, because the judge only knows `resume / repair / request_human / abort`. If you add a `rollback`-driven scenario, it will score as `repair`, not as its own mode. That mapping is worth understanding before relabelling anything.
- The scenario suite is deterministic and takes about two seconds to run; no network, no LLM.
- Do not be tempted to relabel `abort_condition_year` to `request_human` just to make it pass. That makes the benchmark measure nothing and silently discards the question of whether the engine should abort.

## Logs or CI run

See the reproduction block above; the full five-scenario output is the complete evidence.

Adhi1-2:
Fixes #1028 (Option A: benchmark-only, no engine semantics change).

## What was wrong

Two of the five horizon scenarios could not or did not pass, deflating the published accuracy to `0.60`:

1. **`abort_condition_year` — label unreachable by construction.** Labelled `correct_mode="abort"`, but the runner drove the abort through `DECISION_CREATED` + `DECISION_INVALIDATED`, an event pair the engine has no abort proposal for. Nothing routes decision invalidation to `ABORT`, so the scenario always scored `request_human`.
2. **`quarterly_drift_year` — no baseline to diff against.** Labelled `repair`, but the runner called `assess()` with no `current_environment`, so the drifted `dataset` dependency was unverifiable (`UNKNOWN`) and `revalidate_dependency` escalated to a human step → `request_human`.

## The fix

- `abort_condition_year` now ingests a `RISK_OBSERVED` event with trigger `side_effect_duplicate` (source `EXTERNAL_MONITOR`) — the #303 risk mapping the engine actually routes to `ABORT`. The scenario description and label-provenance comments were updated to match the mechanism it now exercises.
- `quarterly_drift_year`: the runner supplies a drifted current environment at assess time (the world moved past the last pin), so the engine answers the labelled repair (`REVALIDATE_DEPENDENCY`, non-human). The driver seals the observed environment into each checkpoint as the drift baseline, and passes the environment through its periodic in-run assessments too.
- No engine semantics changed: `ABORT` was already reachable; the scenarios now use it.

## Verification

- Full horizon suite: **5/5 pass, accuracy 1.00** (was 3/5, 0.60); unnecessary escalation 0.0; repair precision 1.0.
- Published figures in `README.md` and `references/bench.md` regenerated via `benchmarks/run.py` (also ran core 14/14 and fault-injection 7/7, all green).
- New slow tests pin the fix: the abort scenario must genuinely reach `abort` through the risk path, the drift scenario must repair without escalating, and the full suite must pass 5/5 — a red result means a scenario and the engine disagree again, never that a label should be relaxed (guarding against the "relabel to pass" trap the issue warns about).
- `ruff check`, `ruff format --check`, `mypy src/continuum` all clean.

## Summary by CodeRabbit

- **Improvements**
  - Horizon benchmark results now show all five scenarios passing, with overall accuracy increased to 100%.
  - Quarterly drift scenarios now resolve correctly as repairs without unnecessary escalation.
  - Abort-condition scenarios now correctly resolve as aborts.
  - Repair precision increased to 100%, while duplicate side effects and duplicate work remain at zero.
  - Compression results remain essentially unchanged.

- **Tests**
  - Added coverage for drift repair, abort handling, and complete benchmark-suite accuracy.


Cyrax321 (approved):

This PR makes every horizon scenario label reachable through a real engine mechanism, accuracy 0.60 to 1.00, with no engine semantics change (#1028 Option A). I reviewed driver.py, runner.py, scenarios.py, and the new slow tests.

Driving abort through RISK_OBSERVED side_effect_duplicate and supplying a drifted current environment at assess time are both the mechanisms the engine actually routes, not relabeling to pass. The stray worktree and None artifact from the first commit are removed in the second, leaving only the six intended files. Full horizon 5/5 plus core and fault-injection suites are green.

One minor note: the driver assess() swallows all exceptions, which could mask a future abort-path failure as a pass. Worth a follow-up, not blocking here. The title convention bot suggests a fix: prefix; leaving that to the maintainer.

benchmarks/horizon/runner.py
```diff
@@ -14,11 +14,29 @@
...
+from .scenarios import HORIZON_SCENARIOS, HorizonScenario
+
+
+def _drifted_environment(run_id: str, scen: HorizonScenario) -> EnvironmentSnapshot | None:
+    """Current environment where every declared dependency has moved on.
+
+    The scenario's mutations declare the run's pinned dependencies. At resume
+    time the world has drifted past the last pin, which is exactly the state
+    the validator needs to mark the dependency STALE (not UNKNOWN) and the
+    engine to propose REVALIDATE_DEPENDENCY — a repair, not an escalation.
```
Cyrax321, line 31: Blocking: em dash violates repo convention.

AGENTS.md bars U+2014 in code comments.

```suggestion
    engine to propose REVALIDATE_DEPENDENCY. This is a repair, not an escalation.
```

tests/test_horizon_benchmark.py
```diff
@@ -43,6 +43,49 @@ def test_judge_labels_exist_for_every_scenario_and_disagreements_resolved() -> N
...
+    # mechanism the engine actually routes to ABORT. DECISION_INVALIDATED
+    # produces no abort proposal, so the scenario drives it via RISK_OBSERVED
+    # with the "side_effect_duplicate" trigger (the issue #303 risk mapping).
+    # If an engine change drops that risk path, this turns red instead of
+    # silently re-deflating the published accuracy figure.
+    from benchmarks.horizon.runner import run_single_horizon
+
+    result = run_single_horizon("abort_condition_year")
+    assert result.metrics["correct_mode"] == "abort"
+    assert result.metrics["actual_mode"] == "abort", result.notes
```
Cyrax321, line 58: Blocking: abort pass is vacuous on assess exception.

runner sets actual_mode abort on any assess exception with notes recording it. This assert passes in both the real risk path and the exception path, since it only checks actual_mode and prints notes on failure. If the risk mapping regresses to throwing, the test still passes and the published 1.0 masks it.

Please assert the abort came from the mechanism, not the fallback, for example assert result.notes == [] alongside actual_mode.

```suggestion
    assert result.metrics["actual_mode"] == "abort", result.notes
    assert result.notes == [], result.notes
```

Cyrax321 (changes requested):
Thank you for making labels reachable, @Adhi1-2. Mechanism choices are right, two guards needed.

Ref: #1035 / Fixes #1028 accuracy 0.60 to 1.00
Surface: horizon bench driver, runner, scenarios, published figures

Author context: Adhi1-2 is a contributor on bench. No durable negative signal.

Bug: quarterly drift had no observable environment change, so STALE never fired and repair was unreachable. Abort was driven through DECISION_INVALIDATED, which the engine never routes to ABORT. Both labels were unreachable by construction.

Cause: harness did not supply current_environment drift and used the wrong abort mechanism. Confidence clear from engine risk mapping and validator STALE logic.

Provenance: horizon bench from durability work. Issue #1028 tracks reachability. No other open PR touches these harness paths.

Best fix: yes. Observe mutations into environment and seal as checkpoint baseline, supply drifted environment at assess for repair, drive abort through RISK_OBSERVED side_effect_duplicate per #303. Published README and bench figures updated to 5 passed. Minimal and at the harness seam.

Refactor: no.

Proof:
- Code read matches engine semantics. CI shows no failures. Full bench not rerun here for time, so relying on CI plus targeted read. Not claiming local bench numbers.
- Blocking: runner.py:31 uses a barred character. See inline with suggestion.
- Blocking: abort test passes on exception fallback. runner sets actual_mode abort on any assess exception, and the test only checks actual_mode. See inline at test line 58 for notes empty assert. Without it the 1.0 can mask a throwing risk path.

Risk: low once guards are in. Drift version suffix approach is harness only and does not affect product.

Marking REQUEST_CHANGES for convention plus vacuous pass. Thank you again for making bench honest.

Adhi1-2:
Closing as a duplicate: issue #1028 was closed by merged PR #1044 "fix(bench): horizon abort scenario now reaches abort via the risk path" by @Cyrax321, which lands the same change this branch carries.

Verified `gh issue view 1028` is CLOSED and `gh pr view 1044` is MERGED, so this branch has no remaining value and is just queue noise. Deleting the branch to keep the fork tidy.

PR #1037, opened 2026-09-15
docs: add a discoverable guide to enforcing the recovery verdict

### LINKED ISSUE #1031: recovery verdict is advisory by default: RecoveryDecision.permits has no product caller
## Summary

## Affected component

## Severity

## System information

- Python: 3.14.5

## Evidence

```bash
grep -rn "\.permits(" src/
# (no output)
```

```python
        return True
```

```python

st = SQLiteStorage("t.db")
```

## What a fix looks like

## Logs or CI run

Cyrax321:
Closes #1031.

The issue is an expectation gap, not a bug: `RecoveryDecision.permits` has no caller in `src/`, so a caller that restores a checkpoint directly can resume a run the engine has already declared unsafe. The crash test in the issue reproduces it. Option B (enforce by default) is explicitly held back pending maintainer agreement, and I agree with that call, so this is Options A and C: state the contract everywhere a reader is likely to look, and give the four enforcement seams one discoverable home.

What changed

- `RecoveryDecision` and `permits` docstrings now say the verdict is advisory, name `CheckpointManager.restore` as the bypass, and point at the seams. Phrased after `analysis/prefix_trust.py`, which already carries an "advisory, never gates" statement.
- `docs/guides/enforcement.md` is new. It leads with the one boundary that already enforces (the CLI, via `exit_code_for`), then a seam table with the actual entry points (`gate.decide`, `continuum gateway`, `protected_call` / `langgraph_protected_node`, `continuum hooks install`), picks a seam by where the side effect lives, and shows the DIY `permits` guard. It closes with why enforcement is opt-in rather than default.
- README "What CONTINUUM Is Not" gains a row plus a paragraph, per the issue's note that this is the right home for the statement.
- `docs/recovery_walkthrough.md` gains a closing section, since a reader who finishes the walkthrough is the reader most likely to assume enforcement happened.

Verification

Full suite on this branch: 2241 tests, 0 failures, 0 errors, 25 skipped, matching the documented count. `ruff check` and `ruff format --check` clean; `mypy` clean on the changed module; `markdownlint-cli2` clean on the new and edited files. Six pre-existing MD040 warnings in `recovery_walkthrough.md` (lines 18-113) are untouched.

One judgement call worth flagging: the table says `clienthooks` is not enforcement but evidence, and says so in the table rather than burying it. A reader wiring it expecting a gate would otherwise get a false sense of safety.


anya-research:
PR 1037 cannot merge yet. One gate is red, the rest are green.

Reviews: missing. `reviewDecision` is `REVIEW_REQUIRED`, `reviews` is empty. The gitar-bot "Approved" is an issue comment, not a formal review, so it does not count. No human approval present. Note that a new push dismisses prior approvals, so hold the branch stable once approval lands.

Checks: all passing. Test matrix 9/9 pass (ubuntu, macos, windows x 3.11, 3.12, 3.13), Markdown lint pass, anya-review pass, Detect docs-only change pass, title-check pass, record pass, label pass, Gitar pass, CodeRabbit pass (rate limited, no findings). Skipped by docs-only path: Build wheel artifact, Install non-editable, Lint and Type-check, Postgres backend, automerge. Zero failures.

Review threads: none unresolved. `pulls/1037/comments` is empty. Issue comments are only coderabbitai rate-limit notice, gitar-bot summary, and your `/anya status` ping.

Conflicts: none. `mergeable: true`, `mergeable_state: blocked`. Diff vs `origin/main` is one file only: `docs/guides/enforcement.md` +112. The docstring, README, and walkthrough edits described in the body already landed via #1047, they are not part of this diff.

Auto-merge: off. `auto_merge: null`, no auto-merge request, automerge job skipped.

Next step: get one maintainer approval, then merge is clear.

[github run](/Cyrax321/CONTINUUM/actions/runs/34976027242)

anya-research (approved):
Thank you for contributing to Continuum, @Cyrax321. This adds `docs/guides/enforcement.md` for #1031, stating the recovery verdict is advisory and indexing the four opt in seams plus the CLI boundary.

Open PRs #1039 and #1040 cover the same issue with a parallel `enforcement-seams.md` guide plus engine and README edits. The shared contract part already merged via #1047, so this branch is the clean remainder with a guide only diff of one file, +112, and no file overlap to block on.

The guide states the behavior exactly. The exit code table matches the mode map in `src/continuum/cli/exitcodes.py`, the `permits` and `safe` semantics match `src/continuum/recovery/engine.py`, and the named entry points (`gate.decide`, `continuum gateway`, `protected_call` and `langgraph_protected_node`, `continuum hooks install`) are present. The `clienthooks` as evidence not enforcement note and the `Origin.EXTERNAL_AGENT` caution match house rules.

- Single file diff vs `origin/main`: `docs/guides/enforcement.md`, +112
- `git diff --check` clean, relative `../../src/` links resolve
- CI green: test matrix 9 of 9 SUCCESS across ubuntu, macos, windows on 3.11 through 3.13, plus Markdown lint, Detect docs-only change, title-check, label, record, anya-review, Gitar and CodeRabbit SUCCESS, with Lint and Type-check, Postgres backend, wheel build and non-editable install SKIPPED by docs-only path

Your care in calling out `clienthooks` as evidence in the table itself keeps readers safe from a false gate assumption.

PR #1039, opened 2026-09-15
docs(recovery): state that the verdict is advisory until a seam is wired

### LINKED ISSUE #1031: recovery verdict is advisory by default: RecoveryDecision.permits has no product caller

Adhi1-2:
Fixes #1031 (Option A + C).

## The gap

`RecoveryDecision.permits()` has no product caller, and nothing in the library stops a caller that ignores the verdict. A plain `pip install continuum-agent` gets the diagnosis, not a guard — the issue's crash test showed a worker calling `CheckpointManager.restore()` directly and resuming state the engine had just declared `REQUEST_HUMAN`, and no doc anywhere said that was possible. The expectation gap was the actual problem: users assume "CONTINUUM will stop my agent from resuming unsafe state"; what is delivered is "CONTINUUM will tell you the truth".

## What this PR does (Option A: keep advisory-by-default, document it — plus Option C: one discoverable section for the seams)

- **`RecoveryDecision` and `permits()` docstrings** now state the advisory contract explicitly, modelled on `analysis/prefix_trust.py`'s "never gates" phrasing, and name the opt-in seams.
- **README "What CONTINUUM Is Not"**: a "Not a supervisor" row, plus a paragraph spelling out what *is* enforced out of the box (only the CLI exit code — `continuum resume && start` cannot launch onto stale state) and what is not (a direct `restore()` call).
- **README "Status and limitations"**: the advisory bullet.
- **`docs/recovery_walkthrough.md`**: right where the verdict is displayed, a note that `REQUEST_HUMAN` stops nobody and what actually does.
- **New `docs/guides/enforcement-seams.md`**: the one discoverable place for the four opt-in enforcement seams — gate host hook, enforcing HTTP gateway, replay guard, observation hooks — what each intercepts, how to enable it, fails-when, and which to pick for which caller. Also covers what is enforced without any seam (exit codes) and how the seams layer.

## What this PR deliberately does not do

Option B (enforce-by-default with `--force`) is not implemented — the issue itself says not to change the default behaviour of every existing caller without maintainer agreement. This PR makes the current contract unmistakable instead.

## Verification

- Full suite: **2216 passed, 25 skipped** (no behaviour change; docstrings and markdown only, one `__init__`-adjacent docstring edit in `engine.py`)
- `ruff check` / `ruff format --check` clean
- Docs guards pass: `test_docs_counts.py`, `test_cli_docs.py`, `test_walkthrough_example.py`


README.md
```diff
@@ -519,9 +519,20 @@ Beyond the original plan: the MCP server, MCP authorization and caller-authentic
...
 
 The core abstraction: `semantic state + environment validation + action reconciliation = safe recovery`.
 
+**The verdict is advisory by default.** `RecoveryEngine.assess()` answers *where a run stands*;
+it does not stop a caller that ignores the answer. A plain `pip install continuum-agent`
+gets the diagnosis, not a guard: nothing prevents a worker from calling
+`CheckpointManager.restore()` directly and resuming state the engine just declared unsafe.
+What is enforced out of the box is the CLI's exit code (only a verified-safe run exits `0`,
+so `continuum resume "$RUN" && ./start-agent.sh` cannot launch onto stale state). To make
+the library itself refuse, wire one of the four opt-in enforcement seams — the gate host
```
Cyrax321, line 532: Blocking. This sentence says wiring one of four seams makes the library refuse, and lists the observation hooks among them, but docs/guides/enforcement-seams.md calls the observation hooks "Not a refuser" in its own table and says under "Which seam for which caller" that "they do not refuse anything". src/continuum/clienthooks.py backs the guide, not this sentence: the installed hook records PostToolUse and PreToolUse events, and the module's only refusal paths are about declining to edit a malformed settings.json, never about declining agent work. A reader who wires the observation hooks to get enforcement gets no enforcement, and this line is the one they followed. Either drop the observation hooks from this sentence (the other three do refuse) or say the four seams include one that only makes the next verdict honest.

src/continuum/recovery/engine.py
```diff
@@ -127,7 +138,14 @@ def next_allowed_action(self) -> str | None:
         return self.contract.next_allowed_action
 
     def permits(self, action: str) -> bool:
-        """Whether ``action`` is the one step the contract currently allows."""
+        """Whether ``action`` is the one step the contract currently allows.
+
+        This is advice, not enforcement: ``True``/``False`` tells the caller
+        what the contract permits, but nothing in the library blocks a
+        caller that proceeds anyway. ``permits`` is the per-action form of
+        the same fact as ``safe``; enforcement is opt-in via the seams
```
Cyrax321, line 146: Blocking. ``permits`` is not the per-action form of the same fact as ``safe``. ``safe`` is ``mode is RecoveryMode.RESUME`` (line 122), while ``permits`` also returns True when ``action == contract.next_allowed_action`` (lines 140-141). So in any non-RESUME mode that still has a next allowed action, ``permits(...)`` is True while ``safe`` is False, and the two answers genuinely disagree. The rest of this docstring is precise about the advisory contract, and this clause introduces the imprecision it exists to remove. Either describe the disjunction (RESUME permits anything; otherwise only the one next action) or drop the ``safe`` cross-reference and keep the link to the class docstring.

README.md
```diff
@@ -519,9 +519,20 @@ Beyond the original plan: the MCP server, MCP authorization and caller-authentic
...
+hook, the HTTP gateway, the replay guard, or the observation hooks — each documented in
```
Cyrax321, line 533: Non-blocking, style. This paragraph and the new note in docs/recovery_walkthrough.md:147 use em dashes, but every file under docs/ and the README itself contain none before this PR (CHANGELOG.md is the only file in the tree with one). The new guide also uses spaced hyphens in the same commit, so the PR is internally inconsistent as well. Please use the spaced hyphen form already used in enforcement-seams.md, or restructure the sentence, so the prose matches the rest of the docs.

docs/guides/enforcement-seams.md
```diff
@@ -0,0 +1,60 @@
...
+
+This guide is the single place that answers the follow-up question: *if I
+want the verdict enforced, where do I wire it?* There are four opt-in seams.
+Each one refuses work the engine would not permit, at a different boundary.
+Pick by what your agent can bypass.
+
+| Seam | What it intercepts | How to enable | Fails when |
+|:--|:--|:--|:--|
+| Gate host hook | Structured tool calls, before they fire | `continuum hooks install claude-code --with-gate` (also `gemini`, `codex`), or `.continuum/gate.json` for the framework hooks | The call has no live claim, the effect already completed, or the outcome is uncertain |
+| Enforcing HTTP gateway | Outbound HTTP from any language | `continuum gateway --port 8765` with `.continuum/gateway.json` | The request matches a registered route but has no live claim; the proxy settles the claim itself from the real status code |
```
Cyrax321, line 17: Non-blocking. The guide presents `continuum gateway --port 8765` as the gateway's invocation, but `continuum serve --transport http` also defaults to 8765 (src/continuum/cli/main.py:4219). These are two distinct servers sharing one default port, so a user who already runs the sidecar over HTTP gets a bind failure when they follow this row, and nothing here explains the clash. A note that the two default to the same port and that one of them needs `--port` would save that debugging round.

Cyrax321 (changes requested):
Two factual errors block this merge. Both are in the prose the PR adds, which is the only place a change like this can fail; details inline on README.md:532 and src/continuum/recovery/engine.py:146.

Non-blocking, and not anchored because the file is not in this diff: `continuum resume --help` never mentions that its exit code is the enforcement boundary (src/continuum/cli/main.py:3823). Three of the new docs point at that exit code as the one guard that ships enabled, but a user who only runs the CLI never learns it. One sentence in the resume help would carry the contract to the audience that needs it most.

Also non-blocking: no CHANGELOG entry, although `## [Unreleased] > Changed` is open and the repo convention (see 76d9130) is that doc-only work still gets one.

One caveat on the verification claim: the docs guards do not cover this change. test_docs_counts.py asserts documented pytest counts and test_cli_docs.py guards the module-map command count and the subcommand table rows; neither covers the "What CONTINUUM Is Not" table or any guide-page prose, so "docs guards pass" does not exercise the new text. That gap is how the two slips above reached a green CI, and it is the reason this kind of PR still needs a hand read.

Finally, the four seams are now enumerated in five spots: the pre-existing table at README.md:357, the new README paragraph, the new README bullet, the new guide, and the RecoveryDecision docstring. The guide is the right canonical home for it, but the README paragraph restates the list rather than pointing at the guide, so a fifth seam will land in one of them and not the others.

Adhi1-2:
Closing as a duplicate: issue #1031 was closed by merged PR #1047 "docs(recovery): state the advisory verdict contract where a reader can find it" by @Cyrax321, which lands the same change this branch carries.

Verified `gh issue view 1031` is CLOSED and `gh pr view 1047` is MERGED, so this branch has no remaining value and is just queue noise. Deleting the branch to keep the fork tidy.

PR #1040, opened 2026-09-15
docs(recovery): state that the verdict is advisory, and index the enforcement seams

### LINKED ISSUE #1031: recovery verdict is advisory by default: RecoveryDecision.permits has no product caller

Adhi1-2:
Fixes #1031 (Option A + C).

## The decision this PR makes explicit

`RecoveryDecision.permits()` has no product caller, and nothing states that the verdict is **advisory by design**. The gap the issue identifies is an expectation gap, not a missing feature: a user assumes "CONTINUUM will stop my agent from resuming unsafe state" and gets "CONTINUUM will tell you the truth and refuse a duplicate claim."

I deliberately did **not** implement Option B (enforce-by-default with `--force`). The issue's own note: *"Do not implement Option B without maintainer agreement. It changes the default behaviour of every existing caller, and a recovery library that surprises people during a crash is worse than one that is advisory."* Option A+C — document the answer clearly, make the seams discoverable — is what the issue offers as the lowest-risk close, and the docs now state the answer either way.

## What changed

- **`RecoveryDecision` docstring** — `ADVISORY, NOT ENFORCING` up front, phrased after `prefix_trust.py`'s house style ("advisory output never gates"). Names the four enforcement seams and the one built-in enforcement: the CLI exit code (only a verified-safe run exits 0, so `resume && ./start-agent.sh` short-circuits). `permits()` carries the same note.
- **README: new "Enforcement seams" section** — the single discoverable place the issue asks for: a table of the four seams (host gate hooks / HTTP gateway / replay guard / client hooks), what each enforces, and the exact command/API to turn each on. Linked from the nav line.
- **README: "What CONTINUUM Is Not"** — new row: *not a supervisor that stops your process; a library that computes a verifiable verdict; enforcement is the host's decision* (the issue names this section as the right home).
- **`docs/recovery_walkthrough.md`** — "What just happened" now closes with the advisory note and what makes the verdict physical.

## Verification

- All seam details verified against the tree (paths, `hooks install --with-gate`, `continuum gateway`, `protected_call`/`langgraph_protected_node`, fail-closed behaviors).
- README passes `markdownlint-cli2` with the repo's CI config (the four linted files; walkthrough MD040s pre-exist on main and aren't linted by CI).
- `ruff`, `mypy src/continuum`, and the engine tests all green (docstring-only code change).


Cyrax321 (approved):

This PR documents the verdict as advisory and indexes the enforcement seams (#1031, Option A plus C). I reviewed the engine docstrings, README seams table, and walkthrough note.

Choosing documentation over enforcement-by-default respects the issue constraint, and centralizing the four seams with exact turn-on commands is the discoverability fix the issue asks for. Seam paths and behaviors check out against the tree.

One accuracy note for a follow-up: the new permits() wording says it is honoured by the CLI exit code and surfaces, but permits() currently has no product callers and the exit code derives from RecoveryMode. Softening to intended hook not yet wired would keep the advisory statement exact. I do not consider that blocking for these docs.

README.md
```diff
@@ -519,9 +519,38 @@ Beyond the original plan: the MCP server, MCP authorization and caller-authentic
...
 | A RAG system | Verified checkpoints, not retrieval-augmented memory |
 | A workflow engine | A recovery layer, not an orchestrator |
+| A supervisor that stops your process | A library that computes a verifiable verdict; enforcement is the host's decision (see [Enforcement seams](#enforcement-seams)) |
 
 The core abstraction: `semantic state + environment validation + action reconciliation = safe recovery`.
 
+## Enforcement seams
+
+The recovery verdict is **advisory by default**. `RecoveryEngine.assess()` is
+read-only, and `RecoveryDecision.permits()` — the per-action enforcement hook —
```
Cyrax321, line 529: Blocking: em dashes violate repo convention.

AGENTS.md bars U+2014 in docs. Added lines in this file use them at README.md:529, 541, 552. Please use commas, periods, or parentheses with the same meaning across the added Enforcement seams section.

docs/recovery_walkthrough.md
```diff
@@ -153,6 +153,16 @@ docs/guides/liveness-watch.md.
 - No lost rationale: the contract carries its own reason, evidence, and a
   single next allowed action, sealed by hash.
 
+One thing this walkthrough deliberately does not show: the verdict is
+**advisory**. `assess()` is read-only and nothing in the library stops a
+caller that ignores REQUEST_HUMAN and calls `restore()` directly. What makes
+the verdict physical is the surface you read it from: the CLI exit code (only
+a verified-safe run exits 0, so `continuum resume "$RUN" && ./start-agent.sh`
+short-circuits), or one of the four enforcement seams — host gate hooks, the
```
Cyrax321, line 161: Blocking: em dashes violate repo convention.

AGENTS.md bars U+2014 in docs. Added lines use them at recovery_walkthrough.md:161 and 162. Please use commas or parentheses.

src/continuum/recovery/engine.py
```diff
@@ -89,7 +89,18 @@
 
 @dataclass(frozen=True, slots=True)
 class RecoveryDecision:
-    """The engine's verdict, with everything needed to justify it."""
+    """The engine's verdict, with everything needed to justify it.
+
+    ADVISORY, NOT ENFORCING: computing this verdict changes nothing, and
+    nothing in the library stops a caller that ignores it. CONTINUUM is a
+    library, not a supervisor — a caller may legitimately inspect a verdict
```
Cyrax321, line 96: Blocking: accuracy plus convention.

AGENTS.md bars U+2014, used at engine.py:96. More important, the permits docstring says the hook is honoured by the CLI exit code and the README says it is honoured by the surfaces below. permits has zero product callers and the CLI derives from mode via exitcodes, not permits. Since this PR exists to state advisory reality accurately, please soften to intended hook not yet wired (for example available to any seam that wants per step granularity, not yet invoked by a surface).

Cyrax321 (changes requested):
Thank you for stating advisory reality, @Adhi1-2. Direction is right, wording needs tightening.

Ref: #1040 (docs recovery advisory plus seams index)
Surface: README, walkthrough, RecoveryDecision docs

Author context: Adhi1-2 is a contributor on recovery docs. No durable negative signal.

Change: declare assess read only and permits advisory, index four seams with enable paths, note CLI exit code as the one always on enforcement. No behavior change.

Cause: docs and docstrings read as enforcing while the library computes only. Confidence clear from engine read.

Provenance: advisory work from #1031. No other open PR touches this section.

Best fix: mostly yes. Four seams exist as files (gate, gateway, replayguard, clienthooks) and the enable commands match. Advisory framing with override legitimacy is correct.

Refactor: no.

Proof:
- Seam files verified present. Diff intent verified. CI shows no failures.
- Blocking: six added lines use characters barred by AGENTS.md (README 529, 541, 552, walkthrough 161, 162, engine 96). See inline comments.
- Blocking accuracy: permits is described as honoured, but it has zero product callers and the CLI uses mode, not permits. Please soften to intended hook as detailed inline at engine.py:96.

Risk: low once wording is accurate. The seams table itself is a useful index.

Marking REQUEST_CHANGES for convention plus accuracy. Thank you again for making advisory explicit.

Adhi1-2:
Closing as a duplicate: issue #1031 was closed by merged PR #1047 "docs(recovery): state the advisory verdict contract where a reader can find it" by @Cyrax321, which lands the same change this branch carries.

Verified `gh issue view 1031` is CLOSED and `gh pr view 1047` is MERGED, so this branch has no remaining value and is just queue noise. Deleting the branch to keep the fork tidy.

Issue #302, opened 2026-08-24
Novel capability: liveness watchdog - silence as a first-class recovery signal


Cyrax321:
### Affected component

Recovery engine / CLI / serve sidecar (new module proposed: `liveness.py`)

### Summary

Every recovery trigger CONTINUUM ships today is event-driven: a crash, a failed validation, a step-interval revalidation. But the most common way a weeks-long unattended run fails is not an event at all: it is *silence*. A deadlocked tool call, a hung network connection, a wedged worker that never releases its lease, a process stopped by an OOM killer outside our visibility. Nothing in the system notices that nothing is happening. This issue adds a liveness watchdog: wall-clock cadence contracts per run, evaluated continuously, turning sustained silence into a first-class recovery signal.

### Evidence

- Production durable-execution systems treat silence as a primary signal: worker leases are released when a worker stops heartbeating and another worker picks up from the latest checkpoint (LangChain's production deep-agents runtime documents exactly this lease model). CONTINUUM currently has no equivalent at the run level.
- OpenAI's long-horizon safety writeup (*Safety and alignment in an era of long-horizon models*) describes active trajectory monitoring whose core value is the ability to pause a session and alert a human; a pause-on-anomaly path presupposes someone notices anomalies, including non-event anomalies like a stalled run.
- The step-interval periodic revalidation trigger cannot fire if steps stop arriving. Time-based and event-based triggers are complementary by construction; we only ship one of them.
- Empirically, reported agent retry rates of 15-30% make interruption and restart a design regime rather than an accident; every restart is a window in which a resumed process can itself wedge before its next event.
- METR's time-horizon work frames week-long autonomy as the near-term regime (50%-horizons doubling roughly every 7 months); at that duration, mean-time-to-notice is dominated by silence windows, not by detectable failures.

### Proposed design

1. **Cadence contracts.** Per-run configuration declaring the maximum expected wall-clock interval between ledger appends, optionally phase-scoped (for example: 10 minutes while a tool claim is open, 60 minutes otherwise). Stored with run config, editable by operators only.
2. **Watchdog evaluation points (v1, no background process required).** Silence is evaluated whenever anyone looks: `continuum validate`, `continuum resume`, dashboard load, MCP read tools, and the sidecar poll loop all compute `now - last_event_ts` against the contract. This keeps the zero-dependency, host-side posture: the check rides existing read paths.
[...]

PR #1043, opened 2026-09-15
fix(recovery): a liveness breach proposes WAIT once, not twice

### LINKED ISSUE #1042: Duplicated liveness-breach WAIT proposal in RecoveryEngine._decide: rationale and sealed contract reason carry the same sentence twice
## Summary

`RecoveryEngine._decide` appends the liveness-breach WAIT proposal **twice** — the identical `if liveness_advisory is not None and bool(liveness_advisory.get("breached"))` block appears at both `src/continuum/recovery/engine.py:633` and `src/continuum/recovery/engine.py:653` (10 lines apart, same comment, same body). The duplicate is dead weight in the proposals list and, worse, leaks into every user-visible surface: `rationale` is built from *all* proposals matching the winning mode (`engine.py:696`), so a breached run reports the same sentence twice, and the sealed contract's hash-chained `reason` field records the duplication permanently.

## Affected component

Core (recovery engine decision surface)

## Why this matters

The decision itself is unaffected (both duplicates propose WAIT, so `max(proposals, key=SEVERITY)` picks the same mode). What is wrong is everything downstream of the proposal list:

- `RecoveryDecision.rationale` — tuple with the identical string twice.
- `RecoveryContract.reason` — the sealed, hash-chained justification reads `"liveness breach: ...; liveness breach: ..."`. A reader auditing the contract cannot tell whether two distinct breaches were observed or one sentence was pasted twice — and in a component whose entire value proposition is *verifiable provenance of the justification*, that ambiguity is a real defect.
- `continuum resume` / MCP `continuum_resume` `report` and `rationale` fields — operators see the duplicated line.
- The `report` render ("Recovery decision: WAIT / because ...") prints the `because` line twice.

## Evidence: live occurrence

Observed in a real session today (CONTINUUM MCP `continuum_resume` on a run silent for ~33h):

```
"rationale": [
  "liveness breach: silence 118588.8s exceeds threshold 3600s (phase otherwise)",
  "liveness breach: silence 118588.8s exceeds threshold 3600s (phase otherwise)"
],
```

The liveness advisory is computed from wall-clock silence vs. the last event timestamp (`recovery/health.py:advisory_for_storage`), so backdate the last event to force a breach:

```python
import tempfile, os, sqlite3
from datetime import datetime, timedelta, UTC

st = SQLiteStorage(db)
old = (datetime.now(UTC) - timedelta(hours=3)).isoformat()
st.close()
conn = sqlite3.connect(db)   # backdate the last event past the 3600s threshold
conn.execute("UPDATE events SET timestamp = ? WHERE run_id = 'r'", (old,))
conn.commit(); conn.close()

st = SQLiteStorage(db)
print(dec.mode)               # wait  (correct)
print(dec.rationale)          # ('liveness breach: ...', 'liveness breach: ...')  <- duplicated
print(dec.contract.reason)    # 'liveness breach: ...; liveness breach: ...'      <- duplicated, sealed
```

```
wait
('liveness breach: silence 10800.0s exceeds threshold 3600s (phase otherwise)',
 'liveness breach: silence 10800.0s exceeds threshold 3600s (phase otherwise)')
liveness breach: silence 10800.0s exceeds threshold 3600s (phase otherwise); liveness breach: silence 10800.0s exceeds threshold 3600s (phase otherwise)
```

## Root cause

Commit `201219b` ("feat: contract triggering risks", Sep 2) inserted the new risk-trigger block into `_decide` but re-included the entire liveness block that followed it in the diff, leaving two copies. `git log -S` shows the block's text was touched by `201219b`, `923cfff`/`823e5bc` (format revert pair), and `d22bc52` (original, issue #302).

## Expected behaviour

One breach, one proposal, one rationale sentence, one `reason` clause. The second block (at the current line 653) should be removed; the first (at 633, before the risk mapping) is the original position per #302's ordering intent.

## Suggested fix shape

1. Delete the second copy of the block.
2. Add a regression test asserting `len(dec.rationale) == len(set(dec.rationale))` for a breached run (and/or that the WAIT rationale contains exactly one liveness sentence), so a future paste-over can't reintroduce it silently.
3. Optional hardening: since `rationale` is built as `[reason for proposed, reason in proposals if proposed is mode]` (`engine.py:696`), duplicate proposals of the same mode with identical text will always collapse the sentence into the contract twice. A dedup pass (e.g. `dict.fromkeys`) at the rationale build would make the surface robust to this class of bug for any future signal, not just liveness. The block deletion plus the regression test is the minimum; the dedup is a one-line belt-and-braces.

## System information

- OS: macOS 15 (Darwin 25.0.0)
- Python: 3.13
- continuum-agent: from source at upstream/main (`76d9130`), and present since `201219b` (Sep 2) through the current head

## Logs or CI run

Reproduction block above; deterministic, no network, no LLM.

Adhi1-2:
Fixes #1042.

## The bug

`RecoveryEngine._decide` appended the liveness-breach WAIT proposal **twice** — the identical block (same comment, same body) existed at both `engine.py:633` and `engine.py:653`, introduced by commit `201219b` ("feat: contract triggering risks", Sep 2), which inserted the new risk-trigger block but re-included the liveness block that followed it.

The mode was unaffected (both copies propose `WAIT`), but `rationale` is built from *every* proposal matching the winning mode, so every breached run since Sep 2 reported:

```
rationale: ('liveness breach: silence 10800.0s exceeds threshold 3600s (phase otherwise)',
            'liveness breach: silence 10800.0s exceeds threshold 3600s (phase otherwise)')
contract reason: 'liveness breach: ...; liveness breach: ...'
```

The duplicated reason is sealed into the **hash-chained contract**, where a reader auditing the justification cannot tell two observed breaches from one pasted sentence — in the component whose value proposition is verifiable provenance of the justification. It also surfaced live in the MCP `continuum_resume` report (see the issue for a real occurrence from a session silent ~33h).

## The fix

1. **Remove the second copy** of the block. The first (before the risk mapping) is the original #302 position.
2. **Dedup the rationale build** with `dict.fromkeys` (order-preserving), so any future signal that proposes the winning mode twice with identical text reads as one reason — belt and braces for the whole class of paste-over, not just this instance.
3. **Regression test** `test_breach_rationale_states_the_reason_once`: a breached run's rationale has no duplicates, states the liveness reason exactly once, and the sealed contract reason contains exactly one `liveness breach:` clause. **Verified the test fails on the unfixed engine** and passes with the fix.

## Verification

- Full suite: **2217 passed, 25 skipped** (up one: the new regression test)
- `ruff check` / `ruff format` clean
- Bug reproduced deterministically before the fix (backdated event past the 3600s silence threshold), gone after


Cyrax321 (approved):

This PR fixes the liveness breach proposing WAIT twice with a duplicated rationale sentence sealed into the contract (#1042). I reviewed _decide and the new regression test.

Removing the second pasted copy and deduplicating the winning-mode rationale with order-preserving dict.fromkeys is the right shape. It fixes this instance and hardens the whole paste-over class without changing mode selection. The test pins one breach to one sentence in both rationale and sealed reason, and was verified failing on the unfixed engine.

Cyrax321 (approved):
Thank you for the dedup, @Adhi1-2. One breach now reads as one reason.

Ref: #1043 / Regression for #1042 pasted block
Surface: recovery _decide liveness WAIT

Author context: Adhi1-2 is a contributor on recovery. No durable negative signal.

Bug: liveness WAIT block was pasted twice, so a breached run proposed WAIT twice with identical rationale sentences, and the sealed reason joined the duplicate.

Cause: duplicated proposal block. Confidence clear. Worktree retains exactly one liveness proposal at engine.py:640.

Provenance: liveness WAIT from #302. Duplication from #1042. No other open PR touches this block.

Best fix: yes. Remove the duplicate and dedup winning rationale with dict.fromkeys preserving order. Minimal and at the right seam.

Refactor: no.

Proof:
- Worktree tests/test_liveness_events.py, 8 passed, including new rationale once test for mode, unique rationale, and single sealed count.
- Diff has no barred characters. CI has no failures. No prior inline.

Risk: none. Behavior for non breached runs unchanged. Dedup only affects identical sentences for the winning mode.

No blocking issues remain. Marking APPROVED. Thank you again.

Issue #541, opened 2026-08-30
Add Windows PowerShell try-it launcher (try-it.ps1)


Cyrax321:
## Context

`README.md` Quick Start offers zero install paths for macOS and Linux via `uvx`, `docker run` and Codespaces. There is `try-it.sh` for bash, but no equivalent for Windows PowerShell.

Issue #281 tracks the cross platform gap. Windows is the largest contributor pool, but a Windows newcomer who clones the repo has no one command demo. They see `try-it.sh` and have to translate it to PowerShell manually. Stars and forks are 28 and 34 - unlocking Windows would meaningfully grow both.

The fix in `try-it.sh` is simple: `uvx --from git+https://github.com/Cyrax321/CONTINUUM.git continuum --help` or `docker run --rm ghcr.io/cyrax321/continuum`. The same commands work in PowerShell with minor syntax changes.

## How to reproduce

1. On Windows, open PowerShell
2. Try to run `bash try-it.sh` - it fails if bash is not installed
3. Check that `try-it.ps1` does not exist in the repo root

## Expected behavior

Add `try-it.ps1` in the repo root, alongside `try-it.sh`, that does the same demo on PowerShell:

```powershell
uvx --from git+https://github.com/Cyrax321/CONTINUUM.git continuum --help
# fallback to docker if uvx is not available
docker run --rm ghcr.io/cyrax321/continuum --help
```

Update `README.md` Quick Start table to list the Windows path:

| Windows PowerShell | `powershell -ExecutionPolicy Bypass -File try-it.ps1` |

Keep the script friendly for newcomers - add a short comment at the top explaining what it does.

## Files

- `try-it.ps1` (new, copy of `try-it.sh` in PowerShell syntax)
- `README.md` (one row in Quick Start table)

## Acceptance

- `powershell -ExecutionPolicy Bypass -File try-it.ps1` prints `continuum --help` on Windows 11 with uv installed
- `try-it.sh` still works on macOS and Linux
- `ruff` and `markdownlint` pass

## For newcomers

This is a great first issue. You do not need to be a Windows expert - `try-it.sh` is only 20 lines and the PowerShell version is a direct translation. Maintainers will test it on Windows if you cannot.

Amiirhosseini:
Taking this on — opened PR with `try-it.ps1` (PowerShell port of `try-it.sh`) plus README/install docs. Happy to adjust if you'd prefer a different default mode or zero-setup `uvx` fallback in the script itself.

PR #1045, opened 2026-09-15
feat(cli): try-it.ps1 bootstraps its venv, so the Windows demo is one command from a fresh clone

### LINKED ISSUE #281: Cross-platform try-it launcher: Windows has no one-command demo path
## Affected component

CLI

`try-it.sh` (the one-command demo) is bash-only and contains macOS-specific steps (`chflags nohidden`, a symlink hack for uv's `.pth` files under `.venv/lib/python3.14/site-packages`). Windows contributors have no equivalent one-command path to see the crash-recovery demo.

## Proposed solution

Provide a Windows-native equivalent (PowerShell script or a small pure-Python runner invoked via `python tryit.py`) that sets up the venv, installs the dev extra if missing, and runs `examples/crash_recovery_agent.py`.

### Acceptance criteria

- `try-it` demo works on Windows PowerShell from a fresh clone (or a documented `python -m` fallback works everywhere).
- macOS/Linux behavior unchanged; the platform-specific `.pth` workaround stays confined to where it is needed.
- README Quick Start links the cross-platform command.

## Alternatives considered

Telling Windows users to use WSL or Codespaces only; rejected because Codespaces already covers browser use but not local offline evaluation.

## Is this a breaking change?

No

Yes, with guidance

Adhi1-2:
Fixes #281 (good first issue).

## The gap

`try-it.ps1` existed (from the #541 attempt) but never set anything up: it used `.venv` when present and otherwise fell back to bare `python`. On the fresh Windows clone the README's Quick Start row promises, that fallback dies on the first third-party import (`pydantic` via `continuum.models`) — no venv is ever created, nothing is installed. The issue's acceptance criterion ("demo works on Windows PowerShell from a fresh clone") was not met by the script as shipped. `try-it.sh` has the same shape but its users have followed the README's `uv venv` instructions; the issue explicitly asks the Windows launcher to do the setup itself.

## What this PR does

- **`try-it.ps1` bootstraps**: creates `.venv` when missing (`uv venv` when uv is on PATH — the README-recommended tool — else `python -m venv`), then installs `-e ".[dev]"` **only when the venv cannot import `continuum` yet**. Existing environments are untouched; later invocations skip straight to the demo.
- **PowerShell 5.1 hardening**: the import probe runs with `ErrorActionPreference=Continue` — under Windows PowerShell 5.1, a native command writing to stderr while stderr-redirected with EAP=Stop raises a terminating `NativeCommandError`, and a failed import (the expected pre-install state) must not abort the launcher.
- **macOS/Linux unchanged**: `try-it.sh` untouched; the `chflags`/`.pth` workaround stays confined to it (asserted by test — acceptance criterion 2).
- **`tests/test_tryit_launchers.py`** pins the cross-launcher contract: same four modes and default in both scripts, the macOS workaround never leaks into the PowerShell one, and the bootstrap steps are all present. The bootstrap pin **fails on the pre-fix script** (verified).
- **CI**: the Windows test-matrix leg now runs `try-it.ps1 cli --help` from clone conditions (no `.venv`), exercising the real uv bootstrap path on a real Windows runner. The `python -m venv` + pip fallback is covered by the static contract test.

I could not run PowerShell locally (no pwsh/docker on this machine) — the script's runtime behavior is validated by the new CI step; the static contract is validated by the test.

## Verification

- Full suite: **2220 passed, 25 skipped** (one earlier flaky failure re-ran green)
- `ruff check` / `ruff format` clean; `ci.yml` parses
- Bootstrap test verified red on the old script, green on the new


Cyrax321 (approved):

This PR makes try-it.ps1 bootstrap its venv so the Windows demo works from a fresh clone (#281). I reviewed try-it.ps1, the new launcher contract tests, and the Windows CI step.

Creating .venv when missing with uv preferred and python -m venv fallback, installing dev extra only when import continuum fails, and hardening the probe for PowerShell 5.1 EAP behavior is the right shape. Existing environments stay untouched, macOS/Linux behavior is unchanged with the workaround confined to try-it.sh, and the static contract tests plus the real Windows runner step cover both bootstrap paths.

I checked the bot note about a pre-existing venv with base but no dev extras skipping install and then failing test mode. That is a real but minor gap given the leave-existing-env untouched intent, and it does not affect the fresh-clone criterion. A doc line or dev-only probe could be a follow-up.

Cyrax321 (approved):
Thank you for the Windows fix, @Adhi1-2. Fresh clone now works in one command.

Ref: #1045 / Fixes #281 (Windows launcher)
Surface: try-it.ps1 plus CI and launcher contract tests

Author context: Adhi1-2 is a contributor on CLI and bench. No durable negative signal.

Change: try-it.ps1 creates .venv when missing (uv venv preferred, python -m venv fallback) and installs dev extra when import fails. CI adds a Windows step that deletes .venv and runs cli help. README notes the bootstrap. New contract tests pin mode parity, macOS hack confinement, and bootstrap steps statically.

Cause: PowerShell launcher assumed a venv existed, so a fresh clone failed. Confidence clear from script read.

Provenance: try-it launchers for demo. Issue #281 acceptance is fresh clone on Windows with macOS and Linux unchanged. No other open PR touches these launchers.

Best fix: yes. Exit on native failure helper, venv python check (not just Scripts dir), uv then fallback, import probe before install. Static checks are appropriate since Linux runners have no PowerShell. Windows matrix runs the script itself.

Refactor: no.

Proof:
- Worktree tests/test_tryit_launchers.py, 4 passed. Diff has no barred characters. CI shows no failures.
- One prior inline notes import probe checks base package only, so an existing venv without dev extra could miss pytest for test mode. Valid but minor and outside fresh clone scope. Suggest documenting test mode needs dev extra, or probing per mode as follow up.

Risk: low. Existing venvs untouched when import succeeds. New CI step only runs on Windows non docs.

No blocking issues remain. Marking APPROVED. Thank you again.

PR #1046, opened 2026-09-15
fix(recovery): drop duplicated liveness-breach WAIT proposal

### LINKED ISSUE #1042: Duplicated liveness-breach WAIT proposal in RecoveryEngine._decide: rationale and sealed contract reason carry the same sentence twice

Adhi1-2:
## Problem

`RecoveryEngine._decide` carried **two identical copies** of the liveness-breach WAIT proposal block. The second copy was a paste-over introduced in commit 201219b (the first copy — the original block from issue #302, placed before the risk-trigger mapping — is the intended one).

When a run breaches its liveness threshold, both blocks fire, so the proposal list gets the same `(RecoveryMode.WAIT, "liveness breach: silence ...")` tuple twice. Because the rationale build takes *every* proposal of the winning mode, a breached run's output carried the same sentence twice in:

- `RecoveryDecision.rationale` (rendered by `continuum resume` and the MCP report)
- the **sealed, hash-chained contract reason** — an auditor reading the contract could not tell a single breach from a duplicated line

## Fix (per the issue's suggested shape)

1. **Minimum fix:** delete the second copy of the block. The original #302 position (before the risk-trigger mapping) is kept.
2. **Belt and braces:** dedupe at the rationale build with `dict.fromkeys`, which preserves order while collapsing identical texts — so any future paste-over of a winning-mode proposal cannot double-write the rationale or contract reason again.

```python
rationale = tuple(
    dict.fromkeys(reason for proposed, reason in proposals if proposed is mode)
)
```

## Regression test

`test_breach_proposes_wait_once_in_rationale_and_contract` backdates a run's last event 3 hours past the 3600s threshold (the advisory reads wall-clock silence, so an old timestamp is a deterministic breach) and asserts:

- exactly one rationale entry starting with `"liveness breach"`
- the whole rationale is duplicate-free (guards any future signal, not just liveness)
- `"liveness breach"` appears exactly once in the sealed `contract.reason`

Verified the test **fails on unfixed `main`** (AssertionError at the rationale-count assert) and passes with this change.

## Validation

- `ruff format --check` — clean
- `mypy src/continuum` — clean (127 files)
- Full pytest suite — green
- Reproduced with the issue's exact scenario before/after: mode `wait`, single liveness sentence in both rationale and contract reason

Fixes #1042


Adhi1-2:
Closing as a duplicate of #1043 — same root cause (commit 201219b pasted the liveness-breach WAIT block twice into `_decide`), same fix (delete the second copy, dedupe the rationale build with `dict.fromkeys`), both green.

Keeping #1043 because it was filed first and carries the `Fixes #1042` link.

One real hazard worth recording for whoever lands this: upstream still carries **two** copies of the block (`engine.py` lines 671 and 691 at `d10ee63`). Each PR removes one copy and is correct alone, but if both were to merge the second would strip the surviving copy and a liveness breach would no longer propose `WAIT` at all. That is why this duplicate is being closed rather than left for a reviewer to pick.

Either PR resolves it; upstream issue #1152 reports the same defect and would also close with the same change.

Issue #1050, opened 2026-09-15
assess is half archive-aware: liveness, risk and consumed-authority scans read the live tail only


Cyrax321:
## Summary

`RecoveryEngine.assess` is half archive-aware. The confirmation scan and the validator both fold the archived prefix via a shared `read_all_events` fetch (`src/continuum/recovery/engine.py:274`, citing #553 in the comment at `:270`), but three signals computed in the same method still read the live tail only: the liveness breach count at `engine.py:399`, the `RISK_OBSERVED` scan at `engine.py:419`, and the consumed-authorities scan at `engine.py:470`.

`continuum compact` archives the pre-anchor prefix, so all three signals disappear from the sealed contract the moment a run is compacted. A risk that policy maps to `ROLLBACK` stops being proposed and the contract's `triggering_risks` empties; a recorded `LIVENESS_SILENCE_DETECTED` breach count falls to zero; a consumed authority awaiting external probe drops out of the rationale entirely. The engine still returns a verdict, it is just a verdict computed over a deliberately truncated history.

## Affected component

continuum.recovery.engine (liveness, risk, and authority sections of `assess`)

## Severity

Major functionality broken, workaround exists. The contract is the auditable artifact a human or downstream gate acts on, and compaction silently removes the evidence that justified the cautious verdict. The workaround is to never compact a run whose risk or authority state matters, which defeats compaction. The downgrade direction is always toward less caution.

## Evidence

At HEAD `14f8a84`, a run carrying one risk, one consumed authority, and one liveness breach:

```
--- BEFORE compact: request_human
    contract.triggering_risks: ['event_6a5710e5aaa75671f09984b4fcfdd9c8']
    contract.liveness: {'last_append_age': 0.003617, 'breaches': 1, ...}
    rationale: risk governance_decay triggers request_human
    rationale: consumed authority blocks resume: ['cred-1']
compact archived: {'archived': 9}
AFTER compact live RISK_OBSERVED: 0
AFTER compact live AUTHORITY_CONSUMED: 0
AFTER compact live LIVENESS: 0
--- AFTER  compact: request_human
    contract.triggering_risks: []
    contract.liveness: {'last_append_age': 0.000917, 'breaches': 0, ...}
```

The `triggering_risks` list empties and `breaches` drops 1 to 0. Isolating the risk signal with `meltdown`, which policy maps to `ROLLBACK`, makes the downgrade unambiguous:

```
[...]

Issue #554, opened 2026-08-31
[#288] DOT export and compaction survival for provenance graph


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

Parent epic: #288. See execution board #550 for wave context and dependency chain.

### What problem does this solve?

No visual export and no proof the graph survives `continuum compact` (archive rows). Invisible after month-long runs.

### Proposed solution

* `continuum provenance <run> --dot` emits Graphviz DOT with per-node Origin color
* projector queries both `events` and `events_archive` tables, so compacted runs preserve the graph
* digest-auditable: `verify --include-archive` covers caused_by hashes
* test: create graph, compact, reopen, provenance + impact still return same DAG

Depends on: [#288] N-hop propagation.

Files: same projector + cli, `src/continuum/storage/` archive path, tests.

### Alternatives considered

Recompute graph from live log only. Rejected: loses history after compaction, same bug that motivated compaction testing.

### Is this a breaking change?

No

### Would you like to work on this?

Yes, I have a plan

PR #1073, opened 2026-09-15
fix(cli): make the watch liveness episode scan archive-aware (#1072)

### LINKED ISSUE #1072: continuum watch is archive-blind after compaction: duplicate LIVENESS_SILENCE_DETECTED and never-minted LIVENESS_RECOVERED
## Summary

`cmd_watch` decides whether to append `LIVENESS_SILENCE_DETECTED` / `LIVENESS_RECOVERED` by scanning the **live tail only** (`storage.read_events(run_id)` at `src/continuum/cli/main.py:1112`, upstream HEAD `915f1ab`). After `continuum compact` moves a breach episode into `events_archive`, the scan no longer sees the last liveness event, and the episode state machine breaks in both directions:

- **Still-breached run**: `last_liveness` is `None` instead of `LIVENESS_SILENCE_DETECTED`, so watch appends a **second DETECTED for the same episode**. The run's breach history is silently inflated.
- **Recovered run**: `last_liveness` is `None` instead of `LIVENESS_SILENCE_DETECTED`, so the `elif not breached and last_liveness == "LIVENESS_SILENCE_DETECTED"` branch never fires and **`LIVENESS_RECOVERED` is never minted**. The audit trail shows an unterminated breach forever, even though the run demonstrably resumed.

This is the same archive-blindness family as #553 / #554 / #1050, but it lives in `cmd_watch`'s write path, not in `assess` — so fixing #1050 does not fix it. Worse, the two compose: the duplicate DETECTED minted here feeds the (currently live-tail) breach count in `assess`, doubling the same episode even after #1050 makes that count archive-aware.

## Affected component

continuum.cli.main `cmd_watch` (liveness episode state machine); interacts with `SQLiteStorage.compact_run`

## Severity

Major functionality broken, workaround exists. `LIVENESS_*` events are hash-chained audit records — the whole point of #302/#562 is that a breach is an undeniable fact about the run. Compaction currently causes duplicate breach records (false history) and missing recovery records (episode never terminates). Workaround: never run `continuum compact` on a run that has ever breached, which defeats compaction.

## Evidence

At upstream HEAD `915f1ab` (reproduced on a checkout of `76d9130`, the code is identical in this region):

**Symptom 1 — duplicate DETECTED.** A run breaches, watch mints DETECTED, the operator compacts, the run is still silent, watch runs again:

```
DETECTED events before watch: 1
watch rc: 20
DETECTED events after watch:  2
DUPLICATE DETECTED MINTED
```

**Symptom 2 — missing RECOVERED.** A run breaches (DETECTED minted), resumes while watch is not running, the operator compacts, then watch runs (run healthy, no breach):

```
RECOVERED events before watch: 0
watch rc: 0
RECOVERED events after watch:  0
MISSING RECOVERED (episode never terminates)
```

Direct read of the episode state the scan computes vs. the truth after compaction:

```
live events: 2 | all events (archive+live): 6
cmd_watch sees last_liveness (live-only): None
truth (archive-aware):                    LIVENESS_SILENCE_DETECTED
```

```python
import tempfile, sqlite3, argparse, io, sys
from datetime import UTC, datetime, timedelta
from continuum.cli.main import cmd_watch

db = tempfile.mkdtemp() + "/watch.db"
store = SQLiteStorage(db)
rid = "watch-demo"
store.create_run_started(Run(run_id=rid, goal="watch repro"))
old = datetime.now(UTC) - timedelta(hours=3)

store.append_event(rid, EventType.WORK_COMPLETED, {"note": "work 1"})
store.append_event(rid, EventType.LIVENESS_SILENCE_DETECTED,
                   {"silence_seconds": 10800, "threshold_seconds": 3600, "phase": "otherwise"})
store.append_event(rid, EventType.WORK_COMPLETED, {"note": "work 2"})
con = sqlite3.connect(db)
con.execute("UPDATE events SET timestamp=? WHERE run_id=?", (old.isoformat(), rid))
con.commit(); con.close()

print("DETECTED before:", sum(1 for e in store.read_all_events(rid)
                              if e.type is EventType.LIVENESS_SILENCE_DETECTED))
store.compact_run(rid)
con = sqlite3.connect(db)  # keep the live tail old too: run is still silent
con.execute("UPDATE events SET timestamp=? WHERE run_id=?", (old.isoformat(), rid))
con.commit(); con.close()

args = argparse.Namespace(run_id=rid, max_silence="1h", on_breach="exit", webhook_url=None)
cmd_watch(args, store, io.StringIO(), io.StringIO())
print("DETECTED after: ", sum(1 for e in store.read_all_events(rid)
                              if e.type is EventType.LIVENESS_SILENCE_DETECTED))
# prints 2 — the same episode now has two DETECTED records
```

For symptom 2, keep only the pre-compaction events old and let the post-compaction live events carry fresh timestamps; watch then exits 0 with no RECOVERED appended.

## Root cause

The scan at `cli/main.py:1112` arrived with the original watch implementation for #302/#562 and has used `read_events` since. `read_all_events` (which merges `events_archive` with the live tail, base.py) exists and is already the established convention for post-compaction correctness — the DAG command uses exactly the fallback pattern at `cli/main.py:758-762` ("Use read_all_events so compacted runs still show the full DAG (issue #554)"). The liveness episode scan never adopted it. `base.py`'s own invariant — "compaction moves facts but does not revoke their consequences" — is what breaks here.

## Expected behaviour

The episode state machine must consider the full history including the archived prefix: a compacted run whose last liveness event is an archived DETECTED must not mint a second DETECTED for the same episode, and must mint RECOVERED once the run resumes.

## Suggested fix shape

In `cmd_watch`, replace the `read_events` scan with the same archive-aware fetch with fallback the DAG command already uses:

```python
try:
    events = storage.read_all_events(run_id)
except Exception:
    events = storage.read_events(run_id)
```

(If #1050's fix adds a storage-level archive-aware helper, both call sites should converge on it.) A regression test belongs in `tests/test_liveness_events.py`: append DETECTED, `compact_run`, run `cmd_watch` twice more — assert exactly one DETECTED total while breached, and exactly one RECOVERED after the run resumes (both red on current main).

## System information

- OS: macOS 25.0 (Darwin 25.0.0)
- Python: 3.13.5
- continuum-agent: from source at `76d9130` (verified identical at upstream HEAD `915f1ab`)
- Installed via: editable install with dev extra

Adhi1-2:
Fixes #1072

## What

`cmd_watch`'s liveness episode scan read the live tail only (`storage.read_events`), so after `compact_run` moved a breach episode into `events_archive` the last liveness event became invisible and the state machine broke in both directions:

- **still-silent run** → `last_liveness` was `None` instead of `LIVENESS_SILENCE_DETECTED`, so every watch invocation minted **another DETECTED for the same episode** (inflated breach history; also doubles the count `assess` will see once #1050 lands);
- **resumed run** → the `elif not breached and last_liveness == "LIVENESS_SILENCE_DETECTED"` branch never fired, so **`LIVENESS_RECOVERED` was never minted** — an unterminated breach in the hash-chained audit trail forever.

## Change

The scan now uses the archive-aware fetch with the same fallback pattern the DAG command already established for #554:

```python
try:
    events = storage.read_all_events(run_id)
except Exception:
    events = storage.read_events(run_id)
```

One file touched in `src/`; everything else is the two regression tests.

## Tests

Both new tests in `tests/test_liveness_events.py` verified **red on current main** and green with the fix:

- `test_watch_does_not_duplicate_detected_after_compaction` — DETECTED archived by `compact_run`, old live tail, watch runs: exactly one DETECTED total.
- `test_watch_mints_recovered_after_compaction` — DETECTED archived, fresh live tail, watch runs: exactly one RECOVERED minted.

The setup guards that the DETECTED actually lands in the archive (otherwise the tests would pass for the wrong reason), and uses `append_sealed` with explicit timestamps to backdate events rather than raw SQL.

Full check run locally: `ruff check` + `ruff format --check` clean, `mypy src/continuum` clean (127 files), full suite with `-m "not slow"` green.


Cyrax321 (dismissed):

This PR makes the watch liveness episode scan archive-aware (#1072). I reviewed cmd_watch and the new tests in test_liveness_events.py.

Using read_all_events with the read_events fallback, the same pattern the DAG command established for #554, is the right shape. It fixes both directions: no duplicate DETECTED for a still-silent compacted run, and RECOVERED still minted for a resumed one. Both regression tests guard that DETECTED actually lands in the archive and were verified red on main, so they prove the fix.

Note: codecov/patch is at 50 percent on the fallback branches, which is expected for defensive compatibility paths.

src/continuum/cli/main.py
```diff
@@ -1116,9 +1116,17 @@ def cmd_watch(args: argparse.Namespace, storage: Storage, out: Any, err: Any) ->
...
+    # The episode scan walks the full history including the archived prefix
+    # (read_all_events) so compaction cannot break the state machine: without
+    # it, a compacted run whose DETECTED lives in the archive looks like it
+    # never breached, minting a duplicate DETECTED for the same episode and
+    # never minting LIVENESS_RECOVERED (issue #1072).
     try:
-        events = storage.read_events(run_id)
+        try:
+            events = storage.read_all_events(run_id)
+        except Exception:
```
Cyrax321, line 1129: Non-blocking. The fallback is defensible, watch is invoked repeatedly and one unreadable archive page must not kill the episode state machine, and it matches the pattern the DAG command established for #554.

But note what the fallback actually degrades to: `read_events` is the pre-fix scan, so a transient failure here reintroduces the exact bug this PR fixes, silently. A `database is locked` under concurrent access, or any transient error in `read_archived_events`, falls through to the live tail and can mint a duplicate DETECTED for an episode whose marker is in the archive, with no signal that anything was skipped.

The archive-unsupported case and the transient-failure case want different handling: the first is a legitimate permanent limitation, the second is a retryable error being reported as success. Narrowing the catch, or at least emitting a warning to stderr when the fallback fires, would keep a persistent transient failure from quietly corrupting the breach history.

Cyrax321 (approved):
This is the right fix and it's well tested. The two regressions are demonstrated red on main and green with the change, and the setup asserts the DETECTED actually lands in the archive so neither test can pass for the wrong reason. `read_all_events` merges the archived prefix with the live tail by sequence, so the `reversed()` scan still finds the true last liveness event. Approved, with one non-blocking note below.

PR #1081, opened 2026-09-15
fix: reject a reconciler registry missing the probes wrapper

### LINKED ISSUE #1062: load_reconcilers silently returns {} for a registry with no probes key
## Summary

A reconciler registry that maps action types at the top level, instead of nesting them under a `probes` key, is valid JSON and a non-empty file, but `load_reconcilers` returns an empty dict for it without a warning or an error. Every uncertain action is then reported as having no probe registered, and nothing in that message hints that the registry shape is the cause. The module docstring documents the behaviour as intended; the function's own docstring and the runtime signal do not, so an operator with a correctly-written-looking file stays blocked with no path to the mistake.

## Affected component

CLI and integrations (reconciler registry loading)

## Severity

Minor inconvenience or cosmetic issue for the library. The run fails closed, which is the safe direction; the cost is operator time spent on a message that points away from the cause.

## Evidence

`src/continuum/reconcilers.py:91-92` raises `ReconcilerConfigError` only when `raw` is not a dict or `raw.get("probes", {})` is not a dict. A dict with no `probes` key satisfies both conditions, so the probe loop iterates over `raw.get("probes") or {}`, which is empty, and the function returns `{}`. Its own docstring promises "Empty dict when absent; raise when malformed", which is neither of those cases. The module docstring does describe the wrapper as required, but an operator reads the error, not the module.

```python
>>> load_reconcilers(p)   # p holds {"send_invoice": {"command": "check-outbox", "timeout": 5}}
{}
```

Contrast: nesting the same probe under `probes` registers it, and a probe with a non-string `command` raises `ReconcilerConfigError`, so the function is already strict about every shape problem except this one.

```python
import json, tempfile
from pathlib import Path
from continuum.reconcilers import load_reconcilers

p = Path(tempfile.mkdtemp()) / 'reconcilers.json'
p.write_text(json.dumps({'send_invoice': {'command': 'check-outbox', 'timeout': 5}}))
print('load_reconcilers ->', load_reconcilers(p))
```

Output at HEAD `14f8a84`:

```
load_reconcilers -> {}
```

## Expected behaviour

Either the missing wrapper should raise alongside the other shape errors, or `load_reconcilers` should return its empty dict with a warning that names the file and the missing `probes` key, so `skipped_no_probe` on every action has an obvious cause.

## Suggested fix shape

The smallest change that keeps the documented contract is a warning on the empty-but-shaped path: when `raw` is a non-empty dict with no `probes` key, log or return a message naming the resolved path. The strict alternative, raising `ReconcilerConfigError` for a non-empty dict without `probes`, is a breaking change for anyone whose registry is intentionally empty-by-shape, so it should be gated behind the warning first.

## System information

- Python: 3.14.5

InnoxCodes:
## Summary

Fixes #1062.

A reconciler registry file that maps action types at the top level, instead of nesting them under a `probes` key, is valid JSON and a non-empty file — but `load_reconcilers` returned an empty dict for it without a warning or an error. `raw.get("probes", {})` and `raw.get("probes") or {}` both default to `{}` when the `probes` key is entirely absent, so this shape slipped past every existing malformed-input check. Every uncertain action was then reported as having no probe registered, with nothing hinting that the registry shape was the cause.

## Change

- `load_reconcilers` now raises `ReconcilerConfigError` for a non-empty dict missing the `probes` key, naming the top-level keys it actually found so the fix is obvious from the message alone.
- An empty file (`{}`) is unaffected and still loads as a valid empty registry — the new check only fires when the dict is non-empty but has no `probes` key.
- Updated the module docstring, which previously described the old silent-empty-registry behavior as intended.
- Added two tests: one pinning the new error (with the exact repro shape from the issue), one confirming the empty-registry case stays untouched.
- Added a CHANGELOG entry under Unreleased → Fixed.

**Aside, not fixed here:** `gate.py` has the identical `raw.get("tools", {})` / `raw.get("tools") or {}` pattern for its own `tools` wrapper, so it's likely susceptible to the same silent-empty bug for a mis-shaped file. Left out of this PR to stay scoped to the issue; noted in the changelog entry in case it's worth a follow-up.

## Testing

- `pytest tests/test_reconcilers.py -v` — 29 passed.
- `pytest --no-cov` (full suite) — 2222 passed, 27 skipped, 1 failed. The one failure (`test_dashboard_bind.py::test_default_bind_is_loopback`, `OSError: [Errno 48] Address already in use`) is a pre-existing port conflict in my environment, unrelated to this change — it reproduces in isolation on a clean run of just that file, and I never touched `dashboard/`.
- `ruff check src/ tests/` — all checks passed.
- `ruff format --check src/ tests/` — clean.
- `mypy src/continuum` (strict) — no issues in 127 source files.

## AI assistance disclosure

This PR was implemented by Claude (Anthropic's Claude Code) at my direction. I reviewed the diff and the verification above before submitting.

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid non-empty registry files without the required `probes` wrapper now raise a clear configuration error instead of being treated as empty registries.
  - Error messages identify unexpected top-level keys and indicate the required structure.
  - Empty registry files (`{}`) continue to load successfully as valid empty registries.

- **Documentation**
  - Updated configuration documentation and changelog details to reflect the registry validation behavior.


Cyrax321:

This PR fixes #1062. A registry mapping action types at the top level instead of under `probes` loaded silently as empty, leaving every uncertain action with no probe and no signal pointing at the shape error. This change raises `ReconcilerConfigError` naming the top-level keys found, and keeps `{}` as a valid empty registry. Scope is correct, `gate.py` stays out.

Two updates needed:

1. `src/continuum/reconcilers.py`, new error string. Change the punctuation to match repo style. No em dashes in messages:

```
f"{sorted(raw)!r} instead, wrap them under a 'probes' key"
```

2. `tests/test_reconcilers.py`, refusal test. Change `match="probes"` to `match="send_invoice"` to pin the new diagnostic. The current match also passes against the old generic error. Add a one-line docstring to the empty registry test for consistency and coverage.

Once updated, this is ready to merge.

Cyrax321 (changes requested):
Requested changes are outlined in the comments above. Marking this review as changes requested.

InnoxCodes:
Both addressed in 1ce1d98:

1. `reconcilers.py`: dropped the em dash, using a comma now.
2. `test_reconcilers.py`: the refusal test now matches `"send_invoice"` instead of `"probes"` (you're right that `"probes"` also matches the pre-existing generic error, so it wasn't actually pinning the new diagnostic). Added the one-line docstring to the empty-registry test too.

`pytest tests/test_reconcilers.py` (29 passed), `ruff check`, and `ruff format --check` all still clean.

---
Drafted-by: Claude Code (Opus 5)

InnoxCodes:
Heads up on the failing anya-review check, it's not related to this diff. The "Check for prior Anya review on pushes" step errors before doing anything:

gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable.
Process completed with exit code 4.

Looks like GH_TOKEN isn't set in that step's env for the pull_request (synchronize) trigger path. All the checks that actually evaluate this PR (Lint & Type-check, tests, install, build wheel) pass.

---
Drafted-by: Claude Code (Opus 5)

Cyrax321 (dismissed):
Thank you for fixing the silent empty registry case. This correctly rejects a non-empty registry missing the `probes` wrapper with a clear diagnostic.

InnoxCodes:
Addressed CodeRabbit's finding in 07630ad. The three guidance call sites (cli resume, tui recovery view, mcp continuum_resume) all caught every exception around load_reconcilers and fell back to an empty probe list, so the new ReconcilerConfigError from this PR was getting silently absorbed there too. Each now lets it through and surfaces it the way that surface already handles other errors (CLI's existing ValueError handler, an inline error line in the TUI, ToolError for MCP). Added a regression test per site, each confirmed to fail without the corresponding source change.

Full suite still green: 2226 passed, 27 skipped. ruff and mypy clean.

---
Drafted-by: Claude Code (Opus 5)

Issue #1032, opened 2026-09-14
observability metrics API is dead: get_metrics/reset_metrics/collect_from_decision have no caller


Cyrax321:
## Summary

`src/continuum/observability.py` exports a metrics API that nothing uses. Of its public surface, only `render_dashboard` has a product caller, from `src/continuum/cli/main.py:67`. The counters and their collection helpers are dead.

This is the same class of finding as #867 (the unused `observations_evidence_lines` helper that was removed): dead code in a path people read to understand what is measured.

## Affected component

Documentation; Other (observability surface)

## Severity

Minor inconvenience or cosmetic issue.

## System information

- OS: macOS 15.2 (Darwin 25.2.0)
- Python: 3.14.5
- continuum-agent: from source, commit `75b2760` on `origin/main`
- Installed via: `uv pip install -e ".[dev]"`

## The dead symbols

Symbol-level reachability across `src/` and `tests/`:

| Symbol | Callers outside its own module |
|:--|:--|
| `render_dashboard` | `src/continuum/cli/main.py:67` (product) |
| `get_metrics` | `tests/` only |
| `reset_metrics` | `tests/` only |
| `set_metrics` | `tests/` only |
| `collect_from_decision` | `tests/` only |
| `VALIDATIONS_RUN` | `tests/` only |
| `RECOVERIES_RESUMED` | `tests/` only |
| `RECOVERIES_BLOCKED` | `tests/` only |
| `ACTIONS_CLAIMED` | `tests/` only |
| `ACTIONS_COMPLETED` | `tests/` only |
| `UNKNOWN_SIDE_EFFECTS` | `tests/` only |
| `CHECKPOINTS_CREATED` | `tests/` only |

## Command or code to reproduce

```bash
grep -rn "get_metrics\|reset_metrics\|set_metrics\|collect_from_decision" src/
# src/continuum/observability.py only (its own definitions)

grep -rn "continuum.observability\|from continuum import observability" src/
# src/continuum/cli/main.py:67:from continuum.observability import render_dashboard
```

## What a fix looks like

Two directions, and the choice depends on intent.

[...]

PR #1082, opened 2026-09-15
refactor(observability): remove the dead metrics collector API (#1032)

abyyxhek:
observability.py exported a process-wide metrics surface -- Metrics, _Timer, get_metrics/set_metrics/reset_metrics, collect_from_decision and seven counter constants -- that nothing outside the tests ever called. Only render_dashboard has a product caller (the CLI --dashboard path). Dead code in a module people read to understand what is measured misled the next reader into assuming metrics were being collected when they were not.

Removed per Option B in the issue. The recovery ledger is the durable, tamper-evident record of what actually happened and survives the crash, where a process-global counter would not, so 'how many runs resumed and how many blocked' stays answerable. The collector is recoverable from git history if a future observability surface needs it.

## Types of changes

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Removed the unused process-wide observability metrics collector and related public APIs.
  - Recovery outcomes continue to be recorded through the durable recovery ledger.
  - Dashboard rendering remains available for viewing recovery information.

- **Documentation**
  - Updated the changelog and observability documentation to reflect the simplified observability approach.

- **Tests**
  - Updated observability coverage to focus on recovery ledger behavior and dashboard rendering.


abyyxhek:
Closing — superseded. Issue #1032 was resolved the other way by the maintainer's merged #1105 (feat(observability): wire the process-wide counters into the product paths), which gives the collector API real callers. This PR removes that same API, so merging it now would delete wired-in product paths and break the tree. The removal approach is off the table; keeping this open would be misleading.

Issue #394, opened 2026-08-26
feat(mcp): instant resume detection, scoped confirm, and a slim read-only tool subset


Cyrax321:
### Affected component

MCP server

### What problem does this solve?

Every resume pays three taxes that compound across weeks of restarts: the agent must discover and call `continuum resume --json` itself, mutating-tool confirmation is all-or-nothing (a single self-certified component forces full human review even when everything else is trusted), and the full tool surface sits in context whether or not the run needs it. Cold-start latency and token floor both suffer.

### Proposed solution

Three coordinated pieces:

1. Instant detection: write `.continuum/resume.json` on every checkpoint; SessionStart hook reads it out of band and injects an interrupted-run banner only when one exists. Silent otherwise. Target: cold-start banner latency well under a second.
2. Scoped confirm: `confirm --scope <component>` clears self-certified REQUIRES_REVIEW for the named components only (goal, progress), leaving any other uncertainty intact. REVIEW_CONFIRMED gains an optional scope payload; validator clears only what was confirmed. Full confirm remains available.
3. Token floor: `continuum_resume_check` slim tool subset (validate/resume/list_actions) exposed via a minimal registration for contexts that should not carry the mutating surface at all.

COORDINATION CONSTRAINT: this issue touches the confirm flow, which is also the subject of open contributor PR #377. Work MUST build on main's current post-revert behaviour, must not preempt or reproduce #377's diff, and the branch rebasing discipline on the board applies doubly. If #377 merges mid-flight, integrate rather than override.

Acceptance criteria:

- Banner appears only when an interrupted run exists; latency measured and reported in PR.
- Scoped confirm leaves unrelated uncertainty untouched (test: goal confirmed, uncertain side effect still blocks).
- Slim subset lists exactly the read-only trio; mutating calls refuse identically to today.
- Default flows unchanged when features are unused.

Alternatives considered: doing nothing (taxes stay); auto-confirming trusted subsets without an explicit event (rejected: breaks the provenance contract).

### Is this a breaking change?

No

### Would you like to work on this?

No

Cyrax321:
Claiming #394

Cyrax321:
[...]

PR #1084, opened 2026-09-15
fix(tui,dashboard): clear the resume pointer when a run is closed, not just from the CLI

abyyxhek:
## Description

`.continuum/resume.json` is written on **every** checkpoint so a `SessionStart` hook can banner the interrupted run without opening the database (issue #394). A run closed as completed is no longer interrupted, so the pointer must not keep naming it — but only `continuum complete` actually removed the file.

The TUI's and the dashboard HITL button's `complete_run` both document themselves as mirroring `continuum complete` ("the same pairing as `continuum complete`" / "mirroring cmd_complete"), and neither performed the cleanup. So completing a run from either surface left the pointer in place, and the next session's banner surfaced a run the operator had already finished.

The cleanup is now one exported helper, `continuum.checkpoint.clear_resume_pointer(run_id)`, sitting next to `_write_resume_json` in the module that owns the pointer. All three completion paths route through it:

- `continuum/cli/main.py::cmd_complete` — the inline block it already had is replaced by the call (behaviour unchanged).
- `continuum/tui/model.py::complete_run` — now calls it (the bug).
- `continuum/dashboard/hitl.py::complete_run` — now calls it (the same bug).

Semantics are deliberately narrow: a pointer naming a **different** run is left alone, and a missing, unparseable, or undeletable file is not an error — the pointer is a cache, and failing its removal would cost the completion itself.

## Related issues

Self-discovered bug; no issue exists for it (searched open and closed issues for resume pointer / resume.json / stale banner / instant resume — nothing). It sits in the #394 instant-resume feature area. `Refs #394`.

## Types of changes

- Type: `bugfix`

## Breaking changes

No. The new helper is additive; the three call sites behave as the CLI already did.

## How has this been tested?

Reproduced first, then fixed, then guarded:

- **Reproduction before the fix:** checkpoint a run (writes `resume.json` naming `r1`), then `model.complete_run(storage, "r1")` from the TUI — the file survived, still naming `r1`. The same held for the dashboard path. After the fix both remove it.
- **New tests** — `tests/test_resume_pointer.py` (7 tests): the helper's contract (no file → no error; pointer naming another run left in place; matching pointer removed; truncated/non-JSON file tolerated) plus a parametrized test driving the real `continuum complete` command, the TUI verb, and the dashboard verb, asserting the pointer is gone after each.
- **Guard verified:** with the two `clear_resume_pointer` calls temporarily removed, exactly the TUI and dashboard cases of that parametrized test fail (`assert not _POINTER.exists()`), and pass again once restored. The CLI case continued to pass, confirming the fix is what the tests pin.
- **Regression scope:** `pytest tests -m "not slow"` green on the full fast suite (Windows 11, Python 3.11 from the repo venv), including `test_tui.py`, `test_dashboard.py`, `test_dashboard_hitl.py`, `test_cli.py`, `test_mcp_instant_resume.py`, `test_checkpoint.py`, and the `__all__`/docs-count pinning tests.
- **Lint/types:** `ruff check` clean and `ruff format --check` clean on all touched files; `mypy src/continuum/checkpoint/manager.py src/continuum/tui/model.py src/continuum/dashboard/hitl.py` reports no issues.

## Checklist

Tests added or updated for the changed behaviour — yes, `tests/test_resume_pointer.py`, verified to fail without the fix. Documentation updated where the change affects user-facing behaviour — yes, a `clear_resume_pointer` section in `docs/api/checkpoints.md` and a `### Fixed` CHANGELOG entry. CI passes on the latest commit — to be confirmed by CI. Security-relevant changes state known limitations explicitly — not security-relevant; the fail-soft behaviour around the pointer file is documented in the docstring and the docs section.

## Additional context

The existing `test_banner_appears_only_when_interrupted_run_exists` in `tests/test_mcp_instant_resume.py` already asserted "file should be gone or not refer to r1" with an `if resume.exists():` guard, so it tolerated the bug; it passes unchanged here.

One adjacent divergence I deliberately did **not** touch, to keep this PR focused: the dashboard's `complete_run` appends only `RUN_COMPLETED`, whereas the CLI and TUI also append `REVIEW_CONFIRMED` first (both `Origin.HUMAN`, to clear self-certification gates). The dashboard docstring claims "the same pairing as `continuum complete`", which that gap contradicts — worth a separate issue/PR if a maintainer agrees it matters.

## Summary by CodeRabbit

* **Bug Fixes**
  * Completing a run through the CLI, TUI, or dashboard now clears its stale resume pointer.
  * Finished runs no longer appear as active or interrupted in the next session.
  * Resume-pointer cleanup safely handles missing, invalid, or unreadable data without interrupting completion.
  * The TUI now displays database incompatibility details instead of a generic error.
  * Malformed reconciler configuration errors are now reported instead of silently ignored.

* **Documentation**
  * Added API documentation for resume-pointer cleanup behavior.
  * Updated documented test-suite counts across project references.


src/continuum/cli/main.py
```diff
@@ -1809,15 +1814,8 @@ def cmd_complete(args: argparse.Namespace, storage: Storage, out: Any, err: Any)
...
-    try:
-        resume_path = Path(".continuum/resume.json")
-        if resume_path.exists():
-            data = json.loads(resume_path.read_text(encoding="utf-8"))
-            if data.get("run_id") == args.run_id:
-                resume_path.unlink()
-    except Exception:
-        pass
+    # is no longer interrupted, so the pointer must not keep naming it.
+    clear_resume_pointer(args.run_id)
```
Cyrax321, line 1821: Blocking: idempotent retry leaves the stale pointer behind.

cmd_complete has an already completed early return about 30 lines above this cleanup (run.status is COMPLETED returns OK). That branch returns before reaching this clear call. A run completed from the TUI or dashboard before this fix left a stale pointer naming it. Retrying complete from the CLI then reports already completed without clearing, so the next SessionStart banner still surfaces finished work.

This gaps the every path guarantee because the retry is a completion path.

Please call clear_resume_pointer(args.run_id) before that early return as well. No change needed on this line itself, it is already correct.

Cyrax321 (changes requested):
Thank you for unifying the cleanup, @abyyxhek. One path still bypasses it.

Ref: #1084 / Fixes #394 (resume pointer)
Surface: checkpoint resume pointer plus CLI, TUI, dashboard completion

Author context: abyyxhek is a contributor on checkpoint and recovery. No durable negative signal.

Bug: .continuum/resume.json banners the interrupted run. Only continuum complete removed it. TUI and dashboard complete_run claimed to mirror the CLI and did not, so finishing from either left the next session surfacing completed work.

Cause: three completion seams with only one clearing. Confidence clear. CLI cleared inline, TUI model and dashboard hitl did not.

Provenance: resume pointer from SessionStart work. Issue #394 tracks completion parity. No other open PR touches these three seams together.

Best fix: mostly yes. Single helper clear_resume_pointer with tolerant semantics (other run left alone, missing or unreadable or non object JSON tolerated, unlink failure swallowed) routed through all three paths is the right ownership. Docs and CHANGELOG updated. Tests pin helper plus all three paths, 12 passed in worktree.

Refactor: no. Helper in checkpoint manager is the correct seam.

Proof:
- Worktree at 9496b67: tests/test_resume_pointer.py, 12 passed.
- Prior non object JSON crash is already fixed by the second commit (isinstance dict check at manager.py:434). That inline is addressed.
- Diff has no barred characters. CI shows only anya-review config FAILURE.
- Race note on concurrent checkpoint replace between read and unlink is real but rare and banner only. Worth a follow up with a lock, not required here.

Risk:
- Blocking: already completed retry in cmd_complete returns before cleanup (main.py:1786). See inline. A stale pointer from before the fix survives a CLI retry. One line fixes the every path claim.
- Once that branch clears too, this is ready.

Marking REQUEST_CHANGES only for the idempotent path. Thank you again for making completion consistent.

Cyrax321 (dismissed):
The already-completed branch now clears the pointer before its early return, which was the one path still bypassing the helper, so this is approved.

The retry regression test is the right shape: it writes a stale pointer by hand, retries complete on a finished run, and asserts the file is gone. I confirmed it fails without the fix line and passes with it. Replacing the duplicated inline cleanup in cmd_complete with clear_resume_pointer also removes the copy that had no isinstance guard, so the tolerant parsing now covers all three completion surfaces. The docs-count re-sync in 15cc7175 is numbers only, same shape as 76d9130d.

Merging once CI settles.

PR #1085, opened 2026-09-15
fix: error when reconcilers.json omits probes wrapper

### LINKED ISSUE #1062: load_reconcilers silently returns {} for a registry with no probes key

wized2:
## Summary
Fixes #1062.

A registry that maps action types at the top level (no `probes` key) used to load as an empty dict with no warning. Uncertain actions then looked "unregistered" with no hint that the file shape was wrong.

`load_reconcilers` now raises `ReconcilerConfigError` when `probes` is missing, and adds a short hint when top-level values look like probe specs (`command` string present).

## Test plan
- [ ] Registry with `{"probes": {...}}` still loads
- [ ] Registry with top-level probe map raises with a clear message
- [ ] Missing file still returns `{}`

## Summary by CodeRabbit

- **Bug Fixes**
  - Configuration loading now reports clear errors when the configuration has an invalid structure.
  - Missing probe definitions are reported instead of being silently treated as an empty configuration.
  - Invalid probe configuration values are identified more accurately, helping users resolve setup issues faster.


src/continuum/reconcilers.py
```diff
@@ -88,10 +88,24 @@ def load_reconcilers(path: Path) -> dict[str, dict[str, Any]]:
...
+        # registry, and every uncertain action then looks "unregistered" (#1062).
+        looks_like_probes = any(
+            isinstance(v, dict) and isinstance(v.get("command"), str) for v in raw.values()
+        )
+        hint = (
+            " top-level keys look like probe specs — wrap them under a 'probes' object"
+            if looks_like_probes
+            else ""
+        )
+        raise ReconcilerConfigError(f"{location}: expected {{'probes': {{...}}}}{hint}")
```
Cyrax321, line 104: This branch is superseded by #1081 which is already merged, and this hunk breaks two invariants that #1081 pinned.

1. Empty {} must stay valid. Main guards with `if raw and "probes" not in raw` so test_an_empty_registry_file_is_still_a_valid_empty_registry passes. This `if "probes" not in raw` raises for {}, which is why the matrix is red.
2. The merged message names the top-level keys and the test matches on them. This hint text never contains the key name, so test_probes_not_wrapped_under_the_probes_key_is_refused fails even on the raise path.

Please rebase onto main and close this in favor of #1081. If the probe-spec heuristic is still wanted, propose it as a follow-up on top of #1081 that keeps the empty guard and the key listing.

Cyrax321 (changes requested):

The problem statement is correct, and this did raise for a missing probes wrapper with a helpful hint. I checked the diff at 9080ea1, the issue, the loader on main, the two registry tests, and the CI rollup.

What blocks this as written:

1. Issue #1062 is now closed by #1081, merged as 579376d. Main already refuses a non-empty missing wrapper, keeps {} as a valid empty registry, and pins both with tests. This PR was branched before that merge and conflicts with it.
2. src/continuum/reconcilers.py:93 raises for {} because it lacks the `if raw and` guard. That breaks test_an_empty_registry_file_is_still_a_valid_empty_registry and the settle path for an empty placeholder. Gitar flagged the same edge case and it is valid.
3. The hint text here does not name the top-level keys, so test_probes_not_wrapped_under_the_probes_key_is_refused fails. CI shows 10 failing checks across the matrix for these two reasons, with patch coverage at 12.5 percent and no new tests in the PR.

Suggested path: close this as superseded by #1081, or rebase and withdraw the loader change. If you want the looks-like-probes heuristic kept, please re-propose it on top of current main while preserving the empty dict guard and the key listing, with tests for both the {} case and the named-key message.

Thank you again for picking up #1062.

Cyrax321:
Thank you for contributing to Continuum. Closing this as superseded by #1081 (merged as 579376d), which already refuses a non-empty missing probes wrapper, keeps {} as a valid empty registry, and pins both with tests. As noted in the prior review, this branch predates that merge: it raises for {} (breaking the empty-registry test and settle path) and its hint does not name the top-level keys. If the looks-like-probes heuristic is still wanted, please re-propose it on top of current main preserving the empty-dict guard and key listing, with tests for both. Thank you again for picking up #1062.

PR #1086, opened 2026-09-15
fix: wire collect_from_decision into CLI dashboard path

wized2:
## Summary
Advances #1032.

`collect_from_decision` / `get_metrics` were only exercised by unit tests. The CLI recovery dashboard path now calls `collect_from_decision(decision)` immediately before `render_dashboard`, so the process-wide metrics collector reflects the recovery that produced the printed dashboard.

No public API removed; product path now uses the collector.

## Test plan
- [ ] Existing `tests/test_observability.py` still pass
- [ ] CLI recovery dashboard still prints


Cyrax321 (changes requested):
This is a clean step toward #1032. Wiring collect_from_decision into validate --dashboard means the in-process collector now reflects the recovery behind the printed dashboard, with output and exit code unchanged. I verified this directly: base leaves counters empty while this branch records the validation count plus the resume and gauge values.

Thank you @wized2 for picking this up and keeping the scope tight. Leaving the wider Option A work (event-site counters, an operator-readable surface, HTML path) as follow-ups is the right call.

One request before merge: add a small regression test that resets metrics, runs validate --dashboard, and asserts the counter. The existing dashboard test covers the lines without proving the wiring, and the new call updates process-global state that later tests inherit.

That regression test is the only remaining item on my side. Once it is added and passing, this is ready for merge.

Cyrax321:
Thank you for contributing to Continuum. I verified this wiring is now covered on main by 81b59b70 (product-paths wiring), which calls collect_from_decision inside RecoveryEngine.assess(). On current main, validate --dashboard already records validations.run == 1 without this extra CLI call. Merging this as-is would increment twice. Closing this as superseded by 81b59b70. The remaining gap is a regression test for validate --dashboard feeding metrics, which will be added separately to guard the engine wiring. Thank you again for picking this up and keeping the scope tight.

PR #1090, opened 2026-09-15
fix(recovery): report a corrupted checkpoint as corrupted, not missing (#1059)

### LINKED ISSUE #1059: a tampered checkpoint is reported as missing; CorruptedRecord is swallowed by both resolvers
## Summary

A tampered or corrupted checkpoint is reported to the operator as though it did not exist. Both checkpoint resolvers swallow every exception while looking up an id, including `CorruptedRecord`, and fall through to their "no checkpoint" error. The genuine finding, that the stored record failed its integrity check, is discarded and replaced with a message that points the operator at a typo or a missing version.

## Affected component

Core (checkpoint rewind, restore, storage)

## Severity

Minor inconvenience or cosmetic issue for the library, with a security-adjacent edge: the tamper-evidence the storage layer raises is the signal an operator most needs, and it is the one signal these paths convert into noise.

## Evidence

`src/continuum/checkpoint/rewind.py:63-68` wraps `storage.get_checkpoint(to)` in `except Exception: pass`, and `src/continuum/recovery/restore.py:61-66` does the same in `_anchor_for`. `src/continuum/storage/sqlite.py:1058-1065` raises `CorruptedRecord` when the body fails validation or its integrity hash does not match. Both resolvers then fall to their "no checkpoint" error.

```
get_checkpoint raises     : CorruptedRecord - checkpoint 'checkpoint_6979c4658f4d77e0325f77f5d9e17502': integrity hash does not match content
rewind.resolve_checkpoint : RewindError: no checkpoint 'checkpoint_6979c4658f4d77e0325f77f5d9e17502' for run 'r'
restore._anchor_for       : ValueError: no checkpoint 'checkpoint_6979c4658f4d77e0325f77f5d9e17502' for run 'r'
```

```python
import json, os, sqlite3, tempfile
from continuum.environment import StaticProvider, capture
from continuum.models import EnvResource
from continuum.checkpoint.rewind import resolve_checkpoint
from continuum.recovery.restore import _anchor_for

st = SQLiteStorage(db)
st.append_event('r', EventType.DEPENDENCY_DECLARED, {'resource': 'dataset', 'version': 'v1'})
env = capture('r', StaticProvider(resources={'dataset': EnvResource(name='dataset', version='v1')}))
cid = CheckpointManager(st).checkpoint('r', environment=env).checkpoint_id
st.close()

conn = sqlite3.connect(db)
body = conn.execute('SELECT body FROM checkpoints WHERE checkpoint_id = ?', (cid,)).fetchone()[0]
obj = json.loads(body)
obj['state']['goal']['description'] = 'TAMPERED GOAL'
conn.execute('UPDATE checkpoints SET body = ? WHERE checkpoint_id = ?', (json.dumps(obj), cid))
conn.commit(); conn.close()

st = SQLiteStorage(db)
for fn, name in ((lambda: st.get_checkpoint(cid)), 'get_checkpoint raises     ',
                 (lambda: resolve_checkpoint(st, 'r', cid), 'rewind.resolve_checkpoint '),
                 (lambda: _anchor_for(st, 'r', cid), 'restore._anchor_for       ')):
    try:
        fn()
    except Exception as exc:
        print(name, ':', type(exc).__name__, '-', exc)
```

Output at HEAD `14f8a84`:

```
get_checkpoint raises     : CorruptedRecord - checkpoint 'checkpoint_6979c4658f4d77e0325f77f5d9e17502': integrity hash does not match content
rewind.resolve_checkpoint : RewindError: no checkpoint 'checkpoint_6979c4658f4d77e0325d9e17502' for run 'r'
restore._anchor_for       : ValueError: no checkpoint 'checkpoint_6979c4658f4d77e0325f77f5d9e17502' for run 'r'
```

## Root cause

Introduced in `a3b065b` ("feat(rewind): atomic dual-state rewind", 2026-08-29, Refs #292), which mirrored the resolve-by-id-then-version ladder from `restore.py`. Both ladders were written to treat "not found" and "unreadable" as the same case.

## Expected behaviour

`CheckpointNotFound` should fall through to the next lookup strategy and finally to the "no checkpoint" error, while `CorruptedRecord` should propagate, or be re-raised as a distinct error naming the corruption, so the operator sees the real cause.

## Suggested fix shape

Narrow the except to the not-found case in both files: catch `CheckpointNotFound` (and `ValueError` from the int parse) and let `CorruptedRecord` escape. If the CLI surface must stay single-exception, wrap it as `RewindError(f"checkpoint {to!r} is corrupted: {exc}")` so the message distinguishes corruption from absence.

## System information

- Python: 3.14.5

abyyxhek:
## Summary

Both checkpoint resolvers wrapped `storage.get_checkpoint` in a bare `except Exception: pass`. A checkpoint whose stored body failed validation, or whose sealed integrity hash no longer matched its content, was therefore silently retried as a version number and finally reported to the operator as a lookup miss — the exact signal the storage layer raises `CorruptedRecord` to communicate, converted into noise that points at a typo or a missing version.

`resolve_checkpoint` (`src/continuum/checkpoint/rewind.py`) and `_anchor_for` (`src/continuum/recovery/restore.py`) now let `CorruptedRecord` through, wrapping it in the same `RewindError`/`ValueError` the resolvers already raise, but naming the corruption and the underlying integrity failure. A genuine lookup miss still falls through to the version and source-sequence strategies exactly as before, so `rewind --to 2` and version/anchor lookups are unaffected.

This is the security-adjacent edge the issue calls out: on a path whose entire purpose is deciding what state is safe to rewind or restore to, the one record the storage layer refuses to vouch for was the one record the resolvers silently skipped past.

## Root cause

`SQLiteStorage.get_checkpoint` raises `CheckpointNotFound` (a `KeyError`) for an unknown id and `CorruptedRecord` for a record that fails validation or its integrity hash. The resolvers caught `Exception` to implement "try as an id, then as a version, then as a sequence" fallthrough, which correctly swallowed `CheckpointNotFound` and incorrectly swallowed `CorruptedRecord`.

## Reproduction

```python
st = SQLiteStorage(db)
st.append_event('r', EventType.DEPENDENCY_DECLARED, {'resource': 'dataset', 'version': 'v1'})
env = capture('r', StaticProvider(resources={'dataset': EnvResource(name='dataset', version='v1')}))
cid = CheckpointManager(st).checkpoint('r', environment=env).checkpoint_id
st.close()
# rewrite the sealed body so its integrity hash no longer matches
conn = sqlite3.connect(db)
obj = json.loads(conn.execute('SELECT body FROM checkpoints WHERE checkpoint_id = ?', (cid,)).fetchone()[0])
obj['state']['goal']['description'] = 'TAMPERED GOAL'
conn.execute('UPDATE checkpoints SET body = ? WHERE checkpoint_id = ?', (json.dumps(obj), cid))
conn.commit(); conn.close()
```

Before:

```
get_checkpoint raises     : CorruptedRecord - checkpoint '...': integrity hash does not match content
rewind.resolve_checkpoint : RewindError: no checkpoint '...' for run 'r'
restore._anchor_for       : ValueError: no checkpoint '...' for run 'r'
```

After:

```
rewind.resolve_checkpoint : RewindError: checkpoint '...' for run 'r' is corrupted and cannot be trusted: ...
restore._anchor_for       : ValueError: checkpoint '...' for run 'r' is corrupted and cannot be trusted: ...
```

The CLI surfaces this through its existing handler (`error: checkpoint '...' for run 'r' is corrupted and cannot be trusted: ...`, exit 1) rather than a traceback.

## Tests

New `tests/test_checkpoint_corruption.py` (5 tests) pins both resolvers:

- the storage layer flags the tampered body as `CorruptedRecord`,
- each resolver reports corruption instead of a missing checkpoint,
- a genuinely missing checkpoint is still reported as missing,
- version and source-sequence lookups still resolve after an id-shaped miss.

## Checklist

- [x] Full suite green (`pytest tests/`)
- [x] `ruff check src/ tests/` and `ruff format --check` clean
- [x] No em dashes (U+2014) in added prose or comments
- [x] CHANGELOG entry under Unreleased > Fixed
- [x] No behaviour change for genuine lookup misses or version/sequence resolution

Fixes #1059


Cyrax321 (approved):
Thank you for the fix, @abyyxhek. Tamper evidence is now surfaced instead of buried.

Ref: #1090 / Fixes #1059
Surface: checkpoint resolvers (rewind and restore anchor)

Author context: abyyxhek is a contributor on recovery and storage integrity. No durable negative signal.

Bug: both resolvers wrapped get_checkpoint in bare except Exception pass. A body that failed validation or its sealed hash raised CorruptedRecord, which was retried as a version then reported as a lookup miss. The operator was pointed at a typo while storage refused to vouch for the record.

Cause: fallthrough for id to version to sequence correctly needs CheckpointNotFound (a KeyError for unknown id) but incorrectly swallowed CorruptedRecord. Confidence clear. Storage base confirms the two types are distinct, and the new test tampers the body to break the sealed hash.

Provenance: resolver fallthrough predates this PR. Made visible by #1059. No other open PR touches these resolvers.

Best fix: yes. Catch CorruptedRecord first and wrap in the same RewindError and ValueError the resolvers already raise, with corruption named and chained. Catch CheckpointNotFound for the miss path so version and sequence fallthrough is unchanged. Order is specific before general. Unexpected storage errors now propagate instead of being misreported as missing, which is the safer direction.

Refactor: no. Minimal change at the ownership boundary.

Proof:
- Worktree at 13caec1: tests/test_checkpoint_corruption.py, 5 passed. Covers tamper flagged at storage, both resolvers naming corruption, genuine miss still missing, version and sequence fallthrough intact.
- Diff has no barred characters. CHANGELOG entry added for the user visible change. No prior inline to process.
- CI: only anya-review FAILURE remains, which is workflow config, not code.

Risk: negligible. Miss path behavior preserved. Corruption path now fails loudly with the id in the message, which is what an operator needs on a rewind decision.

No blocking issues remain. Marking APPROVED. Thank you again for keeping tamper evidence loud.

PR #1092, opened 2026-09-15
fix(adapters): write the sealed RUN_STARTED payload in the LangGraph checkpointer

### LINKED ISSUE #1088: langgraph_store writes RUN_STARTED with a hardcoded empty payload, so every native-checkpointer run is born with a broken hash chain and cannot be projected
### Affected component

Adapters (LangGraph, OpenAI Agents, LangChain, generic)

### Severity

Blocks my use of CONTINUUM entirely

### Regression check

### System information

OS: Windows 11 22H2
Python: 3.13
continuum-agent: 0.1.2 (commit 36edb6f on `main`, editable install `pip install -e ".[dev]"`)
Installed via: from source
langgraph: installed via the `[dev]` extra

```bash
python - <<'PY'
import tempfile, os
from continuum.adapters.langgraph_store import make_continuum_checkpointer

st = SQLiteStorage(os.path.join(tempfile.mkdtemp(), "t.db"))
saver = make_continuum_checkpointer(st)
saver.put({"configurable": {"thread_id": "t1"}},
          {"id": "ck1", "channel_values": {}, "channel_versions": {}, "versions_seen": {}},
          {"step": 1}, {})

for e in st.read_events("lg-t1"):
    print("EVENT", e.sequence, e.type.value, "payload=", dict(e.payload))
print("verify ok:", st.verify_events("lg-t1").ok)

try:
    RecoveryEngine(st).assess("lg-t1")
    print("assess: OK")
except Exception as exc:
    print("assess RAISED:", type(exc).__name__, str(exc)[:110])
PY
```

### Current behaviour

Every run created by the native LangGraph checkpointer is born with a broken
hash chain and cannot be projected:

```
EVENT 1 RUN_STARTED payload= {}
EVENT 2 STATE_CHECKPOINTED payload= {'checkpoint_id': 'ck1', 'via': 'langgraph'}
verify ok: False
assess RAISED: ProjectionError run 'lg-t1' has no goal: the log never recorded RUN_STARTED
```

`verify_events` reports `TAMPERED_CONTENT` at sequence 1 and `BROKEN_CHAIN`
at sequence 2.

### Expected behaviour

The run's `RUN_STARTED` event should carry the payload the sealed `Event` was
built with (`{"goal": "LangGraph thread t1", "total": 0}`), the chain should
verify, and `assess` should return a verdict. The module's own docstring
promises "any LangGraph app gets CONTINUUM's durability" and "the run inherits
everything else CONTINUUM does (gate, reconcilers, contracts, briefing,
human_steps)". None of that is reachable while the chain is broken at birth.

### Root cause

`src/continuum/adapters/langgraph_store.py:107` hardcodes the payload column
in the `RUN_STARTED` insert:

```python
conn.execute(
    "INSERT INTO events(run_id, sequence, event_id, type, timestamp,"
    " payload, causer_event_id, source, prev_hash, hash)"
    " VALUES (?, ?, ?, ?, ?, ?, NULL, ?, ?, ?)",
    (
        event.run_id, event.sequence, event.event_id, event.type.value,
        event.timestamp.isoformat(),
        "{}",                       # <- hardcoded, event.payload discarded
        event.source.value, event.prev_hash, event.hash,
    ),
)
```

The sealed `Event` two blocks above is constructed with the real payload and
its `hash` is computed over that content, so the stored row cannot match.
The sibling `_log_event` at `langgraph_store.py:143` does it correctly with
`json.dumps(dict(event.payload), sort_keys=True)`, which confirms this is a
copy-paste defect rather than a design choice.

Because `Event.source` is included in `content()`, the stored `source` value
is correct but unhashable-with, so the mismatch is silent until verification.

### Impact

The native checkpointer is the seam the module exists to ship. Every thread
it persists:

- fails `continuum verify`, so the tamper-evidence surface reports the run as
  edited when it was not;
- raises `ProjectionError` from `RecoveryEngine.assess`, so `continuum resume`,
  `assess_graph_recovery` and the whole recovery contract are unreachable;
- cannot be checkpointed, forked, or briefed.

Anyone following the documented LangGraph integration path gets a durability
plane that fails its own integrity check on first use.

### Logs or CI run

Repro output above, run against a fresh temp database on the current `main`.
The existing test `tests/test_langgraph_store.py:142` asserts the
`RUN_STARTED` event exists with the right origin but never inspects its
payload and never calls `verify_events`, so CI passes green over the break.

### Suggested solution

Write the payload the same way `_log_event` does:

```python
json.dumps(dict(event.payload), sort_keys=True),
```

A regression test should assert the stored payload round-trips and that
`verify_events` reports `ok` after a `put`.

Adhi1-2:
Fixes #1088

## What was broken

`_ensure_run` in `src/continuum/adapters/langgraph_store.py` inserted the `RUN_STARTED` row with a hardcoded `"{}"` payload, while the `Event.sealed()` hash two blocks above was computed over the real payload (`{"goal": "LangGraph thread ...", "total": 0}`). Every run created by the native checkpointer was therefore born with a broken hash chain:

- `verify_events` reported `TAMPERED_CONTENT` at sequence 1 and `BROKEN_CHAIN` at sequence 2
- `RecoveryEngine.assess` raised `ProjectionError: run ... has no goal`, making resume, `assess_graph_recovery` and the whole recovery contract unreachable

The sibling `_log_event` writes the payload correctly with `json.dumps(dict(event.payload), sort_keys=True)`, confirming this was a copy-paste defect, not a design choice.

## The fix

One line: write the payload in `_ensure_run` the same way `_log_event` does.

## Verification

- The repro script from the issue now produces the expected output:
  - `RUN_STARTED payload= {'goal': 'LangGraph thread t1', 'total': 0}`
  - `verify ok: True`
  - `assess: OK`
- Two new regression tests in `tests/test_langgraph_store.py`:
  - `test_run_started_payload_round_trips_and_chain_verifies` — the stored payload round-trips and `verify_events(...).ok` after a `put`
  - `test_put_run_can_be_projected_by_recovery_engine` — the `ProjectionError` path from the issue no longer raises
- Full test suite, `ruff check`, `ruff format --check` and `mypy` all clean against the changed files

## Summary by CodeRabbit

* **Bug Fixes**
  * Run-start events now retain their actual goal and total values instead of an empty payload.
  * Run verification and recovery assessment now work correctly for these events.


Cyrax321 (approved):

This PR fixes the LangGraph checkpointer writing a hardcoded empty RUN_STARTED payload while sealing over the real one (#1088). I reviewed langgraph_store.py and the new tests.

Writing the payload with json.dumps(dict(event.payload), sort_keys=True), the same way the sibling _log_event does, is the right one-line fix. It restores hash chain integrity and makes the run projectable again. Both regression tests pin the payload roundtrip with chain verify and the recovery assess path, and CI plus Codecov coverage are green.

PR #1106, opened 2026-09-15
fix(cli): write the attest-keygen private key 0600 instead of umask default

### LINKED ISSUE #1056: continuum attest-keygen writes the Ed25519 private key world-readable (0644)

Adhi1-2:
Fixes #1056

## What was broken

`continuum attest-keygen` wrote the Ed25519 private key with `Path.write_text`, which creates files at 0666 masked by umask — 0644 out of the box. The unencrypted PKCS8 PEM was therefore world-readable while the command's own output told the operator to keep it secret. Overwriting a pre-existing 0644 file kept the old mode as well.

## The fix

- Create the private key through `os.open(..., 0o600)` + `os.fdopen`, so it is never wider than 0600 at creation (umask can only narrow it further)
- Follow with `os.chmod(path, 0o600)` inside `contextlib.suppress(OSError)` so a pre-existing wider mode is narrowed on overwrite — the same compromise `save_budgets` in `src/continuum/budgets.py` already makes for filesystems without permission bits
- The public key stays world-readable
- The text output and the JSON payload now state the applied mode (`mode 0o600` / `private_key_mode`), so an operator on a surprising filesystem can see what they got

## Verification

- The repro from the issue: private key lands `0o600`, public key `0o644`, output reports the mode
- A pre-existing `0644` file is narrowed to `0o600` on overwrite
- Two new regression tests in `tests/test_cli.py`:
  - `test_attest_keygen_writes_private_key_owner_only` — fresh-file mode, reported mode, public key readability
  - `test_attest_keygen_narrows_preexisting_wide_mode` — overwrite path


tests/test_cli.py
```diff
@@ -768,6 +769,32 @@ def test_attest_keygen_writes_pem_files(tmp_path: Path) -> None:
...
 
+def test_attest_keygen_writes_private_key_owner_only(tmp_path: Path) -> None:
+    # Issue #1056: write_text created the unencrypted private PEM at 0644,
+    # world-readable under a standard umask. It must land 0600, and the
+    # output states the mode that was applied so a surprising filesystem
+    # is visible to the operator.
+    priv = tmp_path / "signer.pem"
+    code, out, _ = run("attest-keygen", "--out", str(priv))
+    assert code == ExitCode.OK
+    assert stat.S_IMODE(priv.stat().st_mode) == 0o600
```
Cyrax321, line 787: Blocking: this assertion fails on Windows (CI: Test windows-latest 3.11/3.12/3.13 report S_IMODE 0666 instead of 0600). Windows does not honor POSIX owner-only bits the same way, so stat.S_IMODE cannot prove 0600 there.

Please gate both new permission tests on POSIX, for example:

```suggestion
import sys
@pytest.mark.skipif(sys.platform == "win32", reason="POSIX file modes are not enforced on Windows")
def test_attest_keygen_writes_private_key_owner_only(tmp_path: Path) -> None:
```

and the same guard on test_attest_keygen_narrows_preexisting_wide_mode. The src fix itself (os.open 0600 plus chmod narrow) is correct and should stay as-is.

Cyrax321 (changes requested):

This PR fixes attest-keygen writing the unencrypted private PEM world-readable (#1056). The src change is the right shape: os.open 0600 at creation plus chmod narrow on overwrite, public key left readable, applied mode reported in output. That part looks correct.

Blocking: Windows CI fails on all three versions in test_attest_keygen_writes_private_key_owner_only because S_IMODE returns 0666 on Windows. POSIX mode bits cannot prove 0600 there, so the new permission assertions need a Windows guard (skipif or platform-conditional). I left one inline comment with the exact spot. Once the tests are gated and Windows is green, this is approvable. The fix itself does not need to change.

Thank you again, and sorry for the extra round trip on platform coverage.

Adhi1-2:
The Windows guard you asked for is in the pushed commit and all three windows-latest jobs are green, so I rebased onto current `main` and added the missing CHANGELOG entry (the one thing this PR still lacked).

- `posix_only = pytest.mark.skipif(os.name != posix, ...)` on `test_attest_keygen_writes_private_key_owner_only` and `test_attest_keygen_narrows_preexisting_wide_mode` — the two assertions that read `stat.S_IMODE`. Windows reports 0o666 for a writable file no matter what POSIX mode was requested, so the assertions are meaningless there and are skipped rather than loosened.
- The `src` fix is unchanged: `os.open(..., 0o600)` at creation plus `os.chmod` narrow on overwrite, public key left readable, applied mode reported in both the text output and the JSON payload.
- CHANGELOG entry under `### Fixed`, naming the 0666-masked-by-umask cause and the overwrite path.

CI is re-running on the rebased head; the diff against `main` is the same two files plus the CHANGELOG.

Cyrax321:
Thank you for contributing to Continuum. Closing this as a duplicate of #1083, which reported and fixed the same #1056 exposure about 6 hours earlier with the same os.open 0600 plus narrowing shape. Your branch is green and well tested, and this decision is about authorship order, not quality. One idea worth keeping from your version: reporting private_key_mode in the JSON payload for machine consumers. That is welcome as a small follow-up PR if you want it. Thank you again for the work and for helping close a real key exposure.

PR #1107, opened 2026-09-15
fix(tui): connect the dashboard to current core APIs and harden every read

Adhi1-2:
## Summary

Rebuilds the TUI work on top of current `main` so the dashboard uses the core APIs that now exist, not compatibility shims, and hardens every store read.

- **`action_rows`** folds through the ledger's own archive-aware `folded()` view — the same one the dashboard HITL buttons use — so a compacted run still shows its archived unresolved actions, and the key offered is the key that would settle the action.
- **`budget_rows`** shows configured action types even with no recorded attempts, and type-checks payload fields, so the dashboard never hides an allowance the next claim would not have. Attempts are counted over the whole log (archive included), so compaction does not hand a run a fresh budget (#734).
- **`family_lines`** finds children by the `parent_run_id` column, falling back to the legacy metadata convention so records written before the column existed still show their children.
- **Guarded reads throughout the app layer**: a store that opens but cannot be read (deleted/corrupted out of band) degrades to a retryable message instead of a traceback — a broken store is exactly when the operator reaches for the dashboard. The detail view preserves the selected row across an auto-refresh tick, so a keypress settles the action the operator parked on.
- **Bare invocation against a newer-schema database** lands in the TUI splash with the incompatibility visible, instead of the generic storage error — a newer database is never downgraded or silently replaced, and the branded launcher stays usable.
- **Star-import subprocess test** pins `PYTHONPATH` to the tree under test so a bare subprocess cannot resolve a stale site-packages copy of the package.

Supersedes the work previously on `fix/tui-rebase-891` (that branch had diverged 10 commits behind `main`; this is the same change set rebuilt as a single commit on top of current `main`, keeping `main`'s newer `recovery_lines` advisories, `_human_steps` #1062 fix, and `read_all_events` Storage contract).

## Testing

- `ruff check src/ tests/ examples/` ✓
- `ruff format --check` (301 files) ✓
- `mypy src/continuum` (127 files) ✓
- `pytest tests/` — 2235 passed, 25 skipped ✓

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of databases with unsupported schemas, showing a clear incompatibility message while keeping the interactive interface available.
  * Prevented unreadable database views from crashing the interface; errors now appear within the affected screens.
  * Restored recovery-tree and safety checks for legacy child-run relationships.
  * Preserved selected actions and cursor position across dashboard refreshes.
  * Ensured archived actions remain available for resolution.

* **Usability**
  * Dashboard run counts remain available when detailed storage data cannot be read.
  * Configured budget types without recorded attempts now display their full allowance.


Cyrax321 (dismissed):

This PR reconnects the dashboard to current core APIs and hardens every store read. I reviewed app.py, model.py, the cli splash path, and the TUI tests.

Using the ledger folded() view for action_rows, counting budget attempts over the whole log with configured types shown, finding children by parent_run_id with legacy fallback, guarding every store read to a retryable message, preserving the detail cursor across refresh, and surfacing schema incompatibility in the splash are all the right shapes. I verified both bot findings are addressed in the latest commit: the preserved cursor restores only on table tabs, and a recovered run count clears the stale read error.

Notes, none blocking: anya-review reports failure but the log pattern matches the known missing GH_TOKEN infra issue, not this change; codecov/patch is at 80 percent, expected for guarded error branches.

Adhi1-2:
The third review pass flagged two real defects, both now fixed.

## The selection could settle the wrong action

`_selected_action` re-read the store and indexed by line number, while the actions tab renders `model.action_rows` sorted by key. An action arriving between the render and the keypress sorts ahead of the selected one and shifts the list, so the cursor number resolved to a *different* key: `y` queued a reconcile of the action one line away from the one the highlight marked. An auto-refresh tick had the same failure by construction, which is what made it worth catching rather than theoretical.

The tab now keeps the rows it drew, and `_selected_action` reads that snapshot, so the key settled is the key under the highlight. `_refresh_detail` restores the cursor **by key**, following its row to a new line instead of holding the line and marking whatever displaced it, and clears the selection when the key is gone.

## Legacy children were resolved two different ways

This one was mine, introduced by this PR. `family_lines` resolved children through a helper of its own that knew about the pre-column `Run.metadata` convention, while the recovery roll-up and the CLI tree went through `recovery.family.children_of`, which did not. A child recorded before the `parent_run_id` column existed therefore appeared on the tree tab while the aggregate never saw it, so a parent holding an unreconciled side effect could read RESUME, and `resume` could call that family safe.

`children_of` now resolves the convention itself, making it the single resolution for the roll-up, the CLI tree, and the TUI; the TUI's helper just delegates. That also fixes the CLI tree, which had the same blind spot.

## Verification

Both are pinned by tests that fail on the code this replaces, which I confirmed by stashing the fixes and watching them go red:

- `test_family.py` — an uncertain **legacy** child blocks the parent's resume, costs the zero exit, is listed by the tree, and unblocks once settled. Mirrors the existing `test_a_family_blocked_parent_never_exits_zero` shape.
- `test_tui.py` — the tree tab and the recovery verdict both name a legacy child, so the display and the aggregate agree.
- `test_tui.py` — a second action claimed out of band, sorting ahead of the drawn row, then `y` with no re-render in between settles the drawn key and the highlight holds; a refresh follows the selected key to its new line.

Full non-slow suite **2235 passed / 25 skipped**; `ruff check`, `ruff format --check`, and `mypy` clean.

One housekeeping note: my first push of the last commit inadvertently captured local worktree gitlinks under `.claude/`. Nothing left the intended four files in the final commit, and `.claude/` is now in `.gitignore` so it cannot recur. History was force-pushed to correct it.

src/continuum/tui/model.py
```diff
@@ -326,23 +333,34 @@ def family_lines(storage: Storage, run_id: str) -> list[str]:
...
+    dashboard never hides an allowance the next claim would not have. Limits
+    come from the budget registry, the same one the claim sites enforce
+    through ``evaluate_budget``.
+    """
     storage.get_run(run_id)
     try:
         raw = load_budgets(Path(DEFAULT_BUDGETS_PATH))
     except Exception:
-        raw = {}
+        raw = {}  # a broken registry must not take the dashboard down with it
```
Cyrax321, line 346: Non-blocking, but it contradicts the lesson this PR invokes. `recovery_lines` was fixed in #1062 to surface a broken reconciler registry as `error:` (model.py:197-211) instead of silently degrading, and this PR deletes the test that asserted that. Here the same shape goes the other way: an unreadable `budgets.json` falls back to `raw = {}` and the tab renders `max_attempts` and `remaining` from an empty registry as if they were the run's configured allowance.

The regression is now worse than before this PR, because configured action types are read from `raw` too. A store with a typo'd budgets.json loses every configured type from the budget tab and shows default limits, and the operator gets no hint the config is the problem. Keeping the dashboard alive is right; presenting derived numbers as authoritative while the config that produced them is broken is the part to fix. Surfacing the load failure the way `recovery_lines` does, or at least substituting the error for the row values, would close the loop.

src/continuum/tui/app.py
```diff
@@ -74,83 +82,177 @@ def __init__(self, storage: Storage) -> None:
...
             self.cursor = -1  # the runs list marks its selection itself
             if self.rows:
                 self.index = min(self.index, len(self.rows) - 1)
         else:
             self._refresh_detail()
 
+    def _count_runs(self) -> int:
+        """Count runs without assessing recovery for each: a bare list read."""
+        assert self.storage is not None
+        return len(self.storage.list_runs())
```
Cyrax321, line 127: Non-blocking efficiency note. This is strictly cheaper than the old splash, which assessed recovery per run, so the intent is met. But there is no count API in the storage layer, so `list_runs()` materializes every run row and parses each one's metadata JSON just to take a length, and it runs on every idle tick when `--refresh` is set.

For a store with many runs that is still O(rows) of parsing on the cheapest screen. A `count_runs` method backed by `SELECT COUNT(*) FROM runs` would make the splash constant-time; the current call site is the only place that wants a count without the rows, so it's a natural addition.

Cyrax321 (commented):
This does what it says: every read in the app layer is guarded, the splash survives an incompatible database, and the actions tab now settles the key that was drawn rather than a fresh read at the same line. The legacy-child fix is the real one, and it lands in `children_of` where both the tree and `roll_up_children` resolve from, so the verdict and the display can't disagree again. I would merge it.

Nothing below blocks the merge. The first is a consistency issue with a lesson this PR otherwise leans on; the other two are small.

.gitignore
```diff
@@ -37,7 +37,7 @@ htmlcov/
 # Agents / AI tools
 .agents/
 .gemini/
-.claude/settings.local.json
+.claude/
```
Cyrax321, line 40: Question, not a blocker. Nothing under `.claude/` is tracked today, so nothing is lost by broadening the pattern. But the new pattern would silently exclude any shared agent configuration committed later, and it isn't related to the TUI work in this PR. If the intent is only to keep local agent settings out of the tree, the narrower pattern already did that; if the intent is to ignore the whole directory, that's worth a line of intent somewhere since the reason isn't obvious from the diff.

Cyrax321 (dismissed):
Approving. Every read in the app layer is guarded, the splash survives an incompatible database without downgrading it, and the actions tab now settles the key that was drawn instead of a fresh read at the same line. The legacy-child fix is the substantive one: it lands in `children_of`, which both the tree tab and `roll_up_children` resolve from, so the verdict and the display can't diverge again.

The three notes inline are non-blocking. The one I'd most like followed up is the `budget_rows` registry swallow, which is a consistency issue with the #1062 lesson this PR otherwise leans on.

src/continuum/tui/model.py
Cyrax321: Deferred to a follow-up, not blocking this merge. Keeping the dashboard alive on a broken budgets.json is intended; surfacing the load failure the way recovery_lines does is the right next step.

src/continuum/tui/app.py
Cyrax321: Noted for follow-up, not blocking. A count_runs SELECT COUNT(*) would make the splash constant-time; the current shape is already strictly cheaper than the old per-run assess.

.gitignore
Cyrax321: Accepted as local-settings hygiene, not blocking. Nothing shared is tracked under .claude/ today.

Cyrax321 (approved):
Thank you for the contribution. This PR reconnects the TUI dashboard to current core APIs and hardens every store read, including the ledger folded() view for archived actions, whole-log budget counting with configured types shown, parent_run_id child resolution with legacy fallback, guarded reads, cursor preservation by key, and schema-incompatibility surfacing in the splash. I reviewed app.py, model.py, family.py, the CLI splash path, and the TUI tests, and verified the suite locally: 76 passed with ruff check and format clean. The three open notes are addressed above as deferred follow-ups and their threads are resolved. Required CI is green on this head. No blocking issues remain. Marking ready to pull. Thank you again for the contribution.

PR #1110, opened 2026-09-15
fix(environment): verify the digest while snapshotting, refuse a stale one

### LINKED ISSUE #1077: snapshot_file stores bytes under a caller-supplied digest without verifying them, so a content-addressed snapshot can hold content that does not match its name
`continuum.environment.file_snapshot.snapshot_file` accepts a caller-supplied `sha256` and copies the file's current bytes into `snapshot_path(sha256)` without ever checking that those bytes actually hash to that digest. The content-addressed invariant the store is named for is assumed, not enforced.

The production write path computes the digest and performs the copy in two separate reads. `continuum/clienthooks.py:162` hashes the file in bounded chunks and puts the digest in the `TOOL_COMPLETED` payload, then calls `snapshot_file(path, sha256=payload["sha256"])`, which re-reads the file with `shutil.copyfile`. Any writer that touches the file in that window stores the new content under the old digest:

```python
recorded = hashlib.sha256(b"original content v1").hexdigest()
# ... a racing writer appends ...
snapshot_file("f.txt", sha256=recorded)   # returns a Path, stores the appended bytes
hashlib.sha256(snapshot_path(recorded).read_bytes()).hexdigest() != recorded
```

Verified on disk: the snapshot returns a path, and the stored content hashes to something other than the path it is filed under.

The blast radius is bounded by `rewind_to_checkpoint`, which re-digests after `restore_file` and reports a conflict when the restored bytes do not match. So the visible symptom is a rewind that refuses with "after restore digest ... != expected ..." on a run the operator believed was clean, and a snapshot that stays poisoned for every later rewind of that checkpoint. Silent wrong-content restore is prevented only by that downstream check, and the store itself never learns its invariant is broken.

The fix is to verify at write time: digest the bytes being copied (the copy already streams the file, so this is one pass, not an extra read) and refuse or re-key when the result does not equal the supplied `sha256`. Returning `None` on mismatch matches the existing contract for "could not snapshot faithfully". Callers that supply a stale digest then get a snapshot of nothing rather than a snapshot that lies.

The file also has a redundant block worth folding in the same pass: the leading `if sha256 is not None: dst = snapshot_path(sha256); if dst.exists(): return dst` and the `else` branch eight lines later perform the identical existence check twice.

I found this auditing the dual-state rewind path; `tests/test_rewind_dual_state.py` and `tests/test_file_snapshot_edges.py` both supply the digest the caller already computed and never assert the stored content matches it.

Adhi1-2:
Fixes #1077

## What was broken

`snapshot_file` accepted a caller-supplied `sha256` and copied the file's current bytes into `snapshot_path(sha256)` without ever checking that those bytes actually hash to that digest. The production write path (`clienthooks.py`) hashes the file in one bounded-chunk read and then calls `snapshot_file(path, sha256=...)`, which re-read the file with `shutil.copyfile` — so any writer touching the file in that window stored new content under the old digest. The snapshot then held content that did not match its name, and every later rewind of that checkpoint refused with a bogus "after restore digest != expected" conflict.

## The fix

- Digest the bytes **as they are copied** (the copy already streams the file — one pass, not an extra read) and return `None` when the result does not equal the supplied `sha256`. `None` matches the existing contract for "could not snapshot faithfully"; the only caller ignores the return and swallows exceptions, so a racing writer now gets a snapshot of nothing rather than a snapshot that lies
- Fold the duplicate existence check the supplied-digest and computed-digest branches each performed
- The `sha256=None` path now reuses the existing `file_digest` helper

## Verification

- New regression tests in `tests/test_file_snapshot_edges.py`:
  - stored content hashes to the digest it is filed under, and repeat calls deduplicate
  - a stale supplied digest is refused: no entry, no `.tmp` litter
  - digest-free snapshots are still keyed by content
- `tests/test_rewind_dual_state.py` (the downstream consumer) still passes unchanged


Cyrax321 (approved):

This PR fixes snapshot_file filing raced bytes under a stale caller-supplied digest (#1077). I reviewed file_snapshot.py and the new edge tests.

Digesting as the bytes are copied, in one pass, and refusing with None plus tmp cleanup on mismatch is the right shape. It matches the existing could-not-snapshot-faithfully contract, folds the duplicate existence check, and reuses file_digest for the digest-free path. The tests pin verified write with dedup, stale refusal with no entry and no tmp litter, and content-keyed snapshots, and the downstream rewind tests still pass.

PR #1112, opened 2026-09-15
fix(recovery): keep the caller's timestamp in ingest_risk

### LINKED ISSUE #1075: ingest_risk discards the caller's timestamp: ts-only payloads are overwritten with server now, timestamp-only payloads get no ts at all
`continuum.recovery.risk.ingest_risk` builds the `ts` field of the `RISK_OBSERVED` payload with a condition whose branches cannot both be right:

```python
if "ts" not in event_payload and "timestamp" not in payload:
    event_event_payload["ts"] = _now_iso()
elif "ts" in payload:
    event_payload["ts"] = str(payload["ts"])[:64]
```

`"ts" not in event_payload` is always true at that point (the payload built so far only holds trigger/score/detail/episode_id/step_id), so the first branch reduces to `"timestamp" not in payload` and is checked before the caller's own `ts` is ever consulted. Two inputs are mishandled:

A payload carrying `ts` but no `timestamp` takes the first branch and mints `_now_iso()`, so the caller's timestamp is silently replaced by the server's clock.
A payload carrying `timestamp` but no `ts` takes neither branch, so the recorded event has no `ts` field at all and the caller's timestamp is silently dropped.

Repro (against 849c4f2):

```python
ingest_risk(store, "r1", {"trigger": "latency_anomaly", "ts": "2020-01-01T00:00:00+00:00"})
ingest_risk(store, "r1", {"trigger": "token_runaway", "timestamp": "2020-02-02T00:00:00+00:00"})
```

records `ts` as 2026-09-15 (server now) for the first and `None` for the second.

This is a data-fidelity bug in a monitoring signal whose entire purpose is temporal correlation: a probe reporting when it observed the anomaly cannot get its observation time into the log, and the event that lands instead claims CONTINUUM's ingestion time as if it were the observation time. The module is documented as fail-open on malformed input, but these payloads are not malformed, they are simply misread.

The likely intent is to prefer the caller's timestamp and mint only when neither spelling is present:

```python
if "ts" in payload:
    event_payload["ts"] = str(payload["ts"])[:64]
elif "timestamp" in payload:
    event_payload["ts"] = str(payload["timestamp"])[:64]
else:
    event_payload["ts"] = _now_iso()
```

Whether `timestamp` should be honoured at all is a question for the maintainers; if only `ts` is the supported spelling, the current code still needs fixing so that a `ts`-carrying payload keeps it.

I found this by reading `src/continuum/recovery/risk.py:152` while auditing the ingestion paths; no test covers a payload that supplies either spelling.

Adhi1-2:
Fixes #1075

## What was broken

The `ts`-building condition in `ingest_risk` (`src/continuum/recovery/risk.py`) consulted `"timestamp"` before the caller's own `ts`:

```python
if "ts" not in event_payload and "timestamp" not in payload:
    event_payload["ts"] = _now_iso()
elif "ts" in payload:
    event_payload["ts"] = str(payload["ts"])[:64]
```

`"ts" not in event_payload` is always true at that point, so:

- a payload carrying `ts` but no `timestamp` minted `_now_iso()` — the probe's observation time silently replaced by the server's clock
- a payload carrying `timestamp` but no `ts` took neither branch — the event recorded no `ts` at all

Both are data-fidelity bugs in a monitoring signal whose purpose is temporal correlation, and neither payload is malformed.

## The fix

Prefer the caller's clock in either spelling and mint only when neither is present; an explicit `None` value is treated as absent (recording `str(None)` would be worse than minting):

```python
if payload.get("ts") is not None:
    event_payload["ts"] = str(payload["ts"])[:64]
elif payload.get("timestamp") is not None:
    event_payload["ts"] = str(payload["timestamp"])[:64]
else:
    event_payload["ts"] = _now_iso()
```

## Verification

New regression tests in `tests/test_risk_policy.py`:
- `test_ingest_risk_keeps_caller_supplied_ts` — the `ts` spelling survives unchanged
- `test_ingest_risk_honours_timestamp_spelling` — the `timestamp` spelling lands in `ts`
- `test_ingest_risk_mints_ts_only_when_caller_supplies_none` — server time only when the caller supplied none

Full test suite, `ruff check`, `ruff format --check` and `mypy` all clean.


Cyrax321 (approved):

This PR fixes ingest_risk silently replacing the caller's ts with server now, and dropping the timestamp spelling entirely (#1075). I reviewed risk.py and the new tests in test_risk_policy.py.

Preferring the caller's clock in either spelling and minting only when neither is present is the right shape, and treating explicit None as absent avoids recording str(None). The three regression tests pin each branch, and CI plus Codecov coverage are green.

PR #1117, opened 2026-09-15
fix(cli): stop the briefing banner advertising a run that is not in the database

### LINKED ISSUE #1063: continuum briefing banner advertises a run that is not in the database
## Summary

`continuum briefing` injects an "Interrupted run ... resume pending" banner straight from `.continuum/resume.json` without checking that the run it names still exists. The run-id resolution path a few lines above does check, but the banner does not use its answer, so a stale file advertises a run that is absent from the database while the briefing body beneath it covers a different, live run. An operator following the banner runs `continuum resume <ghost>` and gets a not-found error for their trouble.

## Affected component

CLI (`continuum briefing`, SessionStart hook path)

## Severity

Minor inconvenience or cosmetic issue for the library. The briefing is advisory and read-only, so nothing unsafe happens; the defect is a confidently-stated instruction that cannot succeed.

## Evidence

`src/continuum/cli/main.py:2351-2355` contains the comment "Verify the run is still active (not completed) before surfacing" followed immediately by `pass`, so the verification is a stub. The `run_id` resolution path at `:2366-2373` does check `storage.get_run(candidate)` before accepting it, falling back to `storage.get_active_run()` when the check fails. The banner block at `:2418-2427` then re-reads the same file and emits its line for `_banner_run` with no check at all, using a different variable from the one that was validated.

```
Interrupted run ghost – resume pending
  run: continuum resume ghost --json

CONTINUUM active run: real
recovery verdict (verified, sealed contract)
  recovery verdict: resume (safe=True)
```

(`ghost` is not in the database; `real` is the live run.)

```bash
cd "$(mktemp -d)" && mkdir -p .continuum
printf '{"run_id": "ghost"}' > .continuum/resume.json
uv run python -c "
import os
from continuum.cli.main import cmd_briefing

db = 't.db'
st = SQLiteStorage(db)
st.create_run(Run(run_id='real', goal='live run'))
st.append_event('real', EventType.RUN_STARTED, {'goal': 'live run'})
st.close()

class A:
    run_id = None
    json = False
    hook_event_name = 'SessionStart'
    raw_summary = False
import io
out, err = io.StringIO(), io.StringIO()
cmd_briefing(A(), SQLiteStorage(db), out, err)
print(out.getvalue().split('semantic state')[0])
"
```

Output at HEAD `14f8a84`, with no run named `ghost` anywhere in the database:

```
Interrupted run ghost – resume pending
  run: continuum resume ghost --json

CONTINUUM active run: real
recovery verdict (verified, sealed contract)
  recovery verdict: resume (safe=True)
```

## Root cause

The banner was added for the instant-detection path in #394 as an out-of-band injection before the full briefing, so it was written to trust the file the checkpoint wrote. The existence check was left as a comment with a `pass` body and never wired.

## Expected behaviour

The banner should only appear for a run the database still holds and that is not completed, and it should name the same run the briefing below it covers.

## Suggested fix shape

Replace the `pass` with the check the comment describes, set the validated run id once, and have the banner block reuse that value instead of re-reading the file into `_banner_run`. When the file names a run that is gone, either omit the banner or print a one-line note that the interrupted run is no longer recoverable, which is more useful than a resume command that cannot work.

## System information

- Python: 3.14.5

Adhi1-2:
Fixes #1063

## What was broken

`continuum briefing` injected the "Interrupted run … resume pending" banner straight from `.continuum/resume.json` without checking that the run it names still exists — the existence check was a comment over a bare `pass`, and the banner block re-read the file into its own unvalidated variable (`_banner_run`) instead of reusing the answer the run-id resolution path a few lines above had already computed. A stale file advertised a ghost run while the briefing body beneath it covered a different, live run, and an operator following the banner got a not-found error.

## The fix

- Read `resume.json` **once**, validate the run it names (exists in the database, not completed), and reuse that answer for both run-id resolution and the banner
- The banner now names the same run the briefing below it actually covers, and only for a run the database still holds and that is not completed
- When the file names a run that is gone, the banner is replaced by a one-line note (`Interrupted run <id> is no longer in the database`) — more useful than a resume command that cannot work, as the issue suggests
- Corrupt/non-dict files and non-string run ids fall through to the normal briefing, as before

## Verification

- New regression tests in `tests/test_mcp_instant_resume.py`:
  - `test_banner_notes_a_run_not_in_the_database_instead_of_advertising_it` — the issue's exact ghost-run scenario: note instead of banner, briefing covers the live run
  - `test_banner_omitted_when_interrupted_run_is_completed` — a completed run has no pending interruption to surface
- The existing banner tests (appearance, latency, silence-without-file) still pass
- Full test suite, `ruff check`, `ruff format --check` and `mypy` clean (one pre-existing `no-untyped-def` on an unrelated test)

**Pre-existing flakiness noticed while verifying (not addressed here):** `tests/test_briefing_curation.py` has three tests that only pass when a leftover `.continuum/resume.json` naming an existing run sits in the cwd — they fail on pristine `main` when the file is absent, because the `db` fixture in `tests/test_cli.py` checkpoints relative to cwd. Worth a separate issue for test isolation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resume briefing behavior for completed or unavailable runs.
  - Stale resume information now displays a clear notice instead of an unusable resume command.
  - Malformed resume files no longer prevent the active run from being briefed.
  - Session start remains silent when no resume information is available.


Cyrax321 (dismissed):

This PR stops the briefing banner advertising a ghost run (#1063). I reviewed cmd_briefing in cli/main.py and the new tests in test_mcp_instant_resume.py.

Reading resume.json once, validating existence and completion via get_run, and reusing that answer for both run-id resolution and the banner is the right shape. It removes the duplicate unvalidated read, guarantees the banner names the run the briefing actually covers, replaces the ghost banner with a useful note, and preserves the corrupt-file fallthrough. Both new tests pin the ghost and completed cases, and existing banner tests still pass.

One note: codecov/patch reports 75 percent on this PR, so some defensive branches may lack coverage. I do not consider that blocking given the two regression tests cover the main paths.

src/continuum/cli/main.py
```diff
@@ -2346,40 +2346,44 @@ def cmd_briefing(args: argparse.Namespace, storage: Storage, out: Any, err: Any)
...
         except Exception:
             # Corrupt file is not a blocker; fall through to normal briefing
             # which will do the DB check and report correctly.
-            pass
+            resume_data = None
+        if isinstance(resume_data, dict):
+            candidate = resume_data.get("run_id")
+            if isinstance(candidate, str) and candidate:
+                try:
+                    resume_run_completed = storage.get_run(candidate).status is RunStatus.COMPLETED
```
Cyrax321, line 2375: Non-blocking. `get_run` raises `RunNotFound` for a row that was never created and that is the only missing-row signal it produces (base.py:246). A bare `except Exception` also catches `sqlite3.OperationalError: database is locked` under concurrent access, or a Postgres connection blip, and reports it to the operator as `Interrupted run X is no longer in the database`.

That is a false claim of data loss, and it also withdraws the resume command for a run that is intact. The corrupt-file handler above has the correct instinct for the same reason: a transient failure should degrade to the normal briefing, which does its own DB check and reports accurately.

`RunNotFound` is already imported at main.py:84 and used at :270, so narrowing this to `except RunNotFound:` is a one-word change that keeps the ghost-run path intact.

src/continuum/cli/main.py
```diff
@@ -2420,20 +2424,19 @@ def cmd_briefing(args: argparse.Namespace, storage: Storage, out: Any, err: Any)
...
+    # file named so the SessionStart hook points the agent at resume without
+    # it having to discover the command itself. Only for a run the database
+    # still holds, that is not completed, and that this briefing actually
+    # covers; a file naming a run that is gone gets a note instead of a
+    # resume command that cannot work (issue #1063).
+    if resume_run and not resume_run_completed and run_id == resume_run:
+        lines.append(f"Interrupted run {resume_run} – resume pending")
+        lines.append(f"  run: continuum resume {resume_run} --json")
+        lines.append("")
+    elif unrecoverable_run:
```
Cyrax321, line 2437: Non-blocking. This note also fires when the caller passed an explicit `--run-id`, and on a `SessionEnd` hook: the early return at :2350 only covers `SessionStart` with no resume file, so any other path with a stale `.continuum/resume.json` sitting in the cwd reaches this branch.

The file is read relative to the cwd while `--db` can point anywhere, so under `--db /elsewhere.db` it names a run that has no relationship to the database being briefed, and the note is stated with a confidence the check cannot back. Gating it on the same `not args.run_id` and SessionStart conditions the banner uses keeps the note tied to the run the briefing actually covers.

Cyrax321 (commented):
Fixes the ghost-run banner the right way: the file is read once, the run is validated against the database, and that one answer drives both run-id resolution and the banner, so the banner can no longer advertise a run the briefing beneath it does not cover. The malformed-payload matrix is the strongest part, since every failure mode degrades to the same place. Nothing here blocks the merge; I would take it.

Both findings below are non-blocking and are the same shape: the new validation reaches a conclusion it has not earned, and in both cases the fix is to narrow an existing check rather than add a new one.

Cyrax321 (approved):
Approving. The banner now reads the resume file once, validates the run it names, and reuses that one answer for both run-id resolution and the banner, so the banner and the briefing body can't disagree again. The five malformed-payload cases are good coverage since every failure mode degrades to the same place.

The two notes I left inline are non-blocking and both are the same shape: the new validation reaches a conclusion it has not earned, and narrowing an existing check fixes it. Neither holds up the merge.

PR #1121, opened 2026-09-15
fix(recovery): name every trigger of the winning mode in the risk rationale

### LINKED ISSUE #1057: risk rationale names one trigger while triggering_risks lists every contributor to the verdict
## Summary

When two or more distinct risk triggers map to the same recovery mode, the engine records every contributing event id in `RecoveryContract.triggering_risks` but names only one trigger in the rationale that becomes the sealed `reason`. The hash-chained justification omits contributors to the verdict it justifies: an auditor reads one trigger name and two event ids with no way to learn what the second event was.

## Affected component

Core (recovery engine risk path, recovery contract)

## Severity

Minor inconvenience or cosmetic issue for the library. It matters because the contract's entire value proposition is a verifiable justification, and this one is incomplete rather than wrong.

## Evidence

At HEAD `14f8a84`, `src/continuum/recovery/engine.py:437` sets `best_trigger` only when a strictly more cautious mode arrives, `:442` appends every equally-severe event id to `triggering`, `:445` assigns all of them to `triggering_risks`, and `:446` builds the rationale from `best_trigger` alone. `src/continuum/recovery/contract.py:185` then seals that single-trigger string into `reason` while `:183` carries the full id list into `next_allowed_action`'s sibling field set.

```
mode                     : wait
rationale                : ('risk error_cascade triggers wait',)
contract.triggering_risks: ['event_866fc22cb6828e3ae7e1c8973ac17447', 'event_5a908ec3ec2329603319784f6ec926b5']
contract.reason (sealed) : risk error_cascade triggers wait
```

Both triggers map to `wait` in `src/continuum/recovery/risk.py`, so both are genuine contributors to the verdict; only one is named.

```python
import os, tempfile
from continuum.recovery.risk import ingest_risk

st = SQLiteStorage(db)
ingest_risk(st, 'r', {'trigger': 'error_cascade', 'score': 0.9})
ingest_risk(st, 'r', {'trigger': 'token_runaway', 'score': 0.8})
st.close()

st = SQLiteStorage(db)
print('mode                     :', dec.mode.value)
print('rationale                :', dec.rationale)
print('contract.triggering_risks:', dec.contract.triggering_risks)
print('contract.reason (sealed) :', dec.contract.reason)
```

Output:

```
mode                     : wait
rationale                : ('risk error_cascade triggers wait',)
contract.triggering_risks: ['event_866fc22cb6828e3ae7e1c8973ac17447', 'event_5a908ec3ec2329603319784f6ec926b5']
contract.reason (sealed) : risk error_cascade triggers wait
```

## Root cause

Introduced in `201219b` ("feat: contract triggering risks", 2026-09-02). The loop tracks the id list and the winning trigger name in the same branch but threads only the name into the rationale.

## Expected behaviour

The sealed `reason` should name every trigger that contributed to the winning mode, or the contract should carry a trigger-to-event mapping so each id in `triggering_risks` is self-explaining.

## Suggested fix shape

Collect the trigger names alongside the ids and build the rationale from all triggers that proposed the winning mode, for example joining the sorted trigger set rather than `best_trigger` alone. Deduplicate by trigger so the fix does not reintroduce the duplicated-sentence problem of #1042. `triggering_risks` is already correct and needs no change.

This is the opposite failure from #1042 in the same field: #1042 appends the liveness WAIT proposal twice so one sentence appears twice in `reason`, while this issue is a real contributor going missing. Fix either without closing the other; a naive dedup pass for #1042 would not touch the risk path.

## System information

- Python: 3.14.5

Adhi1-2:
Fixes #1057

## What was broken

When two or more distinct risk triggers mapped to the same recovery mode, the engine recorded every contributing event id in `RecoveryContract.triggering_risks` but built the rationale — and therefore the sealed `reason` — from the first winning trigger alone. An auditor read one trigger name and two event ids with no way to learn what the second event was: the hash-chained justification was incomplete rather than wrong.

The loop tracked the id list and the winning trigger name in the same branch (`engine.py:437-446`) but threaded only the name into the rationale.

## The fix

- Collect the trigger names alongside the ids; reset both when a strictly more cautious mode wins, append both on equal severity
- Build the rationale from the deduplicated, sorted trigger set: `risk error_cascade, token_runaway triggers wait`
- A trigger proposing a less severe mode stays excluded from the reason (it did not contribute to the winning verdict)
- A repeated trigger yields one mention, so the #1042 duplicate-sentence failure mode is not reintroduced — this is the mirror image of #1042 in the same field, fixed without touching that path
- `triggering_risks` is already correct and unchanged

## Verification

- The issue's repro now seals `risk error_cascade, token_runaway triggers wait` with both contributor ids
- New regression tests in `tests/test_risk_policy.py`:
  - multi-trigger reason names both, sorted, in rationale and sealed `contract.reason`
  - a more severe trigger (`meltdown` → rollback) excludes the earlier `wait` contributor
  - a repeated trigger appears exactly once


Cyrax321 (approved):

This PR fixes the risk rationale naming only the first trigger while triggering_risks lists every contributor (#1057). I reviewed the risk loop in engine.py and the new tests in test_risk_policy.py.

Collecting trigger names alongside ids, resetting both on a strictly more cautious winner and deduplicating on equal severity, is the right shape. Less severe triggers stay excluded, repeats stay deduplicated so the #1042 failure mode is not reintroduced, and sorted output keeps the sealed reason deterministic. The three regression tests pin exactly those properties, and CI plus Codecov coverage are green.

Issue #838, opened 2026-09-08
Packaging hygiene: version drift between pyproject.toml, tags, and PyPI


gouthamkrishnak2003:
## Problem

The repository is versioned inconsistently:

- `pyproject.toml` says `version = "0.1.0"`.
- Git tags exist for `v0.1.0` **and `v0.1.2`** (with matching GitHub Releases).
- README and docs advertise only `pip install continuum-agent==0.1.0`.

So PyPI is stale relative to the newest release, or 0.1.2 was released without the version
bump — either way, a user installing "latest" does not get the newest tagged code, and the
docs point at an older version than the release history implies.

## Root cause (evidence)

`git tag -l` shows `v0.1.0` and `v0.1.2`; `pyproject.toml:7` is `0.1.0`;
`README.md:60` advertises 0.1.0. `docs/release-checklist.md` documents the maintainer gate
but has no check that pyproject version == tag == PyPI.

## Proposed solution

1. Decide the actual current version and set `pyproject.toml` to match reality; publish a
   reconciling release if PyPI is stale (maintainer action).
2. Add a release-checklist item (and optionally a lightweight CI check on tag push) that
   fails when `pyproject.toml` version != tag != PyPI version.
3. While in there: the Dockerfile installs with **no extras**, so the published image cannot
   run `continuum-mcp` — either add the extra or document the limitation explicitly.

## Scope

Version reconciliation + checklist/CI guard + Dockerfile extras decision.

## Out of scope

- Any code changes; the docs contradictions beyond version numbers (Issue 7).

## Acceptance criteria

- pyproject version, latest tag, and PyPI agree (or a documented reason they do not).
- The checklist/CI guard exists and would have caught this drift.

## Test plan

A tiny CI or checklist step comparing the three values; no product tests.

## Platform coverage

Platform-independent.

## Dependencies

None.

## Contributor notes

- `pyproject.toml:7`, `docs/release-checklist.md`, `Dockerfile:4`, `release.yml` (the
  opt-in `PUBLISH_PYPI` gate at `release.yml:42-45`).

Cyrax321:
[...]

PR #1122, opened 2026-09-15
fix: sync CITATION.cff version with package 0.1.2

### LINKED ISSUE #1120: CITATION.cff pins version 0.1.0 while the package is 0.1.2, and CONTRIBUTING says the version lives in two places when the #838 guard checks four
## Summary

The release version is out of sync and the contributor docs misdescribe where it lives:

**`CITATION.cff:14` pins `version: 0.1.0` while the package is `0.1.2`** — `pyproject.toml:7` and `src/continuum/__init__.py:167` both say `0.1.2`, and the README install pins (`README.md:60`, `:66`) say `continuum-agent==0.1.2`. Anyone citing the project gets the wrong version from the citation file. (The ORCID placeholder in this file was #792; this is the version field.)

**`CONTRIBUTING.md:217-220` understates the version sites and the guard.** It says the version lives in **two** places (pyproject + `__init__.py`) and that "CI will fail if they diverge **once version-checking is added** to the release workflow." Both halves are wrong:

- Version-checking already exists and already runs in CI's pytest: `tests/test_version_drift.py` (the guard added for #838) enforces `continuum.__version__ == pyproject version` (`test_pyproject_version_matches_package`), that the README carries `continuum-agent=={v}` (`test_readme_pins_match_pyproject`), and that the git tag `v{v}` exists (`test_git_tag_exists_or_skip`).
- So the real count is four places to touch on a bump — pyproject, `__init__.py`, both README pins — plus the tag, and a contributor who follows the paragraph and updates only the two documented sites gets a red CI via the README-pin test.

`CITATION.cff` is a fifth site the guard does not cover, which is how it ended up two releases stale.

## Affected component

Documentation; Other (release hygiene)

## Severity

Minor inconvenience — wrong version in citations; a contributor's bump fails CI for a reason CONTRIBUTING does not explain.

## System information

- Python: 3.13.5

```bash
grep -n '^version' pyproject.toml                    # 0.1.2
grep -n '__version__' src/continuum/__init__.py      # 0.1.2
grep -n 'version' CITATION.cff                       # version: 0.1.0
grep -n 'continuum-agent==' README.md                # two pins at 0.1.2

sed -n '215,221p' CONTRIBUTING.md                    # "two places" + "once version-checking is added"
```

## What a fix looks like

Bump `CITATION.cff` to the current version, rewrite the CONTRIBUTING paragraph to name all the sites the `#838` guard actually checks (pyproject, `__init__.py`, both README pins, the release tag) and drop the "once version-checking is added" clause, and — to keep the citation file from drifting again — extend `tests/test_version_drift.py` with a `CITATION.cff` version check alongside the README-pin test.

wized2:
## Summary
Closes #1120.

`CITATION.cff` still pinned `version: 0.1.0` while `pyproject.toml` is `0.1.2`. Updated the citation file to match the published package version.

## Summary by CodeRabbit

* **Chores**
  * Updated the software version from 0.1.0 to 0.1.2.


anya-research (changes requested):
Thank you for contributing to Continuum, @wized2. This bumps `CITATION.cff` from 0.1.0 to 0.1.2 for #1120. There are some changes needed before this can merge.

`CITATION.cff:14` now reads 0.1.2, matching `pyproject.toml:7` and `src/continuum/__init__.py:167` plus the README pins at 0.1.2. No other open PR touches the citation file or claims #1120.

- diff is one line in `CITATION.cff` with clean whitespace
- `CITATION.cff:14` matches the package version and README pins
- Lint and Type-check is SUCCESS in CI, test matrix is mixed as noted below

`Test (windows-latest, Python 3.12)` reports FAILURE from the benchmark subprocess timeout in `tests/test_horizon_benchmark.py`. That failure guards benchmark runtime behavior, not citation metadata, and it needs a rerun or fix first since it is a required matrix leg.

The remaining gap is scope. #1120 asks for three parts and this diff contains only the version bump. Details are in the two findings below.

`CONTRIBUTING.md:217`

`CONTRIBUTING.md:217-223` still says the version lives in two places and that CI will fail once version checking is added. That description is stale because `tests/test_version_drift.py` already enforces pyproject, package, README pins, and the tag, and it omits `CITATION.cff` entirely.

```suggestion
The version string lives in five places:

- `pyproject.toml` → `version = "X.Y.Z"`
- `src/continuum/__init__.py` → `__version__ = "X.Y.Z"`
- `README.md` → two `continuum-agent==X.Y.Z` pins
- `CITATION.cff` → `version: X.Y.Z`
- release tag `vX.Y.Z`

Update all five when bumping. CI enforces them in `tests/test_version_drift.py` and fails if they diverge.
```

`tests/test_version_drift.py:25`

`tests/test_version_drift.py` still contains only the three #838 checks with no citation assertion. Without a `CITATION.cff` guard the file can drift stale again exactly as #1120 describes.

```suggestion
def test_citation_version_matches_pyproject():
    v = _pyproject_version()
    text = (ROOT / "CITATION.cff").read_text(encoding="utf-8")
    assert f"version: {v}" in text
```

Cyrax321 (changes requested):
Author context: wized2, external contributor. The one-line CITATION.cff version sync direction is right, but the Test matrix is mixed with a Windows failure standing. Rebase onto current main and confirm the full matrix green before merge; a version-sync PR should land clean.

PR #1124, opened 2026-09-15
fix(storage): persist and return runs.parent_run_id on Postgres (#1079)

### LINKED ISSUE #1079: Postgres backend never writes or reads runs.parent_run_id, so fork lineage and the family resume block are silently absent there
The `runs` table in the Postgres schema declares the lineage column (`postgres.py:64`, `parent_run_id TEXT REFERENCES runs(run_id)`), but the backend never populates it and never returns it:

`PostgresStorage.create_run` and `create_run_started` both insert only `(run_id, goal, status, created_at, updated_at, metadata)`, omitting `parent_run_id`, and `PostgresStorage._row_to_run` constructs the `Run` without `parent_run_id`, so the stored value comes back as the model default `None` even for a row that somehow had one.

The SQLite backend does all three: both inserts carry the column and `_row_to_run` reads it.

The consequence is that on the Postgres backend an entire documented safety property is vacuous. `fork_run` builds the child as `Run(..., parent_run_id=parent_run_id, ...)`; that value is dropped on write. `children_of` (recovery/family.py:37) filters `list_runs` by `parent_run_id`, so it always returns empty; `roll_up_children` therefore never reports a family block; and the rule the TUI and CLI both render, "a parent may not RESUME while any child is unsafe", cannot fire. The foreign-key constraint that is supposed to guarantee the referenced parent exists is also never exercised, because the column is always null.

This is the same class of divergence as the missing `compact_run` anchor guard, and it is the more damaging of the two: the compact one needs a direct API caller to reach, while this one breaks a property on every fork as a matter of course, with no error anywhere. A Postgres deployment would look healthy and simply never block a parent over an unsafe child.

The fix is to write the column in both insert methods and read it in `_row_to_run`, matching SQLite. Worth adding to the contract test suite a case that forks on each backend and asserts `children_of` returns the child, since that is the observable behaviour the two backends must share and nothing today catches the gap.

I found this comparing the two storage backends field by field while auditing the fork path.

Cyrax321:
## Summary

Fixes #1079. The Postgres backend declared `runs.parent_run_id` in its schema but never wrote it and never read it back, which made an entire documented safety property vacuous on that backend with no error anywhere.

`create_run` and `create_run_started` inserted only `(run_id, goal, status, created_at, updated_at, metadata)`, and `_row_to_run` constructed the `Run` without the column, so a value stored by some other path still came back as the model default `None`. The SQLite backend does all three.

The consequence is worse than a stale field. `children_of` (`recovery/family.py:37`) filters `list_runs` on `parent_run_id`, so on Postgres it always returned empty, `roll_up_children` therefore never reported a family block, and the rule the CLI and the TUI both render, "a parent may not RESUME while any child is unsafe", could not fire. The `REFERENCES runs(run_id)` constraint was also never exercised, because the column was always null. A Postgres deployment would look healthy and simply never block a parent over an unsafe child. This is the same class of divergence as the missing `compact_run` anchor guard, but more damaging: that one needs a direct API caller to reach, while this one breaks a property on every fork as a matter of course.

## Changes

- `src/continuum/storage/postgres.py`: write `parent_run_id` in both insert methods and read it in `_row_to_run`, matching SQLite exactly. The schema already declared the column, so no migration is needed and existing Postgres databases are fixed on their next write.

## Tests

- `tests/test_storage_postgres.py`: new contract test that forks on the Postgres backend and asserts `children_of` returns the child through the query the roll-up reads, not just the column. This is the observable behaviour the two backends must share, and nothing today caught the gap because the Postgres suite never forked.
- `tests/test_family.py`: the existing SQLite test asserted the column round-trip but not the query that reads it, which is precisely how the Postgres side went unnoticed. It now asserts `children_of` returns the child too.

## Validation

`pytest tests/test_family.py tests/test_fork.py tests/test_fork_preconditions.py tests/test_storage.py tests/test_cli.py` green, `ruff check` and `mypy` clean.

The Postgres contract test skips locally (`CONTINUUM_TEST_POSTGRES_DSN` unset, no local server); CI exercises it for real via the Postgres 16 service container. That is the honest limitation of this verification: the fix is verified by inspection against the SQLite reference and by a test that will run in CI, not by a local Postgres run.

## Summary by CodeRabbit

- **Bug Fixes**
  - PostgreSQL now preserves parent-child relationships for forked runs.
  - Child runs are correctly returned when viewing a run’s family, enabling related safety and recovery rules to work as expected.

- **Tests**
  - Added coverage confirming fork lineage survives PostgreSQL storage and retrieval.
  - Strengthened family relationship checks across storage backends.

- **Documentation**
  - Updated the unreleased changelog with details of the PostgreSQL lineage fix.


Cyrax321 (commented):

Ref: #1124 / Issue #1079
Surface: Postgres storage backend, run lineage

Bug: the Postgres backend declared runs.parent_run_id in its schema but never wrote it on insert and never read it back, so a fork's parent came back as None. Since children_of (src/continuum/recovery/family.py:37) filters list_runs on this column, the family resume block was silently vacuous on Postgres: a parent could always RESUME over an unsafe child, and the FK constraint was never exercised.

Fix: write parent_run_id in both create_run and create_run_started, read it in _row_to_run. This matches the SQLite backend exactly and lives at the right ownership boundary. No schema change needed since the column already exists.

Proof: new regression test test_fork_lineage_survives_the_round_trip asserts the round trip and, importantly, goes through children_of rather than only the column, plus a strengthened SQLite-side assertion in test_family.py. The Postgres CI job runs this file against a real backend and is green, as is the rest of the matrix at head b82e2383.

Note for merge ordering: #1313 (fix/issue-1079-postgres-parent-run-id) addresses the same issue. Whichever lands second will need a rebase; the tests in both target the same seam, so keep the one with the stronger regression coverage (this PR's children_of assertion) when reconciling.

Cyrax321 (commented):
Maintainer re-review at current head. The Postgres parent_run_id round-trip fix is still correct and matches the SQLite backend. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review. Overlaps #1313 on the same issue #1079, which has since merged; rebase and close or reduce to what remains beyond the landed fix.

PR #1125, opened 2026-09-15
fix(recovery): degrade the consumed-authority block closed on an unreadable ledger

### LINKED ISSUE #1066: consumed-authority block fail-opens on a transient read error, clearing a blocked run for resume
## Summary

The consumed-authority block in `assess()` wraps its event read in a bare `except Exception` that substitutes an empty map. When the read fails, the block is skipped entirely and the engine issues a `resume` verdict for a run holding an unreconciled consumed credential. This is the one safety block on the authority-lifecycle path, and it degrades fail-open instead of fail-closed.

## Affected component

Core (recovery engine, authority lifecycle)

## Severity

Major functionality broken, workaround exists. Reaching the failure requires a storage read to fail after earlier reads in the same call succeeded, so it is not common; but the whole point of the block is to be the check that survives a degraded log, and it is the first thing the degradation removes.

## Evidence

`src/continuum/recovery/engine.py:470` reads the event tail and `:471-472` discards any exception into `consumed_authorities = {}`. The `if consumed_authorities:` guard then sees nothing and the mode computed above stands. The comment at `engine.py:463-468` states the block's purpose, that the decision must surface the block so `continuum resume` does not report safe when the gate would deny. A read that raises here makes exactly that misreport.

```
baseline: mode=request_human safety=requires_human
with transient read failure at the authority block:
  mode                      : resume
  safety                    : safe_to_resume
  rationale                 : ('all state verified against the environment',)
  permits("anything")       : True
```

```python
import os, tempfile, inspect
from continuum.storage.base import CorruptedRecord

st = SQLiteStorage(db)
st.append_event('r', EventType.AUTHORITY_CONSUMED, {'authority_id': 'cred-1'}, source=Origin.HUMAN)
st.close()

orig = SQLiteStorage.read_events

def flaky(self, run_id, *a, **k):
    frame = inspect.currentframe().f_back
    src = frame.f_code.co_filename
    line = open(src).read().splitlines()[frame.f_lineno - 1]
    if src.endswith('recovery/engine.py') and 'collect_consumed_authorities' in line:
        raise CorruptedRecord('events table: unreadable page (transient)')
    return orig(self, run_id, *a, **k)

SQLiteStorage.read_events = flaky
dec = RecoveryEngine(SQLiteStorage(db)).assess('r', replay=False)
print('permits(anything)   :', dec.permits('anything'))
```

Output at HEAD `14f8a84`, with the same run returning `request_human` / `requires_human` when the read succeeds:

```
mode                : resume
safety              : safe_to_resume
permits(anything)   : True
```

The wrapper raises only on the call made from the authority block, so every earlier read in `assess()` succeeds; this models a page going unreadable, or a backend connection dropping, mid-assessment.

## Root cause

Introduced in `d4c5ddb` ("engine: block resume while a consumed authority awaits probe", 2026-09-05). The surrounding code already favours degradation over raising (see the `on_unprojectable="degrade"` comment), and this block inherited the pattern without considering that an empty map is the unblocked answer rather than an unknown one.

## Expected behaviour

A failure to read the authority ledger should not convert a blocked run into a resumable one. Either the exception should propagate, or the block should fall back to the most cautious available answer.

## Suggested fix shape

Distinguish "no consumed authorities" from "could not determine". On exception, either re-raise so the caller learns the log is unreadable, or set the mode to `REQUEST_HUMAN` with a rationale naming the failed read, which keeps the degradation-cannot-raise convention without asserting a safety conclusion the engine could not compute. The same shape applies to the sibling bare excepts at `engine.py:274-275`, `:403-405` and `:421-422`, though only this one gates a security-relevant block.

## System information

- Python: 3.14.5

Adhi1-2:
Fixes #1066

## What was broken

The consumed-authority block in `assess()` wrapped its event read in a bare `except Exception` that substituted an empty map — and an empty map is the **unblocked** answer. When the read failed mid-assessment (a page going unreadable, a backend connection dropping), a run holding an unreconciled consumed credential got a `resume` verdict with `permits() == True`. The one safety block on the authority-lifecycle path degraded fail-open instead of fail-closed, exactly when the log became degraded — the condition the block exists to survive.

## The fix

Distinguish "no consumed authorities" (empty map → no block, unchanged) from "could not determine" (`None`): on a failed read the verdict degrades to `REQUEST_HUMAN` with a rationale naming the unreadable ledger. This keeps the degradation-cannot-raise convention the surrounding code follows (`on_unprojectable="degrade"`) without asserting a safety conclusion the engine could not compute.

The sibling bare excepts the issue mentions (`engine.py:274-275`, `:403-405`, `:421-422`) are left as-is per the issue: only this one gates a security-relevant block.

## Verification

New regression test in `tests/test_authority_probe.py` pinning both halves:
- readable ledger → the run stays `request_human` (baseline, as designed)
- unreadable ledger → `request_human`, `recovery_status == requires_human`, `permits("anything") is False`, rationale names the unreadable ledger

Full test suite, `ruff check`, `ruff format --check` and `mypy` clean (one pre-existing `type-arg` on an unrelated test).


Cyrax321 (approved):

This PR fixes the consumed-authority block degrading fail-open on an unreadable ledger (#1066). I reviewed assess() in engine.py, the SEVERITY invariant, the _decide path with risk_mode, and the new regression test.

Distinguishing unknown (None) from empty is the right shape. It keeps the degradation-cannot-raise convention while no longer asserting an unblocked verdict the engine could not compute. The test pins both halves, readable baseline and unreadable degradation to request_human with permits() refusing, and CI plus Codecov coverage are green.

I checked the bot note about overwriting a more severe ROLLBACK or ABORT with REQUEST_HUMAN. That pattern pre-exists in the readable branch, the run still blocks resume, and the downgrade case needs both a risk trigger and a mid-assessment read failure, so I do not consider it blocking. Taking the max with SEVERITY could be a follow-up if you want the label to stay strictly most-cautious.

Issue #705, opened 2026-09-07
compact_run(through_sequence=...) is unbounded: archives and deletes the anchor, forking the hash chain


Adhi1-2:
## Problem

`SQLiteStorage.compact_run` (`src/continuum/storage/sqlite.py:470-529`) guards only `through < 1`. A `through_sequence` at or above the anchor's own sequence is accepted, and the `DELETE ... WHERE sequence <= ?` then removes **the anchor marker the method just appended** — together with the entire live log.

The live log becomes empty, so the next `append_event` finds no head and mints a fresh genesis (`sequence=1`, `prev_hash=None`): the live chain forks away from the archive, which still holds the true chain. `verify_events` then reports permanent `SEQUENCE_GAP` + `BROKEN_CHAIN`, and `read_all_events` (sorted by sequence) interleaves two sequence-1 events, scrambling history for every fold over it (provenance, exactly-once action claims).

The caller cannot know the anchor's sequence in advance (it is appended during the call), so even a plausible-looking value like `last_sequence() + 2` triggers this.

## Repro sketch (mechanism reproduced against `upstream/main` at `f5cb8d0`)

```python
# run r1 with live events at sequences 1..6
store.compact_run("r1", through_sequence=10)
# forced checkpoint -> seq 7, anchor -> seq 8, then DELETE <= 10 removes all 8 live rows
store.append_event("r1", "TASK_UPDATED", {...})
# -> new event gets sequence 1, prev_hash None: forked chain
store.verify_events("r1")   # permanent SEQUENCE_GAP + BROKEN_CHAIN
```

## Why tests miss it

The only in-tree callers passing `through_sequence` are tests with values far below the boundary (`test_dashboard.py:90,138`); the CLI never passes it. No test exercises an over-large value.

## Suggested fix

Clamp `through` to `min(through_sequence, anchor_sequence)` (or reject values that would delete the anchor with a clear `ValueError`), since the anchor marker's sequence is known inside the method before the delete runs. Today no production caller passes `through_sequence`, so this is latent — but it is a public `Storage` API whose contract is silently violable with unrecoverable corruption.

PR #1127, opened 2026-09-15
fix(storage): share the compaction anchor guard between both engines

### LINKED ISSUE #1078: compact_run's through_sequence anchor guard exists on SQLite but is absent on Postgres, so a Postgres run can archive and delete its own anchor marker
`SQLiteStorage.compact_run` refuses an explicit `through_sequence` that would reach the anchor marker, and the refusal is load-bearing (sqlite.py:518):

```python
anchor_sequence = self.last_sequence(run_id) + 1
if through_sequence is not None and through_sequence >= anchor_sequence:
    raise ValueError(
        f"through_sequence {through_sequence} would archive the anchor marker "
        f"at sequence {anchor_sequence}: the live log must retain its anchor"
    )
```

`PostgresStorage.compact_run` (postgres.py:583) has no equivalent check. It computes `through = through_sequence` directly and runs `DELETE FROM events WHERE run_id = %s AND sequence <= %s`, so a `through_sequence` at or above the anchor marker's sequence archives and deletes the marker and every live row after it.

The consequence is the one the SQLite guard documents: the live log no longer retains its anchor, the next append mints a fresh genesis with `prev_hash = None`, and the live chain forks away from the archive. The whole point of the single-transaction marker-plus-move that both backends implement is that the live log always retains its anchor, and the explicit-`through_sequence` path is the only path that can violate it, since the default `through = min(source_sequence, head)` is always below the marker.

The two implementations are otherwise the same function, including the `needs_fresh_anchor` recomputation and the `through < 1` refusal, which is what makes the one missing guard stand out: the Postgres backend kept every other safety check from the SQLite one and dropped this one.

Reachability is the same as it was for SQLite before issue #705: the CLI's `compact` calls `storage.compact_run(args.run_id)` with no `through_sequence`, so only a direct caller of the storage API can hit it. That is still a real hole for a library whose storage is a public interface, and for any future CLI flag that exposes the bound.

The fix is to mirror the guard before the transaction, using the same `anchor_sequence = last_sequence + 1` computation, or to lift the check into a shared helper both backends call so the two cannot drift apart again.

I found this comparing the two `compact_run` implementations while auditing the compaction path.

Cyrax321:
Fixes #1078.

`SQLiteStorage.compact_run` refused an explicit `through_sequence` at or above the anchor marker's sequence (#705), and the refusal is load-bearing: without it the archive and delete remove the marker and every live row after it, the next append mints a fresh genesis with `prev_hash = None`, and the live chain forks away from the archive. `PostgresStorage.compact_run` had no equivalent check. It computed `through` straight from the argument and ran `DELETE FROM events WHERE sequence <= through` over it, so a direct caller of the storage API could delete the anchor on a Postgres run.

The two implementations were otherwise the same function, including the `needs_fresh_anchor` recomputation and the `through < 1` refusal, which is what made the one missing guard stand out: the Postgres backend kept every other safety check and dropped this one.

Rather than mirror the check and leave the same drift open to recur, both engines now resolve the bound through one shared helper, `continuum.storage.compaction.resolve_compaction_bound`. It owns the whole pre-transaction decision (fresh-anchor recomputation, forced checkpoint, the `no projectable state` refusal, the anchor guard, the default bound, the empty-anchor refusal) and returns `(version, through)`. Each engine keeps only its transaction body, which genuinely differs in parameter style and connection handling. This follows the precedent already set by `actionindex.py`, which exists for exactly this reason: shared logic so the two engines "cannot drift apart in what they accept."

Reachability is unchanged from #705. The CLI's `compact` calls `storage.compact_run(run_id)` with no bound, and the default `through = min(source_sequence, head)` is always below the marker, so only a direct caller of the storage API can hit this. That is still a real hole for a library whose storage is a public interface, and for any future CLI flag that exposes the bound.

The Postgres suite gains the anchor rejection test the SQLite suite has had since #705: a `through_sequence` of 10_000 is refused, the rejected call leaves a verifiable log with only the forced checkpoint appended and live rows unmoved, and a bounded value below the anchor still compacts normally. Before this change that test fails on the first assertion, because the unguarded path archived the whole log instead of raising.

Behaviour on SQLite is unchanged: the helper reproduces the existing preamble verbatim, and `tests/test_compaction.py` including the #705 test still passes through it.

Verification: `ruff check` and `ruff format --check` clean, `mypy src/continuum` clean, and the compaction, checkpoint, provenance-compaction, and storage suites pass. The docs-count guard passes with one added test.

The Postgres suite skips locally because this machine has no `psycopg` and no Postgres server, so the new test is exercised by the CI Postgres 16 service container rather than here. I verified the guard logic two ways instead: the shared helper is covered by the SQLite #705 test, which now runs through it, and I confirmed against the pre-fix preamble that `through_sequence=10_000` against a head of 5 resolves to an anchor at sequence 6, which is exactly the violation this prevents.


Cyrax321 (commented):
Thank you.

Sharing the bound resolution instead of mirroring the check is the right fix for a second drift between the same two functions. The failure mode is load-bearing: archiving the marker forks the live chain with a fresh genesis, so the missing guard is not cosmetic. Keeping only the transaction bodies engine-specific follows the actionindex precedent and leaves reachability unchanged.

I reviewed compaction.py, both engine call sites, and the new Postgres anchor test. The rejected call leaving a verifiable log with live rows unmoved plus the safe bounded compact still working covers both sides. CI is green. SQLite behavior is verbatim through the helper.

Cyrax321 (commented):
Maintainer re-review at current head. The shared compaction anchor guard direction is still correct with engine-specific transaction bodies. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review with no blocking issues at last review. Rebase out of dirty state and confirm CI green.

PR #1129, opened 2026-09-15
fix(recovery): stop the contract advertising a permitted step under ROLLBACK and ABORT

### LINKED ISSUE #1058: contract advertises a permitted repair step under ROLLBACK and ABORT, and permits() returns True for it
## Summary

`build_contract` derives `next_allowed_action` from `plan.first` without checking the recovery mode. Under a risk-driven ROLLBACK or ABORT the contract still advertises a repair step as the single permitted action, and `RecoveryDecision.permits()` returns True for it. That contradicts the contract's own rendering fallback, which prints "none (settle required_actions first)" for every status other than SAFE_TO_RESUME, and it hands any caller that does gate on `permits()` a green light on a run the engine has declared unsafe.

## Affected component

Core (recovery contract, recovery decision)

## Severity

Major functionality broken, workaround exists. Today nothing in `src/` calls `permits()`, so the wrong answer is latent; #1031 is exactly the issue that proposes wiring a caller, and wiring one without this fix activates the bug. The contract field is also echoed by `render`, the dashboard, the resume report and the MCP `next_allowed_action` surface, so an operator reads "the next allowed action is to revalidate the dependency" on a run that must not be resumed at all.

## Evidence

At HEAD `14f8a84`, a run driven to risk-driven ABORT with a stale dependency:

```
mode                     : abort
safety                   : unsafe
plan.first               : RepairKind.REVALIDATE_DEPENDENCY target='dataset' reason='v3 -> v4'
next_allowed_action      : revalidate_dependency:dataset
permits(next_allowed)    : True
```

`src/continuum/recovery/contract.py:148` sets `next_action = plan.first.action_name if plan.first else None` and `:183` seals it as `next_allowed_action` with no mode or safety check. `src/continuum/recovery/engine.py:136` `permits()` then returns True for that action because it equals `contract.next_allowed_action`. The contract's own fallback at `contract.py:227` reads "none (settle required_actions first)" for anything but SAFE_TO_RESUME.

```python
import os, tempfile

st = SQLiteStorage(db)
st.append_event('r', EventType.RUN_STARTED, {'goal': 'g', 'total': 1})
st.append_event('r', EventType.DEPENDENCY_DECLARED, {'resource': 'dataset', 'version': 'v3'})
st.append_event('r', EventType.WORK_COMPLETED, {'doc': 0})
st.append_event('r', EventType.RISK_OBSERVED, {'trigger': 'side_effect_duplicate', 'score': 1.0},
st.close()

st = SQLiteStorage(db)
dec = RecoveryEngine(st).assess('r', current_environment=__import__(
    'continuum.environment.snapshot', fromlist=['capture', 'StaticProvider']).capture(
        'r', __import__('continuum.environment.snapshot',
                        fromlist=['StaticProvider']).StaticProvider(dataset='v4')),
    replay=False)
print('plan.first          :', dec.plan.first)
print('next_allowed_action :', dec.contract.next_allowed_action)
print('permits(next_allowed):', dec.permits(dec.contract.next_allowed_action))
```

Output:

```
mode                : abort
safety              : unsafe
plan.first          : RepairKind.REVALIDATE_DEPENDENCY target='dataset' reason='v3 -> v4'
next_allowed_action : revalidate_dependency:dataset
permits(next_allowed): True
```

## Root cause

`build_contract` was written when `REPAIR_AND_RESUME` was the only non-resume mode that carried a plan. The risk path added ROLLBACK and ABORT proposals later, both of which can coexist with a non-empty repair plan, and the contract never gained a mode check for the action it advertises.

## Expected behaviour

`next_allowed_action` is null, or `permits()` returns False for everything, when the mode is not one that permits repair. A contract that declares a run unsafe should not simultaneously name a permitted action.

## Suggested fix shape

Either gate `next_action` in `contract.py:148` on the recovery status being one that permits repair, or make `permits()` at `engine.py:136` consult the safety status before admitting the action. The contract-level fix is preferable because it stops every downstream surface from echoing the wrong action, not just the one caller that gates on it. Add a test asserting `next_allowed_action is None` and `permits(...) is False` under a risk-driven ABORT.

## System information

- Python: 3.14.5

Adhi1-2:
Fixes #1058

## What was broken

`build_contract` derived `next_allowed_action` from `plan.first` without checking the recovery mode. A risk-driven ROLLBACK or ABORT can coexist with a non-empty repair plan, so under ABORT the contract advertised `revalidate_dependency:dataset` as the single permitted action and `permits()` returned `True` for it — on a run the engine had declared unsafe. This contradicts the contract's own rendering fallback ("none (settle required_actions first)" for every status other than SAFE_TO_RESUME) and would hand any caller gating on `permits()` (the seam #1031 proposes wiring) a green light. The field is also echoed by `render`, the dashboard, the resume report and the MCP `next_allowed_action` surface.

## The fix

The contract-level gate, as the issue prefers — it stops every downstream surface from echoing the wrong action, not just one caller:

- `next_allowed_action` is `None` when the safety is `BLOCKED` (ROLLBACK) or `UNSAFE` (ABORT)
- Every other mode keeps its current behaviour: REPAIR_AND_RESUME, WAIT and REQUEST_HUMAN still advertise their first plan step (human-review and reconcile steps are legitimately the permitted action there, as existing tests pin)
- `required_actions` is unchanged — an auditor still sees the work that exists; only the permission is gone

## Verification

- The issue's repro now prints `next_allowed_action : None` and `permits(...) : False` with the plan still present and `required_actions` still listed
- New regression tests in `tests/test_recovery_engine.py` for risk-driven ABORT and ROLLBACK: plan exists, `next_allowed_action is None`, `permits()` refuses the repair step, `required_actions` survive
- All existing `next_allowed_action` assertions still pass (recovery engine, MCP server, provenance, serialization roundtrip, risk bench)
- Full test suite, `ruff check`, `ruff format --check` and `mypy` clean (one pre-existing `arg-type` on an unrelated test)


Cyrax321 (approved):

This PR fixes the contract advertising a permitted repair step under risk-driven ROLLBACK and ABORT (#1058). I reviewed build_contract, the permits() path in engine.py, the safety mapping, the render fallback, and the surrounding tests.

The contract-level gate is the right shape. It stops every downstream surface from echoing the wrong action while keeping required_actions visible for audit, and it preserves behavior for REPAIR_AND_RESUME, WAIT, and REQUEST_HUMAN. The two new regression tests pin the exact invariant from the issue, and CI plus Codecov coverage are green.

I did not find any blocking issues. One minor note for future work: the ROLLBACK test could also assert plan coexistence and required_actions survival the way the ABORT test does, but that does not block this change.

Marking as approved. Thank you again for the fix.

PR #1132, opened 2026-09-15
fix(recovery): a blocked or unsafe run permits no action, not even its own repair (#1058)

### LINKED ISSUE #1058: contract advertises a permitted repair step under ROLLBACK and ABORT, and permits() returns True for it

abyyxhek:
## Summary

`build_contract` derived `next_allowed_action` from `plan.first` with no check on the recovery mode. Under a risk-driven ROLLBACK or ABORT the contract still advertised the repair step as the single permitted action, and `RecoveryDecision.permits()` returned `True` for it — a green light on a run the engine had declared off-limits.

The contract's own rendering fallback already prints "none (settle required_actions first)" for every status other than `SAFE_TO_RESUME`, so the field contradicted the rendering. The field is echoed by `render`, the dashboard, the resume report and the MCP `next_allowed_action` surface, so an operator read "the next allowed action is to revalidate the dependency" on a run that must not be resumed at all.

Nothing in `src/` calls `permits()` today, so the wrong answer is latent — but #1031 proposes wiring a caller, and wiring one without this fix activates the bug.

## Changes

- `src/continuum/recovery/contract.py` — gate the advertised `next_allowed_action` on the recovery status being one that permits repair (`safety not in {BLOCKED, UNSAFE}`). Contract-level, so every downstream surface stops echoing the action, not just one caller. The plan is still recorded verbatim in `required_actions`; it is simply not *permitted*.
- `src/continuum/recovery/engine.py` — `permits()` returns `False` for everything when the contract is blocked or unsafe, guarded explicitly rather than relying on the field being null (a caller can hold a contract whose field predates the rule). Docstring updated.
- `tests/test_recovery_engine.py` — new test asserting `next_allowed_action is None`, `permits(...) is False` for the action and for an arbitrary string, that the repair still appears in `required_actions`, and that `render` agrees — parametrized over risk-driven ABORT/UNSAFE and ROLLBACK/BLOCKED.

## Verification

The issue's repro at HEAD printed `next_allowed_action : revalidate_dependency:dataset` and `permits(next_allowed): True` under a risk-driven ABORT. On this branch:

```
mode                : abort
safety              : unsafe
next_allowed_action : None
permits(next_allowed): False
```

Local gates: `tests/test_recovery_engine.py` plus the downstream surfaces that echo the field (`test_recovery_guidance`, `test_dashboard`, `test_dashboard_hitl`, `test_mcp_server`, `test_contract_forgery`, `test_contract_observations`, and the resume/assess/decision selection) all pass; `ruff check` and `ruff format --check` clean.

Fixes #1058.

## Summary by CodeRabbit

* **Bug Fixes**
  * Blocked or unsafe recovery runs no longer permit any actions, including repairs listed as required.
  * Recovery contracts now show no permitted next action when safety restrictions apply.
  * Recovery displays now clearly indicate that required actions must be settled before proceeding.

* **Tests**
  * Added coverage for blocked and unsafe recovery scenarios, including action permissions and contract output.


src/continuum/recovery/engine.py
```diff
@@ -158,6 +162,13 @@ def permits(self, action: str) -> bool:
         """
         if self.mode is RecoveryMode.RESUME:
             return True
+        # A blocked or unsafe run permits nothing, not even the repair the
+        # contract's plan happens to name (issue #1058): ROLLBACK and ABORT say
+        # the run must not be continued, so no action is a green light. Guarded
+        # explicitly rather than relying on ``next_allowed_action`` being null,
+        # because a caller can hold a contract whose field predates this rule.
+        if self.contract.recovery_status in {RecoverySafety.BLOCKED, RecoverySafety.UNSAFE}:
+            return False
```
Cyrax321, line 171: Blocking: RESUME branch bypasses the safety guard.

permits returns True for every action when mode is RESUME before checking recovery_status. Fresh assess never yields RESUME with BLOCKED or UNSAFE, but interchange reconstructs mode and contract independently, so a stale or crafted document can produce that combination. That violates the guarantee this PR states.

This causes a hole because the check order trusts mode over safety.

Please check safety first so blocked or unsafe permits nothing in all paths.

```suggestion
        # A blocked or unsafe run permits nothing, not even the repair the
        # contract's plan happens to name (issue #1058): ROLLBACK and ABORT say
        # the run must not be continued, so no action is a green light. Guarded
        # explicitly rather than relying on ``next_allowed_action`` being null,
        # because a caller can hold a contract whose field predates this rule.
        if self.contract.recovery_status in {RecoverySafety.BLOCKED, RecoverySafety.UNSAFE}:
            return False
        if self.mode is RecoveryMode.RESUME:
            return True
```

Cyrax321 (changes requested):
Thank you for closing the permitted step hole, @abyyxhek. Core direction is right, one ordering gap remains.

Ref: #1132 / Fixes #1058
Surface: recovery contract and permits

Author context: abyyxhek is a contributor on recovery. No durable negative signal.

Bug: a blocked or unsafe run advertised a permitted step. ROLLBACK and ABORT coexist with a non empty repair plan, yet both mean do not continue. Naming plan.first as next_allowed_action handed callers gating on permits a green light, echoed in render, dashboard, resume report, and MCP.

Cause: build_contract used plan.first unconditionally, and permits compared against it. Confidence clear. New test reproduces both triggers with plan present.

Provenance: risk path proposing ROLLBACK and ABORT while validator offers repair. Issue #1058 is the report. No other open PR touches this seam.

Best fix: mostly yes. Gating next_action on safety not in BLOCKED and UNSAFE, plus an explicit permits guard that does not rely on the field being null for old contracts, is the right ownership. Plan stays in required_actions, only permission is removed.

Refactor: no.

Proof:
- Worktree tests for blocked and unsafe pass. Diff has no barred characters. CI has no failures.
- Verified safety mapping RESUME to SAFE, ROLLBACK to BLOCKED, ABORT to UNSAFE, so fresh assess is consistent.
- One hole remains, see inline blocking comment at engine.py:163 to 171. Interchange rebuilds mode and contract independently, so RESUME with BLOCKED is reachable there. Safety must be checked before mode.

Risk:
- Once the guard order is fixed, this is ready. No other blocking issue.

Marking REQUEST_CHANGES only for the guard order. Thank you again for making blocked mean blocked.

abyyxhek:
Closing as superseded — #1129 (merged) implements the #1058 fix on main: a BLOCKED or UNSAFE run now advertises no next_allowed_action while keeping the plan in required_actions.

PR #1131, opened 2026-09-15
docs: fix doubled docs/ path for crash-recovery image

wized2:
## Summary
Closes #1113.

`docs/release-notes.md` embedded `docs/assets/crash-recovery.svg`, which resolves to a non-existent `docs/docs/assets/...` from that file. Use relative `assets/crash-recovery.svg` so the image renders.

PR #1134, opened 2026-09-15
fix(docs): resolve the crash-recovery embed relative to its page (#1113)

### LINKED ISSUE #1113: docs/release-notes.md embeds the crash-recovery visual with a doubled docs/ prefix, so the image does not render
## Summary

`docs/release-notes.md:85` embeds the crash-recovery visual with a repo-root-relative path:

```markdown
![Crash recovery: hard kill, refusal, reconcile, resume](docs/assets/crash-recovery.svg)
```

But `release-notes.md` lives inside `docs/`, so the relative URL resolves to `docs/docs/assets/crash-recovery.svg`, which does not exist — the image does not render. The asset is at `docs/assets/crash-recovery.svg`, so the correct relative form from this file is `assets/crash-recovery.svg`.

Every other reference gets this right, which confirms the convention: `README.md:182` and the translated READMEs use `docs/assets/crash-recovery.svg` (correct — they sit at the repo root), and `references/architecture-diagram.md:10` uses `../docs/assets/architecture.svg` (correct — it escapes its directory first).

The surrounding prose mentions of the path (`release-notes.md:81`, `:150`) are in backticks, not links, so line 85 is the only broken embed.

## Affected component

Documentation

## Severity

Minor inconvenience or cosmetic issue — a broken image in the release-notes page.

## System information

- Python: 3.13.5

```bash
ls docs/docs/assets/crash-recovery.svg   # No such file or directory — where the link resolves
ls docs/assets/crash-recovery.svg        # the real asset

grep -n 'crash-recovery.svg' docs/release-notes.md
# :81  backtick prose mention (not a link) — fine
# :85  ![...](docs/assets/crash-recovery.svg) — broken embed
# :150 backtick prose mention (not a link) — fine
```

## What a fix looks like

One line: change the embed at `docs/release-notes.md:85` to `Crash recovery: hard kill, refusal, reconcile, resume`. Optionally, a tiny guard in the spirit of `tests/test_cli_docs.py` / `tests/test_docs_counts.py` that resolves every relative image/link target in `docs/**/*.md` and asserts it exists would prevent the class.

abyyxhek:
## Summary

`docs/release-notes.md:85` embedded the crash-recovery visual as `docs/assets/crash-recovery.svg`. The page lives inside `docs/`, so that relative URL resolves to `docs/docs/assets/crash-recovery.svg`, which does not exist — the image does not render.

Every other reference in the tree gets the relative form right, confirming the convention: `README.md` uses `docs/assets/crash-recovery.svg` (correct, it sits at the repo root) and `references/architecture-diagram.md` uses `../docs/assets/architecture.svg` (correct, it escapes its directory first). The surrounding backtick prose mentions at `:81` and `:150` are not links, so `:85` was the only broken embed.

## Changes

- `docs/release-notes.md` — one line: `(docs/assets/crash-recovery.svg)` → `(assets/crash-recovery.svg)`.
- `tests/test_docs_links.py` — new guard. It resolves every relative link and image target in `docs/**/*.md` against the page that links it and asserts the target exists, so a page that moves or a link that drops a `../` fails CI instead of shipping a dead URL. External URLs, fragments and mailto links are out of scope; fenced code blocks are skipped.

## Verification

At HEAD the new guard fails on exactly the reported page and target:

```
AssertionError: docs\release-notes.md links to non-existent targets: [docs/assets/crash-recovery.svg]
```

With the one-line fix applied, all 55 pages under `docs/` pass. `ruff check` and `ruff format --check` clean.

Fixes #1113.


tests/test_docs_links.py
```diff
@@ -0,0 +1,56 @@
...
+# [text](target) and ![alt](target), capturing the target.
+_LINK = re.compile(r"!?\[[^\]]*\]\(([^)]+)\)")
+_FENCE = re.compile(r"^\s*(```|~~~)")
+
+
+def _pages() -> list[Path]:
+    return sorted(DOCS.rglob("*.md"))
+
+
+@pytest.mark.parametrize("page", _pages(), ids=lambda p: str(p.relative_to(ROOT)))
```
anya-research: The new parametrize adds one case per page under docs/, 55 cases here. CI collects 2310 tests while README and related docs state about 2241, which exceeds the tolerance of 30 in test_docs_counts.py:71. The suite fails on test_documented_count_matches_suite until the three documented totals are re-synced.

tests/test_docs_links.py
```diff
@@ -0,0 +1,56 @@
...
+    lines = page.read_text(encoding="utf-8").splitlines()
+    broken: list[str] = []
+    in_fence = False
+    for line in lines:
+        if _FENCE.match(line):
+            in_fence = not in_fence
+            continue
+        if in_fence:
+            continue
+        for target in _LINK.findall(line):
```
anya-research: The link pattern captures everything up to the closing paren, so a valid titled link yields a target with trailing title text. That combined string never resolves to a file, so a future titled link would fail with a confusing non-existent target message. No current page uses that form, so this changes no current result.

```suggestion
for raw in _LINK.findall(line):
            target = raw.strip()
            if target.startswith("<") and target.endswith(">"):
                target = target[1:-1].strip()
            else:
                target = target.split(None, 1)[0] if target else target
```

anya-research (changes requested):
Thank you for contributing to Continuum, @abyyxhek. This fixes the crash-recovery embed for #1113 and adds a docs link guard, and it overlaps #1131 which makes the same one line docs fix without the guard. There are some changes needed before this can merge.

The embed fix is correct. `docs/assets/crash-recovery.svg` exists and `docs/docs/assets/crash-recovery.svg` does not, so `assets/crash-recovery.svg` from inside `docs/release-notes.md` is the right page relative form. `README.md:182` keeps the root relative form, which matches the convention.

All 55 pages under `docs/` resolve with no broken targets with the fix applied:

- `docs/release-notes.md:85` now points at an existing file
- `docs/release-notes.md:81` and `:150` are backtick prose, not links, so `:85` was the only broken embed
- no title style links exist in `docs/` today, so the guard passes on current content
- `git diff --check` is clean

CI is not green. `Test (ubuntu-latest, Python 3.11)`, `Test (ubuntu-latest, Python 3.12)`, `Test (ubuntu-latest, Python 3.13)`, `Test (macos-latest, Python 3.12)` and `Test (macos-latest, Python 3.13)` conclude FAILURE on the documented count guard, while `Lint and Type-check` and `Test (Postgres backend)` conclude SUCCESS. Please fix the count failure first since it blocks merge.

Two issues, most severe first:

1. `tests/test_docs_links.py:34` breaks the documented count guard.
2. `tests/test_docs_links.py:46` treats a link title as part of the path.

abyyxhek:
Closing — superseded. Issue #1113 (doubled docs/ prefix on the crash-recovery embed) was fixed and closed by merged #1131, which corrected the same path in docs/release-notes.md. The remaining piece of this PR was the docs-links guard; happy to re-open that as its own issue/PR if the guard is still wanted, but it does not belong attached to a closed issue.

PR #1137, opened 2026-09-15
docs(health): a liveness breach proposes WAIT and reaches resume as exit 20

### LINKED ISSUE #1076: health.py claims a liveness breach never moves the recovery mode, but the engine maps it to WAIT and the CLI exits 20
`src/continuum/recovery/health.py` states the liveness contract is advisory in terms that are no longer true. The module docstring (line 7) says "Breach is advisory only, it never moves the recovery mode", and `advisory_for_storage` repeats "Breach is advisory only, never moves recovery mode".

`RecoveryEngine.assess` feeds that same advisory into `_decide`, which proposes `RecoveryMode.WAIT` whenever `breached` is true:

```python
if liveness_advisory is not None and bool(liveness_advisory.get("breached")):
    proposals.append((RecoveryMode.WAIT, f"liveness breach: silence {silence:.1f}s ..."))
```

WAIT has severity 3, so when it is the most cautious proposal it wins, and `exit_code_for(WAIT)` maps to `ExitCode.REQUIRES_HUMAN` (20). A run whose only problem is a silent interval therefore fails the documented "only a fully verified, safe-to-resume run exits 0" contract in `cli/exitcodes.py`, and `continuum resume "$RUN" && ./start-agent.sh` short-circuits.

So the breach is not advisory in the sense the docstrings assert. It is advisory only in the narrower sense the engine's own comment states (never auto-rollback, issue #302): it routes to a human instead of rewinding state. That is a real and load-bearing distinction for an operator deciding whether to ship `.continuum/liveness.json`, because the current wording reads as "enabling this cannot change what the CLI allows", while enabling it can in fact turn a green run into a blocked one.

The fix is to the wording, not the behaviour: state that a breach proposes WAIT (never ROLLBACK), that WAIT surfaces as exit code 20 and blocks `continuum resume`, and that the threshold is therefore an operator-set gate on resumption rather than a passive annotation. Naming the exit code is what makes the trade-off legible.

I found this while auditing the liveness path; `docs/guides/liveness-watch.md` may carry the same wording and is worth checking in the same pass.

Cyrax321:
Closes #1076.

The issue is a wording defect, not a behaviour change. `health.py` states the liveness contract is advisory in terms that are no longer true, and the wording actively misleads the operator it is meant to inform.

## What is wrong

`src/continuum/recovery/health.py:6` says "Breach is advisory only, it never moves the recovery mode", and `advisory_for_storage` repeats "Breach is advisory only, never moves recovery mode". But `RecoveryEngine.assess` feeds that advisory into `_decide`, which proposes `RecoveryMode.WAIT` whenever `breached` is true. WAIT has severity 3, so when it is the most cautious proposal it wins, and `exit_code_for(WAIT)` is `ExitCode.REQUIRES_HUMAN` (20).

Reproduced against this branch:

```
mode                : wait
exit code           : 20
```

So a breach can make `continuum resume "" && ./start-agent.sh` short-circuit. An operator deciding whether to ship `.continuum/liveness.json` reads the current wording as "enabling this cannot change what the CLI allows", when a low threshold can in fact turn a green run into a blocked one. That trade-off is exactly what the docs should be making legible.

## What I changed

- `health.py` module docstring and `advisory_for_storage`: replaced the false "never moves the recovery mode" with what the breach actually does. Advisory in the #302 sense (never a rollback, no state rewritten), but not inert: it proposes WAIT, which reaches `continuum resume` as exit 20. Names the exit code and the chaining consequence explicitly, and states that the threshold is an operator-set gate on resumption rather than a passive annotation.
- `docs/guides/liveness-watch.md`: the resume section called the reading "read-only", and the rules list said "resume decisions stay with the recovery engine". The watch half of that sentence is right, but the resume half misleads, because the engine it defers to is the component that proposes WAIT. Both reworded to say the breach contributes a WAIT proposal (never a rollback) and shows up as exit 20 through resume as well as watch.

## What I did not change

- `cli/main.py` and `UPGRADE_SPEC.md` carry "advisory only" language, but they describe `continuum health` (the prefix-trust score), which genuinely never moves the mode. Accurate, left alone.
- `docs/GLOSSARY.md` and `docs/api/recovery.md` both say the breach "informs" the verdict, which is the correct framing. Left alone.
- The duplicated liveness WAIT block at `engine.py:675` and `:695` is the still-open #1042 and is out of scope here. It does inflate the rationale sentence in my reproduction, but it does not change the mode or the exit code, so it does not affect this fix either way.

## Verification

Reproduced the breach-to-WAIT-to-exit-20 path first, then made the wording match it. Extended `test_engine_maps_breach_to_wait` in `tests/test_liveness_events.py` to assert the exit code, since that link is what the new wording rests on and nothing was pinning it before.

`ruff check` and `ruff format --check` clean, `mypy` clean on health.py, `markdownlint-cli2` clean on the guide, and the docs-count guard passes. Full suite: 2255 tests, 0 failures, 0 errors, 25 skipped.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that `continuum resume` evaluates liveness breaches and may exit with code 20 when a run has been quiet beyond its configured limit.
  * Documented that liveness breaches request a wait rather than trigger rollback.
  * Added examples showing healthy runs exiting successfully and breached runs preventing chained agent startup.

* **Tests**
  * Added coverage confirming liveness breaches produce the expected exit code.


anya-research (approved):
Thank you for contributing to Continuum, @Cyrax321. This corrects the false advisory wording for #1076 to state that a liveness breach proposes WAIT and reaches resume as exit 20.

Note on overlap first: #1140 is also open for #1076 and edits the same two surfaces (`src/continuum/recovery/health.py` and `docs/guides/liveness-watch.md`). The two wordings cannot both land, so the merge choice needs one winner. This PR is the narrower docs touch plus the exit code pin, while #1140 rewrites `evaluate` and `advisory_text` as well.

The wording here matches the implementation exactly. The module docstring and `advisory_for_storage` describe WAIT (never ROLLBACK, per #302) winning when most cautious and mapping to exit 20, and the guide states the same gate on resume with the healthy versus breached samples kept distinct.

- `src/continuum/recovery/engine.py:675` and `:695` propose `RecoveryMode.WAIT` when `breached` is true, with comments stating never auto-rollback
- `src/continuum/cli/exitcodes.py:57` maps `WAIT` to `REQUIRES_HUMAN` (20), and `exit_code_for` is the resume path
- `tests/test_liveness_events.py:65-67` pins `exit_code_for(decision.mode) == 20` on a breached run
- `git diff --check` is clean on this head
- CI is green across the matrix plus lint and type-check: `Lint & Type-check` SUCCESS, `Test (ubuntu-latest, Python 3.11/3.12/3.13)` SUCCESS, `Test (macos-latest)` SUCCESS, `Test (windows-latest)` SUCCESS, `Test (Postgres backend)` SUCCESS, `Markdown lint` SUCCESS

No blocking issues found. This is ready to merge once the duplicate with #1140 is resolved in favor of one wording.

Care in scoping the out of scope notes (health versus liveness, and #1042) kept this small and correct. Thank you for tightening a misleading operator contract.

PR #1140, opened 2026-09-15
docs(liveness): state that a breach proposes WAIT and blocks resume (#1076)

### LINKED ISSUE #1076: health.py claims a liveness breach never moves the recovery mode, but the engine maps it to WAIT and the CLI exits 20

abyyxhek:
## Summary

`src/continuum/recovery/health.py` states the liveness contract is advisory in terms that are no longer true. The module docstring (line 7) said "Breach is advisory only, it never moves the recovery mode", and `advisory_for_storage` repeated "Breach is advisory only, never moves recovery mode".

But `RecoveryEngine.assess` feeds that same advisory into `_decide`, which proposes `RecoveryMode.WAIT` whenever `breached` is true:

```python
if liveness_advisory is not None and bool(liveness_advisory.get("breached")):
    proposals.append((RecoveryMode.WAIT, f"liveness breach: silence {silence:.1f}s ..."))
```

WAIT has severity 3, so when it is the most cautious proposal it wins, and `exit_code_for(WAIT)` maps to `ExitCode.REQUIRES_HUMAN` (20). A run whose only problem is a silent interval therefore fails the documented "only a fully verified, safe-to-resume run exits 0" contract in `cli/exitcodes.py`, and `continuum resume "$RUN" && ./start-agent.sh` short-circuits.

The breach *is* advisory, but only in the narrower, original sense the engine's own comment states (issue #302): it routes to a human instead of rewinding state, because silence says nothing about *what* to roll back. That distinction is load-bearing for an operator deciding whether to ship `.continuum/liveness.json` — the current wording reads as "enabling this cannot change what the CLI allows", while enabling it can in fact turn a green run into a blocked one.

## Changes

Wording only, no behaviour change. `health.py` — the module docstring, `evaluate`, `advisory_for_storage` and `advisory_text` now name WAIT, exit 20 and the resume gate, and keep the "never auto-rollback" sense that was the original intent. `docs/guides/liveness-watch.md` — the "breach is advisory, never a gate" line and the "Watch never gates" rule now state that `resume` reads the same advisory and a breach proposes WAIT there.

## Verification

Confirmed the behaviour the docstrings describe, by backstopping the injected clock in `advisory_for_storage` three hours forward and calling `RecoveryEngine.assess` on an otherwise-clean run:

```
mode           : wait
exit code      : 20
safe           : False
```

`tests/test_liveness_cadence.py` and `tests/test_liveness_events.py` pass; `ruff check` and `ruff format --check` clean.

Fixes #1076.


abyyxhek:
Closing — superseded. The maintainer's merged #1137 (docs(health): a liveness breach proposes WAIT and reaches resume as exit 20) rewords health.py and the liveness-watch guide for issue #1076, the same files and the same correction this PR makes. Issue #1076 is closed. No reason to keep a duplicative docs PR open.

PR #1141, opened 2026-09-15
fix(analysis): score a declared dependency by who declared it, not by its version (#1065)

### LINKED ISSUE #1065: prefix_trust scores agent-declared dependencies as DETERMINISTIC, the self-certification it exists to exclude
## Summary

`prefix_trust` counts an agent-declared external dependency as `DETERMINISTIC` purely because the agent supplied a version string. The module's own docstring says trusted means "recorded by deterministic local code or by a human, not by an autonomous agent reporting on itself", and the dependency in question was recorded by the autonomous agent over the MCP surface. An agent can raise its own trust score by declaring versioned dependencies, which is the self-certification the score exists to expose.

## Affected component

Analysis (advisory trust score)

## Severity

Minor inconvenience or cosmetic issue for the library. The score is advisory and never moves recovery mode or an exit code, so nothing unsafe follows from it; the defect is that a metric labelled as anti-self-certification has a self-certification hole in it.

## Evidence

`src/continuum/analysis/prefix_trust.py:80` appends `Origin.DETERMINISTIC if dep.version else Origin.EXTERNAL_AGENT` to the role fold, and the evidence fold adds `0.9` for any dependency with a version. Neither consults `dep.provenance`. The comment that justifies the shortcut, "ExternalDependency does not carry provenance in the model", is no longer true: `src/continuum/models.py:442` declares `provenance: Provenance` and `src/continuum/state/semantic.py:447` populates it from the declaring event. `src/continuum/mcp/server.py:100` sets `AGENT_SOURCE = Origin.EXTERNAL_AGENT` and writes `DEPENDENCY_DECLARED` with that source, so this is the common path for agent-declared dependencies, not an edge case.

```python
import os, tempfile
from continuum.analysis.prefix_trust import trust_over_prefix

st = SQLiteStorage(db)
st.create_run(Run(run_id='r', goal='agent-supplied goal'))
st.append_event('r', EventType.RUN_STARTED, {'goal': 'agent-supplied goal'}, source=Origin.EXTERNAL_AGENT)
st.append_event('r', EventType.WORK_COMPLETED, {'doc': 0}, source=Origin.EXTERNAL_AGENT)
st.append_event('r', EventType.DEPENDENCY_DECLARED, {'resource': 'dataset', 'version': 'v1'}, source=Origin.EXTERNAL_AGENT)
events = list(st.read_events('r'))
st.close()

state = project('r', events)
dep = state.external_dependencies[0]
print('dependency provenance origin :', dep.provenance.origin)
print('goal provenance origin       :', state.goal.provenance.origin)
print('progress provenance origin   :', state.progress.provenance.origin)
print('trust_over_prefix            :', trust_over_prefix(state))
```

Output at HEAD `14f8a84`, every fact in the run self-asserted:

```
dependency provenance origin : external_agent
goal provenance origin       : external_agent
progress provenance origin   : external_agent
trust_over_prefix            : {'trust_score': 0.528, 'breakdown': {'role': 0.333, 'goal': 0.35, 'evidence': 0.9}}
```

The same run without the dependency declaration scores `role` 0.0, so the entire 0.333 comes from the agent having typed a version.

## Root cause

Introduced in `d33351c` ("feat(analysis): advisory prefix-trust score with mode-invariance guarantee", Refs #401). The dependency arms were written against a model that lacked provenance, and the provenance field was added later without revisiting the shortcut.

## Expected behaviour

A dependency declared by an `EXTERNAL_AGENT` origin should be scored by that origin, the same as a goal or a finding, so an agent cannot buy trust by declaring a version.

## Suggested fix shape

Read `dep.provenance.origin` in both `_collect_origins` and `_score_evidence` and delete the now-false comment about the model lacking the field. If a version-bearing dependency still deserves partial credit for freshness, that credit should be multiplied by whether the declaration is trusted rather than awarded unconditionally.

## System information

- Python: 3.14.5

abyyxhek:
## What

`prefix_trust` counted an agent-declared external dependency as `DETERMINISTIC` purely because the agent supplied a version string. The module's own definition of a trusted fact is one *"recorded by deterministic local code or by a human, not by an autonomous agent reporting on itself"* — so the version was exactly the self-certification the score exists to expose.

## Evidence

A run whose every fact is self-asserted by the agent, with a versioned dependency declaration:

```
dependency provenance origin : external_agent
goal provenance origin       : external_agent
progress provenance origin   : external_agent
trust_over_prefix            : {'trust_score': 0.528, 'breakdown': {'role': 0.333, 'goal': 0.35, 'evidence': 0.9}}
```

The same run *without* the dependency declaration scores `role` 0.0. The entire 0.333 came from the agent having typed a version — an agent can raise its own trust score by declaring versioned dependencies.

Two sites were wrong, and the comment justifying the shortcut was false:

- `_collect_origins` appended `Origin.DETERMINISTIC if dep.version else Origin.EXTERNAL_AGENT` — never consulted `dep.provenance`.
- `_score_evidence` awarded an unconditional `0.9` for `dep.version`.
- The comment read *"ExternalDependency does not carry provenance in the model"*. It has carried it since `state/semantic.py` populated `ExternalDependency.provenance` from the declaring event's origin, and `mcp/server.py` writes `DEPENDENCY_DECLARED` with `Origin.EXTERNAL_AGENT`, so this is the common path for agent-declared dependencies, not an edge case.

## Fix

Read `dep.provenance.origin` in both folds and delete the stale comment. Scale the version-freshness credit by whether the declaration itself is trusted:

- agent-declared with a version: `0.2` (was `0.9`)
- agent-declared without: `0.1` (was `0.3`)
- trusted source, either: unchanged

After the fix the same run scores `role` 0.0 whether or not the dependency is declared — no trust purchased with a version — and a dependency declared by a deterministic source still earns the full `0.9`.

The score stays advisory: it never moves recovery mode, gates a tool call, or changes an exit code, and the mode-invariance tests still pass.

Closes #1065.

## Verification

- Full suite: **2218 passed, 38 skipped**, no failures.
- New regression test `test_an_agent_declared_dependency_cannot_buy_trust_with_a_version` fails on the unpatched code (`assert 0.333 == 0.0`) and passes after.


Cyrax321 (approved):
Thank you for the fix, @abyyxhek. This removes a self certification path correctly.

Ref: #1141 / Fixes #1065
Surface: analysis prefix trust (score by declarer, not version)

Author context: abyyxhek is a contributor on analysis and recovery. No durable negative signal.

Bug: a dependency with a version was scored trusted regardless of who declared it. An agent could raise its own role and evidence scores by typing a version string. A version is pinning metadata, not a certificate of who recorded it.

Cause: _collect_origins and _score_evidence in prefix_trust.py used dep.version as a proxy for trust. Confidence clear. The old comment admitted the approximation and fell back to goal trust.

Provenance: trust scoring from prefix work. Made visible by #1065. No other open PR touches prefix_trust.py.

Best fix: yes. Score by dep.provenance.origin like every other fact. Verified ExternalDependency carries provenance with default, populated in state/semantic.py dependency_declared from the declaring event via _provenance_for. Imports load cleanly. Credit is now 0.9 versioned trusted, 0.3 unversioned trusted, 0.2 versioned untrusted, 0.1 unversioned untrusted, so an unversioned trusted declaration still outranks a versioned agent one. Advisory only, never gates, mode invariance preserved.

Refactor: no. Minimal change at the ownership boundary where the bug belongs.

Proof:
- Worktree at 8fd24c8: tests/test_prefix_trust.py passes.
- New test pins the exact bypass plus every origin by version cell, with role invariance and trust ordering. This is the right regression shape.
- Diff has no barred characters. No prior inline to process.
- CI: only anya-review FAILURE remains, which is workflow config in this repo state, not code. Other legs report success.

Risk: none blocking. Score shifts are intended and covered. Old runs that relied on versioned agent dependencies will score lower, which is the fix.

No blocking issues remain. Marking APPROVED. Ready to pull once workflow permits. Thank you again for tightening trust semantics.

PR #1145, opened 2026-09-15
fix(recovery): fail closed when the consumed-authority ledger cannot be read (#1066)

### LINKED ISSUE #1066: consumed-authority block fail-opens on a transient read error, clearing a blocked run for resume

abyyxhek:
## What

The consumed-authority block in `RecoveryEngine.assess()` fail-**opens**. Its event read sits in a bare `except Exception` that substitutes an empty map — and an empty map is the *unblocked* answer, not an unknown one. When the read fails mid-assessment, the block is skipped and the engine issues a `resume` verdict for a run holding an unreconciled consumed credential.

This block is the one check that keeps `continuum resume` from reporting safe when the gate would still deny. A degraded log is precisely when it matters, and it is the first thing the degradation removes.

## Evidence

A run with an unreconciled consumed credential is normally `request_human` / `requires_human`. Break only the authority scan, leaving every earlier read in `assess()` intact:

```
baseline                  : request_human requires_human
with transient read failure: resume safe_to_resume
  rationale                : ('all state verified against the environment',)
  permits(anything)        : True
```

`src/continuum/recovery/engine.py` — the bare `except` discarded the failure into `consumed_authorities = {}`, so the `if consumed_authorities:` guard saw nothing and the mode computed above stood. The comment three lines above states the block's purpose in the terms it then violates.

## Fix

Distinguish *"no consumed authorities"* from *"could not determine"*. On failure, escalate to the most cautious of the computed mode and `REQUEST_HUMAN`, with a rationale naming the unreadable ledger:

```
with transient read failure: request_human requires_human
  rationale: (..., 'consumed authority block could not be evaluated: the ledger is unreadable (...); assume a consumed credential awaits reconciliation until a human confirms otherwise')
  permits(anything): False
```

This keeps the engine's established convention that degradation never raises, without asserting a safety conclusion the engine could not compute. Escalation takes a max over severity, so it never weakens an already-cautious verdict — a risk-driven `ABORT` (severity 6) survives and is not downgraded to `REQUEST_HUMAN` (4).

Scope is deliberately the authority block only. The sibling bare `except`s elsewhere in `assess()` degrade informational folds where an empty answer is genuinely neutral; this one gates a security-relevant block.

Closes #1066.

## Verification

- Full suite: **2220 passed, 38 skipped**, no failures.
- New tests in `tests/test_recovery_engine.py`: the block blocks on a successful read, fails closed on an unreadable ledger (fails on unpatched code), and never downgrades a harder verdict.

## Summary by CodeRabbit

* **Bug Fixes**
  * Recovery assessments now require human review when consumed-authority records cannot be verified.
  * Unreadable authority records no longer incorrectly report a run as safe.
  * Higher-severity outcomes, such as abort decisions, remain unchanged when authority records are unavailable.
  * Recovery contracts now correctly prevent permitted actions when unresolved consumed authorities are detected.


src/continuum/recovery/engine.py
```diff
@@ -468,14 +468,30 @@ def assess(
...
             rationale = (
                 *rationale,
-                f"consumed authority blocks resume: {sorted(consumed_authorities)}",
+                "consumed authority block could not be evaluated: the ledger is "
+                f"unreadable ({exc}); assume a consumed credential awaits "
+                "reconciliation until a human confirms otherwise",
             )
+        else:
+            if consumed_authorities:
+                mode = RecoveryMode.REQUEST_HUMAN
```
Cyrax321: Blocking: success branch can downgrade a harder verdict.

The failure branch above uses max(mode, REQUEST_HUMAN) so escalation never weakens an already cautious verdict. This branch does a blind assignment to REQUEST_HUMAN. When a run has both a readable consumed authority and a risk driven ABORT (severity 6 outranks REQUEST_HUMAN 4), this downgrades ABORT to REQUEST_HUMAN, which is the exact weakening the PR description forbids.

This causes an unsafe downgrade because mode before this block can already be ABORT via risk_mode through _decide.

Please use the same max so a consumed authority never downgrades an ABORT verdict.

```suggestion
                mode = max((mode, RecoveryMode.REQUEST_HUMAN), key=lambda m: SEVERITY[m])
```

Cyrax321 (changes requested):
Thank you for the fix, @abyyxhek. The fail closed direction is right, with one inconsistency to resolve.

Ref: #1145 / Fixes #1066
Surface: recovery engine assess, consumed authority block

Author context: abyyxhek is a contributor on gate and recovery. No durable negative signal.

Bug: the consumed authority block in assess used bare except with empty map fallback. An empty map is the unblocked answer, so a read failure skipped the block and left a resume verdict for a run holding an unreconciled credential.

Cause: try around collect_consumed_authorities in engine.py with except Exception substituting {}. Confidence clear. The PR reproduces baseline request_human flipping to resume safe_to_resume on transient failure, with rationale losing the block.

Provenance: authority lifecycle block from #289c work, fail open pattern predates this PR. No other open PR touches this block. Issue #1066 is still OPEN.

Best fix: mostly yes. Distinguishing unknown from empty and escalating to max(mode, REQUEST_HUMAN) with a rationale naming the unreadable ledger preserves the degrade never raises convention without asserting safety. The max preserves ABORT 6 over REQUEST_HUMAN 4, and the new test guards that for the failure path.

Refactor: no. Scope to the authority block only is correct. Sibling bare excepts elsewhere are out of scope.

Proof:
- Worktree at 7cfd19a: tests/test_recovery_engine.py, 44 passed.
- New tests cover blocked resume, unreadable fails closed with contract REQUIRES_HUMAN and permits false, and unreadable never downgrades ABORT.
- CI: 33 SUCCESS, 0 FAILURE. Diff has no barred characters.
- Ordering verified: risk_mode is computed before _decide, so mode entering the authority block can already be ABORT.

Risk:
- The success branch still downgrades. See inline blocking comment at engine.py:490. A readable consumed authority plus a risk ABORT would be reduced to REQUEST_HUMAN. Please apply the same max there and extend the downgrade test to the readable path.
- Once that line is fixed, this is ready. The core fail closed behavior is correct and well tested.

Marking REQUEST_CHANGES only for the success branch max. Thank you again for catching the fail open.

src/continuum/recovery/engine.py
abyyxhek: Applied in 8ec14b3 — thank you for catching this, and for the fail-closed direction on the rest.

The success branch now takes the same max over severity:

```python
mode = max((mode, RecoveryMode.REQUEST_HUMAN), key=lambda m: SEVERITY[m])
```

so a readable consumed authority raises the floor of the verdict without lowering its ceiling, and the `consumed authority blocks resume` rationale is still appended so the block stays visible in the contract.

Extended the downgrade guard to the readable path as `test_a_readable_ledger_never_downgrades_a_harder_verdict`: a readable ledger plus a risk-driven ABORT stays `ABORT` while the rationale still names the block. It fails on the pre-fix code (`REQUEST_HUMAN` where `ABORT` was expected), so it covers the case the unreadable-path test missed.

abyyxhek:
### Reviewer feedback addressed

The one blocking item from the review (success branch can downgrade a harder verdict) is fixed in 8ec14b3.

- **Blocking inline comment — success branch blindly assigned `mode = RecoveryMode.REQUEST_HUMAN`** → Now `mode = max((mode, RecoveryMode.REQUEST_HUMAN), key=lambda m: SEVERITY[m])`, mirroring the failure branch above it. A readable consumed authority escalates to at least `REQUEST_HUMAN` but can no longer weaken a risk-driven `ABORT` (6) down to `REQUEST_HUMAN` (4). `src/continuum/recovery/engine.py`.
- **"Extend the downgrade test to the readable path"** → Added `test_a_readable_ledger_never_downgrades_a_harder_verdict`: a readable ledger plus a risk `ABORT` stays `ABORT` and the rationale still names the consumed-authority block. `tests/test_recovery_engine.py`.

Everything else in the PR is unchanged: the fail-closed failure branch, its rationale line, and the original three tests are untouched.

### Validation

- `ruff check src/ tests/ examples/` — clean
- `ruff format --check src/ tests/ examples/` — 316 files already formatted
- `mypy src/continuum` — no issues (127 source files)
- Full suite — 2,259 collected, exit 0 (no failures); skips vary by environment as the counts guard notes
- New test verified to **fail** on the pre-fix engine (`REQUEST_HUMAN` where `ABORT` was expected) and pass after, so it guards the exact path the review identified
- Merge check: no conflicts — `recovery/engine.py` and `test_recovery_engine.py` are untouched on `main` since the merge base

Cyrax321 (changes requested):
Author context: abyyxhek, external contributor. Follow-up to the prior review. The fail-closed direction for #1066 is still right, but the readable-path downgrade gap still blocks: the success branch can still reduce a risk ABORT to REQUEST_HUMAN. Apply the same max(mode, REQUEST_HUMAN) escalation there as on the failure path and extend the downgrade test to cover the readable path. Rebase out of dirty state and confirm the recovery engine suite green.

abyyxhek:
Closing as superseded — issue #1066 (fail closed on an unreadable consumed-authority ledger) was fixed on `main` by #1125, merged 2026-09-19. This branch is now 43+ commits behind and conflicts with the merged fix, so I'm dropping it rather than rebasing duplicate work onto a red base. Happy to reopen if #1125 turns out not to cover a case this one did.

Issue #556, opened 2026-08-31
[#289] Gate resurrection check and atomic reconciliation for authorities


Cyrax321:
### Affected component

Gate / Action ledger

Parent epic: #289. See execution board #550 for wave context and dependency chain.

### What problem does this solve?

Gate forwards claims without checking AUTHORITY_CONSUMED before the restore point, so a consumed token is resurrected. Uncertain consumption also blocks neither gate nor ledger atomically.

### Proposed solution

* Gate (+ gateway) loads AUTHORITY_CONSUMED set before forwarding; if `authority_id` consumed at seq < restore point, deny with `Authority <id> consumed at seq <n> by run <r>` and emit audit event
* reconciler path: uncertain action consuming an authority blocks both until `reconcile --authority` or probe settles them atomically (one transaction)
* Restore semantics: `RECOVERY_STARTED` vs checkpoint seq defines the resurrection window

Depends on: [#289] event type.

Files: `src/continuum/gate.py`, `src/continuum/gateway.py`, `src/continuum/actions/ledger.py`, tests `tests/test_authority_resurrection.py` (consume -> checkpoint -> restore -> reuse denied; then reconcile -> passes).

### Alternatives considered

Check only in recovery engine. Rejected: gate is the enforcement boundary, recovery is advisory.

### Is this a breaking change?

No

### Would you like to work on this?

Yes, I have a plan

PR #1148, opened 2026-09-15
fix(gate): refuse a consumed authority nested in the argument structure (#1074)

### LINKED ISSUE #1074: Authority resurrection check scans top-level argument values only: a spent authority nested one level deep passes the claim, the gate, and the gateway
## Summary

The Authority Resurrection defense (issue #289b, shipped via #556) is bypassable by nesting the spent authority one level deep inside any tool argument. The README advertises it as a hard guarantee:

> Consumed-grant tracking: Single-use authority references are marked spent at terminal status; **reuse after restore is refused (`GRANT_DENIED`), defending the checkpoint-restore path against Authority Resurrection**

That refusal does not hold. All three enforcement points only inspect **top-level** argument values:

- **gate** (`src/continuum/gate.py:277`): scans top-level string values of `tool_input` — a value nested in a dict or a list is invisible.
- **ledger claim** (`src/continuum/actions/ledger.py:911-915`): even weaker — it looks for the authority only under four hard-coded top-level argument names (`authority_id`, `authority`, `token`, `approval_id`).
- **gateway** (`src/continuum/gateway.py:189-192`): the same top-level value scan as the gate.

So a call whose spent authority sits at top level is refused at all three points, but the identical call with the authority wrapped as `{"payment": {"auth_token": "cred-1"}}` — a completely ordinary argument shape for a payment/credentials tool — passes the claim, passes the gate, and fires.

The gate check was deliberately made **value-based rather than field-name based** "so that drift in argument names does not resurrect spent authority" (`gate.py:272-275`). Argument-shape drift (nesting) defeats it exactly the way field-name drift was designed not to. The code even admits it: `gate.py:296-298` says the top-level loop "is sufficient for the tested shapes."

## Affected component

`continuum/gate.py` (`decide` authority check), `continuum/actions/ledger.py` (`claim` authority-resurrection check), `continuum/gateway.py` (request-path authority check)

## Severity

Security: a documented safety guarantee is defeated by trivial argument reshaping, no tooling or exploitation required — an LLM agent that naturally structures its arguments nested (or a prompt injection steering it to) resurrects a consumed credential. The denial direction must be fail-closed; this is fail-open on shape.

## Evidence

At upstream HEAD `915f1ab`, one run, one consumed authority `cred-1`:

**1. The claim accepts a nested spent authority** (the ledger scan only reads 4 top-level field names):

```python
record_authority_consumed(store, "r1", "cred-1")
led.claim("invoice_write",
          {"invoice": "INV-2", "payment": {"auth_token": "cred-1"}},   # nested
          key="INV-2")
# -> ACCEPTED, status: started        (top-level {"authority_id": "cred-1"} would raise)
```

**2. The gate allows the claimed call.** Full chain through the real CLI hook (`continuum gate`, as the Claude Code PreToolUse hook invokes it), with `.continuum/gate.json` registering `pay_invoice`:

```
payload {"invoice": "INV-2", "payment": {"auth_token": "cred-1"}}   # nested in a dict
  -> exit=0  [ok] allow: live claim 'INV-2'
payload {"invoice": "INV-2", "payment": {"auth": ["cred-1"]}}       # nested in a LIST
  -> exit=0  [ok] allow: live claim 'INV-2'
payload {"invoice": "INV-2", "token": "cred-1"}                     # top level (control)
  -> exit=2  [!!] deny: Authority 'cred-1' consumed at seq 2 by run 'r1'. Obtain a fresh authority.
```

The side effect fires with the spent authority — the exact scenario #289b exists to prevent.

```python
import tempfile
from pathlib import Path
from continuum.actions.authority import record_authority_consumed
from continuum.actions.ledger import ActionLedger, fold_action_events
from continuum.gate import collect_consumed_authorities, decide

db = tempfile.mkdtemp() + "/auth.db"
store = SQLiteStorage(db)
store.create_run(Run(run_id="r1", goal="bypass repro"))
store.append_event("r1", EventType.RUN_STARTED, {"goal": "bypass repro", "total": 5})
record_authority_consumed(store, "r1", "cred-1")

config = {"pay_invoice": {"key_template": "{invoice}", "action_type": "invoice_write"}}
led = ActionLedger(store, "r1")
led.claim("invoice_write",
          {"invoice": "INV-2", "payment": {"auth_token": "cred-1"}},
          key="INV-2")                                   # ACCEPTED: nested authority invisible

history = store.read_all_events("r1")
d = decide(config, "pay_invoice",
           {"invoice": "INV-2", "payment": {"auth_token": "cred-1"}},
           run_id="r1",
           actions_by_key=fold_action_events(history),
           consumed_authorities=collect_consumed_authorities(history))
assert d.allow is True   # <-- spent authority resurrected; "live claim 'INV-2'"
```

## Root cause

Both scans stop at argument depth 1. The gate's loop `for _key, _value in tool_input.items()` (`gate.py:277`) checks only direct string values, and the ledger's detection (`ledger.py:911-915`) additionally requires the authority to appear under one of four known field names. `render_key`'s docstring says "Only top-level argument fields are supported in v1" — a documented v1 limit for **key rendering**, which the authority checks silently inherited as an undocumented v1 limit on **enforcement**.

## Expected behaviour

A consumed authority must be refused regardless of where in the argument structure it appears. Per the value-based design philosophy already stated in `gate.py:272-275`, any argument string equal to a consumed authority id should deny — including values nested in dicts and lists.

## Suggested fix shape

One shared helper (in `gate.py`, next to `normalize_key_value` which is already shared with the gateway for exactly this reason) that walks the argument structure to a bounded depth and collects all string values; use it in `decide()`, in the gateway's request path, and to broaden the ledger claim's authority detection (replacing the hard-coded four-name top-level scan, which is weaker still). Depth/size-bounded to keep hostile payloads from turning the check into a DoS. Red-first regression tests: the dict-nested and list-nested cases above must deny at the gate and raise at claim.

## System information

- OS: macOS 25.0 (Darwin 25.0.0)
- Python: 3.13.5
- continuum-agent: from source at upstream HEAD `915f1ab`
- Installed via: editable install with dev extra

abyyxhek:
## What

The Authority Resurrection defense (#289b, shipped via #556) is bypassable by nesting the spent authority one level deep inside any tool argument. The README advertises it as a hard guarantee — *"reuse after restore is refused (`GRANT_DENIED`)"* — but all three enforcement points scanned only **top-level** argument values.

## Evidence

One run, one consumed authority `cred-1`, claimed with a live key so the only possible denial is the authority check itself:

```
PRE-FIX                                 WITH FIX
claim (dict-nested) : ACCEPTED  <-- BUG   claim (dict-nested) : refused -> LedgerError
claim (list-nested) : ACCEPTED  <-- BUG   claim (list-nested) : refused -> LedgerError
gate  dict-nested   : allow=True          gate  dict-nested   : allow=False
gate  list-nested   : allow=True          gate  list-nested   : allow=False
gate  top-level     : allow=False         gate  top-level     : allow=False
```

`{"invoice": "INV-2", "payment": {"auth_token": "cred-1"}}` is a completely ordinary argument shape for a payment tool — no exploitation required, just natural argument structuring.

The three sites, all depth-1:
- `gate.py` `decide` — `for _key, _value in tool_input.items()`
- `gateway.py` — the same loop over `body.values()`
- `actions/ledger.py` `claim` — strictly weaker still: four hard-coded top-level names (`authority_id`, `authority`, `token`, `approval_id`)

The checks were made value-based rather than field-name based *"so that drift in argument names does not resurrect spent authority"* (`gate.py`). Argument-**shape** drift defeats them exactly the way field-name drift was designed not to, and the code already conceded the point — the comment it left behind said the top-level loop *"is sufficient for the tested shapes."*

## Fix

One shared, bounded walk in `gate.py`, used by all three enforcers so they cannot disagree about depth:

- `collect_argument_values(arguments)` — every string value at any depth, recursing through mappings and sequences. Bounded (`max_depth=8`, `max_values=10_000`) so a hostile payload cannot turn a safety check into unbounded work.
- `find_consumed_authority(arguments, consumed)` — the first spent authority found anywhere in the structure.
- `consumed_authority_reason(id, event)` — the shared denial message, so it cannot drift between the three sites.

`decide` and the gateway now use the shared scan in place of their top-level loops; the ledger's four-name scan is replaced outright, since it was strictly weaker than the other two. The live-retry carve-out is unchanged: a live mid-flight attempt under the same key and authority still passes, matching the grant rule.

Closes #1074.

## Verification

- Full suite: **2227 passed, 38 skipped**, no failures.
- New regression tests in `tests/test_gate_resurrection.py` (gate + ledger, dict/list/deep nesting, a false-positive guard, and the bounds of the walk) and `tests/test_gateway.py` (gateway, dict/list nesting). All fail on unpatched code.


Cyrax321 (approved):
Thank you for the contribution, @abyyxhek. This closes a real bypass correctly.

Ref: #1148 / Fixes #1074
Surface: gate, gateway, ledger claim (authority resurrection)

Author context: abyyxhek is a contributor focused on recovery and gate fixes. No durable negative signal.

Bug: the authority resurrection defense refused reuse after restore only for top level argument values. Nesting the spent authority one level deep ({"payment": {"auth_token": "cred-1"}}, an ordinary payment shape) passed claim, gate, and gateway. The ledger was weaker still, matching only four hard coded top level names.

Cause: three depth 1 scans. gate.py decide looped tool_input.items, gateway looped body.values, ledger checked four names. Confidence clear. The PR description reproduces all three, and the old comment conceded top level was sufficient for the tested shapes.

Provenance: top level enforcement introduced by 852464a (gate blocks resurrection). Made visible by #1074. Carried forward through the ledger grant checks. No other open PR touches these paths for nesting.

Best fix: yes. One shared bounded walk in gate.py (collect_argument_values, find_consumed_authority, consumed_authority_reason) used by all three enforcers so they cannot disagree about depth. Value based and shape agnostic, which preserves the original design intent from #289b (drift in names must not resurrect) and extends it to shape drift. Bounds (depth 8, 10k values) keep a hostile payload from turning the check into unbounded work. Shared denial message prevents drift between sites. Deferred imports avoid a gate and ledger cycle. Verified imports load cleanly.

Refactor: no larger refactor needed. The ownership is right. Gate owns the scan, gateway and ledger reuse it.

Proof:
- Worktree at c9bb4e3: tests/test_gate_resurrection.py plus tests/test_gateway.py, 36 passed.
- New tests pin the exact gap: parametrized dict nested, list nested, deeply nested, top level for decide, plus ledger claim refusal with GRANT_DENIED audited, plus fresh authority false positive guard, plus bounded walk test for depth, width, and scalar handling.
- CI: 33 SUCCESS, 0 FAILURE. No em dash violation in the diff.
- Blast radius is contained to the three enforcers and their tests.

Risk:
- Residual truncation bypass noted by prior bot review: a payload with more than 10k filler strings visited before the real credential could miss the credential because the walk breaks at the cap (DFS order). This requires attacker control of shape plus a large padded body. Main is bypassable with one nested value, so this PR strictly raises the bar from trivial to expensive. I do not block on it. Follow up options are inline membership testing with early exit, deterministic breadth ordering, or fail closed on truncation with a size guard. None is required for #1074.
- Performance note on ledger reading the full log for any Mapping claim is correct behavior. The old skip was the bug (it skipped exactly the nested cases). If this path becomes hot, cache the consumed set per run. Not required now.
- mergeStateStatus is BLOCKED only for lack of approval. No failing checks.

No blocking issues remain. Marking APPROVED. Ready to pull once required workflow permits. Thank you again for tightening this guarantee.

PR #1151, opened 2026-09-15
fix(recovery): run_with_limits detaches the runaway worker instead of joining it

### LINKED ISSUE #1147: recovery(limits): run_with_limits joins the runaway it just timed out, so the timeout bounds nothing
### Summary

`run_with_limits(fn, timeout=t)` is the mechanism the docs point operators at when a probe might hang, and its own module docstring states the goal plainly: *"A runaway probe must be terminated, not hung."* It is not terminated, and the caller is hung. The `RecoveryTimeoutError` is raised only after the runaway finishes, because the `with ThreadPoolExecutor(...)` block's `__exit__` calls `shutdown(wait=True)`.

### Affected component

### Severity

### System information

- OS: Windows 11 Home 22H2
- Python: 3.11.9

```bash
python - <<'PY'
import sys, time
sys.path.insert(0, "src")
from continuum.recovery.limits import run_with_limits, RecoveryTimeoutError

def slow_probe() -> str:
    time.sleep(6)
    return "done"

start = time.monotonic()
try:
    run_with_limits(slow_probe, timeout=1.0)
except RecoveryTimeoutError as exc:
    elapsed = time.monotonic() - start
    print(f"{exc}  -- caller saw it after {elapsed:.1f}s, timeout was 1.0s")
PY
```

### Current behaviour

```
recovery operation timed out after 1.0s  -- caller saw it after 6.0s, timeout was 1.0s
```

The error message is correct and the exception type is right, so nothing downstream can tell the difference between "the probe was stopped at 1s" and "the probe ran to completion at 6s and then we reported a timeout". Instrumenting the two moments makes the gap explicit:

```python
timeout raised at 4.00s   # probe was scheduled to finish at 4.00s; timeout was 0.5s
```

The mechanism is the context manager, not the `future.result(timeout=...)` call, which behaves correctly. That call raises `TimeoutError` on time; then leaving the `with` block runs `ThreadPoolExecutor.shutdown(wait=True)`, which joins the worker:

```python
# src/continuum/recovery/limits.py:39
with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
    future = executor.submit(fn, *args, **kwargs)
    try:
        return future.result(timeout=timeout)
    except concurrent.futures.TimeoutError as exc:
        raise RecoveryTimeoutError(...) from exc
# <-- __exit__ runs shutdown(wait=True) here, joining the still-running worker
```

Python cannot kill a thread, so the worker finishes its sleep regardless — that is unavoidable. What is avoidable is the caller joining it: the exception should be raised and the worker left to die on its own.

### Expected behaviour

The timeout should bound *when the caller gets control back*, which is the only thing the API promises and the only thing the docstring advertises: *"a `RecoveryTimeoutError` is raised if it does not complete in time. The caller can decide how to handle the timeout."* Today the caller cannot decide anything until the probe is done.

Detaching rather than joining is the whole fix — the worker is a daemon-less orphan either way, but the caller is free:

```python
executor = concurrent.futures.ThreadPoolExecutor(max_workers=1)
future = executor.submit(fn, *args, **kwargs)
try:
    return future.result(timeout=timeout)
except concurrent.futures.TimeoutError as exc:
    executor.shutdown(wait=False)   # do not join the runaway
    raise RecoveryTimeoutError(...) from exc
finally:
    executor.shutdown(wait=False)
```

Leaked workers are the known and accepted cost of stopping a thread you cannot kill, and it is strictly better than the current behaviour: a probe written to hang forever (`lock.acquire()` with no timeout, an infinite loop) blocks the caller forever *through the very API that exists to bound it*. Note the `wait=False` in both branches — the `finally` must not join either, or a successful fast call followed by a slow second one in the same interpreter would still hang.

A test is worth adding alongside it, because the suite currently cannot catch this. `tests/test_recovery_limits.py:8` is named `test_runaway_op_is_terminated_not_hung` and asserts only that the exception is raised — which passes today, since it is raised, late. Timing the interval between the deadline and the raised exception is what pins the behaviour:

```python
start = time.monotonic()
with pytest.raises(RecoveryTimeoutError):
    run_with_limits(lambda: time.sleep(2.0), timeout=0.1)
assert time.monotonic() - start < 1.0
```

### Why this matters

`docs/threat_model.md` lists *"Unbounded resource exhaustion by adversarial probes"* under what CONTINUUM does not protect against, and its mitigation sentence is: *"Callers that run untrusted probes should pass an explicit timeout."* An operator who follows that advice and passes one is no safer than one who does not — the pause is the same, only the exit code differs. The `timeout` argument reads as a bound on the operation, and it is a bound on nothing.

The reconciler path does not use this helper (it runs `subprocess.run(..., timeout=...)` directly at `reconcilers.py:159`, which terminates correctly), so `continuum reconcile` is unaffected. The exposure is the public `continuum.recovery.run_with_limits` export plus any caller that wires a probe through it, and it is the documented answer for the threat the threat model names.

### Logs or CI run

All output above is from direct invocation on commit `2e835ad`.

Parthipashok04:
## Summary

`run_with_limits(fn, timeout=t)` did not bound anything: the `with ThreadPoolExecutor(...)` block's `__exit__` ran `shutdown(wait=True)`, so the caller joined the very runaway the timeout was meant to stop and saw `RecoveryTimeoutError` only after the probe finished — or never, for a probe that hangs forever (`lock.acquire()` with no timeout, an infinite loop).

The fix is to detach rather than join. Python cannot kill the worker thread, so it finishes (or hangs) on its own either way; what changes is that the caller now gets control back at the deadline, which is the only thing the API promises and the docstring advertises.

## Changes

- `src/continuum/recovery/limits.py`: create the executor bare instead of as a context manager, and run `executor.shutdown(wait=False, cancel_futures=True)` in a `finally` so **neither** the timeout path nor the success path joins the worker. Docstring updated to state the worker is detached, not killed.
- `tests/test_recovery_limits.py`: two new tests —
  - `test_timeout_bounds_when_caller_gets_control_back` times the interval between the deadline and the raised exception; the old suite could not catch this bug because it only asserted the exception was raised (which it was, late).
  - `test_success_path_does_not_join_worker` covers the fast-success path.

## Verification

Issue reproducer, before → after:

```
before: recovery operation timed out after 1.0s  -- caller saw it after 6.0s, timeout was 1.0s
after:  recovery operation timed out after 1.0s  -- caller saw it after 1.0s, timeout was 1.0s
```

- `tests/test_recovery_limits.py`: 5 passed
- full recovery test selection (`pytest tests/ -k recovery`): 136 passed

The reconciler path is unaffected (it uses `subprocess.run(..., timeout=...)` directly).

Fixes #1147

## Summary by CodeRabbit

- **Bug Fixes**
  - Recovery operations now return a timeout error promptly when a probe exceeds its configured limit, instead of waiting for the underlying operation to finish.
  - Successful recovery operations complete without unnecessary delays caused by lingering background work.

- **Tests**
  - Added coverage confirming timeout deadlines are respected and successful operations return promptly.


src/continuum/recovery/limits.py
```diff
@@ -29,16 +29,29 @@ def run_with_limits(
     and a ``RecoveryTimeoutError`` is raised if it does not complete in time.
     The caller can decide how to handle the timeout. Using a thread avoids
     signal restrictions on non main threads and works on all platforms.
+
+    On timeout the worker thread is not killed — Python cannot do that — it is
```
anya-research: The docstring uses em dashes on lines 33 to 34. AGENTS.md bars that character in code comments and docs. Lint and type-check stay green with it present, so CI does not catch this convention.

```suggestion
On timeout the worker thread is not killed (Python cannot do that), it is
    detached. The exception reaches the caller at the deadline, not when the
    runaway happens to finish.

This is ready once the windows leg goes green. Your timing test pins the exact deadline behavior that the old suite missed.
```

anya-research (dismissed):
Thank you for contributing to Continuum, @Parthipashok04. This makes `run_with_limits` detach the timed out worker instead of joining it for #1147.

No duplicate work applies. #1147 is still OPEN and no other open PR touches `src/continuum/recovery/limits.py` or `tests/test_recovery_limits.py`, the closest recovery PRs touch engine, contract, risk and restore instead.

The implementation matches its docstring. Both the timeout path and the success path fall through to `shutdown(wait=False, cancel_futures=True)`, so the caller gets control at the deadline and the worker stays detached as documented.

- diff is 2 files, 42 insertions and 6 deletions, confined to limits and its tests
- both public names in `limits.py` carry docstrings
- `git diff --check` is clean
- grep shows no product caller of `run_with_limits` outside its tests, only a reexport

CI is mostly green. `Lint & Type-check` is SUCCESS and the ubuntu, macos and Postgres test legs are SUCCESS. `Test (windows-latest, Python 3.11)` is FAILURE, it guards the suite on that platform, please rerun or fix that leg before merge.

One low item remains. 1. `src/continuum/recovery/limits.py:33` uses characters barred by AGENTS.md.

Parthipashok04:
The windows-latest py3.11 failure was not caused by this PR: `test_table_regenerates_from_runner_no_invented_numbers` timed out because it runs `benchmarks/run.py` (the full horizon + fault-injection + crash-recovery suite) with a 60s subprocess timeout — too tight on a cold Windows runner. The same failure appeared on unrelated PRs (e.g. CI run 35004733889 on `fix/issue-1068-readme-counts`).

Pushed f9e6f63, which raises that timeout to 600s — the same budget the `--publish` test in the same file already grants the identical script. Verified locally: the test plus the full `test_recovery_limits.py` suite pass.

src/continuum/recovery/limits.py
Cyrax321: Blocking: em dashes violate repo convention.

AGENTS.md bars U+2014 in code comments and docs. This line uses two em dashes ("not killed -- Python cannot do that -- it is detached" in the current text uses the barred character). Lint and type-check stay green with it present, so CI does not catch it.

This causes a convention failure because the project explicitly forbids that character in this surface.

Please change the line so that it uses commas or parentheses and preserves meaning.

```suggestion
    On timeout the worker thread is not killed (Python cannot do that), it is
```

Cyrax321 (changes requested):
Thank you for the fix, @Parthipashok04. This addresses the timeout bound correctly.

Ref: #1151 / Fixes #1147
Surface: Core recovery limits (src/continuum/recovery/limits.py)

Author context: Parthipashok04 is an external contributor (CONTRIBUTOR). Co-authored with Claude Code. No durable negative signal. Prior bot reviews flagged only style and a design suggestion.

Bug: run_with_limits(fn, timeout=t) did not bound anything. future.result(timeout) raised on time, but leaving the with ThreadPoolExecutor block ran shutdown(wait=True), which joined the runaway worker. The caller saw RecoveryTimeoutError only after the probe finished, or never for a hung probe.

Cause: the context manager exit path in src/continuum/recovery/limits.py:39 (main). Confidence: clear. Reproduced locally on main: timeout 0.1s, caller blocked 3.01s. Fixed logic returns in 0.11s. The mechanism matches the issue analysis exactly.

Provenance: introduced by c442f52 (2026-08-21, Cyrax321) which added run_with_limits with the with executor pattern. Made visible by #1147. No other open PR touches limits.py.

Best fix: yes. Creating the executor bare and calling shutdown(wait=False, cancel_futures=True) in finally is the smallest correct change. It restores the documented promise (caller gets control at the deadline) and correctly notes the worker is detached, not killed. Python cannot terminate a thread, so detach is the only available semantic in this seam.

Refactor: no. CodeRabbit suggested a shared executor with a bounded semaphore. That would be worse here. A hung probe would permanently consume a slot, so repeated timeouts would exhaust capacity and eventually block new calls. Process isolation would truly terminate a probe, but that is a larger design change and out of scope for #1147. The current orphaned thread is the known accepted cost, already stated in the docstring and comments.

Proof checked:
- Local reproducer on main: 3.01s delay confirmed. Fixed logic: 0.11s.
- New tests pin the exact gap the old suite missed: test_timeout_bounds_when_caller_gets_control_back times deadline to exception, test_success_path_does_not_join_worker covers the fast path.
- Codecov reports all modified lines covered. limits.py is 100 percent in CI coverage output.
- CI: Lint and type-check SUCCESS, ubuntu (3.11, 3.12, 3.13) SUCCESS, macos SUCCESS, Postgres SUCCESS, windows 3.11 and 3.13 SUCCESS after f9e6f63. Remaining windows 3.12 FAILURE is tests/test_serve_http.py::test_the_server_keeps_serving_after_a_non_object_body (TimeoutError, socket read), unrelated to this diff. Same class of flake noted by gitar-bot. The anya-review FAILURE legs are config (missing GH_TOKEN), also unrelated.
- No product caller of run_with_limits outside tests, only a reexport in recovery/__init__.py, so blast radius is minimal.

Risk and notes:
- Residual: each timeout orphans one worker thread until it exits. For hung forever probes this leaks a thread. Acceptable for the current call volume (no product callers yet), but worth a follow up note in docs if this path becomes hot.
- tests/test_horizon_benchmark.py timeout 60 to 600 is out of scope for #1147, but acceptable here. It matches the 600s the --publish test already grants the same script and unblocks the windows leg. In future, please split CI stability bumps into a separate PR.
- No milestone is set (repo has no open milestones). Labels benchmark and tests are present.

1. src/continuum/recovery/limits.py:33 uses characters barred by AGENTS.md. See inline comment with suggestion. Once that line is fixed, and the windows 3.12 leg is rerun green (or confirmed flake with a passing rerun), this is ready.

I am marking REQUEST_CHANGES only for the convention fix and a clean CI signal. The core logic is correct and the tests are the right shape.

Parthipashok04:
@Cyrax321 Thanks for the review. Pushed 392e8e9 with the convention fix: the two em dashes on the `run_with_limits` docstring line are now parentheses, using the wording from your suggestion.

```diff
-    On timeout the worker thread is not killed — Python cannot do that — it is
+    On timeout the worker thread is not killed (Python cannot do that), it is
```

A grep for U+2014 across `src/continuum/recovery/limits.py` and this PR's test files now returns nothing. Ruff and mypy on limits.py are clean. The commit is docstring wording only, so the core diff is unchanged.

Verified locally on Windows / Python 3.12, the leg that was failing:
- `tests/test_recovery_limits.py`: 5 passed
- `tests/ -k recovery`: 155 passed
- `tests/test_serve_http.py::test_the_server_keeps_serving_after_a_non_object_body` passes here too, consistent with the earlier windows 3.12 failure being a socket-read flake rather than something in this diff.

The push re-ran the full matrix including the windows 3.12 leg.

Cyrax321 (approved):
Thank you for the fix, @Parthipashok04. This closes the timeout gap in #1147 correctly.

Ref: #1151 / Fixes #1147
Surface: Core recovery limits (src/continuum/recovery/limits.py)

Author context: Parthipashok04 is an external contributor. No durable negative signal.

Bug: run_with_limits(fn, timeout=t) did not bound the caller. future.result(timeout) raised on time, but exiting the with ThreadPoolExecutor block ran shutdown(wait=True), which joined the runaway worker. The caller saw RecoveryTimeoutError only after the probe finished, or never for a hung probe.

Cause: the context manager exit path in src/continuum/recovery/limits.py:39 on main. Confidence: clear. I reproduced the shape locally with plain ThreadPoolExecutor semantics: old shape, 0.1s timeout on a 3.0s probe, caller blocked 3.01s. New shape with shutdown(wait=False, cancel_futures=True) in finally, caller returned in 0.10s.

Provenance: introduced by c442f52 (2026-08-21, Cyrax321) which added run_with_limits with the with executor pattern. Made visible by #1147. Confidence: clear via git log -S and blame.

Best fix: yes. Creating the executor bare and calling shutdown(wait=False, cancel_futures=True) in finally is the smallest correct change. It restores the documented promise (caller gets control at the deadline) and the updated docstring states the worker is detached, not killed, which is the only available semantic since Python cannot terminate a thread.

Refactor: no. CodeRabbit proposed a shared executor with a bounded semaphore. I am declining that here. A hung probe would hold a slot forever, so repeated timeouts would exhaust capacity and block new calls. Process isolation would truly terminate a probe but is a larger design change out of scope for #1147. The orphaned thread is the known accepted cost, already noted in the code comment.

Proof checked:
- Live timing shape above (3.01s vs 0.10s).
- New tests pin the gap the old suite missed: test_timeout_bounds_when_caller_gets_control_back times deadline to exception, test_success_path_does_not_join_worker covers the fast path.
- Read limits.py, recovery __init__ reexport, callers (no product caller outside tests), neighbouring tests, and the full three commit diff.
- CI on latest commit 392e8e9: Lint and Type-check SUCCESS, all ubuntu, macos, windows Test legs SUCCESS, Postgres SUCCESS, codecov/patch SUCCESS, Gitar SUCCESS. The anya-review FAILURE leg is config (missing GH_TOKEN), unrelated to this diff. No open milestones in the repo.

Risk and notes:
- Residual: each timeout orphans one worker thread until it exits. For hung forever probes this leaks a thread. Acceptable at current call volume (no product callers yet). Worth a docs note if this path becomes hot.
- tests/test_horizon_benchmark.py 60s to 600s is out of scope for #1147 but acceptable here. It matches the 600s the publish test already grants the same script and unblocks the Windows leg. Please split CI stability bumps into a separate PR in future.
- Prior bot items are resolved: the em dash convention item is fixed in 392e8e9 (verified no U+2014 in the PR file), the shared executor item is declined above with reason.

No blocking issues remain. The implementation satisfies #1147 as described. Marking ready to pull once required reviews are recorded. Thank you again for the contribution.

PR #1162, opened 2026-09-15
fix(docs): drop the invented monitored_commands key from the Codex guide

### LINKED ISSUE #1161: docs: embed-codex guide offers a monitored_commands key in .continuum/config that no code reads
## Summary

`docs/guides/embed-codex.md:111` offers operators a durability alternative when they cannot route file writes through Bash:

> the durability alternative is `continuum observe` via the adapter's `checkpoint_node` or via **`monitored_commands` in `.continuum/config`** (not covered here)

Neither half of that second option exists. No code reads a `.continuum/config` file, and no `monitored_commands` key is defined or consulted anywhere in `src/`, `tests/`, or the other docs:

```bash
grep -rn "monitored_commands" src/ tests/    # no hits
grep -rn "\.continuum/config" src/           # no hits
```

The `.continuum/*.json` files the code actually reads are `budgets.json`, `gate.json`, `gateway.json`, `liveness.json`, `reconcilers.json`, `resume.json`, and `webhooks.json` — no `config`.

The first half of the sentence is real: `checkpoint_node` exists on the LangChain/LangGraph adapters (`src/continuum/adapters/langgraph.py:263`, `langchain.py:248`). So the guide is half-right, and the invented key is the one an operator reaches for when the shell-routing workaround above it does not apply — exactly the case the surrounding paragraph is about, since Codex cannot route `apply_patch` through Bash.

This is the same family as #1144 (a doc naming a surface the CLI does not have), but a different file and a different mechanism, so it is filed separately rather than folded in.

## Affected component

Documentation

## Severity

Minor inconvenience or cosmetic issue — an operator following the guide's fallback reaches for a config key that was never implemented.

## System information

- Python: 3.13.5

```bash
grep -rn "monitored_commands" src/ tests/ docs/
# docs/guides/embed-codex.md:111: ... or via `monitored_commands` in `.continuum/config` ...

grep -rn "\.continuum/config" src/
# (no hits — no loader exists for this file)
```

## What a fix looks like

Either remove the `monitored_commands` clause so the guide offers only the `checkpoint_node` route that works, or — if the key was intended as a real feature — that is an enhancement rather than a docs fix, and the guide should stop naming it until it ships. A `git log -S monitored_commands` on the guide will show whether the key was ever wired and later dropped, or only ever written into the doc.

Cyrax321:
Closes #1161.

## What was wrong

`docs/guides/embed-codex.md:111` offered an operator who cannot route a file write through Bash a durability alternative in two parts:

> the durability alternative is `continuum observe` via the adapter's `checkpoint_node` or via `monitored_commands` in `.continuum/config`

The second half never existed. No code reads a `.continuum/config` file, and `monitored_commands` is defined and consulted nowhere in `src/`, `tests/`, or the other docs. `git log -S monitored_commands` on the guide returns a single commit, the embeddability sprint (9e655be), so the key was written into the doc and never wired, never wired and later dropped.

The `checkpoint_node` half is real, verified at `src/continuum/adapters/langgraph.py:263` and `langchain.py:248`.

This is the fallback an operator reaches for precisely when the shell-routing workaround above it does not apply, which is the `apply_patch` case the surrounding paragraph is about. Codex cannot route `apply_patch` through Bash, so the invented key is the one the guide hands the reader in the case it was written for.

## The fix

Rewrote the sentence to offer only the `checkpoint_node` route, and named the two adapter source files so the claim is checkable rather than asserted. Also corrected the mechanism: `checkpoint_node` captures state from inside your agent's own code, so it does not depend on Codex traversing the tool call at all, which is the actual reason it works as a fallback here.

## Guard

This is the second docs-names-a-surface-that-does-not-exist issue, so I added a regression guard in `tests/test_docs_counts.py` rather than fixing the one line and leaving the class of bug open. Every `.continuum/<file>` path named in `README.md`, `docs/`, and `references/` must appear as a literal somewhere under `src/`. Every real file does (budgets, gate, gateway, liveness, reconcilers, resume, webhooks, risk-policy, mcp-policy, the precompact pair). I verified the guard is not vacuous in both directions: it fails when the removed sentence is reintroduced, and passes on the fixed guide.

A failure has exactly two honest resolutions, and only one is a code change: wire the loader the prose promises, or fix the prose. An illustrative rather than real path can go in the test's `ALLOWLIST` with a reason, since an unguarded illustration is how this read as a feature.

## Verification

`pytest --no-cov -m "not slow"`: 2222 passed, 26 skipped. `ruff check`, `ruff format --check`, and `mypy src/continuum` all clean.

Prose-only change to the guide; no runtime or behaviour change. `CHANGELOG.md` entry added under Unreleased / Fixed.

## Summary by CodeRabbit

- **Documentation**
  - Updated the Codex embed guide to recommend the functional `checkpoint_node` route for durable state handling.
  - Removed references to the unsupported `monitored_commands` fallback.
  - Refreshed documented test-suite counts across the main documentation, translated READMEs, and testing references.
  - Recorded the documentation updates in the changelog.

- **Tests**
  - Added validation for documented configuration paths and expanded checks for test-count references across documentation.


Cyrax321 (commented):
Thank you.

Removing the invented key instead of wiring it is the right fix. Provenance checks out: git log shows the key arrived with the embed sprint and no loader ever read it, while checkpoint_node is real in both adapters. The guard is the durable part, every .continuum path in prose must appear in src, verified to fail on the removed sentence and pass on the fix. That keeps the second occurrence from becoming a third.

I reviewed the guide edit, the CHANGELOG entry, and the new guard. One Windows check reports failure in test_table_regenerates_from_runner_no_invented_numbers with a 60s timeout running benchmarks/run.py. That test is unrelated to this docs-only change and the other 33 checks are green, so I read it as a slow-runner flake, not a merge blocker. Worth a retry on CI.

Minor note from the bot on guard precision (any literal counts, even a write-only or comment mention) is fair hardening for a follow-up, not blocking here.

No blocking issues in this change. Ready when you are, pending a green retry on that Windows leg. (Note: GitHub does not allow approving your own PR, so this is filed as a comment review.)

Cyrax321 (commented):

Ref: #1162 / Issue #1161
Surface: docs/guides/embed-codex.md, docs count guards

Bug: the guide offered monitored_commands in .continuum/config as a durability alternative. No code reads that file and the key is defined nowhere, and git log -S shows it arrived with the embeddability sprint and was never wired. The other half of the same sentence (checkpoint_node) is real and verified in the LangGraph and LangChain adapters.

Fix verified at head 0492a0a: the invented key is gone from the guide; the only remaining mention is the regression test that guards its absence, which is correct. The earlier ~0 passed, ~0 skipped defect introduced by the re-sync commit is also fixed at head (README.md:102 now reads ~2,406 collected, ~2,379 passed, ~27 skipped), and the guard in tests/test_docs_counts.py now covers the pass and skip figures. All three bot threads are resolved with verdicts recorded.

Residual, non-blocking: the translated READMEs still carry stale counts in places (ja, ko, and zh still say ~2,402 in three spots each, and es retains older pass/skip figures). This predates the PR's re-sync scope and does not contradict the removed key, but the count guard should be widened to translations in a follow-up; #1259 already does exactly that.

Cyrax321 (commented):
Maintainer re-review at current head. The invented monitored_commands removal plus the prose-path guard is still the right shape. Overlaps #1209 on the same issue #1161; only one should land, prefer the one with the docs guard plus changelog. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review. Rebase out of dirty state and retry the Windows slow-runner leg.

Issue #1144, opened 2026-09-15
docs: README and references/cli.md say every command accepts --json, but the trailing position is rejected with exit 2


abyyxhek:
### Summary

`README.md` and `references/cli.md` state **"Every command accepts `--json`"**. That sentence is the one a reader reaches for when they want machine-readable output, and following it literally fails: `--json` is a global flag that must precede the subcommand, and the trailing position is rejected with exit code 2. The translated READMEs carry the same sentence, so the misstatement is not language-specific.

### Affected component

Documentation

### Severity

Minor inconvenience or cosmetic issue

### Regression check

- [x] This worked in an earlier version or commit

### System information

- OS: Windows 11 Home 22H2
- Python: 3.11.9
- continuum-agent: from source, commit `2e835ad` on `main`, editable install `uv pip install -e ".[dev]"`

### Command or code to reproduce

```bash
# The form the docs' own command block invites, next to `continuum resume <run_id> --env dataset=v4`
continuum resume run_1 --json
echo "exit=$?"

# What actually works
continuum --json resume run_1
echo "exit=$?"
```

Tested across every read-only command the docs name:

```bash
for c in "runs" "status run_1" "inspect run_1" "history run_1" "events run_1" \
         "validate run_1" "resume run_1" "budget run_1" "tree run_1" "briefing" \
         "actions run_1" "show-contract run_1" "replay run_1" "verify run_1" \
         "provenance run_1" "gate" "impact run_1 --evidence e1"; do
  continuum --db t.db $c --json >/dev/null 2>&1; echo "trailing: $c -> $?"
done
```

### Current behaviour

Every one of those commands exits 2:

```
usage: continuum [-h] [--version] [--db DB] [--json] [--color | --no-color] ...
continuum: error: unrecognized arguments: --json
```

Only `health` and `watch` accept the trailing form, because they were given their own subparser-level `--json` with an `argparse.SUPPRESS` default (fix for #677, merged as #678). The comment at `cli/main.py:3847` states the purpose directly: *"accepts trailing --json without shadowing the global flag"*.

So the sentence is true for 2 of ~40 commands in the position a reader naturally tries it, and the docs never say the position matters.

### Expected behaviour

The project already knows the answer and wrote it down — in `docs/api/cli.md:100-103`:

> `--db` ... and `--json` ... are global flags, so they go before the command: `continuum --json runs`, **not `continuum runs --json`, which is rejected as an unrecognised argument.**

[...]

PR #1165, opened 2026-09-16
docs: clarify --json is global and must precede the subcommand

wized2:
## Summary
Advances #1144.

README / `references/cli.md` implied every command accepts trailing `--json`. Clarified that `--json` is a **global** flag that must appear **before** the subcommand (e.g. `continuum --json resume RUN`), matching CLI parsing.


anya-research (approved):
Thank you for contributing to Continuum, @wized2. This corrects the `--json` placement in README and `references/cli.md` for #1144, from trailing to global before the subcommand.

The new wording matches the parser and the canonical doc.

- `src/continuum/cli/main.py:3707` defines `--json` on the top level parser before subcommands, with trailing copies only on health and watch for #677
- `docs/api/cli.md:100-103` already states global before the command with the same example form, so both edited files now agree with it
- `git diff --check` is clean and the diff touches only the two intended lines
- No open PR duplicates these two files for the same issue, and #1144 remains open so this advances it

CI is green across the test matrix plus Markdown lint, title check and docs-only detection, with lint and type-check skipped as docs-only.

Appreciated the tight example that mirrors the real parser behavior.

PR #1167, opened 2026-09-16
docs: remove duplicated scope paragraph in idempotency_key

### LINKED ISSUE #1156: docs: the scope paragraph of idempotency_key's docstring is duplicated verbatim
## Summary

The `scope` paragraph of `idempotency_key`'s docstring appears twice, word for word, at `src/continuum/actions/idempotency.py:111-113` and again at `:115-117`:

```python
    """Derive a stable key identifying this operation.

    ``scope`` narrows the key, typically to a run, so two runs performing
    the same logical operation do not deduplicate against each other within that
    ledger. A narrower scope (for example, a run id) is what most callers want.

    ``scope`` narrows the key, typically to a run, so two runs performing
    the same logical operation do not deduplicate against each other within that
    ledger. A narrower scope (for example, a run id) is what most callers want.

    Note: ``scope`` only shapes the derived key. ...
```

Cosmetic, but it is the docstring of the function that defines deduplication semantics, and a duplicated paragraph suggests a merge or rebase left two copies of an edit — which makes a reader wonder what else diverged.

## Affected component

Documentation

## Severity

## System information

- Python: 3.13.5

```bash
sed -n '109,118p' src/continuum/actions/idempotency.py
```

## What a fix looks like

Delete one copy, keeping the first, and check whether the two paragraphs were ever *intended* to differ — if one was meant to describe `scope` and the other the `volatile` or `key` parameter, the fix is to write the missing paragraph rather than delete it. `volatile` and `key` are both documented only by the trailing `Note:` paragraph, so the function may genuinely be under-documented rather than merely over-copied. A quick `git log -L` on the docstring will say which.

wized2:
## Summary
Closes #1156.

The `scope` paragraph in `idempotency_key`'s docstring was repeated twice. Kept a single copy.


Cyrax321 (approved):

This PR removes the duplicated scope paragraph in idempotency_key and closes #1156. I verified the duplication on main at src/continuum/actions/idempotency.py:111-117, checked the diff at 02bcde9 which deletes only the second copy, and traced the duplication to 76a14bc which inserted a second identical paragraph when rewriting the Note for unscoped claims.

No behavior change, docstring only. CI is fully green including the test matrix, Postgres backend, and lint and type-check, and Codecov reports the touched lines covered.

No blocking issues. Marking as ready to pull.

Thank you again for cleaning this up.

PR #1176, opened 2026-09-16
fix(cli): resume --pinning drift reads full history, not the live tail (#1126)

### LINKED ISSUE #1126: resume --pinning drift reads the live tail only, so a compacted run reports every pin as newly pinned
## Summary

`continuum resume --pinning` computes its drift display from the live tail only. `cli/main.py:1413` feeds `latest_pinning(storage.read_events(run_id))`, and `latest_pinning` (`src/continuum/pinning.py:78`) scans `ACTION_RECORDED` payloads for the newest non-empty `pinning` dict. After `continuum compact` moves the pre-anchor prefix into `events_archive` (`storage/sqlite.py:486`), every pinning-carrying `ACTION_RECORDED` event is gone from the live tail — the anchor marker that remains carries only `anchored_through`/`version`, no pinning — so `latest_pinning` returns `{}` and `pinning_drift({}, current)` reports **every key as newly pinned**, even when nothing changed.

Verified end-to-end on a run whose recorded pinning is identical to the pinning passed at resume:

```
BEFORE compact:
  recorded: {'model_id': 'm-2024-09', 'prompt_sha256': 'aaaa…'}
  drift: []          # correct — nothing drifted
AFTER compact:
  recorded: {}
  drift: ['pinning drift: model_id newly pinned (m-2024-09...)',
          'pinning drift: prompt_sha256 newly pinned (aaaaaaaaaaaaaaaa...)']
```

The masked directions are the ones an operator would act on: a genuinely *changed* hash renders as "newly pinned", and a key that was unpinned never renders at all (the `unpinned (was ...)` line needs the old value from the empty record). The display is informational-only (issue #241), but it is wrong in the direction of hiding drift — the opposite of what a drift report is for.

This is the same class as #1050 (assess) and #1072 (watch), but a separate read site: the pinning fold in `cmd_resume`.

## Affected component

continuum.cli (cmd_resume pinning drift display); continuum.pinning

## Severity

Minor inconvenience — informational output, no gating effect, but silently wrong after compaction.

## System information

- Python: 3.13.5

```python
import tempfile, os
from continuum.pinning import latest_pinning, pinning_drift

store = SQLiteStorage(os.path.join(tempfile.mkdtemp(), "pin.db"))
store.append_event("run_1", EventType.ACTION_RECORDED, {
    "key": "deploy-1", "action_id": "act-1", "action_type": "external",
    "status": "started", "external_id": None,
    "pinning": {"prompt_sha256": "a" * 64, "model_id": "m-2024-09"},
})
current = {"prompt_sha256": "a" * 64, "model_id": "m-2024-09"}

print(pinning_drift(latest_pinning(store.read_events("run_1")), current))
# [] — correct

store.compact_run("run_1")

print(pinning_drift(latest_pinning(store.read_events("run_1")), current))
# ['pinning drift: model_id newly pinned (m-2024-09...)', ...] — wrong

print(pinning_drift(latest_pinning(store.read_all_events("run_1")), current))
# [] — the fix
```

## What a fix looks like

One-word change at `cli/main.py:1413`: `latest_pinning(storage.read_all_events(run_id))` — `read_all_events` (`storage/base.py:135`) folds the archived prefix back in, ordered by sequence, so "newest non-empty pinning" stays correct across compaction. A regression test in the style of the #1050/#1072 ones — record an action with pinning, compact, assert `resume --pinning` with the identical dict reports no drift — pins it.

Cyrax321:
## Description

`continuum resume --pinning` computed its drift display by folding `latest_pinning` over `storage.read_events(run_id)`, the live tail only. After `continuum compact` moves the pre-anchor prefix into `events_archive`, every `ACTION_RECORDED` event carrying a pinning is gone from the live tail, and the anchor marker that remains carries only `anchored_through`/`version`. So the fold returned `{}` and `pinning_drift({}, current)` reported every key as newly pinned, even on a run whose recorded pinning was identical to the pinning passed at resume.

The masked directions were the ones an operator would act on: a genuinely changed hash rendered as "newly pinned", and a key that had actually been unpinned never rendered at all, because the "unpinned (was ...)" line needs the old value from the record the fold could no longer see. The display is informational only (issue #241) and gates nothing, but it was wrong in the direction of hiding drift, which is the opposite of what a drift report is for.

The fix is one read site in `cmd_resume`: `read_all_events` folds the archived prefix back in, ordered by sequence, so "newest non-empty pinning" stays correct across compaction. This is the same read-site class as #1050 (assess) and #1072 (watch), but a separate call site, which is why it was missed when those were fixed.

## Related issues

Fixes #1126

## Types of changes

- Type: `bugfix`

## Breaking changes

No. The display is informational only, and the corrected output strictly matches what a non-compacted run already reported.

## How has this been tested?

Reproduced the exact scenario from the issue first, then confirmed the fix removes it:

```
python /tmp/repro_1126.py      # issue's repro script
live tail: []
after compact, read_events:     ['pinning drift: model_id newly pinned (m-2024-09...)', ...]
after compact, read_all_events: []
```

Then, on `origin/main` in an isolated worktree, Python 3.13.5, installed via `pip install -e ".[dev]"`:

```
pytest tests/test_pinning.py -q                                    # 11 passed
pytest tests/test_pinning.py -k after_compaction                   # 2 passed
pytest tests/test_compaction.py tests/test_compact_twice.py \
  tests/test_provenance_compaction.py tests/test_cli.py \
  tests/test_cli_docs.py tests/test_pin_surface.py \
  tests/test_docs_counts.py -q                                     # all passed
pytest -q                                                          # full suite, exit 0
ruff check src/ tests/ examples/                                   # all checks passed
ruff format --check src/ tests/ examples/                          # 316 files already formatted
mypy src/continuum                                                 # no issues in 127 source files
```

Two new regression tests in `tests/test_pinning.py` pin both directions over a compacted run: identical pinning reports no drift, and a changed hash still reports "changed" rather than "newly pinned". They assert the pinning event really landed in `events_archive` and not the live tail before resuming, so the test cannot pass vacuously on a tree where compaction never moved the event. Both fail on the pre-fix code with the issue's exact symptom ("newly pinned" where "changed" is correct) and pass after.

## Checklist

Tests added or updated for the changed behaviour. Documentation updated where the change affects user-facing behaviour (CHANGELOG `[Unreleased]` under Fixed). CI passes on the latest commit. Security-relevant changes state known limitations explicitly in this description (not security-relevant: informational output only, no gating, no writes).

## Additional context

The CHANGELOG entry sits under `[Unreleased] > Fixed`. The documented collected-test figure in README and CONTRIBUTING_ONBOARDING moves by two, within the `test_docs_counts.py` tolerance of 30, so those files were not re-synced.


Cyrax321 (commented):
Thank you.

Reading full history in cmd_resume is the right fix for the same read-site class as assess and watch. Folding the live tail alone returns empty after compaction, which both false-positives identical pinning and hides the real changed and unpinned directions. That the display is informational does not soften it, hiding drift is the opposite of the report purpose.

I reviewed the one read-site change and the two regression tests. Asserting the pinning event is actually in the archive guards against a vacuous pass, and both directions are pinned. CI is green.

Cyrax321 (commented):
Maintainer re-review at current head. The resume pinning full-history read is still correct with both drift directions pinned. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review with no blocking issues at last review. Rebase out of dirty state and confirm CI green.

Issue #731, opened 2026-09-07
GLOSSARY.md code pointers drifted from the lines they cite


Cyrax321:
## Context

`docs/GLOSSARY.md` pins definitions to implementation lines (`src/continuum/models.py:555` style), but several pointers have drifted as the code moved: for example `models.py:555` now points at a `@classmethod` line, `models.py:163` at a blank line, and `models.py:601` at unrelated code. Spot-checked on current `main`; the full list needs auditing.

## Why this matters

The glossary's whole promise is that definitions are verifiable, not inferred. A pointer aimed at the wrong line breaks that promise and teaches readers to distrust the rest.

## How to reproduce

```bash
python - <<EOF
import re
from pathlib import Path
text = Path("docs/GLOSSARY.md").read_text(encoding="utf-8")
for m in sorted(set(re.findall(r"(src/continuum/[a-z_/]+\.py):(\d+)", text)))[:40]:
    path, line = m
    src = Path(path).read_text(encoding="utf-8").splitlines()
    print(f"{path}:{line} :: {src[int(line)-1].strip()[:60]}")
EOF
```

Compare each printed line against what the entry claims it points to.

## Expected outcome

Every `path:line` pointer in the glossary resolves to the code the entry describes. Entries whose targets moved get corrected line numbers; entries whose targets no longer exist get re-anchored to the right symbol.

## Step by step for the fix

1. Run the script above on `main` and check all pointers, not just the examples.
2. Fix the drifted ones.
3. Run markdownlint on the file.

## Files to touch

- `docs/GLOSSARY.md` (pointers only)

Mechanical and verifiable, ideal for a first contribution. Pairs well with #681 (new glossary terms).

PR #1177, opened 2026-09-16
docs(glossary): re-anchor 16 stale code pointers and guard them (#1069)

### LINKED ISSUE #1069: docs/GLOSSARY.md code pointers drifted from the lines they cite (16 of 25)
## Summary

`docs/GLOSSARY.md` pins each definition to a `src/.../file.py:LINE` pointer, and 16 of the 25 pointers no longer land on the code the entry describes. The `models.py` cluster is the worst: `models.py:1060` is cited for `RecoveryContract` but holds `class EnvResource`, `models.py:1133` is cited for `StateCheckpoint` but holds a field assignment, `models.py:172` is cited for `Origin` but holds `class ApprovalStatus`, and the `EnvironmentSnapshot` and `EnvResource` pointers at 1024 and 1014 are off by 48 and 46 lines respectively. `RecoveryLedger` is cited at `recovery/ledger.py:205`, which now holds `_UNSAFE_FILENAME_CHARS`.

This is a recurrence of #731, which fixed the same class of drift and asked for a full audit rather than a spot check. The audit is what found these.

## Affected component

Documentation

## Severity

Minor, but it defeats the file's purpose. The glossary's promise is that every definition is verifiable in the code, not inferred; a reader who follows a pointer to an unrelated line stops trusting the ones that are right.

## Evidence

Each row: what the entry claims, what the cited line actually contains, and where the symbol really is.

| GLOSSARY line | Cited pointer | Actually at that line | Real location |
|---|---|---|---|
| 6 | `recovery/ledger.py:88` | `run_id: str` | `anchor: bool` at `:95` |
| 9 | `recovery/ledger.py:208` | filename-sanitising comment | `RecoveryLedger.__init__` at `:284`, `lock` at `:286` |
| 12 | `models.py:1060` | `class EnvResource(BaseModel):` | `RecoveryContract` at `models.py:1114` |
| 12 | `recovery/contract.py:83` | `mode="json", exclude={...}` | `def build_contract` at `:88` |
| 15 | `models.py:172` | `class ApprovalStatus(StrEnum):` | `class Origin` at `models.py:192` |
| 15 | `models.py:94` | `FAILED = "failed"` (an enum member) | `class StateStatus` at `models.py:97` |
| 18 | `models.py:1133` | `evidence: list[str] = Field(...)` | `class StateCheckpoint` at `models.py:1188` |
| 18 | `checkpoint/manager.py:168` | comment on replay scope | `def checkpoint(` at `:174` |
| 21 | `models.py:1024` | `raise ValueError("caused_by ...")` | `class EnvironmentSnapshot` at `models.py:1072` |
| 21 | `models.py:1014` | comment on caps | `class EnvResource` at `models.py:1060` |
| 24 | `state/validator.py:218` | `lines.append(f"Reason: ...")` | `class StateValidator` at `:222` |
| 24 | `recovery/impact.py:47` | blank line | `class DependencyGraph` at `:48` |
| 27 | `recovery/ledger.py:205` | `_UNSAFE_FILENAME_CHARS = ...` | `class RecoveryLedger` at `:281` |
| 30 | `recovery/planner.py:111` | blank line | `class RepairPlan` at `:112` |
| 42 | `recovery/engine.py:320` | `)` | `check_admissibility(...)` call at `:324` |
| 48 | `reconcilers.py:289` | blank line | `def probe_authority_verdict` at `:291` |

Still correct and left alone: `checkpoint/policy.py:62` and `:52`, `recovery/cleanup.py:20`, `concurrency/lease.py:58`, `security/provenance.py:19`, `provenance_map.py:60`, `adapters/actions.py:24` and `:53`, `models.py:227`.

    python - <<'EOF'
    import re
    from pathlib import Path
    text = Path("docs/GLOSSARY.md").read_text(encoding="utf-8")
    for path, line in sorted(set(re.findall(r"(src/continuum/[\w./]+\.py):(\d+)", text))):
        src = Path(path).read_text(encoding="utf-8").splitlines()
        n = int(line)
        print(f"{path}:{line} :: {src[n-1].strip()[:70] if 0 < n <= len(src) else '<out of range>'}")
    EOF

Compare each printed line against what the surrounding glossary prose claims it points to.

## Expected behaviour

Every `path:line` pointer in the glossary resolves to the code the entry describes. Targets that moved get corrected line numbers; targets that no longer exist get re-anchored to the right symbol.

## Suggested fix shape

Re-anchor the 16 drifted pointers using the table above, then add a small test that resolves every pointer in `docs/GLOSSARY.md` and asserts each lands within a few lines of the symbol named in its entry's backticks. Pointers have now drifted twice (#731 and here), so a mechanical check is what stops a third round.

## System information

- Python: 3.14.5

Cyrax321:
Closes #1069.

16 of the 25 code pointers in docs/GLOSSARY.md no longer landed on the code their entry describes. models.py:1060 was cited for RecoveryContract but holds EnvResource, recovery/ledger.py:205 was cited for RecoveryLedger but holds _UNSAFE_FILENAME_CHARS, and the EnvironmentSnapshot and EnvResource pointers were off by 48 and 46 lines. Every definition in that file promises the reader a line they can open and verify, so a pointer to an unrelated line defeats the file's purpose and trains readers to distrust the pointers that are still right.

This is the second occurrence (#731 fixed the same class of drift and asked for a full audit, which is what found these), so the fix is not just the re-anchor. tests/test_docs_glossary.py parses every src/continuum/*.py:LINE pointer in the file, asserts the pointer set matches an expected table exactly, and asserts each cited line lands within three lines of the symbol its entry names. A pointer edited or added without a table entry fails on the set comparison; a symbol that moves in the source fails on the per-pointer check. I confirmed both failure modes by deliberately re-staleing a pointer locally.

The expected table stores symbols and how each is declared (class/def, assignment, or a bare reference such as a call site), never line numbers: the resolver re-finds the symbol each run, so code moving does not rot the test the way it rotted the glossary.

Three notes for review. The issue's evidence table named engine.py:324, reconcilers.py:291, and a ledger lock at :286, but those were measured against HEAD 14f8a84; on current main the real targets are engine.py:352, reconcilers.py:297, and :288, so I anchored to what the code actually holds today. The 9 pointers that were still correct were left untouched. And this PR adds 26 tests, so the documented counts in README.md, docs/CONTRIBUTING_ONBOARDING.md, and CHANGELOG.md are re-synced from 2,241 to 2,267 collected (2,216 to 2,242 passed); the new tests need no optional extras and pass on a minimal env. I also corrected a separate pre-existing inaccuracy in the line I was already editing: README said 161 test files and there are 171.


Cyrax321 (commented):
Thank you.

Re-anchoring plus a guard is the right response for the second occurrence of this drift. Storing symbols and kinds in the expected table instead of line numbers is the durable shape, the resolver re-finds each run so code moves do not rot the test the way they rotted the glossary. Anchoring to current main line numbers instead of the issue HEAD values is correct, and leaving the 9 good pointers untouched keeps the diff honest.

I reviewed GLOSSARY.md, the new test_docs_glossary.py, and the count resync. CI is green.

Cyrax321 (commented):
Maintainer re-review at current head. The glossary re-anchor plus symbol guard is still the right shape. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review with no blocking issues at last review. Rebase out of dirty state and confirm CI green.

PR #1178, opened 2026-09-16
docs(cli): correct the read-only classification of resume and watch (#1175)

### LINKED ISSUE #1175: cli: `continuum resume` is documented read-only, but a blocked run appends NOTIFICATION_FAILED to the event log
## Summary

`references/cli.md:55-57` lists `resume` among the read-only commands and states they "never write, so they are safe against a live database while an agent is mid-run". `README.md:498` makes the same guarantee. `cmd_resume`s own docstring opens "Read-only unless ``--repair`` is given".

That is not true. When a `resume` returns `REQUEST_HUMAN` and `.continuum/webhooks.json` configures an endpoint for that mode, `cmd_resume` calls `_notify_blocked_run` **before** the `--repair` branch, and `notify_blocked` appends a `NOTIFICATION_SENT` or `NOTIFICATION_FAILED` event to the runs event log (`src/continuum/recovery/webhooks.py:297` and `:304`). The write happens on the plain, flag-less invocation — `--repair` is not required.

Reproduced at `d10ee63`:

```
$ continuum --db run.db resume blocked
warning: http://127.0.0.1:1/never-listens: delivery failed after 3 attempt(s); dead-letter recorded
$ # event log: RUN_STARTED, ACTION_RECORDED, NOTIFICATION_FAILED
```

The run went from 2 events to 3 on a command the docs promise never writes.

## Why it matters

The guarantee is load-bearing, not cosmetic: the docs whole point is that an operator can run `continuum resume` against a live database while an agent is mid-run, trusting the read to be side-effect free. A second append from the same command is exactly the collision the promise exists to rule out — and the appended event lands in the hash-chained log the agents own recovery will later fold and verify. The design intent in the comment ("the verdict above is already final", "never raises") is sound; only the read-only classification is wrong.

Also worth noting: `notify_blocked`s dedup key is `(run_id, mode, contract hash)`, so a *second* identical `resume` correctly stays silent — the write happens once per distinct verdict. The first one still writes.

## Affected component

CLI — `src/continuum/cli/main.py` (`cmd_resume`, `_notify_blocked_run`); documented in `references/cli.md` and `README.md`.

## Severity

Minor correctness / trust. No data is corrupted and the appended event is a legitimate audit record; the defect is that a command classified as read-only mutates the log it claims not to touch.

## System information

- Python: 3.13.5

```bash
mkdir -p /tmp/ronly/.continuum
cat > /tmp/ronly/.continuum/webhooks.json <<EOF
{"endpoints": [{"url": "http://127.0.0.1:1/never-listens", "events": ["request_human"]}]}
EOF
cd /tmp/ronly
python - <<PY
from continuum import Run, SQLiteStorage
from continuum.actions.ledger import ActionLedger
store = SQLiteStorage("run.db")
store.create_run(Run(run_id="blocked", goal="ship it"))
store.append_event("blocked", "RUN_STARTED", {"goal": "ship it"})
# A claimed-but-never-completed action makes the verdict REQUEST_HUMAN.
ActionLedger(store, "blocked").claim(action_type="curl",
                                     arguments={"url": "http://x/y"},
                                     key="invoice:INV-001")
print("before:", len(store.read_events("blocked")))
PY
python -m continuum.cli --db run.db resume blocked   # no --repair
python -c "from continuum import SQLiteStorage; \
print(after:, len(SQLiteStorage(run.db).read_events(blocked)))"
```

The other read-only-listed commands (`inspect`, `status`, `history`, `events`, `diff`, `validate`, `verify`, `actions`, `show-contract`, `replay`, `budget`, `tree`, `gate`, `briefing`, `health`, `impact`, `provenance`, `export-evidence`, `watch`) were spot-checked and do not write.

## What a fix looks like

1. **Fix the classification.** Move the notification out of `cmd_resume` into the mutating path — either gate it behind `--repair`, or (better, since a blocked operator should not need a mutating flag to get the bell) drive it from `continuum watch`, which is already the polling surface for liveness and is *not* on the read-only list. Then `resume` stays genuinely read-only and the docs hold as written.
2. **Fix the docs.** Remove `resume` from the read-only list in `references/cli.md` and note that a blocked run records its notification outcome in the event log. This matches todays behaviour but weakens a useful guarantee.

Option 1 preserves the property operators rely on; option 1 via `watch` also matches the comments own framing — dedup on contract hash already exists because the command is expected to be re-run.

Cyrax321:
Fixes #1175.

`references/cli.md` listed `resume` and `watch` among commands that "never write", and `README.md` made the same blanket guarantee. Both are false. A plain `continuum resume` (no `--repair`) whose verdict is `REQUEST_HUMAN` calls `_notify_blocked_run` at `main.py:1488`, before the `--repair` branch, and `notify_blocked` appends `NOTIFICATION_SENT` or `NOTIFICATION_FAILED` to the run's event log. `watch` appends `LIVENESS_SILENCE_DETECTED` / `LIVENESS_RECOVERED` when the liveness verdict flips, which its own docstring already says. Reproduced against origin/main: a blocked run went from 2 events to 3 on a bare `resume`, with `NOTIFICATION_FAILED` in the chain.

I kept the behaviour and fixed the classification, rather than the reverse, for three reasons measured in the code.

The behaviour is intentional and tested. Issue #305 built this bell so "nobody has to poll to learn a run is parked", the dedup key `(run_id, mode, contract hash)` exists precisely because `resume` is expected to be re-run, and `test_resume_notifies_on_request_human` asserts the plain invocation delivers and dedups across processes. Gating the bell behind `--repair` would defeat #305; moving it to `watch` would not help, since `watch` sits in the same read-only list and would need to run a full recovery assessment to know the verdict.

The appends cannot change a verdict. `NOTIFICATION_SENT` and `NOTIFICATION_FAILED` hit `case _: return False` in `_dispatch`, so the projection folds them into no state at all. The new `test_resume_notification_append_is_audit_only` pins this: folding the log with and without the dead-letter lands on identical projected state, the event cursor advanced by exactly one, and a second `resume` returns the same verdict and exit code.

The mid-run property still holds. The append fires only when the verdict is `REQUEST_HUMAN`, meaning the run is already parked. A `resume` of a run an agent is actively working returns RESUME and writes nothing, so the collision the guarantee exists to rule out does not occur.

Changed: `README.md`, `references/cli.md`, the module docstring, and `cmd_resume`'s docstring now carry the precise wording, and the two commands are removed from the never-write list. No code path changed. The MCP tool `continuum_resume` has no notification wiring, so the tool split documented in `README.md` stays accurate and is untouched.

Verified with `tests/test_webhooks.py` (27, including the new regression test), `test_cli_docs.py`, `test_docs_counts.py`, a 128-test sweep over resume/webhook/notify/CLI-main, plus `ruff check` and `mypy src/continuum/cli/main.py` clean.

One scope note for review: I included `watch` in the correction even though the issue only names `resume`. It shares the exact defect in the same sentence I was editing, so leaving it would have kept the paragraph internally false. Happy to split it into its own PR if you would rather keep the two separate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented PostgreSQL compaction from deleting the chain anchor and creating a fork.
  * Preserved parent-run relationships in PostgreSQL.
  * Improved action retry handling, including interrupted and drifted claims.
  * Webhook failures during resume now record an audit event without changing run state or verdicts.
  * Improved latency benchmark handling for malformed baselines, invalid thresholds, percentile boundaries, and generator inputs.

* **Documentation**
  * Clarified CLI audit behavior and updated project statistics and architecture references.


Cyrax321 (commented):
Thank you.

Correcting the classification instead of the behavior is the right call. The bell from #305 is intentional and pinned by test_resume_notifies_on_request_human, and gating it behind repair or moving it would defeat the purpose. The three properties stated here check out: NOTIFICATION events fall through _dispatch so they cannot move the verdict, they fire only for an already parked run, and dedup keeps it to once per verdict.

I reviewed README, references/cli.md, the docstring updates, and the new audit-only test. Keeping watch in scope is correct since it shares the same sentence and the same defect. CI is green.

Cyrax321 (commented):
Maintainer re-review at current head. The resume and watch read-only reclassification is still correct with the #305 bell preserved. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review with no blocking issues at last review. Rebase out of dirty state and confirm CI green.

PR #1179, opened 2026-09-16
fix(cli): replay --upto windows over the archived prefix on a compacted run (#1172)

### LINKED ISSUE #1172: cli: replay --upto fails on every compacted run and blames the wrong cause
### Summary

`continuum replay <run> --upto N` fails on every compacted run, for every value of `N`. The command reads only the live event tail, where `RUN_STARTED` no longer lives after compaction, so its guard rejects the request and advises the operator to "increase `--upto`" — advice that cannot help, because the event is in the archive, not the live log.

### Expected Behaviour

`--upto` should window the replay over the run's full history, the way `continuum events` does: `cmd_events` (`src/continuum/cli/main.py:701-713`) deliberately reads `read_all_events` so "a compacted run reads the same as one that was never compacted". A replay of a compacted run should honour the same contract.

### Actual Behaviour

On a compacted run, every `--upto` value exits 1 with a misleading message:

```
error: --upto 11 excludes the RUN_STARTED event for run 'r1'; increase --upto or omit it to replay from the beginning
error: --upto 999 excludes the RUN_STARTED event for run 'r1'; increase --upto or omit it to replay from the beginning
```

`--upto 999` is far beyond the run's last sequence (13) and still fails, which is what proves the message is wrong about the cause.

### Steps to Reproduce

```bash
python repro_replay_upto_compacted.py
```

The script builds a run, replays with `--upto` successfully, compacts it, then retries.

### Root Cause

`src/continuum/cli/main.py:3381` reads the live tail only:

```python
events = storage.read_events(args.run_id, upto=args.upto)
```

`SQLiteStorage.read_events` (`src/continuum/storage/sqlite.py:679-695`) queries the `events` table, never `events_archive`. After compaction, `RUN_STARTED` is archived, so the guard at `src/continuum/cli/main.py:3436-3441` fires for every `N`.

The same `--upto is None` gating at line 3387 means the anchored-replay branch (which does handle compaction correctly) is skipped whenever `--upto` is supplied, so the request falls through to the plain path that cannot work.

`cmd_events` on the same run works correctly, because it reads `read_all_events`. The two commands disagree about what "the event log" is.

### Impact

`--upto` is documented (`references/cli.md:21`: `continuum replay <run_id> [--upto N]`) and is the natural tool for bisecting where a run's state diverged — a diagnostic for suspected corruption. On a compacted run it is unavailable entirely, and the failure message points the operator at the wrong fix, so they will raise `N` until they give up. Compaction is the feature that exists for long-lived runs, so the diagnostic is missing precisely where it is most likely to be needed.

The anchored-replay path (`--upto` omitted) handles these runs correctly, so the run itself is fine — only the windowed diagnostic is broken.

### Evidence

```
== before compaction ==
  replay --upto 11 rc: 0
== compact ==
  archived: ['RUN_STARTED', 'WORK_COMPLETED' x5, 'STATE_CHECKPOINTED', 'WORK_COMPLETED' x4]
  live:     ['STATE_CHECKPOINTED', 'EVENT_LOG_ANCHORED']
== after compaction ==
  replay (no upto) rc: 0
  replay --upto   11 rc: 1
  replay --upto   12 rc: 1
  replay --upto   13 rc: 1
  replay --upto  999 rc: 1
```

Full script: `repro_replay_upto_compacted.py`. Baseline suite green before this change.

### Duplication checks

- #648 (second compact fails, missing archived `RUN_STARTED`) shares the underlying "projection ignores the archive" theme but is a different command and code path; it was fixed by PR #650, which anchored over full history and did not touch this guard.
- #532 is the archive-inclusion principle this command violates.
- Searched open and closed issues for `replay upto`, `anchored replay`, `compacted`.

### Suggested Fix

Read the window from `read_all_events` when the run is anchored (as `cmd_events` does), and route a windowed request on an anchored run through the existing anchored-replay branch instead of gating that branch on `args.upto is None`.

Cyrax321:
Fixes #1172.

`continuum replay <run> --upto N` failed on every compacted run, for every value of N, and told the operator to increase `--upto`. That advice could not work: compaction archives `RUN_STARTED`, the command read only the live event tail, and no value of N can reach an archived event there. `--upto 999` failed on a run whose last sequence was 13, which is the evidence that the message was wrong about the cause.

The window now spans the run's full history, archived prefix included, by reading `read_all_events`. That is the read `cmd_events` already uses, so the two commands stop disagreeing about what "the event log" is. Three things changed in `cmd_replay`: the window source, the branch gate (windowed requests reach the anchored-replay path instead of being locked out by `--upto is None`), and `_verify_against_stored`, which now reads the stored version's own prefix from full history so a window ending inside the archive verifies instead of reporting a sound version as corrupt.

One subtlety worth calling out, since it is where my first attempt was wrong. Anchoring is detected from the **live** tail only, not from any anchor in the window. An anchor that has itself been archived belongs to a compaction a later one superseded (#648), and a window can contain that archived anchor while still ending short of the current boundary. Detecting from the window restored the current checkpoint and folded an empty tail over it, so `--upto 25` on a twice-compacted run whose boundary was 34 answered the state at 34. A window at or below the boundary now folds the archived prefix from scratch, which is what an uncompacted run does for the same window.

The `RUN_STARTED` guard still fires, and now truthfully, for a window that genuinely excludes it (`--upto 0`). On a compacted run that used to be the failure for every N.

The window bound is pushed into the engine rather than filtered in Python: `read_all_events` and `read_archived_events` take an optional `upto`, so a narrow window on a heavily compacted run does not materialize the whole archive to discard most of it. Compaction exists to bound replay cost, and an unbounded archive read would undo that for every narrow window.

Verification: the repro from the issue goes green at every window, and a compacted run now reads the same as the never-compacted equivalent at every window, for both a single and a double compaction. The corruption contract survives the new path: tampering the stored version, or poisoning an archived event, still exits `CORRUPTED` with `verified: false` for every window, so the windowed path does not retire the check the way a hardcoded pass would. Seven tests pin this, including one that catches the boundary-state trap above and one that the engine bound reaches SQL; all seven fail on the pre-fix code.

Full suite (2,262 collected), `ruff check`, `ruff format --check`, and `mypy src/continuum` all pass on a clean checkout of this branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `continuum replay --upto` for compacted runs.
  * Windowed replays now include archived history, report the requested endpoint and state, and verify correctly across compaction boundaries.
  * Corruption detection remains enabled, while windows that exclude the run start continue to be rejected with an accurate boundary message.

* **Performance**
  * Bounded replay requests now limit archived-history reads to the requested endpoint.


Cyrax321 (commented):
Thank you.

This fixes replay upto on compacted runs by windowing read_all_events, the same read cmd_events uses, so the two commands agree on the event log. Letting windowed requests reach the anchored branch and verifying the stored prefix from full history are the right shapes, and detecting anchoring from the live tail only avoids serving the boundary state for an earlier window. The RUN_STARTED guard now fires only for a genuinely empty window, which makes the error truthful.

I reviewed main.py and the six new tests in test_compaction.py and test_compact_twice.py. The tests pin every window on a compacted run, parity with a live run, tamper detection through the new path, and the between-anchors trap. CI is green (33 checks).

I checked the bot performance note about loading the full archive for a narrow window. That is real but expected here since correctness requires the archived prefix for a diagnostic command. A bounded read on read_all_events could be a follow-up for very large runs, not blocking.

Cyrax321:
Addressed the performance finding: `read_all_events` and `read_archived_events` now take an optional `upto` that reaches the SQL, so `--upto 1` on a heavily compacted run no longer materializes the whole archive. `cmd_replay` and `_verify_against_stored` pass the bound through instead of filtering in Python. Added a test that the bound reaches the engine and that an unbounded read still returns the archive plus the live tail.

Issue #1080, opened 2026-09-15
retry budget is counted under the derived key, so argument drift lets an exhausted budget suppress the completed-result answer the dedup path exists to return


Cyrax321:
`continuum_intercept_action` guards the run-level retry budget (issue #240) by looking the action up under the *derived* idempotency key and skipping the budget when the record is already settled (mcp/server.py:1265):

```python
existing = ledger.get(idempotency_key(action_type, arguments, scope=..., key=key))
settled = existing is not None and existing.status in (COMPLETED, UNKNOWN)
if not settled:
    events = ctx.storage.read_all_events(run_id)
    claim_key = str(idempotency_key(action_type, arguments, scope=..., key=key))
    attempts = attempts_by_key(events, action_type).get(claim_key, 0)
    allowed, used, maximum = evaluate_budget(budgets, action_type, attempts)
    if not allowed:
        raise ToolError("retry budget exhausted ...")
```

The settled-check is exact-key only. `ActionLedger.claim` has a second, drift-tolerant lookup, `_identity_match`, that recognises an already-recorded action by shared identity tokens when the argument hash misses, and it can return a COMPLETED record under a *different* stored key. The budget gate runs before `claim`, so that match never gets the chance to answer.

Concrete sequence:

1. An operation succeeds with argument spelling A and is recorded COMPLETED under key K1.
2. The same operation is later attempted with a drifted spelling A' (a renamed field, a reformatted path). It fails repeatedly; each failed attempt derives K2 and is counted there, until K2's budget is exhausted. `attempts_by_key` omits K1 from the count because it completed, and K1 is a different key anyway.
3. The agent calls `continuum_intercept_action` with spelling A' again. `ledger.get(K2)` is None or FAILED, so `settled` is false, the budget check sees an exhausted K2 and raises ToolError.

The tool answers "retry budget exhausted ... either settle it a different way or raise the limit" when the correct answer is "proceed=false, already performed, here is the previous result". That is exactly the failure the `settled` guard exists to prevent, and the issue it cites (#309) states the invariant in those terms: gating a settled action "would make an exhausted budget suppress the dedup and reconciliation paths a recovering agent depends on".

The danger is not a duplicate side effect; the tool still refuses. It is that the refusal is misleading in a way that points the operator at the wrong knob: the run is told to raise a budget that is working as intended, when the real situation is that the work is already done and the identity lookup missed it.

[...]

PR #1184, opened 2026-09-16
fix(actions): gate the retry budget only on claims that open a new slot (#1080)

Cyrax321:
## What

`continuum_intercept_action` ran its retry-budget gate against an exact-key lookup of its own, so it could refuse a state `claim` would have answered. The gate now resolves the claim through the same lookup `claim` uses, and only gates when that resolution would actually open a new attempt slot.

## The reachable defect

The state that actually bites is an **interrupted attempt**. A claim is recorded STARTED, the process dies between claim and complete, and the record is never settled. That slot counts against its own key, so with `max_attempts: 1` the re-claim hits an exhausted budget and the tool answers:

> retry budget exhausted ... raise the limit by setting action_types.charge.max_attempts

The ledger's answer was `UnknownSideEffect`: the outcome is unknown and a reconciliation is owed. An operator pointed at the budget is pointed at the wrong knob. Nothing was retried, and the work may already have happened.

This is exactly the invariant the code comment already claimed to hold, quoting #309: gating a settled action "would make an exhausted budget suppress the dedup and reconciliation paths a recovering agent depends on". The comment named COMPLETED and UNKNOWN; STARTED is the same class, since `claim` opens no slot for it either.

## On the drift framing in the issue

I built the issue's literal sequence (complete under spelling A, exhaust the budget on drifted spelling A', re-claim A') and it does not reach the gate, and I want to say so plainly rather than ship a test that pretends it does. The identity-token fallback in `claim` runs on the first drifted call and answers with the completed record, so the exhausting failures never start. The reason is structural rather than lucky: a key with counted attempts always has a folded record, and a settled or uncertain record is either excluded from the count or never reaches the gate. The two readers can only disagree about status, and STARTED is the one status where they did.

The fix still removes the underlying cause the issue names. The gate no longer re-derives what `claim` will find; it asks `claim`'s own resolution, which includes the drift-tolerant and cross-run lookups. Those two lookups cannot drift apart again, and a test pins the drifted-completed case so a future change to either lookup surfaces immediately.

## Change

`ActionLedger.resolve_claim` is the resolution `claim` already performed, extracted as a read-only method: the exact argument-hash key, then another run holding the same unscoped key, then the identity-token fallback. It preserves `claim`'s order and short-circuits exactly, so a claim that returns or raises resolves to the same key it would have on its own. It returns a `ClaimResolution` whose `opens_slot` reports whether `claim` would record a new attempt slot; that, and only that, is what the budget gates.

`claim` now calls it instead of inlining the lookup, and the MCP handler calls it instead of its own `ledger.get(idempotency_key(...))`. The budget is counted under the key the resolution returns, which for a slot-opening claim is the derived key.

Callers passing an explicit `key` are unaffected: `idempotency_key` hashes it verbatim, so no drift is possible and the derived key is the stored key.

## Tests

- `test_an_interrupted_action_still_reconciles_at_budget` — the regression. Fails on `main` with `retry budget exhausted`, passes here. Also asserts the gate still holds for a genuinely fresh key that fails, so the fix is not the gate switched off.
- `test_a_drifted_reclaim_of_a_completed_action_answers_at_budget` — pins the issue's headline invariant: a drifted re-claim answers with the stored result, not a budget refusal.
- Seven `resolve_claim` unit tests covering no-record, FAILED, drifted COMPLETED, interrupted, explicit key, and both cross-run cases.

Full suite passes. Two tests in `tests/test_adapter_result_fuzz.py` fail intermittently on this machine on `hypothesis` `DeadlineExceeded` (323ms against a 200ms deadline); they fail the same way on unmodified `main` and are timing flakes, not related to this change.


src/continuum/actions/ledger.py
```diff
@@ -843,53 +941,41 @@ def claim(
...
         resolves it.
         """
-        explicit_key = key is not None
         _explicit_rendered = key
-        idem = idempotency_key(
+        # The resolution is shared with the retry-budget gate in the MCP
+        # intercept handler, which has to decide *before* a slot is opened
+        # whether to allow it. Doing the lookup twice in two places let the two
+        # readers disagree about what a claim would find (issue #1080).
+        resolved = self.resolve_claim(
```
Cyrax321, line 949: Blocking: Lint and Type-check fails on this PR. This resolved (ClaimResolution) is rebound later in claim() by resolved = on_unknown(existing), which returns ActionOutcome. Mypy reports the incompatible assignment plus resolved.action missing on ClaimResolution.

Please rename the later binding, for example decision = on_unknown(existing) with decision.key, decision.action, return decision. Design itself is unaffected.

Cyrax321 (commented):
Thank you for the fix.

The shared resolve_claim design is the right shape: the gate reads through claim's own resolution (exact key, cross-run, identity fallback) and only gates opens_slot, which closes the STARTED interrupted-attempt defect and the drift divergence at the cause. The regression plus drift pin plus seven resolution unit tests cover the behavior well.

Blocking (cannot merge while red): Lint and Type-check fails on this PR (mypy in ledger.py, ClaimResolution versus ActionOutcome reuse of resolved, plus resolved.action). I left one inline comment with the rename. Once the type gate is green this is approvable. Note: GitHub does not allow requesting changes on your own PR, so this is filed as a comment, but the type failure blocks merging the same way.

Non-blocking for a follow-up: the intercept path now folds the log twice per slot-opening call (resolve_claim in the handler, then again inside claim) plus read_all_events for the budget count. Threading the computed resolution into claim would halve the hot-path fold cost.

Thank you again.

Cyrax321:
Self-review (non-blocking, approve). This PR is my own work from an earlier session, so treat this as the author's second pass rather than an independent vetting. GitHub does not allow an approve verdict on my own PR, which is the correct guard; the substance is below.

I verified the central invariant empirically instead of reading the diff. The whole change rests on one claim: `ClaimResolution.opens_slot` is true exactly when `claim` would record a new attempt slot. If it were wrong in either direction the gate would either refuse a state the ledger answers, or permit a retry it should have refused. I drove both `resolve_claim` and `claim` against the same ledger across all six record states and they agree in every case: COMPLETED returns `fresh=False` with `opens_slot=False`, UNKNOWN and STARTED raise `UnknownSideEffect` with `opens_slot=False`, and FAILED, COMPENSATED and a fresh key all give `fresh=True` with `opens_slot=True`.

The specific bug the issue names reproduces as described and is fixed. An interrupted attempt sits STARTED, its own slot counts against it, and the old exact-key gate answered an exhausted budget with a retry refusal when `claim`'s answer was `UnknownSideEffect`. I confirmed the post-fix gate skips that state and the reconciliation path surfaces.

The refactor is behavior-preserving rather than a reimplementation, which is what makes it safe. `claim` now delegates to `resolve_claim` instead of duplicating the lookup, and the resolution order (exact key, then foreign, then drift-tolerant identity) is the same sequence with the same short-circuits. The subtle part is that identity matching is skipped when a foreign record already settles the claim; that is what keeps a claim that returns or raises from resolving to a different key than it would have alone.

Verified in isolation and after merging current main: the branch is 24 commits behind, main has touched all three files, but the merge is clean and the merged tree passes the action-ledger and MCP suites (186 tests). One note for the merge, not a blocker: both sides edited the same Fixed section of CHANGELOG.md, so that conflict hunk deserves a read even though git resolves it.

All 185 tests in the touched suites pass; ruff and mypy clean on the changed files.

PR #1187, opened 2026-09-16
test(checkpoint, recovery): add direct tests for RewindError and LedgerLockError (#1116)

### LINKED ISSUE #1116: RewindError and LedgerLockError error paths have no direct test: the rewind suite is happy-path only and the ledger lock-contention raise is never asserted
## Summary

Two documented error contracts have no direct test — the exception types are raised in product code but never asserted anywhere in the suite:

**`RewindError`** (`src/continuum/checkpoint/rewind.py:20`, in the module's `__all__` at `:17`) is raised from three sites:

- `:79` — `no checkpoint {to!r} for run {run_id!r}` (unknown checkpoint id)
- `:124` — wrapping an `EditPreconditionError` from the edit gate
- `:208` — `rewind to {to!r} has N conflict(s) and M unrecoverable file(s); use --force...` (environment conflicts without `--force`)

There are two dedicated rewind test files — `tests/test_rewind_carry_forward.py` and `tests/test_rewind_dual_state.py` — plus rewind mentions in `tests/test_admissibility_e2e.py` and `tests/test_grants.py`. None of them ever asserts `RewindError`; every test exercises the happy path.

**`LedgerLockError`** (exported from `src/continuum/recovery/ledger.py:55`) is raised at `recovery/ledger.py:305` inside `_locked()` when the run lease cannot be acquired — the single-writer guarantee of the recovery ledger (`#345`, lease integration). No test references it; the only in-repo consumer of the condition outside the module is the benchmark scenario `benchmark/phase6/scenarios.py:246`, which is not a test of it.

This is the same shape as #902 (file snapshot edge cases with no direct test). It is distinct from open #1053 (rewind fails on every compacted run — a behavior bug) and from closed #672 (a flaky lease-concurrency test): those are about wrong/flaky behavior, this is about the documented error paths having zero coverage, which is exactly why #1053 could land unnoticed.

## Affected component

Tests

## Severity

Minor inconvenience — but these are the paths that guard against unsafe edits and concurrent recovery.

## System information

- Python: 3.13.5

```bash
grep -rn "RewindError" tests/ | cat
# (no output)

grep -rn "LedgerLockError" tests/ | cat
# (no output)

grep -n "raise RewindError" src/continuum/checkpoint/rewind.py
# :79, :124, :208 — three untested raise sites
```

## What a fix looks like

Add direct tests, one per raise site:

- `rewind_to_checkpoint` with an unknown checkpoint id → `pytest.raises(RewindError, match="no checkpoint")`
- a rewind whose environment has conflicting files, without `--force` → `pytest.raises(RewindError, match="conflict")` (and the `--force` escape accepting it)
- a `RecoveryLedger` constructed with a lease stub whose `acquire()` returns `False` → `pytest.raises(LedgerLockError)` on `append_decision`/`record_attempt`

All three fit the existing fixtures in `tests/test_rewind_dual_state.py` and `tests/test_lease.py`.

soorajsuru2004-sudo:
## Description

Adds direct unit test coverage for the documented error paths of `RewindError` and `LedgerLockError`, which were previously unasserted across the suite (#1116).

For `RewindError` in `src/continuum/checkpoint/rewind.py`:
- `test_rewind_unknown_checkpoint_raises_rewind_error`: queries with unknown checkpoint IDs or versions in `resolve_checkpoint` and `rewind_to_checkpoint` assert `RewindError(match="no checkpoint")`.
- `test_rewind_conflicts_without_force_raises_rewind_error`: external modifications leading to file digest conflicts assert `RewindError(match="conflict")` when `force=False`, and verify that `force=True` bypasses the raise and records conflicts in `RewindResult`.
- `test_rewind_precondition_unexpected_error_raises_rewind_error`: unexpected exceptions from precondition checks are verified to be caught and wrapped in `RewindError`.

For `LedgerLockError` in `src/continuum/recovery/ledger.py`:
- `test_ledger_lock_contention_raises_ledger_lock_error`: exercises lock contention on a shared coordinator, asserting `LedgerLockError` on `append_decision`, `record_attempt`, `record_gate`, and `compact`, and verifies that operations succeed once the lock is released.
- `test_ledger_lock_stub_refusal_raises_ledger_lock_error`: asserts that a coordinator whose `acquire()` returns `False` raises `LedgerLockError`.

## Related issues

Fixes #1116

## Types of changes

- Type: `tests`

## Breaking changes

No

## How has this been tested?

```powershell
pytest tests/test_rewind_dual_state.py tests/test_recovery_ledger.py -v
pytest tests/test_docs_counts.py -v
ruff check src/ tests/
```

Environment: Windows 11, Python 3.12.10.

## Checklist

## Additional context

None

## Summary by CodeRabbit

- **Tests**
  - Added coverage for recovery ledger lock contention, including successful operations after locks are released.
  - Added coverage for lock acquisition failures and clear ledger lock errors.
  - Added rewind error-path tests for unknown checkpoints, digest conflicts, forced rewinds, and unexpected validation failures.


Cyrax321 (approved):
Approve. This is test-only coverage for two existing error paths, and I verified the tests are not vacuous.

The regexes the tests match are the literal production messages, so a regression that stopped raising, or raised a different exception type, would fail the assertion rather than pass silently. I confirmed both against the source: `ledger.py:305` raises `LedgerLockError(f"could not acquire ledger lock for run {run_id!r}")` and `rewind.py:79` raises `RewindError(f"no checkpoint {to!r} for run {run_id!r}")`.

Two things I liked. First, the lock test covers all four mutating operations (`append_decision`, `record_attempt`, `record_gate`, `compact`) rather than one and assuming the rest share the guard, which is how a guard silently drifts off one method. Second, the stub test covers the `acquire() -> False` contract directly instead of only the real coordinator's held-lease path, so a coordinator that returned False for a different reason would still be caught. The release-then-succeed half of the first test is what makes it a genuine round trip rather than a one-way assertion.

All 26 tests in the two touched files pass.

PR #1188, opened 2026-09-16
refactor(actions, concurrency): remove dead DuplicateAction and LeaseError exceptions (#1115)

### LINKED ISSUE #1115: DuplicateAction and LeaseError are exported exceptions that no code path can raise
## Summary

Two exported exceptions name conditions that product code handles through other signals, so the exceptions are never raised anywhere:

**`DuplicateAction`** (`src/continuum/actions/ledger.py:251`, a `LedgerError` subclass, exported in `actions/__init__.py:27` and the ledger's `__all__` at `:109`). The duplicate-claim path it presumably names is enforced three other ways in `claim()`: a completed record under the same identity is returned as `ActionOutcome(fresh=False)` (store-wide dedup, `ledger.py:836` onward), an unresolved foreign attempt raises `UnknownSideEffect`, and a spent grant raises `GrantDenied`. `DuplicateAction` has no raise site, no `except` site, and no reference in tests, docs, or the CHANGELOG.

**`LeaseError`** (`src/continuum/concurrency/lease.py:47`, in the module's `__all__` at `:32`). Lease contention is signaled by `acquire()` returning `False` — which is exactly how the one runtime consumer branches (`recovery/ledger.py:304`: `if not self._lock.acquire(...)` → raises `LedgerLockError`, a different exception). `LeaseError` has no raise site anywhere and no test reference.

Same class of finding as the dead helpers in #1094/#1095: exported surface that no code path can produce.

## Affected component

Other (actions ledger / concurrency API surface)

## Severity

## System information

- Python: 3.13.5

```bash
git grep -n "DuplicateAction" | cat
# only: two __all__ entries, one import, one class def — no raise, no except, no test

git grep -n "LeaseError" | cat
# only: __all__ entry and class def in concurrency/lease.py — no raise, no test

git grep -n "raise DuplicateAction\|raise LeaseError" -- src/
# (no output)
```

## What a fix looks like

**Option A: remove both** — delete the classes, their `__all__` entries, and the re-export imports. The named conditions already have enforced, tested signals (`fresh=False` outcomes, `UnknownSideEffect`, `GrantDenied`, `acquire() -> False`).

**Option B: raise them where the condition occurs** — e.g. `claim(..., on_duplicate="raise")` for callers that prefer exception semantics over the `fresh=False` outcome, and `LeaseError` from a blocking `acquire_or_raise` helper — only if a caller actually wants exception semantics; none currently does.

soorajsuru2004-sudo:
## Description

Removes the exported exception classes `DuplicateAction` and `LeaseError`, which have no raise sites, handlers, or callers across the codebase (#1115).

- `DuplicateAction` (`src/continuum/actions/ledger.py`, exported from `actions/__init__.py` and `ledger.py`): duplicate attempts are already handled through `ActionOutcome(fresh=False)` (completed actions), `UnknownSideEffect` (unresolved attempts), and `GrantDenied` (spent grants).
- `LeaseError` (`src/continuum/concurrency/lease.py`, exported from `concurrency/__init__.py` and `lease.py`): lease contention is signaled via `acquire() -> False`, which callers (e.g. `RecoveryLedger._locked`) handle directly with contextual errors like `LedgerLockError`.

Also documents the removals under `[Unreleased]` -> `### Removed` in CHANGELOG.md.

## Related issues

Fixes #1115

## Types of changes

- Type: `refactor`

## Breaking changes

No

## How has this been tested?

```powershell
pytest tests/test_action_ledger.py tests/test_lease.py tests/test_module_all_exports.py -v
pytest tests/test_docs_counts.py -v
ruff check src/ tests/
```

Environment: Windows 11, Python 3.12.10.

## Checklist

## Additional context

None


Cyrax321 (approved):
Approve. I verified the central claim independently rather than trusting the description.

On origin/main both names appear only in their own definitions and re-exports: no `raise DuplicateAction`, no `except DuplicateAction`, no `except LeaseError` anywhere in src/, tests/, examples/ or docs/. I also checked all twelve other open PR heads; the only hits are the same still-present definitions each branch inherited from main, so nothing in flight depends on either name.

The removal is the right kind of dead-code deletion: these were exported public exceptions that no code path could raise, so any downstream `except DuplicateAction` was already silently unreachable. Removing them surfaces that instead of preserving a false seam. The CHANGELOG entry correctly points at the live mechanisms that replaced them (`fresh=False`, `UnknownSideEffect`, `GrantDenied`, `acquire() -> False`), which is what a reader needs to know.

Verified on the PR's tree: `continuum.actions` and `continuum.concurrency` import cleanly with both names absent, and the action-ledger, lease and action-index suites pass (36 tests).

Issue #765, opened 2026-09-07
feat(plugins): dispatch registered ActionReconcilers during recovery reconciliation


Cyrax321:
## Affected component

Core (state, checkpoints, recovery)

## What problem does this solve?

CONTINUUM defines an `ActionReconciler` plugin seam, but the seam is not a general consumer in the recovery path. `docs/ARCHITECTURE_EVOLUTION.md` identifies plugin seams as declared but not load-bearing. This leaves integrations that can reliably inspect an external effect without a supported way to contribute that evidence to reconciliation.

This is distinct from #268, which proposes evidence-backed reconciliation through OpenTelemetry spans and system-level observations. This issue establishes the generic, explicit plugin-dispatch contract that can support those and other reconcilers without making any particular telemetry backend mandatory.

## Proposed solution

Allow reconciliation to receive an explicitly supplied or registry-resolved collection of `ActionReconciler` implementations. For each uncertain action, select applicable reconcilers deterministically, collect structured evidence, and merge results into the existing reconciliation and recovery contract surfaces. Reconciler output must be provenance-stamped, bounded, and traceable to the action and plugin that produced it.

Registration remains explicit. Reconciler code must not be auto-discovered or allowed to mutate the recovery ledger outside the existing controlled reconciliation path.

## Acceptance criteria

- The public reconciliation path accepts an explicit reconciler collection or registry-backed equivalent while preserving default behavior when none is configured.
- A documented reconciler result contract represents confirmed occurrence, confirmed non-occurrence, unavailable evidence, and conflicting evidence.
- Selection and result merge order are deterministic and exposed in JSON/text diagnostics with reconciler identifiers and provenance.
- A reconciler can provide evidence but cannot bypass ledger settlement rules, lower a more cautious verdict, or mutate storage directly.
- Missing, crashing, malformed, or conflicting reconcilers fail closed to an explicit unresolved or requires-review outcome.
- Reusable conformance tests cover result categories, exceptions, conflicting results, ordering, provenance, and no-op behavior without registered reconcilers.
- Documentation explains the plugin trust boundary and how #268-compatible OTel or system-observation reconcilers can be implemented on this seam.

## Alternatives considered

Add each external evidence source directly to the core reconciler. Rejected because it duplicates integration code and prevents framework-neutral extensions. Automatically loading installed plugins is also rejected because a resume or reconcile operation must not implicitly execute untrusted code.

## Is this a breaking change?

No

## Would you like to work on this?

Yes, with guidance

Issue #244, opened 2026-08-23
Open problems in agent durability: nine research-grounded directions (umbrella)


Cyrax321:
### Affected component

Other / not sure

## Open problems in agent durability - and CONTINUUM's plan to close them

This project's differentiator is provenance-gated recovery: an agent can never certify its own work, every external effect is claimed before it fires and settled from reality afterwards, and resume decisions are sealed contracts with executable next steps. Everything below extends that one idea to where production agent systems actually live. Each linked problem states its purpose, cites the papers that define it, gives a step-by-step plan inside our scope guardrails, and defines acceptance criteria so any contributor can pick one up.

### The field has converged; the missing piece has not been built

The 2026 durable-execution survey (Zylos Research, 2026-04-24) documents the convergence: Temporal/Restate/DBOS/Inngest expose durable boundaries; LangGraph ships strong per-superstep checkpointers but its own docs and maintainer-confirmed issue langchain-ai/langgraph#6208 admit nodes re-execute after interrupt resume; ACRFence (arXiv:2603.20625) formalises this as the semantic-rollback hazard; AgentRewind shows recovery must align *agent context* with *environment*, or resumed sessions restart cognition from a progress bar; RetryGuard (arXiv:2511.23278) shows local retries amplify into run-level storms.

What nobody ships is the combination: **a side-effect control plane with mandatory claims, provenance-gated trust, reconciliation of unknowns, and context rehydration - portable across harnesses rather than owned by one runtime.** Runtime owners cannot build it without abandoning their neutrality; framework owners will not build it because it constrains their users. A harness-neutral layer can.

### The open problems

| # | Problem | Why it matters |
|:--|:--|:--|
| #235 | **Reasoning-context rehydration** | Resume inherits the plan stack and open threads, not just "47/100 done" (AgentRewind) |
| #236 | **Native LangGraph BaseCheckpointStore** | Provenance + ledger inside the persistence interface the biggest framework already uses |
| #237 | **Replay-safety guard as a portable primitive** | Close the semantic-rollback class (ACRFence) with proofs in CONTINUUM-Bench |
| #238 | **Production server mode** | Live-Postgres CI, HTTP API for non-Python agents, worker leases |
| #239 | **Log compaction + payload offloading** | Month-long runs without linear replay cost (Temporal codec discipline) |
| #240 | **Run-level retry budgets** | Stop LLM-invented retry storms at claim time (RetryGuard) |
[...]

PR #1189, opened 2026-09-16
feat(reconcile): settle uncertain actions from OTel spans and artifacts (#268)

### LINKED ISSUE #268: Evidence-backed reconciliation: settle uncertain actions from OTel spans and system-level observations
### Affected component

Detector / drift detection

An action left `STARTED` by a crash between intercept and complete is the ledger's most expensive state: resume blocks until a human reconciles it or a registered probe settles it from reality. Today reconciliation depends on `.continuum/reconcilers.json` probes the operator writes by hand. At fleet scale that is the human bottleneck for exactly-once.

Meanwhile system-level observation of agents matured on two fronts:

- AgentSight (arXiv:2508.02736) shows eBPF boundary tracing reliably captures what an agent actually did (syscalls, process events, decrypted LLM traffic) independently of agent self-report, and its case studies include detecting retry loops from kernel evidence alone.
- Cilium Tetragon ships production-grade eBPF observation/enforcement, and practitioner writeups describe correlating self-reported OTel telemetry against kernel-observed syscalls as a discrepancy-detection pattern.
- The agent's own telemetry cannot be trusted to settle outcomes: Codex CLI has open gaps where `codex exec` emits no OTel metrics at all (openai/codex#12913).

Nobody connects this to a durability ledger. Kernel-observed truth exists; uncertain claims sit in our ledger waiting; nothing closes the loop between them.

## Proposed solution

Treat externally observed facts as just another reconciler input, so uncertainty is settled from evidence instead of human memory:

1. Define a probe seam extension: built-in probe types alongside today's command probes. First two:
   - `otel_span`: queries recorded OTel spans (the `continuum.otel` span processor already captures tool-call spans with zero code changes) for a span matching the claim's action type plus identity tokens after the claim sequence; match yields occurred=true with the span's result attributes attached.
   - `artifact_check`: filesystem mtime/hash existence checks for path-scoped actions (did the claimed write actually land), reusing environment snapshot primitives.
2. eBPF collectors stay out of process: document recipes that pipe Tetragon/AgentSight JSON through a tiny adapter into `continuum reconcile` using the same envelope. We consume standard outputs; we do not ship BPF programs. Linux-only, guarded-skip in tests like ContainerAdapter/BrowserAdapter.
3. Discrepancy detector: when evidence contradicts the ledger (action completed but no observed artifact; artifact exists with no completed record), the validator raises a finding (`REQUIRES_REVIEW`) rather than silently picking a side.
4. Provenance: settlements carry the evidence source; strict mode still refuses auto-settlement of unknowns and degrades to human, unchanged.

Acceptance criteria:

- A crash-between-intercept-and-complete scenario is settled automatically by an `otel_span` probe in non-strict mode, with the settlement event citing span ids.
- An `artifact_check` probe proves occurred=false for a claimed-but-absent write.
- Contradictory evidence produces a review finding, not a settlement.
- All collector-dependent tests skip cleanly without the collector present.

## Alternatives considered

- Embed eBPF loading in-process. Rejected: kernel-version matrix, platform lock-in and maintenance swamp the value; consumption of standard outputs gets 90 percent of it.
- Trust agent-emitted telemetry alone for settlement. Rejected outright: self-report is what the ledger exists to distrust; the whole point is settling from outside the agent's control plane.
- Wait for MCP-level attestation standards. Worth revisiting when they exist; the OTel and artifact probes are useful regardless and are protocol-neutral.

No

Yes, I have a plan

Cyrax321:
## Description

An action left STARTED by a crash between intercept and complete is the ledger's most expensive state: resume blocks until a human reconciles it or a registered probe settles it from reality. Probes used to mean commands the operator writes by hand. The registry now also carries two built-in evidence probe types, plus a check that escalates when the ledger disagrees with the world.

New `src/continuum/evidence.py`:

- `OtelSpanReconciler` settles from a tool-call span already in the event log as a `TOOL_COMPLETED` / `TOOL_FAILED` observation (`continuum.otel`). It finds a span matching the claim's action type and identity tokens after the claim sequence and cites the span, trace and event ids in the settlement. `record_span` now carries `span_id` / `trace_id` into the payload, so a settlement names the exact trace it was based on instead of only the event id.
- `ArtifactCheckReconciler` settles a path-scoped action from existence plus a SHA-256 digest, reusing the environment snapshot primitives. A missing file is evidence of absence. A digest mismatch against `expect_sha256` declines to attribute the write, because the file may be somebody else's.
- `detect_discrepancies` runs after probing and escalates two contradictions to `REQUIRES_REVIEW` instead of re-settling: the ledger says completed but the artifact is missing, and the artifact exists but the ledger never recorded the completion. Picking a side silently is how work gets dropped.

A failed span settles nothing. Treating one as `occurred=false` would license a retry against a write that may have landed partially, so it stays in the human queue where an artifact check or a person can settle it.

`reconcilers.py`: specs gain a `type` field. `command` stays the default when `type` is absent, so an existing registry keeps working unchanged. Per-type required keys are enforced and unknown keys are refused, so a typo'd spec fails loudly instead of registering a probe that settles nothing. `settle_run` records the evidence as the settlement's provenance. The new `--strict` flag escalates anything a probe could not settle to `REQUIRES_REVIEW` rather than leaving it pending, so a run tolerating no uncertainty degrades to a human instead of a silent retry.

The eBPF collectors stay out of process. `examples/ebpf_reconciler_adapter.py` reads the JSON a collector such as Tetragon or AgentSight emits, matches it against the action on stdin, and prints the probe verdict. It consumes standard outputs and ships no BPF programs, as the issue's "alternatives considered" section asks for.

## Related issues

Fixes #268
Refs #244 (umbrella), #765 (the ActionReconciler seam this extends)

## Types of changes

- Type: `feature`

## Breaking changes

No. An existing registry with no `type` on any spec parses identically to before, and command specs keep their exact `{command, timeout}` shape, which `test_the_documented_registry_example_registers_its_probe` continues to pin. The one behaviour change is additive: `reconcile --strict` is a new flag, and the discrepancy pass reports findings the command previously did not look for.

## How has this been tested

The four acceptance criteria, against the real ledger and the real event log, in `tests/test_evidence.py`:

1. A crash-between-intercept-and-complete scenario is settled by an `otel_span` probe in non-strict mode, with the settlement event citing span and trace ids.
2. An `artifact_check` probe proves `occurred=false` for a claimed-but-absent write.
3. Contradictory evidence produces a review finding, not a settlement.
4. The collector-dependent test skips cleanly without the collector present.

Also asserted: a failed span does not settle as not-occurred, a digest mismatch declines rather than attributing, `--strict` escalates an unsettled action, settlement provenance is `DETERMINISTIC` with span ids attached, and every registry validation arm (unknown type, unknown key, missing required key) raises `ReconcilerConfigError`.

Commands run, Python 3.13.5, installed via `pip install -e ".[dev]"`, on an isolated worktree off `origin/main`:

```
pytest tests/test_evidence.py tests/test_reconcilers.py tests/test_otel.py \
  tests/test_cli_docs.py                                  # 63 passed, 2 skipped
pytest tests/test_evidence.py tests/test_evidence_injection.py \
  tests/test_reconcilers.py tests/test_reconciliation.py tests/test_action_ledger.py \
  tests/test_ledger_replay.py tests/test_action_index.py tests/test_otel.py \
  tests/test_recovery_engine.py tests/test_recovery_guidance.py tests/test_recovery_ledger.py \
  tests/test_storage.py tests/test_authority_probe.py tests/test_authority_consumed.py \
  tests/test_cli.py tests/test_cli_docs.py                # 425 passed, 5 pre-existing failures
ruff check src/ tests/ examples/                          # all checks passed
ruff format --check <my files>                            # already formatted
mypy src/continuum                                        # no issues in 128 files
```

The 5 failures are pre-existing and reproduce on a tree without this change: 4 `test_attest_*` tests raise `RuntimeError` from the keygen path and 1 `test_mcp_resume_includes_human_steps` needs the `mcp` package, neither installed in this environment.

A second run under a Python 3.13 interpreter that has `pytest-asyncio` collected the full suite and ran it clean: 2,279 collected, 2,252 passed, 27 skipped, 0 failed, which is what CI compares. That run also caught what the first push missed: the 24 new tests pushed the collected total past the 30-test tolerance of `tests/test_docs_counts.py`, and all nine CI Test jobs failed on the guard itself even though every test passed. The second commit re-syncs the documented figures in `README.md`, `docs/CONTRIBUTING_ONBOARDING.md` and `CHANGELOG.md` to that real run, and CI is green since.

## Checklist

Tests added or updated for the changed behaviour. Documentation updated where the change affects user-facing behaviour: `docs/api/cli.md` gains the new probe types and `--strict`, and `docs/guides/evidence-reconciliation.md` walks through both probes and the contradiction check with the Tetragon and AgentSight recipes. CI runs on the latest commit. Security-relevant changes state known limitations explicitly: settlement from a span or a file is still settlement from evidence the host controls, so a compromised host can forge either, exactly as it could forge a command probe's stdout; the discrepancy check catches a ledger that disagrees with the filesystem, not a filesystem that lies.

## Additional context

The acceptance criteria are met without shipping eBPF programs, which is the tradeoff the issue's "alternatives considered" preferred. In-process BPF loading was rejected there for the kernel-version matrix and platform lock-in; consuming standard outputs gets most of the value and keeps the Linux-only surface in the collector, where it already lives. The `otel_span` probe is protocol-neutral, so it is still useful if MCP-level attestation standards arrive later, and `artifact_check` needs no telemetry at all.

Auto-settlement still only ever shrinks what a person must look at: a definitive verdict settles, everything else leaves the action pending or, under `--strict`, escalates. Strict mode refuses to auto-settle unknowns and degrades to human, unchanged.


Cyrax321:
Self-review (blocking finding). This PR is my own work from this session, so this is the author's second pass, not an independent gate. GitHub does not allow request-changes on one's own PR, so I am recording it as a comment with the blocking label explicit.

**Blocking: `OtelSpanReconciler._matches` matches an action with no identity tokens to any span of that tool.**

At `src/continuum/evidence.py:133`, the loop at 125-132 returns `False` only on an explicit disagreement between tokens present on both sides. When `agreed` stays 0 because the action carries no `path` argument, execution falls through to `return True`, and the probe then matches the first completed span of that tool regardless of which file it wrote.

That contradicts the comment two lines above it, which states the intent: "an action with no path to compare must not match every write the tool ever made." The comment describes the invariant; the code violates it. The observable consequence is a false settlement: an unpathed `write_file` action would be recorded as occurred, cited to a span that wrote an unrelated file.

Reproduced against the shipped class:

```
action has NO path arg; span for a DIFFERENT file:
  matches? True
```

The existing tests miss it because every one claims an action with a path argument.

Fix is one line, and it is what the comment already promises: `return agreed > 0`. A covering test asserting a no-path action does not settle from an unrelated span lands with it. Pushing both now.

Everything else re-verified and holds: the discrepancy detector skips action types without an `artifact_check` probe, a failed span settles nothing rather than licensing a retry, the command-spec shape is unchanged for existing registries, and the docs-count re-sync matches the real 2,279/2,251/27 figures.

Cyrax321:
Fixed and pushed in c5ce2f86.

`_matches` now returns `agreed > 0`, so a match requires at least one identity token to actually constrain it. An action with no path argument can no longer settle from the tool's next unrelated span.

The covering test deserves a note, because its first draft was wrong: I placed the span before the claim, and it passed against the buggy matcher for an unrelated reason (the sequence guard rejected it, not the identity logic). Reordered so the span follows the claim, it now fails on the pre-fix code with the exact false settlement and passes after.

Counts re-synced to 2,280 collected (2,253 passed, 27 skipped, 0 failed) across the three guarded files, and the full suite is green.

Cyrax321 (commented):

Ref: #1189 / Issue #268
Surface: reconcilers, new evidence probes, CLI docs

What this adds: two built-in evidence probe types (otel_span, artifact_check) that let the reconciler registry settle uncertain actions without a hand-written command probe, plus detect_discrepancies, which escalates ledger-versus-filesystem contradictions to REQUIRES_REVIEW instead of silently re-settling.

Design reviewed and agreed:
- The safety posture is the strongest part. A failed span settles nothing (returning None keeps the human queue intact, since a failed call is not evidence of absence), a digest mismatch declines to attribute the file to anyone, and discrepancies escalate rather than pick a side. Settling and contradicting are separate passes, which is the right separation.
- Claim-sequence anchoring scans full history so pre-compaction claims still reconcile (issue #647 noted), and span citations carry event, span, and trace ids, so a settlement is auditable back to the trace.
- The identity-token requirement (at least one token must agree, commit c5ce2f86) closes the sibling-write false-match.
- Integration is clean: build_probe composes with load_reconcilers, the command probe path is untouched, and no BPF ships in process.

Proof:
- Ran tests/test_evidence.py at head 1ee5474c: 24 passed, 1 skipped (the real-collector test, skipped by design off Linux).
- CI fully green on the head commit, including lint and strict type-check.
- One non-blocking observation for a follow-up: in reconcilers.py _run_probe, the generic exception handler references probe.name, but if build_probe itself raised a non-ReconcilerConfigError (for example a malformed identity spec that tuple() rejects), probe would be unbound and the handler would raise NameError instead of returning the error tuple. Extremely narrow, registry-malformation only. Worth a one-line guard someday, not worth holding this PR.

Cyrax321 (commented):
Maintainer re-review at current head. The OTel and artifact probe design with fail-closed safety posture is still sound. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review with no blocking issues at last review. Rebase out of dirty state and confirm the evidence suite green.

PR #1190, opened 2026-09-16
fix(recovery): a consumed authority escalates the verdict, never downgrades it (#1146)

### LINKED ISSUE #1146: recovery: a consumed authority downgrades an abort or rollback verdict to request_human, breaking the max-severity invariant
### Summary

A run holding a consumed authority has its recovery mode **unconditionally overwritten** with `request_human` (`recovery/engine.py:446`). When the risk policy had already proposed `abort` or `rollback` — both strictly more cautious — the override throws the stricter verdict away. The module's own documented invariant, two lines above the table, is *"The engine always returns the maximum proposed mode"* (`SEVERITY`, `engine.py:69`). This is the one place it does not.

### Affected component

Core (state, checkpoints, recovery) — recovery engine, authority lifecycle (#289c)

### Severity

### Regression check

### System information

- OS: Windows 11 Home 22H2
- Python: 3.11.9

```bash
python - <<'PY'
import sys; sys.path.insert(0, "src")
from continuum.models import Origin, RecoveryMode, Run
from continuum.recovery.engine import SEVERITY, RecoveryEngine

def build(with_authority: bool) -> SQLiteStorage:
    st.create_run(Run(run_id="run_1", goal="Analyze 100 documents"))
    st.append_event("run_1", EventType.RUN_STARTED, {"goal": "g", "total": 100})
    st.append_event("run_1", EventType.DEPENDENCY_DECLARED, {"resource": "dataset", "version": "v3"})
    st.append_event("run_1", EventType.EVIDENCE_ADDED, {"evidence_id": "paper_1", "summary": "s", "source": "dataset"})
    st.append_event("run_1", EventType.FINDING_ADDED, {"finding_id": "f1", "claim": "X", "evidence": ["paper_1"]})
    # The default risk policy maps side_effect_duplicate -> ABORT
    st.append_event("run_1", EventType.RISK_OBSERVED,
                    {"trigger": "side_effect_duplicate", "score": 0.9, "detail": "dup"},
    if with_authority:
        # The consumed credential the duplicate resurrected
        st.append_event("run_1", EventType.AUTHORITY_CONSUMED,
                        {"authority_id": "authz:stripe-1", "consumer_run_id": "run_1", "via_action_id": "a1"})
    return st

both = RecoveryEngine(build(True)).assess("run_1")
risk_only = RecoveryEngine(build(False)).assess("run_1")
print("risk + consumed authority :", both.mode.value, "severity", SEVERITY[both.mode])
print("risk only                :", risk_only.mode.value, "severity", SEVERITY[risk_only.mode])
print("expected (max of the two):", max(SEVERITY[risk_only.mode], SEVERITY[RecoveryMode.REQUEST_HUMAN]))
PY
```

### Current behaviour

```
risk + consumed authority : request_human severity 4
risk only                : abort severity 6
expected (max of the two): 6
```

The run that policy classified `abort` comes back as `request_human`. The rationale reports both signals honestly — `('risk side_effect_duplicate triggers abort', "consumed authority blocks resume: ['authz:stripe-1']")` — and then the code keeps the weaker of the two.

The site is a hard override, not a floor:

```python
# src/continuum/recovery/engine.py:445
if consumed_authorities:
    mode = RecoveryMode.REQUEST_HUMAN     # discards ROLLBACK (5) and ABORT (6)
    rationale = (*rationale, f"consumed authority blocks resume: {sorted(consumed_authorities)}")
```

The downgrade propagates to everything downstream that reads the mode:

| | risk-only verdict | with consumed authority |
|:--|:--|:--|
| `RecoveryDecision.mode` | `abort` | `request_human` |
| `SEVERITY` | 6 | 4 |
| contract `safety` (`_SAFETY_FOR_MODE`) | `unsafe` | `requires_human` |
| `continuum resume` exit code (`exit_code_for`) | **30 `UNSAFE`** | **20 `REQUIRES_HUMAN`** |

`request_human` also changes what the operator is told to do. `cli/main.py:1489` prints *"Run with `--repair` to record the repair plan"* for every non-`resume` mode, so on an `abort`-classified run the CLI now points someone at `--repair` — the escape hatch for recoverable state — for a run the risk policy said must not be recovered in place. `meltdown -> ROLLBACK` is the same shape, one step down: the automatic rollback signal is disarmed and becomes a parked run waiting for a person.

### Expected behaviour

The block's stated purpose, in its own comment, is that *"the recovery decision must also surface the block so that `continuum resume` does not report safe when the gate would still deny."* That is an escalation requirement — *at least* `request_human` — and the engine already has the machinery to express it: the `SEVERITY` table exists precisely so the most cautious proposal wins, and the risk loop ten lines above uses it (`engine.py:409`, `if SEVERITY[candidate] > SEVERITY[best_mode]`).

So the block should raise rather than replace:

```python
if consumed_authorities and SEVERITY[mode] < SEVERITY[RecoveryMode.REQUEST_HUMAN]:
    mode = RecoveryMode.REQUEST_HUMAN
```

A run already at `rollback` or `abort` keeps its verdict and gains the rationale line; a run at `resume`/`repair_and_resume`/`replan`/`wait` escalates exactly as it does today. Today's behaviour is preserved for every mode the block was written to catch, and the two stricter modes stop being weakened by the check that was supposed to strengthen them.

### Why this matters

Both signals firing at once is the realistic case, not a contrived overlap: the way an agent resurrects a spent credential is by repeating the side effect that used it, which is exactly what `side_effect_duplicate` names. The combined run is the one that most needs `abort` to survive, and it is the one that loses it. The CLI's documented contract is that *"only a verified-safe run exits 0"* with `UNSAFE` (30) reserved for runs that must not resume; a run the operator's own risk policy put in that tier now exits 20, and the `&& ./start-agent.sh` idiom the docs publish reads 20 as "a person must decide" rather than "do not resume."

### Logs or CI run

Output above, from `main.py`'s own `RecoveryEngine` on commit `2e835ad`. Related but distinct open issues: #1066 (the same block degrades fail-open on a read error) and #1050 (the same block reads the live tail only). Neither covers the mode override itself.

abyyxhek:
## Summary

Closes #1146.

`RecoveryEngine.assess` overwrote the recovery mode with `REQUEST_HUMAN`
whenever a consumed authority blocked resume (`recovery/engine.py`), even when
the risk policy had already proposed `ABORT` or `ROLLBACK`. Both are strictly
more cautious than `REQUEST_HUMAN` — `SEVERITY` is ascending caution and the
module documents that "the engine always returns the maximum proposed mode" —
so the block was the one place the engine did not. The rationale kept naming
the abort the verdict no longer delivered.

## The change

```python
-            mode = RecoveryMode.REQUEST_HUMAN
+            mode = max((mode, RecoveryMode.REQUEST_HUMAN), key=lambda m: SEVERITY[m])
```

A consumed authority now raises the floor of the verdict without weakening its
ceiling. The `REQUEST_HUMAN` rationale line is still appended, so the contract
still names the block as a contributing reason.

## Verification

The issue's reproducer, before vs after:

```
mode with BOTH abort-risk and consumed authority: request_human   (severity 4)
mode with abort-risk ONLY (no consumed authority): abort          (severity 6)
DOWNGRADE BUG: 4 vs 6

# after
mode with BOTH abort-risk and consumed authority: abort           (severity 6)
```

- New regression test
  `tests/test_authority_probe.py::test_consumed_authority_does_not_downgrade_a_stricter_verdict`
  asserts the mode stays `ABORT` and the rationale names both the trigger and
  the authority.
- `tests/test_authority_probe.py` (8), plus `test_gate_resurrection`,
  `test_recovery_engine`, `test_cli_gate`, `test_restore_merge_gate`,
  `test_recovery_policy` — 102 passed.
- `uvx ruff@0.16.4 check src/ tests/` and `format --check` clean;
  `mypy src/continuum/recovery/engine.py` clean.

## Checklist

- [x] CHANGELOG `### Fixed` entry added
- [x] Gates run locally (ruff 0.16.4, mypy, pytest)

## Summary by CodeRabbit

- **Bug Fixes**
  - Recovery decisions now preserve the strictest applicable action when a consumed authority blocks resumption.
  - High-severity outcomes such as abort or rollback are no longer weakened to a human-review request.
  - Recovery explanations continue to identify both the consumed authority and the risk trigger, improving clarity when reviewing why an action was selected.


Cyrax321 (dismissed):
Approve. This fixes a real downgrade bug on the recovery verdict path, and the fix is the correct one.

The old line did `mode = RecoveryMode.REQUEST_HUMAN` unconditionally whenever a consumed authority blocked resume. That overwrote whatever the risk policy had proposed. Since `SEVERITY` is ascending caution and the module documents at line 69 that "the engine always returns the maximum proposed mode", an already-proposed `ROLLBACK` (5) or `ABORT` (6) was being silently weakened to `REQUEST_HUMAN` (4). The rationale string still named the abort the verdict no longer delivered, so the output was self-contradictory, not merely suboptimal.

`max((mode, REQUEST_HUMAN), key=lambda m: SEVERITY[m])` is exactly right: it raises the floor without lowering the ceiling, and it matches how the engine already picks the winning proposal at line 728 (`max(proposals, key=...)`). The fix makes this block consistent with the engine's own invariant instead of being the one exception to it.

Verification: I ran `tests/test_authority_probe.py` on this branch (8 passed) and then checked the new test against the pre-fix engine. It fails there with `assert <RecoveryMode...request_human> == <RecoveryMode.ABORT>`, so it is a real regression test and not a tautology. It exercises the actual path through `assess` rather than monkeypatching `SEVERITY`.

Two non-blocking notes:

The docstring quotes the invariant from the module, which is good, but the comment in the engine and the docstring between them restate the rationale three times. That is fine for a safety path where the reasoning is the point, and I would not trim it.

More substantively: `max()` over `SEVERITY` means a consumed authority can no longer force `REQUEST_HUMAN` at all when the risk policy proposed `ABORT`. That is the intended behaviour and it is strictly more cautious, but it does change what an operator sees in the `ABORT` case, where the consumed-authority block now contributes to the rationale without changing the outcome. Worth confirming the dashboard and MCP surfaces render the combined rationale sensibly in that case, though nothing here suggests they would not.

Cyrax321 (approved):

Ref: #1190 / Issue #1146
Surface: RecoveryEngine.assess, consumed-authority block

Bug: the consumed-authority block overwrote the mode with REQUEST_HUMAN, discarding a stricter ABORT or ROLLBACK the risk policy had already proposed, violating the module's own invariant at src/continuum/recovery/engine.py:68 (ascending caution, engine returns the maximum proposed mode) while the rationale still named the discarded verdict.

Fix: escalate with max((mode, REQUEST_HUMAN), key=SEVERITY) instead of overwriting. This is the minimal correct change at the ownership boundary, and the rationale line is kept so the contract still names the block as a contributing reason.

Proof:
- Read the SEVERITY table and the max-proposal logic (engine.py:437, 731); the change is consistent with how every other proposal is combined.
- New regression test test_consumed_authority_does_not_downgrade_a_stricter_verdict asserts ABORT survives a consumed authority and the rationale names both causes.
- Ran the full test_authority_probe.py at head 9605d35f: 9 passed.
- Verified the test fails against the old overwrite behavior (locally reverted the one line, test failed as expected, restored), so it is a genuine regression test.

Note: #1222 wires the related consumed-authority check through is_authority_consistent; whichever of the two merges second should rebase, they touch adjacent lines in the same engine path.

PR #1191, opened 2026-09-16
fix(recovery): raise specific precondition error subclasses for merge and restore (#1114)

### LINKED ISSUE #1114: the edit gate raises ForkPreconditionError for forks but the base class for everything else, so MergePreconditionError and RestorePreconditionError are never raised
## Summary

The edit-precondition gate raises a type-specific exception for forks but the plain base class for every other edit type. `src/continuum/recovery/gate.py:686-688`:

```python
err_cls: type[EditPreconditionError] = (
    ForkPreconditionError if edit_type == "fork" else EditPreconditionError
)
```

So `MergePreconditionError` (`recovery/merge.py:33`) and `RestorePreconditionError` (`recovery/restore.py:33`) — both `EditPreconditionError` subclasses, both exported through their modules' `__all__` and re-exported by `recovery/__init__.py:48-49` — are **never raised anywhere**. `git grep` for a raise of either comes up empty; the only raised precondition error is the base class at `gate.py:569` via the `err_cls` selection above.

Consequences:

- A caller cannot distinguish a merge refusal from a restore refusal by exception type — the entire point of having per-edit-type subclasses.
- The CLI catches the base class everywhere (`cli/main.py:1860`, `:1953`, `:2031`, `:2134`), so nothing would change for it.
- The tests already hedge around the gap: `tests/test_restore_merge_gate.py:81`, `:147`, `:234`, and `:236` all use `pytest.raises((ForkPreconditionError, EditPreconditionError))` — the tuple is only necessary because merge and restore raise the base while fork raises the subclass. (`tests/test_fork_preconditions.py` asserts the specific `ForkPreconditionError` throughout, showing what the merge/restore tests should look like.)

This looks like the #407/#408 "route restore and merge through the same precondition gate" work stopping one edit_type short of the fork handling.

## Affected component

Other (recovery edit gate)

## Severity

Minor inconvenience — no wrong verdicts, but an exported API distinction that cannot be observed at runtime.

## System information

- Python: 3.13.5

```bash
git grep -n "raise MergePreconditionError\|raise RestorePreconditionError" -- src/
# (no output)

sed -n '686,688p' src/continuum/recovery/gate.py
# err_cls: type[EditPreconditionError] = (
#     ForkPreconditionError if edit_type == "fork" else EditPreconditionError
# )
```

## What a fix looks like

Extend the `err_cls` selection at `gate.py:686` to map `edit_type` to its subclass — `"fork" -> ForkPreconditionError`, `"merge" -> MergePreconditionError`, `"restore" -> RestorePreconditionError` (a small dict or `if/elif`) — and tighten the hedged `pytest.raises((...))` tuples in `tests/test_restore_merge_gate.py` to the specific types. The exceptions already carry `edit_type` in their payload, so no constructor changes are needed.

soorajsuru2004-sudo:
## Description

`check_preconditions` in `src/continuum/recovery/gate.py` previously raised `ForkPreconditionError` when `edit_type == "fork"`, but defaulted to base `EditPreconditionError` for all other edit types. Consequently, `MergePreconditionError` (in `recovery/merge.py`) and `RestorePreconditionError` (in `recovery/restore.py`) were never raised anywhere in the runtime codebase. Callers could not catch edit-specific precondition failures for restore and merge operations by class.

This PR:
1. Defines `RestorePreconditionError` and `MergePreconditionError` directly in `src/continuum/recovery/gate.py` and exports them through `__all__`.
2. Updates `recovery/merge.py` and `recovery/restore.py` to import and re-export `MergePreconditionError` and `RestorePreconditionError` respectively, preserving backwards-compatible import paths without circular imports.
3. Updates `check_preconditions` to dispatch `err_cls` based on `edit_type` (`fork` -> `ForkPreconditionError`, `restore` -> `RestorePreconditionError`, `merge` -> `MergePreconditionError`), falling back to `EditPreconditionError` for unknown types.
4. Tightens assertions in `tests/test_restore_merge_gate.py` to assert the specific subclasses instead of accepting the hedged `(ForkPreconditionError, EditPreconditionError)` tuple, and adds explicit inheritance assertions.
5. Documents the fix in `CHANGELOG.md` under `[Unreleased]` -> `Fixed`.

## Related issues

Fixes #1114

## Types of changes

- Type: `bugfix`

## Breaking changes

No. `RestorePreconditionError` and `MergePreconditionError` inherit from `EditPreconditionError`, so existing call sites catching `EditPreconditionError` continue to work without modification.

## How has this been tested?

Ran the following suites on Windows (Python 3.12.10):

```
pytest tests/test_restore_merge_gate.py tests/test_fork_preconditions.py tests/test_merge_gate_source_494.py tests/test_docs_counts.py -v
ruff check src/ tests/
```

All 34 tests passed cleanly and all ruff checks passed.

## Checklist

## Additional context

None.


Cyrax321 (approved):
Approve. The move is clean and the compatibility surface is intact.

The change relocates `RestorePreconditionError` and `MergePreconditionError` from `restore.py`/`merge.py` (where they were empty `pass` subclasses) into `gate.py` alongside `ForkPreconditionError`, then has `check_preconditions` dispatch on `edit_type` through a dict instead of the old `fork ? ForkPreconditionError : EditPreconditionError` ternary. `EditType` is exactly `Literal["fork", "restore", "merge"]`, so the three-way dict covers the whole space and the `.get(..., EditPreconditionError)` fallback is a reasonable belt-and-braces default rather than dead code.

Nothing regresses for existing callers, and I verified the two things that could have:

- `continuum/recovery/__init__.py` still re-exports both names from `merge` and `restore`, so the public import path is unchanged.
- The four `except EditPreconditionError` handlers in `cli/main.py` and the one in `checkpoint/rewind.py` still catch the new subclasses, because inheritance is preserved. That was the load-bearing property: the base class stays the catch-all.

Tests: `tests/test_restore_merge_gate.py` passes 15/15 on this branch. The new `test_precondition_error_subclasses_inherit_from_base` pins the inheritance contract, and the parametrized refusal tests now assert the exact expected subclass per edit type rather than the old `(ForkPreconditionError, EditPreconditionError)` tuple, which is a strict strengthening.

One non-blocking note: the two new classes are structural copies of `ForkPreconditionError`, including the same four keyword-only parameters. If a fourth edit type ever joins `EditType`, the copy-paste cost grows. A factory or a `_subclass_for(edit_type)` helper would absorb it, but for three members the explicit classes read more clearly, and they keep the names importable and documentable, which is what the callers in `__init__.py` actually depend on.

src/continuum/recovery/gate.py
```diff
@@ -683,9 +727,11 @@ def check_preconditions(
             + "; ".join(parts)
             + ". Pass carry_forward with the identifiers you intend to carry, or reconcile first."
         )
-        err_cls: type[EditPreconditionError] = (
-            ForkPreconditionError if edit_type == "fork" else EditPreconditionError
-        )
+        err_cls: type[EditPreconditionError] = {
+            "fork": ForkPreconditionError,
+            "restore": RestorePreconditionError,
+            "merge": MergePreconditionError,
```
Cyrax321, line 733: The dispatch is right, but only one merge path reaches it. The single-run arm of `check_merge_preconditions` delegates here through `check_preconditions(edit_type="merge")`, while the cross-run arm raises base `EditPreconditionError` directly at gate.py:613. A caller catching `MergePreconditionError` therefore sees single-run refusals but not cross-run ones, which is the path with the union rationale. Raising the subclass there is the missing half of the fix.

src/continuum/recovery/gate.py
```diff
@@ -106,6 +108,48 @@ def __init__(
...
+            message,
+            edit_type=edit_type,
+            derivation=derivation,
+            unaccounted=unaccounted,
+            rationale=rationale,
+        )
+
+
+class MergePreconditionError(EditPreconditionError):
+    """Alias for :class:`EditPreconditionError` when ``edit_type`` is ``merge``."""
```
Cyrax321, line 133: This is not an alias. The class is raised distinctly by the dispatch dict above, which is the entire point of the PR, and the wording was already wrong for `ForkPreconditionError` before this change. Saying "alias" invites a reader to treat the three as interchangeable, which is the assumption the issue exists to correct.

Cyrax321 (changes requested):
Requesting changes, on one gap between what the PR promises and what it delivers. The wiring is correct where it reaches: I confirmed `approve_restore` and the single-run `approve_merge` both call `check_preconditions` with `edit_type="restore"` and `edit_type="merge"`, so the dispatch dict now raises the specific subclass on those paths. The re-exports preserve class identity (`recovery.merge.MergePreconditionError is recovery.gate.MergePreconditionError`), so a caller importing from either module gets one class object, and backwards compatibility holds because both subclass `EditPreconditionError`.

The gap is the cross-run merge. `check_merge_preconditions` has two arms: when `source_run_id` is None it delegates to `check_preconditions(edit_type="merge")` and now raises `MergePreconditionError`, but when a source run is supplied it computes the union derivation itself and raises base `EditPreconditionError` directly at gate.py:613. That is the richer refusal path, the one whose rationale names both sides sequences, and the one a caller is more likely to want to catch by class. So `except MergePreconditionError` catches a single-run merge refusal and misses a cross-run one. The summary says callers can distinguish refusal types by exception class, which is true for fork and restore and half-true for merge. Raising `MergePreconditionError` at that line closes it, and a cross-run test asserting the subclass would pin the behaviour the tests currently only cover on the single-run path.

Non-blocking: the docstrings on the new classes say "Alias for EditPreconditionError when edit_type is X", carried over from the empty `pass` bodies they replace. They are not aliases, and were not before this PR either, since `ForkPreconditionError` was already raised distinctly; the wording is now wrong for three classes instead of one. The two `__init__` bodies are also identical apart from the default `edit_type`, which a shared implementation would remove.

Cyrax321 (changes requested):
Thank you for contributing to Continuum, @soorajsuru2004-sudo. The subclass relocation itself is correct: MergePreconditionError and RestorePreconditionError now live in recovery/gate.py with class identity preserved through the re-exports, backwards compatibility holds through the EditPreconditionError base, and the dispatch dict raises the specific subclass on the single run paths, which I verified through approve_restore and the single-run approve_merge arm. The tightened assertions are the right pins.

One item still blocks merge, unchanged since the last review. The cross-run merge arm in check_merge_preconditions computes its union derivation itself and raises base EditPreconditionError directly at src/continuum/recovery/gate.py:613, with edit_type merge. That is the richer refusal path, the one whose rationale names both sides, and the one a caller is most likely to catch by class. So except MergePreconditionError catches a single-run merge refusal and misses a cross-run one, while the PR summary promises callers can distinguish refusal types by exception class. Please raise MergePreconditionError at that line and add a cross-run test asserting the subclass, mirroring the single-run coverage.

Non-blocking for a follow up: the new class docstrings say alias for EditPreconditionError, which understates them now that they are raised distinctly. Thanks again for taking this on.

Cyrax321 (changes requested):
Author context: soorajsuru2004-sudo, external contributor. Follow-up to the prior reviews. The subclass relocation and dispatch-dict direction is still correct with the compat surface intact, but the cross-run merge gap still blocks: check_merge_preconditions with a source run raises base EditPreconditionError at gate.py:613 instead of MergePreconditionError, so except MergePreconditionError misses the richer refusal path. Raise MergePreconditionError there and add a cross-run test asserting the subclass.

PR #1192, opened 2026-09-16
fix(webhooks): dedup against the archived prefix so compaction can't re-ring the bell (#1186)

### LINKED ISSUE #1186: webhooks: the re-notify dedup window reads the live tail only, so a compaction inside the window re-rings the bell
## Summary

`recovery/webhooks.py:_within_dedup_window` scans `storage.read_events(run_id)` — the live tail only — for the `NOTIFICATION_SENT` / `NOTIFICATION_FAILED` rows the dedup state lives in. `continuum compact` archives exactly those rows, so a compaction inside an endpoint's re-notify window (default `3600`s) makes the next blocked-run notification round deliver the same verdict again.

The module's own docstring states the invariant this breaks:

> the dedup that keeps a polling cron from spamming: one notification per *transition* into a blocked state, not per re-assessment, keyed on `(run_id, mode, contract hash)` with a re-notify interval. The dedup state is durable because it lives in the event log itself (`NOTIFICATION_SENT` / `NOTIFICATION_FAILED` rows), so a restart does not re-ring the bell

The dedup state is durable across a *restart*. It is not durable across a *compaction*, which is the operation the docs prescribe for a long blocked run in the first place.

## Affected component

`continuum.recovery.webhooks` (`_within_dedup_window`, `src/continuum/recovery/webhooks.py:234`), reached from `continuum resume` via `_notify_blocked_run` (`src/continuum/cli/main.py:1280`) and by every polling cycle of `continuum watch`.

## Severity

Major functionality broken, workaround exists. This is the exact spam the dedup exists to prevent, and the failure direction is always toward more noise — an operator who was paged once and then compacted the run to shrink the log gets paged again for the same standing blockage, and once archived, on *every subsequent* blocked assessment until the window finally expires. The workaround is to never compact a blocked run, which is not a realistic constraint for a run that blocks for hours.

Against `upstream/main` at `cef019d`, with the network stubbed so delivery always succeeds:

```python
import sys
from datetime import timedelta

import continuum.recovery.webhooks as webhooks
from continuum.actions.authority import record_authority_consumed
from continuum.models import Run, utcnow

RUN_ID = "run-webhook-dedup"
ENDPOINT = webhooks.WebhookEndpoint(url="https://hooks.example.com/blockage",
                                    re_notify_seconds=3600)
REGISTRY = webhooks.WebhookRegistry(endpoints=(ENDPOINT,))

posts = []
webhooks.post_webhook = lambda url, body, secret=None, timeout=5.0: (posts.append(url), True)[1]

st.create_run_started(Run(run_id=RUN_ID, goal="charge the card"))

# A consumed, unreconciled authority blocks the run (REQUEST_HUMAN).
record_authority_consumed(st, RUN_ID, "approval-7", via_action_id="act-1")
decision = RecoveryEngine(st).assess(RUN_ID)
mode, contract = decision.mode.value, decision.contract
print("verdict:", mode)

webhooks.notify_blocked(st, RUN_ID, mode=mode, payload={"run_id": RUN_ID},
                         contract=contract, registry=REGISTRY)
print("round 1 POSTs:", len(posts))                          # 1

# The operator reacts the way the docs suggest for a long blocked run.
CheckpointManager(st).checkpoint(RUN_ID, trigger="manual", reason="anchor")
st.compact_run(RUN_ID)

webhooks.notify_blocked(st, RUN_ID, mode=mode, payload={"run_id": RUN_ID},
                         contract=contract, registry=REGISTRY,
                         now=utcnow() + timedelta(minutes=20))
print("round 2 POSTs (20 min later, inside the window):", len(posts))
```

Output:

```
verdict: request_human
round 1 POSTs: 1
round 2 POSTs (20 min later, inside the window): 2
```

Between the two rounds the only state change is the compaction: the `NOTIFICATION_SENT` row moves from the live tail into the archived prefix (`live=0, archived=1`), and `_within_dedup_window` no longer sees it. `read_all_events` still returns it, with its original timestamp, so the window computation itself is unaffected by the archived prefix — the scan just never looks there.

## Root cause

`_within_dedup_window` was written when the dedup state's only enemy was a process restart. It predates the archive-aware convention established for the other event-history scans in `assess` — the same convention `reconcilers.py:369` already states explicitly for `AUTHORITY_CONSUMED` ("Full history, not the live tail: the AUTHORITY_CONSUMED row can predate a compaction"). Every other consumer that derives durable state from the event log (`ledger._replay`, `cli/main.py:2947`, `gateway.py:460`, `provenance_for_run`) merges the archived prefix; this one does not.

## What a fix looks like

Scan the merged history instead of the live tail:

```python
from heapq import merge

stream = merge(
    storage.read_archived_events(run_id),
    storage.read_events(run_id),
    key=lambda e: e.sequence,
)
for event in stream:
    ...
```

The `timestamp` comparison is already the only thing that decides the window, and archived events keep their original timestamps, so no other change is needed. `notify_blocked` is documented to never raise, and the surrounding `re_notify_seconds` arithmetic is unchanged.

Two adjacent notes, deliberately filed separately rather than folded in, since both are already on the tracker: the same blindness sits behind #1072 (`continuum watch` re-firing liveness events) and #1175 (`continuum resume` appending `NOTIFICATION_FAILED` to the log it then dedups against).

## System information

- OS: macOS 26.0.1 (Darwin 25.0.0, arm64)
- Python: 3.13.5
- continuum-agent: from source at `upstream/main` `cef019d`

Adhi1-2:
## Summary

Fixes #1186.

`_within_dedup_window` (`src/continuum/recovery/webhooks.py`) scanned only the live tail — `storage.read_events(run_id)` — for the `NOTIFICATION_SENT` / `NOTIFICATION_FAILED` rows the webhook dedup state lives in. `compact_run` archives exactly those rows, so a compaction inside an endpoint's re-notify window (default 3600s) made the next blocked-run assessment deliver the same verdict again. This is the precise spam the dedup exists to prevent, and it fails toward more noise every time: an operator who was paged once and compacted the long blocked run to shrink the log — as the docs prescribe — gets paged again for the same standing blockage, and once archived, on every subsequent assessment until the window finally expires.

## The change

One-line behavioural fix: the scan now reads `storage.read_all_events(run_id)`, the merged archive-aware history that every other consumer deriving durable state from the event log already uses (`ledger._replay`, `cli/main.py:2947`, `gateway.py:460`, `provenance_for_run`, `reconcilers.py:372`). This is the archive-aware convention `reconcilers.py` states explicitly for `AUTHORITY_CONSUMED`; the dedup scan predates it. The issue's suggested `heapq.merge` is already what `read_all_events` does internally, so this reuses the shared helper rather than open-coding a second merge.

Archived rows keep their original timestamps, so the window computation itself is unaffected — only the scan had to look there. `notify_blocked` remains documented to never raise, and the `re_notify_seconds` arithmetic is untouched.

## Verification

The issue's repro, run against this branch with delivery stubbed to succeed:

```
verdict: request_human
round 1 POSTs: 1
round 2 POSTs (20 min later, inside the window): 1   # was 2 before
```

- `tests/test_webhooks.py` gains `test_dedup_survives_compaction_of_the_notification_row`: notifies, compacts the run (asserting the `SENT` row has actually left the live tail), then re-notifies 20 min inside the window and asserts `skipped` with no second POST and no new event row — and that past the window the archived timestamp still lets it ring again, so the fix didn't deaden the expiry path.
- The new test fails on the pre-fix code (`['sent'] != ['skipped'`], confirming it guards the regression rather than passing vacuously.
- Full suite green; `ruff check`, `ruff format --check`, `mypy src/continuum`, and `markdownlint` on the touched CHANGELOG entry all clean.

## Out of scope

Two adjacent archive-blindness issues are deliberately left alone here since both are already tracked: #1072 (`continuum watch` re-firing liveness events) and #1175 (`continuum resume` appending `NOTIFICATION_FAILED` to the log it then dedups against).

## Summary by CodeRabbit

* **Bug Fixes**
  * Webhook notification deduplication now remains effective after run compaction.
  * Prevents duplicate verdict notifications during the configured re-notification window, including when notification history has been archived.
* **Tests**
  * Added coverage verifying deduplication both within and after the re-notification window.


Adhi1-2:
Both review findings addressed in 784cb4b:

- **@gitar-bot perf** (re-reading the full archive per endpoint): correct. `_within_dedup_window` now takes the already-merged history as a `Sequence[Event]`, and `notify_blocked` does one `read_all_events` per round that serves every subscribed endpoint. This is the reuse `storage/base.py` already prescribes for callers folding the same history more than once.
- **@coderabbitai style** (em dash): correct. The CHANGELOG entry had a real U+2014, replaced with sentence punctuation, and the dash-like hyphen in the new test docstring became a comma. Confirmed no U+2014/U+2013 anywhere in the diff.

Rerun of the issue's repro still gives `round 2 POSTs: 1`; local ruff, ruff format, mypy, and the full suite are green.

anya-research:

yuki-fuyutsuki:

[github run](/Cyrax321/CONTINUUM/actions/runs/35075231662)

src/continuum/recovery/webhooks.py
```diff
@@ -271,9 +282,12 @@ def notify_blocked(
         return []
     now = now or utcnow()
     key = verdict_key(mode, contract)
+    # One archive-aware read serves every endpoint; re-reading per endpoint
+    # would re-deserialise a long compacted run once per subscription.
+    history = storage.read_all_events(run_id)
```
Cyrax321, line 287: This is the load-bearing line of the fix and it is in the right position. Reading once here instead of once per endpoint keeps the CLI path at one archive deserialization regardless of how many endpoints an operator configured, which is the difference between a fix and a fix that slows down the exact long compacted runs it targets.

src/continuum/recovery/webhooks.py
```diff
@@ -217,21 +218,31 @@ class DeliveryRecord:
 
 
 def _within_dedup_window(
-    storage: Storage,
-    run_id: str,
+    history: Sequence[Event],
```
Cyrax321, line 221: Taking the merged history rather than the storage is the better signature for a second reason: it makes the archive-awareness a caller responsibility that a test can supply directly, and it removes a hidden storage read from a function whose name promises a predicate. The docstring now states the precondition where the next reader will look for it.

tests/test_webhooks.py
```diff
@@ -275,6 +275,70 @@ def test_new_verdict_and_expired_window_ring_again(store: SQLiteStorage) -> None
         server.shutdown()
 
 
+def test_dedup_survives_compaction_of_the_notification_row(
```
Cyrax321, line 278: The two halves of this test are what make it trustworthy. Asserting the SENT row left the live tail proves the fixture really compacted rather than assuming it; then asserting `skipped` inside the window and a second hit past it proves the fix preserved the expiry instead of deadening the bell.

Cyrax321 (dismissed):
Approving. This is the same archive-blindness class as #1193 and it is fixed the right way, and the refactor is better than the one-line change the issue asked for. Hoisting the history read above the endpoint loop and handing `_within_dedup_window` a `Sequence[Event]` means N subscribed endpoints cost one archive deserialization instead of N, which matters precisely because the runs this fixes are the long compacted ones an operator compacted on purpose. A naive `read_all_events` inside the per-endpoint call would have fixed the bug and quietly multiplied the cost of the CLI path it sits on.

The expiry path is what I most wanted confirmed and the test covers it. A fix that only asserted "no second POST" could pass while quietly making dedup permanent, which would be a worse bug than the one being fixed: an operator would stop being paged for a blockage that never cleared. Asserting `skipped` at twenty minutes and then a second hit at two hours proves the archived timestamp still governs the window, so the bell rings on time and not before.

On the checks: `anya-review` is red, but it failed on its own plumbing step and is not in the `main` branch protection set, which requires only Lint and Type-check plus the three Ubuntu Test jobs, all green here. Nothing merge-blocking is failing.

One consistency note worth recording: #1203, which I approved, adds a second notification round carrying its own `verdict_key`, and that round SENT row is just as archivable as this one. This fix is what keeps the review-round dedup alive after a compaction too, so the two want to land together or in this order.

PR #1193, opened 2026-09-16
fix(tui): fold the full log in the Actions tab so a compacted run still lists its blocked action

### LINKED ISSUE #1182: tui: the Actions tab folds the live log only, so a compacted run shows no actions while it is blocked
## Summary

`src/continuum/tui/model.py:278` builds the TUI's Actions tab by folding `storage.read_events(run_id)`, which is the **live tail only**. Every other consumer of the same fold reads the archived prefix too, because compaction moves action events into the archive without revoking them:

- `ActionLedger._replay` (`src/continuum/actions/ledger.py:533-549`) merges `read_archived_events` with `read_events`, and its docstring states why: *"a claim settled before compaction must keep protecting afterwards, or exactly-once would quietly reset at the anchor boundary."* `continuum actions`, `continuum reconcile`, and `RecoveryEngine.uncertain_actions` all go through this fold.
- `src/continuum/dashboard/hitl.py:126` (`pending_actions_with_keys`) folds `read_all_events`.
- The TUI's own `event_rows` (`model.py:301`) and `budget_rows` (`model.py:335`) use `read_all_events`; `budget_rows` documents the reason — *"attempts are counted over the whole log, so compaction does not hand a run a fresh budget (#734)."*

So on a compacted run the TUI shows a **different, smaller** action set than `continuum actions`, the web dashboard, and the recovery verdict. The worst case is visible in one place: the run row correctly reports a blocked run while the Actions tab below it is empty, leaving the operator staring at `safe: no` with nothing to act on.

This is the same archive-blindness family as **#809**, which fixed `pending_actions_with_keys` on the dashboard side. That issue was filed while reviewing the TUI PR (#803) and says of this exact line: *"model.action_rows copies the pattern; that PR carries the blocking finding."* The dashboard half was fixed; the TUI half is still live on `upstream/main`.

## Affected component

`continuum tui` (`src/continuum/tui/model.py`); consistency with `continuum actions`, the dashboard HITL surface, and the recovery verdict.

## Severity

Medium. It does not write anything, so exactly-once is not at risk — but it is the surface an operator reaches for *because* a run is blocked, and it silently disagrees with the verdict shown three lines above it.

## System information

- Python: 3.13.5

```python
from continuum.models import Origin, Run

run_id = "demo"
st.create_run(Run(run_id=run_id, goal="demo"))
st.append_event(run_id, EventType.RUN_STARTED, {"goal": "demo"}, source=Origin.HUMAN)

led = ActionLedger(st, run_id)
outcome = led.claim("charge_card", arguments={"amount": 100}, origin_digest="0" * 64)
led.fail(outcome.key, "timeout: response lost", certain=False)   # -> UNKNOWN

# Anchor after both action events, then compact: the action events are archived.
CheckpointManager(st).checkpoint(run_id, trigger="manual", reason="anchor")
st.compact_run(run_id)
print("live events:", [e.type.value for e in st.read_events(run_id)])
print("full events:", [e.type.value for e in st.read_all_events(run_id)])
print("continuum actions :", [a.status.value for a in led.all()])
print("tui action_rows  :", tui.action_rows(st, run_id))
print("tui run row      :", tui.run_rows(st)[0].safe, tui.run_rows(st)[0].mode)
print("recovery verdict :", RecoveryEngine(st).assess(run_id).safe,
      len(RecoveryEngine(st).assess(run_id).uncertain_actions), "uncertain action(s)")
```

Output against `cef019d`:

```
compact report: {'archived': 4}
live events: ['STATE_CHECKPOINTED', 'EVENT_LOG_ANCHORED']
full events: ['RUN_STARTED', 'ACTION_RECORDED', 'ACTION_RECORDED', 'STATE_CHECKPOINTED', 'STATE_CHECKPOINTED', 'EVENT_LOG_ANCHORED']
continuum actions : ['unknown']
tui action_rows  : []
tui run row      : no request_human
recovery verdict : False 1 uncertain action(s)
```

The TUI's run row is right (the run *is* blocked) and its Actions tab is empty, so the one thing the operator could settle from the keyboard is not listed.

## What a fix looks like

```diff
-    folded: dict[str, Action] = fold_action_events(storage.read_events(run_id))
+    folded: dict[str, Action] = fold_action_events(storage.read_all_events(run_id))
```

That makes `action_rows` agree with `event_rows` and `budget_rows` in the same module, with `pending_actions_with_keys`, and with `ActionLedger`. A regression test worth pinning is the shape above: an uncertain action whose events all pre-date the anchor must survive `compact_run` in the TUI's action list (the existing `budget_rows` archive-awareness test is the template).

Adhi1-2:
## Description

The TUI's Actions tab folded only the **live tail** of the event log, so after compaction it rendered **empty** for a run that was still blocked by an uncertain action — while the run row directly above it still reported the run as unsafe, and while `continuum actions`, the dashboard HITL buttons, and the recovery verdict all still listed that same action.

`src/continuum/tui/model.py:278` used:

```python
folded = fold_action_events(storage.read_events(run_id))   # live tail only
```

The two views beside it in the same module (`event_rows`, `budget_rows`) already read `read_all_events`. Compaction moves an unresolved action's `ACTION_RECORDED` / `ACTION_FAILED` events into `events_archive` **without settling the action**, so the live tail holds only the anchor and later events, and the fold saw nothing. An operator looking at the tab would conclude there is nothing to settle, exactly when there is one thing to settle.

The fix is the one-liner the issue proposes — fold the full log, matching what `ActionLedger._replay` itself does (it merges `read_archived_events` + `read_events` by sequence before folding), so the key the tab offers is the key that would actually settle the action:

```python
```

Read-only view model; no new writes, no change to any other tab or to the CLI/dashboard paths.

## Related issues

Fixes #1182

## Type of change

- [x] **Bugfix** — compaction hides the only uncertain action an operator could settle from the keyboard

## Breaking changes

No. A view model reads a wider (already-stored) event set; the rows it returns are a superset on compacted runs and unchanged on uncompacted ones.

## How has this been tested

**Environment:** macOS 15 (Darwin 25.0.0), Python 3, worktree on `upstream/main` at `cef019d` (the commit the issue reports), SQLite backend. Postgres tests need a live `continuum` database that is not present here.

```bash
# the regression test below, proven to fail without the fix (assert 0 == 1) and pass with it
python3 -m pytest tests/test_tui.py -q

# full non-slow suite (the Postgres backend test is deselected: environmental, no live DB;
# it fails identically on a pristine upstream/main checkout)
python3 -m pytest tests/ -m "not slow" \
  --deselect tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend -q

# quality gates
python3 -m ruff check
python3 -m ruff format --check
python3 -m mypy src/continuum/tui
```

Results: `tests/test_tui.py` 33 passed; full non-slow suite 2222 passed / 25 skipped / 9 deselected; ruff check and ruff format clean; mypy clean over `src/continuum/tui`.

The issue's own reproduction script was also replayed byte-for-byte: before the fix `tui action_rows` printed `[]` while `continuum actions` printed `unknown`; after the fix both show the `unknown` row with the key a reconciliation needs.

## Checklist

- [x] Tests added or updated — `test_action_rows_survive_compaction_when_the_action_stays_unresolved`, which claims an action, fails it into `UNKNOWN`, anchors after it, compacts, asserts the live tail no longer holds the action, then asserts the tab still lists it with the right key and agrees with the ledger.
- [x] Documentation updated — `CHANGELOG.md` Unreleased → Fixed.
- [x] CI passes locally (commands above).
- [x] Security limitations stated — none; this widens a read of already-stored events in a view model. No new write path, no new network call, no change to provenance or exactly-once handling. The archive is part of the run's own recorded history, which is what `read_all_events` exists to expose.

## Additional context

This is a focused standalone fix off `upstream/main`. The same underlying line was also corrected (via `ActionLedger(...).folded()`) in a larger in-flight branch; this PR carries just the issue's scoped change so it can land independently.


anya-research:

src/continuum/tui/model.py
```diff
@@ -271,11 +271,16 @@ def checkpoint_rows(storage: Storage, run_id: str) -> list[CheckpointRow]:
...
+    the dashboard HITL buttons and ``continuum actions`` fold, so the key
+    offered here is the key that would settle the action.
     """
     storage.get_run(run_id)
-    folded: dict[str, Action] = fold_action_events(storage.read_events(run_id))
+    # The archived prefix included, matching event_rows and budget_rows in this
+    # module: compaction moves an unresolved action into the archive without
+    # settling it, so the live tail alone would hide the one thing an operator
+    # could settle from the keyboard (#1182).
+    folded: dict[str, Action] = fold_action_events(storage.read_all_events(run_id))
```
Cyrax321, line 283: This is the whole fix and it is in the right place. `event_rows` and `budget_rows` above and below it already read `read_all_events`; this was the one view in the module that could go blind after compaction, on the one tab whose entire purpose is showing the action an operator can settle.

tests/test_tui.py
```diff
@@ -114,6 +114,41 @@ def test_action_rows_flag_uncertain_actions_with_their_ledger_key(
     assert rows[0].action_type == "send_invoice"
 
 
+def test_action_rows_survive_compaction_when_the_action_stays_unresolved(
```
Cyrax321, line 117: The ordering here is what makes the test trustworthy: asserting the live tail lost `ACTION_RECORDED` before asserting the tab still lists the row means an uncompacted run cannot satisfy it by accident. That is the difference between a regression test and a test that happens to pass.

Cyrax321 (dismissed):
Approving, and every claim the PR makes checks out. The sibling views already read the full history: `event_rows` and `budget_rows` both call `read_all_events`, so `action_rows` was the outlier in its own module. The dashboard does not have the bug, which matters because the new docstring promises parity with it: `hitl.py` folds `read_all_events` too. And the ledger own `_replay` merges `read_archived_events` with `read_events` by sequence, so `read_all_events` is exactly the history the key would be settled against, not an approximation of it.

The one-line fix is the right size for this bug, and the CHANGELOG entry names the failure mode an operator would have seen: the tab empty while the run row above it still says blocked, which is the specific kind of wrong that makes a person distrust the whole dashboard.

The test is doing the thing that makes it a real regression guard rather than a happy path: it asserts the live tail genuinely lost the action before asserting the tab still lists it. Ordering that way means the test cannot pass on an uncompacted run by accident, and it proves the fixture archived the events instead of assuming it. The cross-check against `ActionLedger.all()` at the end pins the agreement the issue asked for.

Adhi1-2:
Closing in favour of #1287, which fixes the same #1182 without the noise this branch picked up: the diff carries 13 `.worktrees/*` submodule-pointer entries and a stray `repro_replay_upto_compacted.py` scratch file that should never have been committed. The actual change (one line in `action_rows`, plus CHANGELOG and tests) is identical in intent to #1287's, and that branch is clean. Deleting this one so a reviewer sees a single tidy option.

PR #1194, opened 2026-09-16
fix(dashboard): offer a settle button for REQUIRES_REVIEW actions (#1183)

### LINKED ISSUE #1183: dashboard: pending_actions_with_keys omits REQUIRES_REVIEW, so escalated actions cannot be settled from the UI
## Summary

Three surfaces decide "this action's outcome is unresolved and a human may settle it," and they do not agree:

| Surface | Statuses treated as unresolved |
|:--|:--|
| `continuum actions` (`src/continuum/cli/main.py`, `cmd_actions`) | `UNKNOWN`, `STARTED`, **`REQUIRES_REVIEW`** |
| `continuum tui` (`src/continuum/tui/model.py:56-58`, `UNCERTAIN_STATUSES`) | `UNKNOWN`, `STARTED`, **`REQUIRES_REVIEW`** |
| dashboard HITL (`src/continuum/dashboard/hitl.py:133`, `pending_actions_with_keys`) | `UNKNOWN`, `STARTED` |

The TUI's constant is even documented as matching the other two:

```python
#: The statuses an operator may still settle from the keyboard, matching the
#: set `continuum actions` flags and `pending_actions_with_keys` offers.
UNCERTAIN_STATUSES = frozenset(
    {ActionStatus.UNKNOWN, ActionStatus.STARTED, ActionStatus.REQUIRES_REVIEW}
)
```

`ActionStatus.REQUIRES_REVIEW` is reachable in production via `ActionLedger.flag_for_review` (`src/continuum/actions/ledger.py:1278`), which is the escalation path — "escalate an action a human must judge." So an escalated action:

- is reported as unresolved by `continuum actions` (and drives that command to exit `REQUIRES_HUMAN`),
- is flagged `uncertain=True` in the TUI's Actions tab,
- is offered as settleable by `continuum reconcile` (`reconciliation.py:242` lists `REQUIRES_REVIEW`),
- but is **absent from the dashboard's pending list**, so the web UI renders no settle button for the one status that exists precisely because a human must judge it.

The console and the terminal both point at work; the dashboard says there is none.

## Affected component

`src/continuum/dashboard/hitl.py` (dashboard HITL settle surface); consistency with `continuum actions`, `continuum reconcile`, and `continuum tui`.

## Severity

Low–Medium. Nothing is lost — the action is still on the ledger and `continuum reconcile` still settles it — but the dashboard is the surface built around "press a button to settle," and for this status it silently offers no button.

## System information

- Python: 3.13.5

```python
from continuum.dashboard.hitl import pending_actions_with_keys
from continuum.models import ActionStatus, Origin, Run

st.create_run(Run(run_id="demo", goal="demo"))
st.append_event("demo", EventType.RUN_STARTED, {"goal": "demo"}, source=Origin.HUMAN)

led = ActionLedger(st, "demo")
outcome = led.claim("payout", arguments={"amount": 5}, origin_digest="a" * 64)
led.flag_for_review(outcome.key, "compliance hold: needs a human")
status = led.all()[0].status

print("action status                       :", status.value)
print("`continuum actions` flags unresolved:", status in {
    ActionStatus.UNKNOWN, ActionStatus.STARTED, ActionStatus.REQUIRES_REVIEW})
print("tui action_rows marks uncertain     :", tui.action_rows(st, "demo")[0].uncertain)
print("dashboard HITL offers a button      :", bool(pending_actions_with_keys(st, "demo")))
```

Output against `cef019d`:

```
action status                       : requires_review
`continuum actions` flags unresolved : True
tui action_rows marks uncertain      : True
dashboard HITL offers a button      : False
```

## What a fix looks like

The minimal fix is to make the dashboard's set match the other two:

```diff
-        if action.status in (ActionStatus.STARTED, ActionStatus.UNKNOWN):
+        if action.status in (ActionStatus.STARTED, ActionStatus.UNKNOWN, ActionStatus.REQUIRES_REVIEW):
```

The cleaner option is one shared constant — move `UNCERTAIN_STATUSES` out of `tui/model.py` (it is the settle set, not a TUI concept) and have `cmd_actions`, `pending_actions_with_keys`, and `reconciliation.py:242` all import it, so the four surfaces cannot drift apart again. Either way, a test asserting the three sets are equal would pin it.

Adhi1-2:
## Summary

Fixes #1183.

Three surfaces decide "this action's outcome is unresolved and a human may settle it," and the dashboard disagreed with the other two. An action escalated via `flag_for_review` — the status that exists *precisely* because a human must judge it — rendered no settle button on the one surface built around "press a button to settle":

```
action status                       : requires_review
`continuum actions` flags unresolved : True
tui action_rows marks uncertain      : True
dashboard HITL offers a button       : False   <- the bug
```

## What changed

Rather than patching the dashboard's literal in place, the four sites that mean the same thing now read one shared constant, `UNCERTAIN_STATUSES`, relocated to `continuum.actions.ledger` — the module that owns action-status semantics and the one every consumer already imports from:

| Site | Before | After |
|---|---|---|
| `dashboard/hitl.py` `pending_actions_with_keys` | `(STARTED, UNKNOWN)` ← the bug | `UNCERTAIN_STATUSES` |
| `cli/main.py` `cmd_actions` | inline triple | `UNCERTAIN_STATUSES` |
| `recovery/engine.py` `all_uncertain` | inline triple | `UNCERTAIN_STATUSES` |
| `tui/model.py` | defined the constant locally | imports it from the ledger |

The set itself is unchanged — it was already correct in three of the four places; only the dashboard had drifted.

## One deliberate non-change

The issue suggested folding `reconciliation.unresolved_actions` into the same constant. I did not, because it is a genuinely different predicate, not a drift:

```python
def unresolved_actions(ledger) -> tuple[Action, ...]:
    """Actions a human must judge before the run can safely continue."""
    return tuple(a for a in ledger.all()
                 if a.status in (ActionStatus.UNKNOWN, ActionStatus.REQUIRES_REVIEW))
```

It excludes `STARTED` on purpose — an in-flight action may still settle itself when the run resumes, so it is the recovering *agent's* to reconcile, not yet a human's to judge. `ActionLedger.pending()` is narrower still for the same reason. Both now carry a docstring saying so, so the next person to "unify" them sees why the two sets differ. Unifying them would have been a behaviour change dressed as a refactor.

## Tests

- `tests/test_dashboard_hitl.py::test_requires_review_action_is_offered_a_button` — the escalated action now appears in `pending_actions_with_keys`.
- `tests/test_action_ledger.py::test_every_uncertain_surface_agrees_on_one_settle_set` — drives all four surfaces from one escalated action and asserts they agree; this is the drift guard the issue asked for.

Both were verified to fail without the fix (the dashboard assertion on the old `(STARTED, UNKNOWN)` predicate; the ledger test on the missing import) and pass with it.

## Verification

- `ruff format --check` — clean
- `mypy src/continuum` — clean (127 files)
- `pytest` — **2232 passed, 25 skipped** (skips are the optional-service suite, unchanged)

## Summary by CodeRabbit

- **Improvements**
  - Standardized how actions with uncertain outcomes are identified across the command interface, dashboard, recovery flow, and terminal interface.
  - Actions requiring review are now consistently surfaced for human attention, while started actions remain eligible to settle when a run resumes.
  - Clarified which unresolved actions require human review versus continued processing.
  - Added coverage to verify consistent behavior across interfaces.


src/continuum/dashboard/hitl.py
```diff
@@ -125,11 +125,11 @@ def reconcile_action(
...
 def pending_actions_with_keys(storage: Storage, run_id: str) -> list[tuple[str, Any]]:
     """Uncertain actions paired with their full ledger key (for buttons)."""
-    from continuum.actions.ledger import fold_action_events
+    from continuum.actions.ledger import UNCERTAIN_STATUSES, fold_action_events
 
     folded = fold_action_events(storage.read_all_events(run_id))
     out: list[tuple[str, Any]] = []
     for key, action in folded.items():
-        if action.status in (ActionStatus.STARTED, ActionStatus.UNKNOWN):
+        if action.status in UNCERTAIN_STATUSES:
```
Cyrax321, line 133: This is the fix for #1183 in one line, and it is the line most worth keeping stable. The old literal matched `STARTED` and `UNKNOWN` but not `REQUIRES_REVIEW`, so an escalated action rendered no button on the one surface built around pressing one. Reading the constant instead of a literal is what stops the fifth surface from drifting the same way.

tests/test_action_ledger.py
```diff
@@ -1107,3 +1107,39 @@ def test_reconciling_an_unknown_outcome_as_occurred_needs_no_prior_receipt(
     assert settled is not None
     assert settled.status is ActionStatus.COMPLETED
     assert settled.external_id == "found-it"
+
+
+# --- the one settle set (issue #1183) ------------------------------------- #
+
+
+def test_every_uncertain_surface_agrees_on_one_settle_set(ledger: ActionLedger) -> None:
```
Cyrax321, line 1115: This test is the durable part of the fix. Driving all four surfaces from one escalated action and asserting they agree turns the original bug from a one-time fix into a property the suite holds; the next surface to hand-roll a status list fails here instead of shipping a fifth answer.

src/continuum/actions/ledger.py
```diff
@@ -83,6 +83,25 @@
...
+#:
+#: ``STARTED`` is in because an interrupted run left the effect in flight and
+#: the record never advanced past intent; ``UNKNOWN`` because the ledger
+#: refused to guess; ``REQUIRES_REVIEW`` because :meth:`ActionLedger.flag_for_review`
+#: escalated exactly when a human must judge it. Note this is deliberately a
+#: wider predicate than :meth:`ActionLedger.pending` and
+#: :func:`~continuum.actions.reconciliation.unresolved_actions`, which exclude
+#: ``STARTED``: an in-flight action may still settle itself, so it is the
+#: recovering *agent's* to reconcile, not yet a human's to judge.
+UNCERTAIN_STATUSES = frozenset(
```
Cyrax321, line 101: The distinction worth preserving is visible here: the wide set says an operator may settle, the narrow predicates say a human must. Including `STARTED` in the button set is pre-existing behaviour this PR did not change, and the comment at the constant names the difference, so a future narrowing is a deliberate decision rather than an accidental one.

Cyrax321 (dismissed):
Approving. The bug is real and the fix is the right shape: the dashboard read a literal `(STARTED, UNKNOWN)` and so missed `REQUIRES_REVIEW`, the one status that exists precisely because a human must judge it, while three other surfaces already had the right set. Unifying on one constant in the module that owns action-status semantics fixes the drift and stops the fourth surface from diverging again, and relocating it to `continuum.actions.ledger` is the correct home given every consumer already imports from there.

The most valuable line in this PR is the refusal to unify `reconciliation.unresolved_actions`. That predicate excludes `STARTED` on purpose, because an in-flight action may still settle itself when the run resumes and is therefore the recovering agent responsibility, not a human obligation. Folding it in would have been a behaviour change dressed as a refactor, and the docstrings now say so where the next person to unify them will read it.

On the red Windows check: it is unrelated to this change. The failing test is `tests/test_dashboard.py::test_dashboard_post_body_too_large_returns_413`, which raises `ConnectionAbortedError: [WinError 10053]`. That is Windows-specific socket behaviour when a server answers 413 and closes before draining the request body; on Linux the same code passes. The PR touches `dashboard/hitl.py` and adds tests to `test_action_ledger.py` and `test_dashboard_hitl.py`; it never touches the 413 path or `test_dashboard.py`. Windows is also not in the `main` branch protection set, which requires only Lint and Type-check plus the three Ubuntu Test jobs, all green here. So this is a pre-existing platform flake, not something to block on, though it is worth a separate issue since it makes the Windows job unreliable as a signal.

One thing to keep an eye on, not to change: `UNCERTAIN_STATUSES` includes `STARTED`, and the dashboard settle-button list reads it, so an in-flight action is offered a settle button even though the narrow predicates say a human is not yet obliged to judge it. That is consistent, not contradictory, because "may settle from a surface" and "must judge before the run continues" are different questions, and the button set was already this wide before this PR; this PR only adds the missing `REQUIRES_REVIEW`. The comment at the constant states the distinction, which is what keeps it readable.

PR #1195, opened 2026-09-16
fix(cli): a missing --attest/--key/--payload-file reports an error line, not a traceback

### LINKED ISSUE #1143: cli: a missing --attest, --key or --payload-file file exits with a raw traceback instead of the CLI's error: line
### Summary

Four CLI commands read a caller-supplied file and let `FileNotFoundError` escape as a raw traceback, instead of the single-line `error: ...` + nonzero exit the CLI uses for every other operator mistake.

### Affected component

CLI

### Severity

### Regression check

### System information

- OS: Windows 11 Home 22H2
- Python: 3.11.9

```bash
continuum init --db t.db
continuum start run_1 --goal "g" --db t.db
continuum checkpoint run_1 --db t.db

# attestation document the operator named from memory, or a typo
continuum attest-verify run_1 --attest missing.json --db t.db

# signing key path typed wrong
continuum attest run_1 --key signer.pem --db t.db   # when the file is elsewhere

# hook payload the harness promised to write
continuum gate --payload-file missing.json --db t.db
continuum observe --payload-file missing.json --db t.db
```

### Current behaviour

A multi-frame Python traceback is printed and the process exits 1 from an uncaught exception:

```
  File ".../continuum/cli/main.py", line 4434, in main
    return int(args.func(args, storage, out, err))
  File ".../continuum/cli/main.py", line 3646, in cmd_attest_verify
    doc = json.loads(Path(args.attest).read_text(encoding="utf-8"))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../pathlib.py", line 1058, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../pathlib.py", line 1044, in open
    return io.open(mode, buffering, encoding, errors=errors)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'missing.json'
```

The same shape comes out of all four sites:

- `attest-verify` — `main.py:3646`, `Path(args.attest).read_text(...)`
- `attest` — `main.py:3614`, `Path(key_path).read_text(...)`
- `gate` — `main.py:2892`, `Path(args.payload_file).read_text(...)`
- `observe` — `main.py:2281`, `Path(args.payload_file).read_text(...)`

### Expected behaviour

`continuum` already converts operator mistakes to a one-line message and a clean exit code, and does so deliberately. `main()` catches `RunNotFound`/`CheckpointNotFound` (exit 2, `NOT_FOUND`), `CorruptedRecord` (exit 3), `CheckpointError`, and `(ProjectionError, StorageError, ValueError)` (exit 1), and the storage-open block catches `sqlite3.Error` with a comment explaining exactly why: *"an unreadable or unwritable database is an ordinary operator mistake (bad path, no permission). A traceback would bury the useful part."*

A missing `--attest`, `--key` or `--payload-file` file is the same class of mistake, so it should read the same way:

```
error: no such file: 'missing.json'
```

and exit 1 (or 2, matching `NOT_FOUND`).

Two further reasons this one is worth fixing rather than leaving as a raw traceback:

- `attest`/`attest-verify` are the commands an operator runs to investigate a suspected integrity problem. A traceback on a typo'd path looks like the tool itself is broken at the moment it is most needed.
- `gate` and `observe` are wired into host hooks (`continuum hooks install`). A hook whose payload file is late or missing gets a traceback instead of a diagnostic, and the traceback is the whole signal a hook script has to decide what to do next.

### Logs or CI run

Confirmed by direct invocation on commit `2e835ad`; the four call sites are listed above. The CLI's own convention is visible at `main.py:4435-4446`.

abyyxhek:
## Description

Four CLI commands read a file the caller names on the command line — `attest-verify --attest`, `attest --key`, and `gate`/`observe --payload-file`. When that file did not exist, `FileNotFoundError` was not in the tuple of exceptions `main()` catches around `args.func(...)`, so it escaped as a multi-frame traceback ending in `FileNotFoundError: <path>` and an exit code of 1.

Every other operator mistake in this CLI (a missing run, a missing checkpoint, an unreadable database) is already rendered as a single `error: ...` line with a nonzero exit code. A mistyped input file is the same class of mistake, but it got the noisiest possible presentation, burying the one piece of information the operator actually needed: the path they typed.

`main()` now catches `FileNotFoundError` alongside those other operator-mistake exceptions and prints:

```
error: file not found: <path>
```

with `ExitCode.ERROR`. `exc.filename` is used so the path is reported verbatim rather than repr-escaped — the same reasoning already documented on the `sqlite3.Error` handler directly above this one (issue #94's class of bug, where repr() doubled every backslash on Windows).

## Related issues

Fixes #1143

## Types of changes

- Type: `bugfix`

## Breaking changes

No. Previously the process exited 1 with a traceback; now it exits with `ExitCode.ERROR` and a single error line. Both are nonzero failures, so no script relying on failure detection breaks, and the error message is now strictly more readable.

## How has this been tested?

```
pytest tests/test_cli_missing_input_file.py --no-cov          # 5 passed
pytest tests/test_cli.py --no-cov                             # 110 passed
pytest tests/test_docs_counts.py --no-cov                     # 3 passed
uvx ruff@0.16.4 check src/continuum/cli/main.py tests/test_cli_missing_input_file.py   # All checks passed
uvx ruff@0.16.4 format --check src/continuum/cli/main.py tests/test_cli_missing_input_file.py
mypy src/continuum/cli/main.py                                # Success: no issues found
uv build --wheel --no-build-isolation                         # Successfully built
```

The new test file was verified to be a genuine regression test: with the fix stashed, all 5 cases fail with an uncaught `FileNotFoundError`; with the fix in place, all 5 pass. Each of the four affected command/flag combinations is covered, plus one case asserting a path containing a space is reported verbatim rather than escaped.

## Checklist

## Additional context

The handler is intentionally scoped to the `args.func(...)` call, which is command dispatch — a `FileNotFoundError` raised there can only come from reading a file the operator named, so catching it cannot mask an internal failure. Placing it before the `RunNotFound`/`CheckpointNotFound` arms is safe: those are distinct types and are matched exactly.

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI commands now show a clear `error: file not found` message when a specified input file is missing.
  * Missing files no longer produce a traceback or unexpected standard output.
  * Error handling works consistently for attestation, key, and payload file inputs, including paths containing spaces.

* **Tests**
  * Added coverage for missing-file scenarios across supported command variants.


src/continuum/cli/main.py
```diff
@@ -4407,6 +4407,13 @@ def main(
 
     try:
         return int(args.func(args, storage, out, err))
+    except FileNotFoundError as exc:
```
Cyrax321, line 4410: The catch is broader than the four documented flags, and that is a strength worth claiming. `--webhooks-config`, the four `--config` flags (budgets, gateway, gate, reconcilers) and `--file` also read caller-named files, and this single arm covers all of them for free. The tests sample four cases; the fix is more complete than the test list suggests.

tests/test_cli_missing_input_file.py
```diff
@@ -0,0 +1,73 @@
...
+    # The one-line contract every other operator mistake already follows.
+    assert err.strip().startswith("error:")
+    # The path is named, which is what the reader was looking for.
+    assert str(missing) in err
+    # No Python traceback: a missing file is not an internal failure.
+    assert "Traceback (most recent call last)" not in err
+    assert out == ""
+
+
+def test_missing_file_error_quotes_the_exact_path(db: str, tmp_path: pathlib.Path) -> None:
```
Cyrax321, line 67: `exc.filename or exc` is the right choice and the second test pins the reason. A plain `str(exc)` or repr would have escaped the path on Windows exactly the way #94 described for the database handler directly above, so inheriting that convention instead of reintroducing the bug is the correct call.

Cyrax321 (approved):
Approving. This is the right fix for the reported problem and it is smaller than the benefit it delivers. `FileNotFoundError` subclasses `OSError` and none of the existing arms, so it previously escaped the whole chain as a traceback; placing the new arm first cannot shadow anything, since `RunNotFound`, `CheckpointNotFound`, `CorruptedRecord` and `CheckpointError` are exact-type matches of unrelated types. Using `exc.filename` rather than the exception repr inherits the #94 reasoning about Windows backslash doubling, and the verbatim-path test pins it.

One correction to the justification, and one thing I would decide before merge. Neither blocks the fix.

The scope comment says a `FileNotFoundError` raised in `args.func` "can only come from reading a file the operator named". That is stronger than the code guarantees. Commands read internal files too: rewind and restore pull file snapshots, and several commands read their own state. The printed message is neutral enough that it does not mislead in those cases, but the comment invites a future reader to believe the catch is narrower than it is. Worth either narrowing the catch to the four flags or softening the comment to say the message is accurate for operator-named files and merely terse for internal ones.

The exit code. `ExitCode.NOT_FOUND` is documented as "no such run, version or checkpoint" and `CheckpointError` already routes there, so a missing input file is the same family; `ERROR` (1) works, and no script breaks either way, but `NOT_FOUND` (2) would keep the not-found class consistent across the CLI. Pick whichever you prefer; if you take NOT_FOUND the parametrized test needs its expected code moved with it.

PR #1196, opened 2026-09-16
fix(checkpoint): rewind on a compacted run restores instead of failing or deleting (#1053)

### LINKED ISSUE #1053: continuum rewind fails on every compacted run, and would delete instead of restore once fixed
## Summary

`continuum rewind` cannot run on any compacted run, and once that is fixed it destroys files instead of restoring them. Both bugs are the same live-tail read in `src/continuum/checkpoint/rewind.py`.

The projection at `rewind.py:125` reads `storage.read_events(run_id)`, the live tail only. Compaction archives `RUN_STARTED` into `events_archive`, so the projection concludes the run never started and raises `ProjectionError: run 'run_1' has no goal`. Every rewind on a compacted run dies there before it does anything, including `--dry-run`, which is the read-only command an operator runs to check whether a rewind is safe. `CheckpointManager.project_current` already fixed this exact class of bug in `src/continuum/checkpoint/manager.py:165-172` by folding `read_all_events`; rewind did not get the same treatment.

`_collect_tool_completed` at `rewind.py:83` reads the live tail too. It populates the before/after file sets, and the restore-or-delete decision keys off `before_digest is None`. After compaction the pre-checkpoint `TOOL_COMPLETED` event is archived, the before-set is empty, and a file that should be restored from its checkpoint-time snapshot takes the delete branch instead. With the projection bug patched out of the way, a dry run over a compacted run reports `deleted_files: ('.../out.txt',)` where the correct answer is `reverted_files: ('.../out.txt',)`. That is a data-loss path, not merely a failure to run.

## Affected component

continuum.checkpoint.rewind, continuum.cli (`continuum rewind`)

## Severity

Critical, no workaround. The first bug makes the command unusable on exactly the runs most likely to need a rewind, and the second, once reachable, deletes workspace content that the command exists to restore. There is no way to rewind a compacted run today.

## Evidence

At HEAD `14f8a84`, real CLI, run seeded with a checkpoint then compacted:

```
$ continuum --db a5d.db compact run_1 --force
Archived 8 event(s); the live log now starts at the anchor.
$ continuum --db a5d.db rewind run_1 --to 0 --dry-run
error: rewind failed: run 'run_1' has no goal: the log never recorded RUN_STARTED
$ echo $?
1
```

The identical rewind succeeds before compaction. Event distribution after compaction: `live RUN_STARTED: 0`, `archived RUN_STARTED: 1`. For the second bug, with only the projection corrected to fold the archive, the dry run over a compacted run with a pre-checkpoint file write reports `reverted_files: ()` and `deleted_files: ('/private/tmp/a5c/ws/out.txt',)`, with no conflict or unrecoverable warning, so nothing would stop it.

```python
import os, tempfile

d = tempfile.mkdtemp()
db = os.path.join(d, "r.db")
store = SQLiteStorage(db)
for i in range(3):
cp = CheckpointManager(store).checkpoint("run_1", reason="cp1")
for i in range(3, 6):
store.compact_run("run_1")                       # archives RUN_STARTED

from continuum.checkpoint.rewind import rewind_to_checkpoint, RewindError
try:
    rewind_to_checkpoint(store, "run_1", str(cp.version), dry_run=True)
except RewindError as exc:
    print("rewind failed:", exc)
    # rewind failed: run 'run_1' has no goal: the log never recorded RUN_STARTED
```

Or with the installed CLI: `continuum compact run_1 --force` then `continuum rewind run_1 --to 0 --dry-run`.

## Root cause

`a3b065b` (2026-08-29, "feat(rewind): atomic dual-state rewind (Refs #292)") wrote both the projection and the tool-event collection against the live log. Compaction landed in `65bb275` (2026-08-23), predating rewind, and #648 later established the archive-aware projection convention in `project_current` without updating rewind. No test covers rewind after compaction; the suite's rewind tests all run against uncompacted logs.

## Expected behaviour

`continuum rewind` should work on compacted runs, restoring pre-checkpoint file content from the snapshots recorded at checkpoint time and deleting only files created after the checkpoint.

## Suggested fix shape

Change the projection at `rewind.py:125` to fold `read_all_events(run_id)` (matching `manager.py:172`) and `_collect_tool_completed` at `rewind.py:83` to do the same. The before/after split by sequence needs no other change, since archived events still carry their original sequence numbers. Add a regression test that compacts first, then rewinds with `--dry-run` and asserts the pre-checkpoint file lands in `reverted_files` rather than `deleted_files`.

## System information

- Python: 3.14.5

abyyxhek:
## Description

`continuum rewind` could not run on any compacted run, and once unblocked it destroyed workspace content instead of restoring it. Both bugs were the same live-tail read in `src/continuum/checkpoint/rewind.py`.

- The projection at `rewind.py:125` read `storage.read_events(run_id)` — the live tail only. Compaction archives `RUN_STARTED` into `events_archive`, so the projection concluded the run never started and raised `ProjectionError: run '...' has no goal`. Every rewind on a compacted run died there before doing anything, **including `--dry-run`**, the read-only command an operator runs to check whether a rewind is safe.
- `_collect_tool_completed` at `rewind.py:83` read the live tail too. It builds the before/after file sets that drive the restore-or-delete decision. After compaction the pre-checkpoint `TOOL_COMPLETED` is archived, the before-set comes back empty, and a file that should be restored from its checkpoint-time snapshot takes the **delete** branch instead — a data-loss path with no conflict or unrecoverable warning that would stop it.

Both reads now fold `read_all_events(run_id)`, matching the convention `CheckpointManager.project_current` already established for exactly this class of bug (issue #648, `manager.py`). The before/after split by sequence needs no other change: archived events keep their original sequence numbers, so the `> checkpoint_seq` / `<= checkpoint_seq` partition is unchanged.

## Related issues

Fixes #1053

## Types of changes

- Type: `bugfix`

## Breaking changes

No. Previously the command failed outright on compacted runs; now it performs the rewind it was asked for. On uncompacted runs behaviour is unchanged (covered by a regression test in the new file).

## How has this been tested?

Windows 11, Python 3.11.15, the repo `.venv`, SQLite backend.

```
pytest tests/test_rewind_compacted.py --no-cov                        # 4 passed
pytest tests/test_rewind_dual_state.py tests/test_rewind_carry_forward.py \
       tests/test_rewind_compacted.py tests/test_checkpoint_manager.py \
       tests/test_compaction.py tests/test_compact_twice.py \
       tests/test_cli_precompact.py --no-cov                          # 82 passed
pytest tests/test_cli.py tests/test_docs_counts.py --no-cov           # 113 passed
uvx ruff@0.16.4 check src/continuum/checkpoint/rewind.py tests/test_rewind_compacted.py
uvx ruff@0.16.4 format --check src/continuum/checkpoint/rewind.py tests/test_rewind_compacted.py
mypy src/continuum/checkpoint/rewind.py                              # Success: no issues found
uv build --wheel --no-build-isolation                                # Successfully built
```

The new tests were verified to be genuine regression tests: with the fix stashed, 3 of the 4 fail (the projection error), and the fourth — an uncompacted-run guard asserting the archive-aware reads did not change existing behaviour — passes either way, as intended. The two failure modes were also reproduced individually against the CLI and the library before the fix: `error: rewind failed: run '...' has no goal` with exit 1, and a dry run reporting `reverted_files: ()` where the correct answer restores the pre-checkpoint content.

## Checklist

## Additional context

The data-loss consequence is order-sensitive in a way worth stating plainly. The CLI's `compact` always anchors at the current head, so it archives the post-checkpoint writes as well as the pre-checkpoint ones — there the live tail has *no* `TOOL_COMPLETED` events at all and rewind silently reported success having restored nothing (and deleted nothing). The literal delete-instead-of-restore path needs a partial compaction that leaves the post-checkpoint write live, reachable via the library's `through_sequence`; the new test builds exactly that fixture and asserts the file lands in `reverted_files`. Both shapes share the one root cause and the one fix.


src/continuum/checkpoint/rewind.py
```diff
@@ -122,7 +126,7 @@ def rewind_to_checkpoint(
         raise
     except Exception as exc:
         raise RewindError(str(exc)) from exc
-    _ = project(run_id, storage.read_events(run_id), upto=target.state.source_sequence)
+    _ = project(run_id, storage.read_all_events(run_id), upto=target.state.source_sequence)
```
Cyrax321, line 129: This is the second `read_all_events` in the same function, one statement after the first, and `storage/base.py` explicitly warns callers folding the same history more than once to reuse the returned sequence instead of rescanning the archive. Reading the history once and passing it into `_collect_tool_completed` would halve the archive scans on the exact long compacted runs this fix is for. Not blocking, but it is the one place this PR makes an existing O(n) cost twice as expensive.

src/continuum/checkpoint/rewind.py
```diff
@@ -80,7 +80,11 @@ def resolve_checkpoint(storage: Storage, run_id: str, to: str) -> StateCheckpoin
 
 
 def _collect_tool_completed(storage: Storage, run_id: str) -> list[Any]:
-    return [e for e in storage.read_events(run_id) if e.type is EventType.TOOL_COMPLETED]
+    # Full history, not the live tail: after compaction the pre-checkpoint
+    # TOOL_COMPLETED events live in the archive. Reading the tail alone leaves
+    # the before-set empty, so a file that should be restored from its
+    # checkpoint-time snapshot takes the delete branch instead (issue #1053).
+    return [e for e in storage.read_all_events(run_id) if e.type is EventType.TOOL_COMPLETED]
```
Cyrax321, line 87: Folding the whole history here is the fix, and it is worth a line on why the result is discarded: this is a gate, checking that the checkpoint prefix projects cleanly before anything is discarded. Reading only the tail meant a compacted run failed this check on a missing `RUN_STARTED` and never reached the restore logic at all, `--dry-run` included.

tests/test_rewind_compacted.py
```diff
@@ -0,0 +1,153 @@
...
+        f"unexpected conflicts: {result.conflicts} unrecoverable: {result.unrecoverable}"
+    )
+    assert str(kept) in result.reverted_files
+    # The pre-checkpoint content is what a real rewind would restore.
+    assert str(kept) not in result.deleted_files
+    # A file created after the checkpoint is still deleted, archive or not.
+    assert str(created) in result.deleted_files
+
+
+def test_partial_compaction_never_takes_the_delete_branch(tmp_path: Path) -> None:
```
Cyrax321, line 101: This is the test that makes the PR trustworthy. The literal delete-instead-of-restore path needs a partial compaction, which the CLI never produces, so it is only reachable through the library `through_sequence` parameter; asserting that both a live and an archived `TOOL_COMPLETED` exist confirms the fixture actually straddles the boundary instead of passing vacuously.

Cyrax321 (approved):
Approving. This is a data-loss fix and the root cause is correctly identified: both reads were live-tail reads, and compaction archives exactly the prefix they needed. The convention is documented in `storage/base.py:135`, which names full-history callers and says compaction moves facts without revoking their consequences, and `CheckpointManager.project_current` already folds `read_all_events` for this same class of bug (#648), so rewind was the outlier rather than a new pattern.

Ordering is the property I most wanted to check, and it holds. `read_all_events` merges the archived and live streams by sequence with a stable merge, so the projection folds a correctly ordered log and the `> checkpoint_seq` / `<= checkpoint_seq` partition in the tool-event split is unchanged, as the body claims. Archived events keep their original sequence numbers, which is what makes the whole fix a two-line change.

The one thing I would change before or soon after merge is a double scan, noted on the lines below. It is a performance issue, not a correctness one, and it does not block this fix: the data-loss path is closed either way, and the tests are genuine regression tests. I verified the partial-compaction fixture is the real shape, since `compact_run` accepts `through_sequence` and the fixture asserts the tool events straddle the archive boundary rather than assuming it.

Issue #1171, opened 2026-09-16
docs: README says "36 event types" and "44 event types in v0.2" — the EventType enum has 51


Adhi1-2:
## Summary

`#1100` covers the "29 event types" claim across `references/` and the docs site. The README itself carries **two more** numbers, both wrong, and they disagree with each other in the same file:

| Location | Claim | Actual |
|:--|:--|:--|
| `README.md:199` (Features — "Tamper-evident log") | "Hash-chained event log (36 event types)" | 51 |
| `README.md:418` (Storage architecture) | "Hash chained append only log (44 event types in v0.2)" | 51 |

The second is wrong twice over: the count is 51, and the released package is `continuum-agent` 0.1.2, so "in v0.2" cites a specific count for a version that has not shipped.

The adjacent claims in those tables are correct — "Seven recovery modes" matches `RecoveryMode` (7 members) and "Twelve tools" matches the 12 `@server.tool` registrations — which leaves the event count as the isolated outlier, exactly as in `#1100`. Nothing guards it: `tests/test_docs_counts.py` pins test-count figures only.

## Affected component

Documentation — `README.md`. Also lint-relevant: `README.md` is one of the four files the markdownlint gate enforces.

## Severity

Minor / cosmetic.

## System information

- OS: macOS 26.0.1 (Darwin 25.0.0)
- Python: 3.13.5
- continuum-agent: from source, commit `d10ee63` on `upstream/main`
- Installed via: `uv pip install -e ".[dev]"`

## Command or code to reproduce

```bash
python -c "from continuum.events import EventType; print(len(EventType))"   # 51
git grep -nE "(36|44) event types" README.md
# README.md:199: | Tamper-evident log | Hash-chained event log (36 event types) with integrity verification |
# README.md:418: | `events` | Hash chained append only log (44 event types in v0.2) |
```

## What a fix looks like

Set both rows to 51, or describe the vocabulary without a hard count. The guard test proposed in `#1100` — asserting the documented count equals `len(EventType)` — should extend to `README.md` so the README and the references cannot drift apart again. `#1068` already tracks the READMEs stale module and test-file counts (124/161 stated; the tree now has 127 modules and 170 test files), so a README-wide count resync is the cleaner unit of work.

Issue #1100, opened 2026-09-15
docs and references claim 29 event types; the EventType enum has 51 (seven sites, two asserting a complete list)


Adhi1-2:
## Summary

"29 event types" is stated in seven places across `references/` and the docs site, but the `EventType` StrEnum (`src/continuum/events.py:47`) has **51** members. Two of the sites assert completeness while omitting 22 event types, including `RUN_FORKED`, `RUN_RESTORED`, `RUN_MERGED`, `CONSTRAINT_PINNED`, `CONSTRAINT_RETRACTED`, `REVIEW_CONFIRMED`, `GRANT_DENIED`, `LIVENESS_SILENCE_DETECTED`, `LIVENESS_RECOVERED`, `RISK_OBSERVED`, `AUTHORITY_CONSUMED`, `AUTHORITY_RECONCILED`, `ATTEMPT_LESSON`, `TRAJECTORY_REPORT`, `PLAN_UPSERT`, `MEMORY_TOMBSTONED`, `NOTIFICATION_SENT`, and `NOTIFICATION_FAILED`.

The sites:

| Location | Claim |
|:--|:--|
| `references/architecture.md:39` | "29 event types covering the full lifecycle" |
| `references/architecture-spec.md:106` | "29 event types span the full lifecycle. **The complete set:**" + a 29-name list |
| `references/architecture-data.md:115` | "29 event types (`EventType` StrEnum, `events.py:46`). Complete list:" + a 29-name list |
| `references/architecture-drawing-prompt.md:72` | "hash-chained, 29 event types" |
| `references/architecture-drawing-prompt.md:131` | "The 29 event types" |
| `docs/index.html:325`, `:334`, `:337` | "29 event types" (rendered on the landing page) |

And a third number in the same doc set: `docs/assets/architecture.svg:88` says "**32** event types".

The adjacent counts in those same sentences are correct — "7 recovery modes" matches `RecoveryMode` (7 members, `models.py:122`) and "6 checkpoint policies" matches the six `CheckpointPolicy` subclasses in `checkpoint/policy.py` — which makes the event count the isolated outlier. Nothing guards it: `tests/test_docs_counts.py` pins only test-count figures.

## Affected component

Documentation

## Severity

Minor inconvenience or cosmetic issue.

## System information

- OS: macOS 26.0.1 (Darwin 25.0.0)
- Python: 3.13.5
- continuum-agent: from source, commit `579376d` on `upstream/main`
[...]

PR #1198, opened 2026-09-16
docs(readme): correct the event-type count and guard it against the enum (#1171)

Cyrax321:
## What

README stated the event vocabulary size twice, and both figures were wrong and mutually inconsistent inside one file: 36 in the feature table and 44 in the storage table, against an `EventType` enum of 51 members. Both rows now say 51, and a guard asserts the README figure against the enum so it cannot drift again.

The storage row was wrong twice over. Beyond the count, it cited a figure "in v0.2" for a version that has not shipped: the released package is `continuum-agent` 0.1.2, and "v0.2" appears nowhere else in the README. The version qualifier is dropped; the row describes the log as it exists today.

## Why nothing caught it

`tests/test_docs_counts.py` pinned the collected-test totals only. An event-type count is a different kind of claim: it is verifiable against the enum in one line, so the new test asserts against `len(EventType)` rather than pinning another number that would rot on its own schedule. It fails on the pre-fix figures, and it will fail the next time the vocabulary grows and the README is not updated.

## Scope

Deliberately README-only. `references/` and `docs/index.html` still carry an older "29 event types" figure, and that resync is #1100's work; extending the assertion there now would fail on figures this issue does not own. The module/test-file counts (124/161) are likewise left alone for #1068 and its open PR #1138.

One adjacent stale figure was left in place deliberately: `docs/assets/architecture.svg` renders "32 event types". It is a generated diagram rather than prose, so regenerating it belongs with whoever owns that asset, not with a README text fix.

## Summary by CodeRabbit

- **Documentation**
  - Updated the README to reflect the current total of 51 supported event types.

- **Tests**
  - Added validation ensuring documented event-type counts stay synchronized with the available event types.


README.md
```diff
@@ -196,7 +196,7 @@ Full walkthrough with code is in `docs/recovery_walkthrough.md` (`examples/recov
 | Framework adapters | Generic Python, OpenAI Agents SDK, LangGraph, and LangChain integrations |
 | Secure planning loop | Two-signal observation verification escalates high-risk branches to REQUIRES_REVIEW |
 | Periodic revalidation | Environment re-checked on a schedule, catching mid-run drift within one cycle |
-| Tamper-evident log | Hash-chained event log (36 event types) with integrity verification |
+| Tamper-evident log | Hash-chained event log (51 event types) with integrity verification |
```
Cyrax321, line 199: Verified: this is 51, and the count is consistent with the enum. The row reads correctly, and pairing it with the storage table below keeps one file from contradicting itself, which is what the issue reported.

tests/test_docs_counts.py
```diff
@@ -64,6 +68,29 @@ def test_documented_counts_agree() -> None:
...
+    only. The count is a claim about an enum a reader can verify in one line, so
+    it is asserted against the enum rather than against a number.
+
+    Scoped to README because that is where this issue's figures live. The same
+    claim in ``references/`` and ``docs/index.html`` is still stale and belongs
+    to #1100; asserting there now would fail on figures that issue owns.
+    """
+    from continuum.events import EventType
+
+    text = COUNTED_FILES[0].read_text(encoding="utf-8")
```
Cyrax321, line 86: Positional access into `COUNTED_FILES` rather than naming the file. The test is correct today, but if the list order ever changes it silently asserts the wrong document while still passing on an untouched README. `ROOT / "README.md"` says what it means; #1214 does this and it is worth taking.

README.md
```diff
@@ -415,7 +415,7 @@ Schema v6. SQLite is primary, Postgres is CI verified. One log, many projections
 
 | Table | Purpose |
 |:--|:--|
-| `events` | Hash chained append only log (44 event types in v0.2) |
+| `events` | Hash chained append only log (51 event types) |
```
Cyrax321, line 418: Removing the version qualifier is the more important half of this row. A count attached to a version that has not shipped cannot be checked against anything, which is how 44 survived here while the feature table said 36 in the same file. Describing the log as it exists today is what makes the guard possible.

Cyrax321 (commented):
The counts are correct and the fix is the right kind. I verified `len(EventType)` by counting the member assignments in `events.py`: exactly 51, the 52nd pattern match being the module-level `CAUSED_BY_TYPES` tuple and not an enum member. Both README rows now say 51, and dropping the "in v0.2" qualifier is right, since v0.2 has not shipped and the string appears nowhere else in the README. The guard is designed to fail forward: it reads the live enum rather than pinning another number, so the next vocabulary growth breaks it on purpose instead of rotting quietly.

The thing that needs resolving is duplication. #1214 is a twin of this PR: same issue, same two files, a near-identical guard, by the same author. One of the two has to close, and the two guards differ in ways worth picking deliberately. This one handles comma-grouped counts in its regex where #1214 does not, and its docstring records the scoping decision (references/ and index.html belong to #1100, the svg is a generated asset). #1214 reads `ROOT / "README.md"` explicitly where this one reaches for `COUNTED_FILES[0]` positionally, and its failure message tells the reader the exact command that prints the true count, which is what an operator hitting a red CI actually needs. Fold the explicit path and the remediation message into whichever one survives.

The broader count story also needs a single owner. #1210 fixes the same 29-to-51 figure in architecture-data, and #1204 rewrites that row too as part of a larger bundle, so `references/` and `docs/index.html` staying at 29 is tracked but not yet resolved anywhere that is merged.

soorajsuru2004-sudo:
This PR's README fix is already on `main` (both rows read 51), and the broader resync it deferred to #1100 is now up in #1281 (and #1280) — the stale "29"/"32" figures across `docs/` and `references/`, plus the translated READMEs, plus a guard that pins every documented figure to `len(EventType)`. Either of those supersedes this one; safe to close.

Cyrax321 (commented):

Ref: #1198 / Issue #1171
Surface: README event-type figures, docs count guard

The bug is real and verified: EventType holds 51 members (confirmed by importing the enum) while README stated 36 in the feature table and 44 in v0.2 terms in the storage table, figures that also contradicted each other. The PR corrects both rows to 51 and adds a guard in tests/test_docs_counts.py asserting any documented event-type count equals len(EventType), scoped to README because references/ is owned by #1100.

Merge-order note: #1214 is the same work (identical intent, near-identical diff) and both target #1171. Merge exactly one; the second will rebase to an empty README diff and only the guard matters then. As of today main already carries the corrected 51 figures (landed with the #1033 merge from an earlier docs fix), so what main still lacks is the guard itself. Whichever PR lands second should be closed or reduced to just the guard.

CI green at head 6b23538a6653d1e73858c910c0d56b2ad43483b6. No blocking issues.

Cyrax321 (commented):
Maintainer re-review at current head. The 51-count correction and fail-forward guard are still correct. This is a twin of #1214 on the same issue with near-identical guards; merge exactly one and fold the best parts (comma-grouped regex and scoping note here, explicit path and remediation message there). Coordinate with #1210 and #1281 on overlapping count ground. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review. Rebase out of blocked state first.

PR #1185, opened 2026-09-16
docs(architecture-data): refresh stale counts, pointers, and add a guard (#1108)

Cyrax321:
Closes #1108.

`references/architecture-data.md` bills itself as source-verified and is the reference a contributor opens to locate the code behind a concept. It had drifted far enough to contradict its own header. Fixed every verified claim and added the guard the issue asked for.

What was wrong, and is now derived from the live tree:

- Section 15 documented **14 CLI commands**; `build_parser()` accepts **46**. It listed `init` through `benchmark` and was missing 32, including `start`, `tree`, `fork`, `merge`, `gate`, `watch`, `dashboard`, `tui`, and the `attest*` group.
- Section 7 listed **29 of the 51** `EventType` members under a heading that says "Complete list".
- Section 8 listed **7 of 8** `ActionStatus` values, omitting `EXPIRED`.
- The mode rank table cited `_ORDER` at `engine.py:63`, which does not exist there and never ranked recovery modes: `_ORDER` lives in `planner.py:74` and ranks `RepairKind`. The real rank source is `SEVERITY` at `engine.py:69`. The mode and safety values themselves were correct; only the citation was fiction.
- Every pointer cluster had moved: all six checkpoint policies, the two reconcilers, the three adapters, both extractors, the storage classes, the SemanticState block (`models.py:398` to `:740`), and the MCP tool table, which is now 3 read-only / 9 mutating after `record_plan` was added.

Guard: `tests/test_architecture_data.py` re-derives the countable claims from introspection (never hardcoded expectations) and checks that every symbol the doc cites is still declared where it is cited. I confirmed the guard bites by deliberately re-staling the counts and the `SEVERITY` citation; all three drift classes failed with actionable messages.

Two judgement calls worth your review. I deliberately do **not** assert line numbers anywhere in the guard: they move on every refactor and the doc cites them for a human reader, not for a test. Asserting them would make the suite brittle without catching the failure that matters, which is a symbol being renamed or deleted while its citation survives. And #1108's suggested fix of pointing the doc at symbols instead of lines would be more durable but changes the file's shape and every cross-reference, so I kept line numbers and guarded the symbols instead.

Sequencing: this branch is rebased on top of #1177 (the glossary fix), because both touch the same three test-count lines. It documents 2,299 collected / 2,274 passed / 172 test files, which is 2,241 base + 26 (#1177) + 32 (this PR). If this lands first, #1177 will need a rebase on the same three lines. Either order keeps the docs-count guard inside its 30-test tolerance.

PR #1200, opened 2026-09-16
fix(state): account for the eleven event types the fold silently ignored (#1169)

### LINKED ISSUE #1169: state: the projection fold reports 11 of its own event types as not understood
## Summary

`project_incremental` in `src/continuum/state/semantic.py` folds events through `_dispatch`, a `match` with 20 explicit cases, and treats the rest as recorded facts via the `_NON_PROJECTING` frozenset (20 members). 11 of the 51 `EventType` members are in **neither** list: they fall through to `case _: return False` and are then counted in `report.ignored_types` — the field `ProjectionReport.complete` defines as "the fold understood every event type it consumed".

All 11 read as recorded facts rather than state (authority audit, notification receipts, the perception/branch ledger, the briefing summary surface, and the restore/merge markers), so the folds **behaviour** is right and its **bookkeeping** is wrong. Any run that restored, merged, confirmed, or notified reports as partially unprojectable: `report.applied` understates, `report.complete` is `False`, and `ignored_types` — which `tests/test_projection_edges.py:60` reserves for a deliberately bogus `QUANTUM_ENTANGLED` — fills with the names of events the codebase legitimately emits. `tests/test_compaction.py:346` asserts `not report.ignored_types`, so a compaction fixture touching any of these 11 types would see that assertion erode.

Reproduced at `d10ee63`: 6 events consumed, 2 applied, `ignored_types = {REVIEW_CONFIRMED: 1, RUN_RESTORED: 1, AUTHORITY_CONSUMED: 1, NOTIFICATION_SENT: 1}`, `complete = False`.

## Affected component

Core library — `src/continuum/state/semantic.py`

## Severity

Minor correctness / observability. The projected state itself is correct; the report and the `complete` contract it backs are not.

## System information

- Python: 3.13.5

```bash
python -c "from continuum.events import EventType; print(len(EventType))"   # 51
```

```python
from continuum import EventType, Run, SQLiteStorage

store = SQLiteStorage("t.db")
store.append_event("r1", EventType.WORK_COMPLETED, {"item": 1})
for et, payload in [
    (EventType.REVIEW_CONFIRMED, {"confirmed_by": "human"}),
    (EventType.RUN_RESTORED, {"anchor": "a1"}),
    (EventType.AUTHORITY_CONSUMED, {"grant": "g1"}),
    (EventType.NOTIFICATION_SENT, {"channel": "webhook"}),
]:
    store.append_event("r1", et, payload)
state, report = project_incremental("r1", store.read_events("r1"))
print(report.consumed, report.applied, report.ignored_types, report.complete)
# 6 2 {REVIEW_CONFIRMED: 1, RUN_RESTORED: 1, AUTHORITY_CONSUMED: 1, NOTIFICATION_SENT: 1} False
```

The static check that finds the full gap (cases and set parsed from `semantic.py` against the enum):

```
enum: 51 | match cases: 20 | _NON_PROJECTING: 20
gap: [AUTHORITY_CONSUMED, AUTHORITY_RECONCILED, BRANCH_RESOLVED, MEMORY_TOMBSTONED,
      NOTIFICATION_FAILED, NOTIFICATION_SENT, PERCEPTION_OBSERVED, REASONING_SUMMARY,
      REVIEW_CONFIRMED, RUN_MERGED, RUN_RESTORED]
```

## What a fix looks like

Add the 11 to `_NON_PROJECTING` (they are recorded facts by design), or add fold cases if any of them should project — `RUN_MERGED` and `RUN_RESTORED` are the ones worth a second look, since restore and merge change what a runs state *is*, even if the marker event itself carries no payload. Then pin the invariant with a guard test asserting `set(EventType) == cases | _NON_PROJECTING`, so every enum member is either folded or declared. Worth confirming against the closed bookkeeping issues (#364, #382, #383, #386) that none of the 11 was deliberately left unhandled.

abyyxhek:
## Description

`project_incremental` in `src/continuum/state/semantic.py` folds events through `_dispatch`, a `match` with one case per state-bearing type, and treats the rest as recorded facts via the `_NON_PROJECTING` frozenset. Eleven of the 51 `EventType` members were in **neither** list, so they fell through to `case _: return False` and were counted in `report.ignored_types` — the field `ProjectionReport.complete` defines as *"the fold understood every event type it consumed"*.

The eleven: `AUTHORITY_CONSUMED`, `AUTHORITY_RECONCILED`, `BRANCH_RESOLVED`, `MEMORY_TOMBSTONED`, `NOTIFICATION_FAILED`, `NOTIFICATION_SENT`, `PERCEPTION_OBSERVED`, `REASONING_SUMMARY`, `REVIEW_CONFIRMED`, `RUN_MERGED`, `RUN_RESTORED`.

So any run that restored, merged, confirmed, notified, consumed an authority, or recorded a perception reported itself as **partially unprojectable**: `report.applied` understated, `report.complete` was `False`, and `ignored_types` — which `tests/test_projection_edges.py` reserves for a deliberately bogus `QUANTUM_ENTANGLED` — filled with the names of events the codebase legitimately emits. `tests/test_compaction.py` asserts `not report.ignored_types`, so any compaction fixture touching one of these types would erode that assertion. The projected state itself was always correct; only the bookkeeping lied.

All eleven are recorded facts rather than state, so they join `_NON_PROJECTING` rather than gaining match cases:
- `NOTIFICATION_SENT` / `NOTIFICATION_FAILED` are documented in the enum itself as *"Recorded facts, not state"*.
- `RUN_RESTORED` / `RUN_MERGED` sit in the enum's lineage block beside `RUN_FORKED`, which is already in the set.
- The rest are authority audit, the perception/branch ledger, the briefing summary surface (read straight off the log by the CLI), and the memory tombstone. Nothing under `src/continuum/state/` consumes any of them.

## Related issues

Fixes #1169

## Types of changes

- Type: `bugfix`

## Breaking changes

No. `ProjectionReport.complete` and `ignored_types` now describe the fold accurately; the projected `SemanticState` is byte-for-byte unchanged, since these types never had a dispatch case and so never applied state.

## How has this been tested?

```
pytest tests/test_projection_ignored_types.py --no-cov          # 3 passed
pytest tests/test_projection.py tests/test_projection_bookkeeping.py \
       tests/test_projection_pins.py tests/test_projection_edges.py \
       tests/test_projection_ignored_types.py --no-cov          # 104 passed
pytest tests/test_cli.py tests/test_recovery_engine.py --no-cov # 151 passed
pytest tests/test_docs_counts.py --no-cov                       # 3 passed
uvx ruff@0.16.4 check src/continuum/state/semantic.py tests/test_projection_ignored_types.py
uvx ruff@0.16.4 format --check src/continuum/state/semantic.py tests/test_projection_ignored_types.py
mypy src/continuum/state/semantic.py                            # Success: no issues found
uv build --wheel --no-build-isolation                           # Successfully built
```

The new tests are genuine regression tests: with the fix stashed, all three fail (eleven types land in `ignored_types`, `complete` is `False`); with the fix in place all three pass. The issue's own repro was also run before and after — before: `consumed 6, applied 2, ignored_types = {REVIEW_CONFIRMED, RUN_RESTORED, AUTHORITY_CONSUMED, NOTIFICATION_SENT}, complete = False`; after: `complete = True`, `ignored_types = {}`.

The suite collects 2,258 tests, 17 above the documented ~2,241 — inside the guard's TOLERANCE of 30, so no docs re-sync is needed.

## Checklist

## Additional context

The enum sweep is written as one loop over `EventType` rather than 52 parametrized cases deliberately: 52 rows would have pushed the suite 52 past the documented total, past the docs-count guard's tolerance of 30, and forced a re-sync of three unrelated doc files with passed/skipped splits that vary by environment. The loop exercises every member identically and keeps this PR to the one issue.


src/continuum/state/semantic.py
```diff
@@ -765,6 +765,32 @@ def _dispatch(acc: _Accumulator, event: Event) -> bool:
         EventType.LIVENESS_RECOVERED,
         # risk (issue #303): real-time risk signal, never state
         EventType.RISK_OBSERVED,
+        # lineage (issue #259): restore/merge markers recording that an edit
+        # happened, sibling to RUN_FORKED above. The fold reads their effect
+        # off the log boundary itself, not out of these events.
+        EventType.RUN_RESTORED,
```
Cyrax321, line 771: Verified: identical eleven types to #1204, all present in the enum, none in the fold match list. The `RUN_RESTORED` and `RUN_MERGED` placement is the one that matters, and it is right: both are markers carrying anchor metadata, and the state a restore represents is carried by the events the anchor points back to, which the fold already reads. Same reasoning as the existing `EVENT_LOG_ANCHORED` entry.

tests/test_projection_ignored_types.py
```diff
@@ -0,0 +1,96 @@
...
+    the fold report a run it fully understood as incomplete. Driving the fold
+    with one event of each type pins the contract where it lives, on the enum.
+    """
+    ignored: list[str] = []
+    for event_type in EventType:
+        log = started(EventLog())
+        log.append("run_1", event_type, {})
+        try:
+            _state, report = project_incremental("run_1", log.events("run_1"))
+        except Exception:
```
Cyrax321, line 71: The `continue` is doing real work and is correct, but it is worth a line saying why: a type with a fold case that validates its payload raises on the empty payload here, and skipping it is fine because such a type is by definition not ignored. Without that note a future reader may read the bare `except Exception` as carelessness.

Cyrax321 (approved):
Approving. The code change is the same eleven types as #1204, and I verified all eleven exist in the enum and none of them appear in the fold match list. The mechanism is also the right one for the bug: `_dispatch` returned False for these eleven either way, so this corrects `ProjectionReport.ignored_types` and `complete` and never changes a projected state.

The guard is sound and I prefer one thing about its shape: it drives the fold with a real event of each type rather than parsing the module, so it pins the contract on the enum where the docstring says it lives. The `except Exception: continue` means a type with a fold case that validates its payload is skipped on an empty payload, so the sweep proves something only about types that fall to `case _`. That is exactly the bug class here, so it is correct; #1204 static parse would additionally catch a type whose case matches but silently no-ops. Neither is wrong, they are looking at slightly different things.

The one real gap is the CHANGELOG. This changes user-visible output: a run that restored, merged, confirmed or notified now reports `complete` where it reported partially unprojectable, with `applied` no longer understated. One paragraph under Unreleased > Fixed and this is done.

Pick one of this and #1204. Both restructure the same frozenset and are not merge-compatible, so whichever lands the other should close. This one has the cleaner scope and the tighter diff; #1204 carries the CHANGELOG entry and the architecture-data event count, but bundles a GLOSSARY and pointer refresh that overlaps #1177 and #1185.

PR #1201, opened 2026-09-16
fix(actions): count a claim's budget slot only once it is recorded (#1168)

### LINKED ISSUE #1168: actions: a rejected claim still consumes the authorization-bound retry budget
### Summary

`ActionLedger.claim()` draws down the authorization-bound retry budget **before** it validates the claim's own inputs. A claim that is ultimately rejected — never recorded, never performed — still increments the durable `authorization_bound` counter and saves the registry. The budget now reflects an attempt that the ledger has no record of, and a later, valid claim for the same resource is refused as `budget exhausted`.

### Expected Behaviour

Budget drawdown should be coupled to a recorded `ACTION_RECORDED` event. A rejected claim should consume no budget slot, so the counter and the log agree on how many attempts exist.

### Actual Behaviour

A rejected claim consumes a slot. With `max_attempts: 1`, a single bad `origin_digest` permanently blocks the resource: the second, valid claim is refused even though no attempt was ever recorded or performed.

### Steps to Reproduce

```python
import json, os, sys, tempfile
sys.path.insert(0, "src")
from continuum.actions.ledger import ActionLedger, LedgerError

d = tempfile.mkdtemp()
budgets = os.path.join(d, "budgets.json")
open(budgets, "w").write(json.dumps({"action_types": {"send_invoice": {"max_attempts": 1}}}))
os.environ["CONTINUUM_BUDGETS_PATH"] = budgets

storage = SQLiteStorage(":memory:")
storage.create_run(Run(run_id="run_1", goal="g"))
storage.append_event("run_1", "RUN_STARTED", {"goal": "g"})

try:
    ledger.claim("send_invoice", {"invoice": "INV-001"}, origin_digest="not-hex")
except LedgerError as exc:
    print("rejected:", exc)

print("events recorded:", storage.last_sequence("run_1"))  # 1 — only RUN_STARTED
print("registry:", json.load(open(budgets))["authorization_bound"])

try:
    ledger.claim("send_invoice", {"invoice": "INV-001"})   # valid this time
except LedgerError as exc:
    print("rejected again:", exc)
```

### Root Cause

`src/continuum/actions/ledger.py:979-991` — inside `claim()`, the `existing is None` branch calls `_budget_consume_claim()` (which calls `increment()` + `save_budgets()`) *before* the `origin_digest` format check that raises `LedgerError`. Any validation or storage failure after that point leaves a counter increment with no corresponding event.

The same ordering affects every later guard in the branch: the `Action(...)` construction and `self._record(...)` can also raise after the drawdown has already been persisted.

### Impact

The authorization-bound budget is the mechanism that stops fresh-key minting from resetting a cap (#390/#413). A rejected claim silently inflating the counter means the operator's configured cap is consumed by attempts that never happened, and the cap can be reached without a single side effect ever being attempted. Because the registry is the durable source of truth and the event log is the audit record, the two now disagree with no way to detect the discrepancy from the log alone.

Severity is moderate rather than high: it requires the authorization-bound registry to be configured (absent it, no budget applies at all) and a claim input to be invalid, and the run is not corrupted — only the accounting drifts.

### Evidence

```
== claim with an invalid origin_digest ==
  rejected: origin_digest must be 64 lowercase hex, got 'not-hex'
  events recorded: 1 (expected 1: RUN_STARTED)
  budget entries created by the REJECTED claim: {'23c5e66b...': {'counter': 1, 'max_attempts': 1}}
== retry the same claim with a valid digest ==
  rejected: budget exhausted for 'send_invoice' / '23c5e66b...' (1 of 1 used, 0 remaining; ...)
```

Full script: `repro_budget_rejected_claim.py`.

Baseline: full suite green before this change (`pytest -q`, 0 failures).

### Suggested Fix

Move the budget drawdown to be the last step of the claim's `existing is None` branch, after `self._record(...)` has returned — so the counter only advances once an `ACTION_RECORDED` event is durably in the log. Alternatively, wrap the drawdown and the record so a raise between them rolls the increment back.

Cyrax321:
Fixes #1168.

`ActionLedger.claim()` drew down the authorization-bound retry budget before it validated the claim's own inputs. A claim rejected after that point had already incremented the durable `authorization_bound` counter and saved the registry, so the budget reflected an attempt the ledger has no record of. With `max_attempts: 1` a single bad `origin_digest` permanently blocked the resource: the next, valid claim was refused as budget exhausted though no attempt was ever recorded or performed. Reproduced against origin/main exactly as reported, with the counter at 1 and the log holding only `RUN_STARTED`.

The fix splits the drawdown into the two phases the `budgets` helpers already document in `ensure_authorization_entry`: "callers ensure, then check `would_refuse`, then `increment` and `save_budgets`".

`_budget_refuse_if_exhausted` is the gate. It keeps the position and the fail-closed behaviour of the old single call, so an over-cap attempt still never reaches the event log and a malformed registry still refuses rather than proceeding unaccounted. It does not touch the counter.

`_budget_commit_claim` is the count. It runs once `_record` has returned, so a claim rejected by validation between the two consumes no slot. This half is fail-open where the gate is fail-closed: by then the event is already durable, and the counter is accounting that follows the audit record, not the other way round. Surfacing an error there would leave a recorded event paired with a caller-visible failure, which is worse than lost accounting. This mirrors the settlement paths (`complete`, `reconcile`), which already record first and count afterwards.

All three claim branches that consumed early are fixed, not just the one named in the issue: the fresh claim plus the `COMPENSATED` and `FAILED` retries shared the same ordering. The `complete` and `reconcile` settlement paths were already correct and are untouched.

One thing I did not change, and want to flag: two runs claiming the same authorization can still interleave between the gate and the count, because the registry is a file with no cross-process lock. That window predates this fix and widening it slightly (the record now sits inside it) does not make it worse in the case the lease covers, which is the one this bug was in. Within a single run the `_single_writer` lease still serializes the whole read-modify-write. Making the cap strict across processes is a separate, larger change.

Verified: `tests/test_budget_drawdown.py` and `tests/test_retry_budgets.py` (84) pass, a 500-test sweep over ledger, action, idempotency, authority and MCP suites passes, `test_bench_fresh_key_reissuance.py` still sees its `budget exhausted` refusal, and `ruff check` and `mypy` are clean on the changed source.

The new test `test_a_rejected_claim_consumes_no_budget_slot` was confirmed to fail against the pre-fix code with the bug's exact signature (counter 1, entry created, after a rejected claim) and to pass with the fix.


src/continuum/actions/ledger.py
```diff
@@ -977,8 +1012,12 @@ def claim(
         budget_auth_id = self._budget_authorization_id(action_type, None, arguments, volatile)
 
         if existing is None:
+            # Refuse an over-cap attempt before anything is recorded, then
+            # count it only once the record is durable: a claim rejected by
+            # the digest check below must consume no slot, or the counter and
+            # the log disagree on how many attempts exist (issue #1168).
             if budget_auth_id is not None:
-                self._budget_consume_claim(action_type, budget_auth_id)
+                self._budget_refuse_if_exhausted(action_type, budget_auth_id)
```
Cyrax321, line 1020: The gate reads but no longer reserves. Two runs can both pass here on counter 0, both record, and both commit to a counter of 2 under `max_attempts: 1`. Restoring the increment at the gate and releasing it on the rejection paths below would keep the #1168 fix and close this.

Cyrax321 (commented):
Blocking, on a concurrency point the PR body gets backwards. The split is the right instinct and the ordering is genuinely correct for the bug it fixes: I confirmed all three branches run gate, validation, `Action(...)`, `_record(...)`, then `_budget_commit_claim`, so a rejected claim consumes no slot and the coupling the docstrings claim actually holds. But the gate no longer reserves, and that weakens the cap under the exact load pattern it exists to guard.

Before the fix the increment sat at the gate, so the increment was also a reservation: a second run claiming the same authorization loaded a registry that already showed the higher counter and was refused. After the fix the gate only reads. Two runs can both gate on counter 0, both record, and both commit, ending at counter 2 under a `max_attempts` of 1 with two `ACTION_RECORDED` events for one permitted attempt. The body says widening the window "does not make it worse in the case the lease covers", but the lease is per-run (`_single_writer` acquires the run lease) while an authorization-bound budget is by definition shared across runs claiming the same authorization, which is precisely the case the lease does not cover. The cross-run race was closed by the old increment position and is open now.

Reserve-then-release keeps both invariants. Let the gate increment as it did before, and add a `_budget_release_claim` called on every path that rejects or fails after the gate: the `origin_digest` error, `Action(...)` validation, and a `self._record` that raises. Then a rejected claim still consumes nothing, and a concurrent claim still faces an incremented counter at its own gate. It is more code than the split, and it needs a test with two ledgers sharing one registry to pin it, but it is the only shape that is correct on both axes.

Non-blocking: the `except Exception: pass` in `_budget_commit_claim` is defensible once the event is durable, but lost accounting becomes undiscoverable. The webhook delivery path dead-letters its best-effort failures into the log rather than passing, and a warning here would match that convention.

src/continuum/actions/ledger.py
```diff
@@ -492,8 +498,37 @@ def _budget_consume_claim(
...
+        """
+        path = self._budget_path()
+        if not path.exists():
+            return
+        try:
+            raw = load_budgets(path)
+            ensure_authorization_entry(raw, action_type, authorization_id)
+            increment(raw, action_type, authorization_id)
+            save_budgets(path, raw)
+        except Exception:
```
Cyrax321, line 529: Defensible, given the event is already durable by this point, but a bare `pass` makes lost accounting invisible. The notification path records its best-effort failures; a warning log here would let an operator notice a budget file that has stopped matching the log.

Cyrax321 (commented):
Maintainer re-review at current head. The claim-budget concurrency gap flagged in the prior review still stands: the gate-only-reads shape opens a cross-run race the old increment position closed, since the run lease does not cover authorization-shared budgets. Reserve-then-release with a release path on every reject-or-fail branch is still the shape that holds both invariants, plus a two-ledger shared-registry test. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review.

PR #1202, opened 2026-09-16
fix(actions): a rejected claim consumes no authorization budget (#1168)

### LINKED ISSUE #1168: actions: a rejected claim still consumes the authorization-bound retry budget

abyyxhek:
## Summary

Closes #1168.

`ActionLedger.claim()` drew down the authorization-bound retry budget **before** it validated the claim's own inputs. A claim that was ultimately rejected — never recorded, never performed — still incremented the durable `authorization_bound` counter and saved the registry. With `max_attempts: 1`, a single bad `origin_digest` permanently blocked the resource: the second, valid claim was refused as `budget exhausted` even though no attempt had ever been recorded or performed.

The drawdown now runs **last** in each claim branch — after the `origin_digest` check, `Action(...)` construction, and `model_copy` — immediately before the `ACTION_RECORDED` event is appended, which is the event the budget is meant to account for.

## The change

`src/continuum/actions/ledger.py` — in all three claim branches (`existing is None`, `COMPENSATED`, `FAILED`), the `_budget_consume_claim()` call moved from the top of the branch to just before `self._record(...)`:

```python
-if budget_auth_id is not None:
-    self._budget_consume_claim(action_type, budget_auth_id)
 action = Action(...)          # or existing.model_copy(...)
+self._budget_consume_claim(...)   # now last, immediately before the event
 self._record(...)
```

No guard was removed and no check reordered relative to the others — only the drawdown moved, and it moved past every line that can raise `LedgerError` or a model/validation error. The budget is still enforced with the same fail-closed semantics (`would_refuse` still raises before any increment), and the refusal message is unchanged.

## Why this ordering

The issue's own framing is that drawdown should be *coupled to a recorded `ACTION_RECORDED` event*. Placing it immediately before the append is the closest that ordering gets: every check that can reject the claim has already run, and the increment lands one statement away from the event it accounts for. Consuming *after* the append would instead let a storage failure leave an event with no matching drawdown.

The `COMPENSATED` and `FAILED` branches had the identical shape; the issue named the `origin_digest` path specifically, but a `model_copy`/`_record` failure there produced the same drift, so all three now match.

## Verification

The issue's reproducer, before vs after:

```
== claim with an invalid origin_digest ==
  rejected: origin_digest must be 64 lowercase hex, got 'not-hex'
  events recorded: 1 (expected 1: RUN_STARTED)
- budget entries created by the REJECTED claim: {'23c5e66b...': {'counter': 1, 'max_attempts': 1}}
+ budget entries created by the REJECTED claim: {}
== retry the same claim with a valid digest ==
- rejected: budget exhausted ... ^ the claim that never recorded anything already spent the whole budget.
+ (succeeds; ACTION_RECORDED lands at seq 2, counter goes 0 -> 1)
```

The cap still holds once an attempt is real: a third claim under a fresh key is still refused `budget exhausted`, with `get_remaining == 0`.

A regression test, `test_a_rejected_claim_consumes_no_budget` in `tests/test_budget_drawdown.py`, asserts the full contract — no budget entry after a rejected claim, the valid retry still gets its slot, and the cap is still enforced afterwards. It fails on the unpatched code (`assert not {'23c5e66b...': {'counter': 1, ...}}`) and passes with the fix.

Local gate: full suite `2218 passed, 38 skipped`; `ruff check` + `ruff format --check` clean; `mypy src/continuum/actions/ledger.py` clean; `tests/test_docs_counts.py` passes (within the documented-count tolerance).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected budget tracking for rejected action claims.
  * Invalid claims no longer consume an available budget attempt, allowing a subsequent valid claim to proceed.
  * Budget usage is now recorded only after a claim passes validation and is successfully recorded.


src/continuum/actions/ledger.py
```diff
@@ -999,6 +997,13 @@ def claim(
                 started_at=utcnow(),
                 origin_digest=origin_digest,
             )
+            # Budget drawdown is coupled to the recorded event (issue #1168):
```
Cyrax321, line 1000: This comment states a coupling the code does not deliver. The drawdown runs before `self._record` on the next line, so a storage failure there leaves an increment with no `ACTION_RECORDED`, the same drift class as #1168. Narrowing the window is a real improvement over drawing down first, but the wording should say that, or the call should be split so the increment lands after `_record` returns; #1201 implements the split.

tests/test_budget_drawdown.py
```diff
@@ -176,3 +177,45 @@ def test_budget_file_absent_is_unbound(tmp_path: Path, monkeypatch: pytest.Monke
         ledger.fail(outcome.key, "boom", certain=True)
     # No exception, and no file was created
     assert not budgets_path.exists()
+
+
+def test_a_rejected_claim_consumes_no_budget(
```
Cyrax321, line 182: The right shape for a regression test: it asserts the full contract (no entry after a rejected claim, the retry still gets its slot, the cap still holds once an attempt is real) and it fails on unpatched code. One gap it leaves is the `_record` failure path, the remaining window above. A test that patches `_record` to raise and asserts no increment would pin that, and would fail against this PR as written.

Cyrax321 (changes requested):
Requesting changes, though the fix is correct for the reported case and I verified its mechanics. `_budget_consume_claim` calls `would_refuse` (ledger.py:479) before it increments, so the fail-closed gate and the over-cap refusal survive the move intact, and the origin_digest path now rejects before any counter moves. The regression test is genuine: it asserts the full contract rather than only the symptom, and it fails on unpatched code.

What stops me approving is that the invariant the new comments assert is not the one the code delivers. Both comments say the drawdown is coupled to the recorded `ACTION_RECORDED` event, but placing it immediately before the append narrows the window without closing it: if `_record` raises on a storage failure the counter has already been incremented against an event that never landed. That is the same drift class as #1168, behind a rarer door. The body frames this as a deliberate either-or against consuming after the append, which is a fair call to have made, but the comments read as though the coupling is achieved.

#1201 closes the window by splitting the call into the two phases `ensure_authorization_entry` already documents: the gate keeps the old position and the fail-closed behaviour without touching the counter, and the commit runs once `_record` has returned. That is the coupling these comments describe. The two PRs restructure the same lines and are not merge-compatible, so either fold the split in here or pick #1201 and close this one.

abyyxhek:
Closing this in favour of #1201, which closes the window the comments here describe: it splits the drawdown into the gate and commit phases `ensure_authorization_entry` documents, so the counter moves only once `_record` has returned rather than immediately before the append. Keeping both open leaves two PRs restructing the same lines. Thanks for the review, @Cyrax321 — your fix is the better one here.

PR #1203, opened 2026-09-16
fix(webhooks): fire the requires_review filter instead of silently accepting it (#1180)

### LINKED ISSUE #1180: webhooks: the documented `requires_review` event filter can never fire
## Summary

`notify_blocked` (`src/continuum/recovery/webhooks.py:245`) selects endpoints by matching the `mode` string it is handed against each endpoint's `events` set (`registry.for_event(mode)`, line 269). The only production call site, `cmd_resume` at `src/continuum/cli/main.py:1487-1488`, is gated on `presented_mode == RecoveryMode.REQUEST_HUMAN.value` and forwards that same string as `mode`. So in production `mode` is always a `RecoveryMode` value — and `RecoveryMode` has no `requires_review` member:

```
resume, repair_and_resume, rollback, wait, request_human, replan, abort
```

Meanwhile `requires_review` is a first-class, documented, load-accepted event filter:

- `webhooks.py:57` defines `EVENT_REQUIRES_REVIEW = "requires_review"`, documented as *"opt-in for operators who also want to hear about self-certified runs awaiting confirmation."*
- `webhooks.py:63` includes it in `_KNOWN_EVENTS`, so the registry **accepts** an endpoint configured with `"events": ["requires_review"]`.
- `docs/guides/webhooks.md` advertises it: *"`events` filters what the endpoint hears about: `request_human` (the default) and optionally `requires_review`. An unknown name is refused at load time, because a typo'd filter that silently never fires is the quietest failure there is."*

The correctly-spelled filter is accepted by the loader and then never fires — exactly the silent failure the strict validation was written to prevent. An operator who configures it gets zero notifications, forever, with no warning.

The mismatch went unnoticed because `notify_blocked`'s `mode` parameter is typed `str` (`webhooks.py:249`), not `RecoveryMode`, so the unreachable value type-checks fine.

## Affected component

`src/continuum/recovery/webhooks.py`, `src/continuum/cli/main.py`, `docs/guides/webhooks.md`

## Severity

Medium. Silent no-op of a documented, validated configuration. The notification path is fire-and-forget and never affects the recovery verdict, so this is a missed signal, not a safety violation — but the signal it misses is the one that tells an unattended operator a run is parked on self-certified state.

## System information

- Python: 3.13.5

A run whose every component is `requires_review`, decided `REQUEST_HUMAN`, with both documented filters configured:

```python
import os, tempfile, threading
from http.server import BaseHTTPRequestHandler, HTTPServer
from continuum.models import Run, Origin, RecoveryContract, RecoverySafety
from continuum.state.validator import StateValidator
from continuum.recovery.webhooks import WebhookRegistry, WebhookEndpoint, notify_blocked

hits = []
class H(BaseHTTPRequestHandler):
    def do_POST(self):
        hits.append(self.path)
        self.send_response(200); self.end_headers(); self.wfile.write(b"ok")
srv = HTTPServer(("127.0.0.1", 0), H)
port = srv.server_address[1]
threading.Thread(target=srv.serve_forever, daemon=True).start()

tmp = tempfile.mkdtemp()
st = SQLiteStorage(os.path.join(tmp, "run.db"))
rid = "review-run"
st.create_run(Run(run_id=rid, goal="g"))
# Every state-writing event from a self-certified origin.
st.append_event(rid, EventType.RUN_STARTED, {"goal": "g"}, source=Origin.EXTERNAL_AGENT)
st.append_event(rid, EventType.TASK_UPDATED,
                {"completed": 1, "pending": 0, "failed": 0, "total": 1},
                source=Origin.EXTERNAL_AGENT)

state, _ = project_incremental(rid, list(st.read_events(rid)))
report = StateValidator().validate(state).report
print("component statuses:", [(c.component.value, c.status.value) for c in report.statuses])
# -> [("goal", "requires_review"), ("progress", "requires_review")]

reg = WebhookRegistry(endpoints=(
    WebhookEndpoint(url=f"http://127.0.0.1:{port}/human",
                    events=frozenset({"request_human"}), re_notify_seconds=3600),
    WebhookEndpoint(url=f"http://127.0.0.1:{port}/review",
                    events=frozenset({"requires_review"}), re_notify_seconds=3600),
))
contract = RecoveryContract(run_id=rid, recovery_status=RecoverySafety.REQUIRES_HUMAN,
                            reason="at least one repair needs a person")
notify_blocked(st, rid, mode="request_human",
               payload={"mode": "request_human"}, contract=contract, registry=reg)
print("endpoints hit:", hits)   # -> ["/human"]  .../review never fires
srv.shutdown()
```

Output:

```
component statuses: [("goal", "requires_review"), ("progress", "requires_review")]
endpoints hit: ["/human"]
```

The run's defining characteristic is that its components require review, the operator asked to hear about exactly that, and the bell did not ring.

Confirming no other path supplies the mode — these are the only call sites in `src/`:

```
$ git grep -n "notify_blocked(" -- src/
src/continuum/cli/main.py:1286:    records = notify_blocked(
src/continuum/recovery/webhooks.py:245:def notify_blocked(
$ git grep -n "_notify_blocked_run(" -- src/
src/continuum/cli/main.py:1256:def _notify_blocked_run(
src/continuum/cli/main.py:1488:        _notify_blocked_run(storage, run_id, presented_mode, payload, decision, args, err)
```

## What a fix looks like

Two coherent directions; pick one rather than leaving the filter half-implemented.

**(a) Wire it.** The condition the filter describes is real and common: `StateStatus.REQUIRES_REVIEW` is produced for self-certified origins at `src/continuum/state/validator.py:733` (goal), `:764` (progress) and `:867` (plan). In `cmd_resume`, when the verdict is `REQUEST_HUMAN` and any component carries `REQUIRES_REVIEW`, also notify with `mode="requires_review"` for endpoints subscribed to it. Because `mode` is a plain string, this needs no enum change — but the mode should be derived from the validation report, not from `RecoveryMode`, which is why the current call site cannot produce it.

**(b) Remove it until it is wired.** Drop `EVENT_REQUIRES_REVIEW` from `_KNOWN_EVENTS` and from `docs/guides/webhooks.md`, and let the loader refuse it with the existing "unknown event" error naming the known set. That converts the silent never-fires into the loud refusal the docs already argue for — and it is a one-line revert when (a) lands.

Option (a) preserves the documented capability; option (b) makes the code honest today. Either is strictly better than a validated filter that cannot fire.

Note the interaction with #1169: a wired `requires_review` path would append `NOTIFICATION_SENT`/`NOTIFICATION_FAILED` rows, which the projection fold currently reports as not understood. Worth fixing both together.

Adhi1-2:
## Summary

Fixes #1180.

`notify_blocked` selects endpoints by matching the `mode` string it is handed against each endpoint's `events` set. The only production caller, `cmd_resume`, forwards a `RecoveryMode` value, and that enum has no `requires_review` member. So `requires_review` was a first-class, documented, load-accepted event filter that could never fire: the registry accepted it, the docs advertised it, and an operator who configured it got zero notifications forever with no warning. That is precisely the silent failure the strict `"unknown event"` validation was written to prevent.

## The change

I took **option (a)** from the issue, wiring the filter rather than removing it, because the condition it describes is real and common: `StateStatus.REQUIRES_REVIEW` is produced for self-certified origins on goal, progress, and plan.

In `cmd_resume`, when the verdict is `request_human` and any component of the validation report carries `REQUIRES_REVIEW`, a second notification round is delivered with `mode="requires_review"`. The mode is derived from the validation report, not from `RecoveryMode`, which is exactly why the existing call site could not produce it.

Design points worth calling out:

- **Second signal, not a substitute.** The human-gate round still fires; an endpoint subscribed to both filters hears both.
- **Own dedup key.** `verdict_key` includes the mode, so the two rounds share a contract but not a key, and a cron re-running resume re-rings neither.
- **Payload reports its own mode.** The review round overrides `payload["mode"]` so a recipient can tell which bell rang.
- **Negative case stays silent.** The derivation is the report, not the mode: `request_human` alone fires the review filter not at all, and no notification rows are appended.

`docs/guides/webhooks.md` now states when the filter fires; previously it advertised a capability whose firing condition was specified nowhere.

## Verification

The issue's repro, with both filters configured, now hits `/review` as well as `/human`.

Two new tests in `tests/test_webhooks.py`, at the CLI level through `_run_cli`:

- `test_resume_fires_the_requires_review_filter_for_self_certified_state` seeds a run that is both `request_human` (unreconciled side effect) and `requires_review` on goal and progress (`EXTERNAL_AGENT` origins), and asserts both endpoints hear their own mode, plus two `NOTIFICATION_SENT` rows.
- `test_requires_review_filter_stays_silent_on_a_plain_human_gate` reuses the existing `_seed_blocked_run` (plain human gate, nothing to review) and asserts no POST and no notification rows.

The first test fails on the pre-fix code (1 POST instead of 2), confirming it guards the regression rather than passing vacuously. Full suite green; `ruff check`, `ruff format --check`, `mypy src/continuum`, and `markdownlint` on the touched files all clean, and no em dashes in the diff.

## Out of scope

The issue notes an interaction with #1169: a wired `requires_review` path appends more `NOTIFICATION_SENT` / `NOTIFICATION_FAILED` rows, which the projection fold currently reports as not understood. That is a separate issue with its own scope (it equally affects the pre-existing `request_human` path) and is deliberately not folded in here.


src/continuum/cli/main.py
```diff
@@ -1485,7 +1485,27 @@ def cmd_resume(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
...
+        # knows a component was downgraded for a self-certified origin.
+        if any(
+            entry.status is StateStatus.REQUIRES_REVIEW
+            for entry in decision.validation.report.statuses
+        ):
+            _notify_blocked_run(
+                storage,
+                run_id,
+                EVENT_REQUIRES_REVIEW,
+                {**payload, "mode": EVENT_REQUIRES_REVIEW},
```
Cyrax321, line 1504: This override is load-bearing and worth one line saying so: `notify_blocked` sets `body["event"]` from the mode but never touches `body["mode"]` (webhooks.py:287), so without this the review payload would arrive with `event: requires_review` and `mode: request_human`, which is exactly the ambiguity the filter exists to remove.

src/continuum/cli/main.py
```diff
@@ -1485,7 +1485,27 @@ def cmd_resume(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
...
+
         _notify_blocked_run(storage, run_id, presented_mode, payload, decision, args, err)
+        # A run parked on self-certified state is a distinct signal from a
+        # plain human gate, and the requires_review filter exists to carry it
+        # (issue #1180). RecoveryMode has no such member by design: the mode
+        # is derived from the validation report, which is the only place that
+        # knows a component was downgraded for a self-certified origin.
+        if any(
+            entry.status is StateStatus.REQUIRES_REVIEW
+            for entry in decision.validation.report.statuses
```
Cyrax321, line 1498: The derivation is the report rather than the mode, which is the right call: `request_human` alone says nothing about whether state was self-certified, so gating on the mode would fire the filter on every human gate or on none. Testing `any` over every component is broader than the docs wording; see the note on webhooks.md.

src/continuum/cli/main.py
```diff
@@ -1485,7 +1485,27 @@ def cmd_resume(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
...
+        ):
+            _notify_blocked_run(
```
Cyrax321, line 1500: This re-loads and re-validates the registry, so a malformed file now prints the warning twice. Harmless, and hoisting the load out to one call is a follow-up rather than a blocker.

docs/guides/webhooks.md
```diff
@@ -31,7 +31,12 @@ data-beside-code convention as the gate and gateway registries:
 * `events` filters what the endpoint hears about: `request_human` (the
   default) and optionally `requires_review`. An unknown name is refused at
   load time, because a typo'd filter that silently never fires is the
-  quietest failure there is.
+  quietest failure there is. `requires_review` fires only for a blocked run
+  (`continuum resume` reporting `request_human`) whose own state was written
+  by a self-certified origin, so goal or progress carries
```
Cyrax321, line 36: The code fires on any component carrying `REQUIRES_REVIEW`, but this says goal or progress specifically. `ComponentValidationEntry.status` is a `StateStatus` and any component can be downgraded for a self-certified origin, so the prose understates the condition. Either narrow the `any` to the goal and progress components or reword this to "any component of its state".

Cyrax321 (dismissed):
Approving. The diagnosis is correct and I traced the wiring end to end: `WebhookRegistry.for_event` matches on the mode string (webhooks.py:96), the only production caller forwards a `RecoveryMode` value, and that enum has no `requires_review` member, so the documented filter could not fire. Deriving the mode from the validation report is the right fix, and the two rounds are correctly independent: `verdict_key` hashes the mode (webhooks.py:189), so neither dedups the other while a cron re-running resume stays quiet on both.

Two nits on the lines below. The bigger decision is a coordination one: #1213 closes the same issue with a much smaller change by making `for_event("request_human")` also return review-subscribed endpoints. It is simpler, but the recipient gets a payload labeled `event: request_human` and cannot tell a human gate from review-worthy state, which is exactly the distinction the docs paragraph added here promises. If the filter should be its own signal, this PR is the one to take; #1213 turns it into an alias for request_human. They are not merge-compatible, so pick before either lands.

The interaction with #1169 is handled correctly and is correctly out of scope: the extra NOTIFICATION_SENT rows this produces are among the eleven types #1204 and #1200 are declaring non-projecting, so the projection report will not be marred by them.

PR #1204, opened 2026-09-16
fix(state): declare 11 audit event types non-projecting, they are understood not ignored (#1169)

### LINKED ISSUE #1169: state: the projection fold reports 11 of its own event types as not understood

Cyrax321:
Closes #1169.

`_dispatch` in `src/continuum/state/semantic.py` folds 20 `EventType` members into state and `_NON_PROJECTING` declares 20 more as recorded facts the fold is right to skip. 11 members were in neither list. Each fell through to `case _: return False` and was counted in `report.ignored_types`, which `ProjectionReport.complete` defines as "the fold understood every event type it consumed". So any run that restored, merged, confirmed, or notified reported as partially unprojectable: `complete = False`, `report.applied` understated, and `ignored_types` filling with names the codebase legitimately emits.

I reproduced the issue's exact scenario before the fix (6 consumed, 2 applied, 4 ignored types, `complete = False`) and confirmed after the fix it reads 6 consumed, 2 applied, empty `ignored_types`, `complete = True`.

The 11 were: `RUN_RESTORED`, `RUN_MERGED`, `REVIEW_CONFIRMED`, `AUTHORITY_CONSUMED`, `AUTHORITY_RECONCILED`, `PERCEPTION_OBSERVED`, `BRANCH_RESOLVED`, `REASONING_SUMMARY`, `NOTIFICATION_SENT`, `NOTIFICATION_FAILED`, `MEMORY_TOMBSTONED`.

On the issue's open question, whether any of the 11 should fold rather than be declared: I traced all 11 to their emitters and readers and none carries state. `RUN_RESTORED` and `RUN_MERGED` were flagged as worth a second look, but the restore and merge paths replay the events that carry the actual state change; the marker event records that the splice happened and is read by the CLI lineage views. Declaring them non-projecting is correct, and it keeps `report.ignored_types` reserved for genuinely unknown types, which is what `test_projection_edges.py:60` asserts it is for.

The codebase already has the precedent and the convention. `test_compaction.py:473` asserts that `EVENT_LOG_ANCHORED` "must be declared non-projecting, not silently ignored", and every existing entry in `_NON_PROJECTING` carries a comment naming the issue that reasoned about it. The 11 new entries follow that same shape.

Guard: `tests/test_projection_coverage.py` parses `semantic.py` the same way the gap was found and asserts `set(EventType)` equals the match cases plus the declarations. I confirmed it bites by removing two declarations and watching it fail with an actionable message naming the unhandled members and the two ways to fix it. A second test exercises every declared member end to end and asserts the report comes back complete with nothing ignored, so a future member added to the set is covered the day it lands rather than when someone happens to emit it.

Two notes for review. The fix is behaviour-preserving for the projected state itself; only the report and the `complete` contract change, and `report.applied` for a stream of nothing but declared facts is now 1 rather than 0, matching what the fold actually did. And this branch is rebased on top of #1177 and #1185, because all three touch the same test-count lines in README, CONTRIBUTING_ONBOARDING, and CHANGELOG. It documents 2,302 collected / 2,277 passed, which is 2,241 base + 26 (#1177) + 32 (#1185) + 3 (this PR). If the landing order differs, whichever PR lands last carries the rebasing; the 30-test tolerance keeps the guard green in any order.


Cyrax321 (commented):
Changes wanted before merge, and a decision to make first. The fix is correct, and I verified the mechanism: `_dispatch` returns False for these eleven either way, so this alters `ProjectionReport.ignored_types` and `complete` only and never a projected state. All eleven names exist in the enum, none of them appear in the fold match list, and the restore/merge reasoning holds. #1200 makes the identical eleven-type change, differing only in comment wording, so one of the two should close.

The scope is what I would block on. The PR re-anchors roughly sixteen GLOSSARY code pointers and rewrites 58 lines of architecture-data (adapter classes, MCP server entries, models.py offsets), none of which has anything to do with #1169. That is #1177 glossary work and #1185 architecture-data work. If either lands first, this conflicts across whole files rather than a few hunks. Either split the pointer sweep into its own PR, or mark this as superseding both and close them; leaving all three open puts three PRs on the same lines.

Non-blocking, on the lines below: the coverage test is the reason to prefer this over #1200, but it has one limit worth stating. It verifies every enum member is folded or declared; it cannot verify the declaration is correct. A future type that should fold but gets declared non-projecting passes the test and stops being counted as applied. The per-type reasoning in the comments is therefore the real guarantee, and new entries should keep needing it.

src/continuum/state/semantic.py
```diff
@@ -765,6 +765,34 @@ def _dispatch(acc: _Accumulator, event: Event) -> bool:
         EventType.LIVENESS_RECOVERED,
         # risk (issue #303): real-time risk signal, never state
         EventType.RISK_OBSERVED,
+        # restore/merge lineage (issue #1169): markers that a run's history was
+        # spliced, read by the CLI lineage views; the splice itself replays
+        # events that carry the state change, the marker carries none.
+        EventType.RUN_RESTORED,
```
Cyrax321, line 771: Verified: `RUN_RESTORED` is emitted by `restore.py:142` carrying anchor metadata, and the state change a restore represents is carried by the events at and before the anchor, which the fold already reads. The marker adds no field to `SemanticState`. Same reasoning as the existing `EVENT_LOG_ANCHORED` entry, so the eleven belong here rather than as fold cases.

tests/test_projection_coverage.py
```diff
@@ -0,0 +1,106 @@
...
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Match):
+            for case in node.cases:
+                pattern = case.pattern
+                if isinstance(pattern, ast.MatchValue) and isinstance(pattern.value, ast.Attribute):
+                    cases.add(pattern.value.attr)
+    return cases | {member.name for member in _NON_PROJECTING}
+
+
+def test_every_event_type_is_folded_or_declared() -> None:
```
Cyrax321, line 44: This is the strongest part of the PR: it parses the module, enumerates `EventType`, and fails on any member in neither list, which is exactly how #1169 shipped. The one thing it cannot catch is the opposite error, a type that should fold but is declared non-projecting. Worth one line in the module docstring making the reasoning comment mandatory for new entries, since that comment is the actual guarantee.

references/architecture-data.md
```diff
@@ -102,27 +103,33 @@ source of truth, and events persist to storage.
...
-RUN_STARTED, RUN_COMPLETED, RUN_ABORTED, TASK_UPDATED, TOOL_CALLED,
-TOOL_COMPLETED, TOOL_FAILED, DECISION_CREATED, DECISION_INVALIDATED,
-EVIDENCE_ADDED, FINDING_ADDED, FINDING_INVALIDATED, WORK_ADDED,
-WORK_COMPLETED, DEPENDENCY_DECLARED, APPROVAL_REQUESTED, APPROVAL_GRANTED,
-APPROVAL_REVOKED, MODEL_CHANGED, MODEL_ASSUMPTION_RECORDED,
-STATE_CHECKPOINTED, STATE_VALIDATED, ENVIRONMENT_CHANGED, RECOVERY_STARTED,
-RECOVERY_COMPLETED, RECOVERY_BLOCKED, ACTION_RECORDED, ACTION_RECONCILED,
-ACTION_COMPENSATED
+  event (`events.py:401`).
+- 51 event types (`EventType` StrEnum, `events.py:47`). Complete list:
```
Cyrax321, line 116: In scope and correct: the fold coverage is this PR subject, and 51 is the enum real size. The adapter and MCP pointer rewrites further down this file are not in scope and are what overlaps #1185.

docs/GLOSSARY.md
```diff
@@ -3,31 +3,31 @@
 Every term here is used in issues and in code. Definitions point to the implementation so they can be verified, not inferred.
 
 - **Anchor**  
-  A checkpoint or ledger entry that must survive cleanup and compaction. In code a checkpoint with `trigger == CheckpointTrigger.RECOVERY` (`src/continuum/checkpoint/policy.py:62`) and a ledger entry with `anchor == True` (`src/continuum/recovery/ledger.py:88`) are anchors. `CheckpointManager.prune` and `RecoveryLedger.compact` keep anchors while dropping older ephemeral entries. See `src/continuum/recovery/cleanup.py:20` for the cleanup rule that preserves referenced anchors.
+  A checkpoint or ledger entry that must survive cleanup and compaction. In code a checkpoint with `trigger == CheckpointTrigger.RECOVERY` (`src/continuum/checkpoint/policy.py:62`) and a ledger entry with `anchor == True` (`src/continuum/recovery/ledger.py:95`) are anchors. `CheckpointManager.prune` and `RecoveryLedger.compact` keep anchors while dropping older ephemeral entries. See `src/continuum/recovery/cleanup.py:20` for the cleanup rule that preserves referenced anchors.
```
Cyrax321, line 6: Out of scope for #1169. This re-anchor is #1177 work; merging both means one conflicts on nearly every entry. Either fold the two deliberately or keep the pointer sweep in its own PR.

Cyrax321 (commented):

Ref: #1204 / Issue #1169
Surface: state/semantic.py projection coverage, glossary pointer guard

Bug: 11 of 51 EventType members were neither folded by _dispatch nor declared in _NON_PROJECTING, so any run emitting restore, merge, confirm, authority, perception, reasoning, notification, or tombstone events reported ProjectionReport.complete = False with the names filling ignored_types. The complete flag exists to mean the fold understood every type it consumed, so legitimate audit events were breaking that contract.

Fix verified at head b9cbd04: the 11 types are declared in _NON_PROJECTING with a rationale comment per group, matching the existing pattern for refusal-audit and log-maintenance types. The PR body traces all 11 to their emitters and shows none carries state, and it answers the issue's open question on RUN_RESTORED and RUN_MERGED explicitly rather than silently declaring them. Bonus: the glossary pointer table in tests/test_docs_glossary.py is now bidirectional (untested and stale pointers both fail), which is a real improvement over one-sided checks.

Proof: ran tests/test_projection.py, test_projection_bookkeeping.py, test_projection_coverage.py, test_projection_edges.py, and test_docs_glossary.py at head: 93 passed. CI green on the head commit.

Cyrax321 (commented):
Maintainer re-review at current head. The 11-type non-projecting declaration is still correct. Prior review flagged scope overlap with the glossary and architecture-data PRs (#1177, #1185): either split the pointer sweep out or mark this as superseding both. Coordinate with #1200, which makes the identical change; only one should land. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review. Rebase out of dirty state first.

PR #1205, opened 2026-09-16
fix(cli): resume on a terminal run reports not-resumable instead of exit 0 (#1197)

### LINKED ISSUE #1197: cli: resume on a terminal run (completed/aborted/failed) exits 0 with mode resume, breaking the only-a-safe-run-exits-0 contract
### Summary

`continuum resume <run_id>` evaluates a **terminal** run (completed, aborted, or failed) exactly like an active one and returns recovery mode `resume` with exit code `0`. The exit-code contract in `references/cli.md` and `src/continuum/cli/exitcodes.py` states that `0` is "the only code that permits a pipeline to continue" and means "verified safe to resume" — but a terminal run has nothing to resume.

### Expected behaviour

`continuum resume "$RUN" && ./start-agent.sh` must never launch an agent onto a run that is already closed. A terminal run should be reported as not resumable (for example `REQUEST_HUMAN`/a dedicated non-zero code, or an explicit "run is completed" answer), mirroring the guard `get_active_run()` already applies: terminal statuses are deliberately excluded from the implicit target because "a completed run is terminal: it can never be offered for resume" (`tests/test_cli.py:952`).

### Actual behaviour

`resume` reads the run's state, folds it, finds nothing wrong, and reports `RESUME` / `Next permitted action: continue` with exit code `0`.

```shell
git clone https://github.com/Cyrax321/CONTINUUM && cd CONTINUUM
uv pip install -e . --extra dev

DB=$(mktemp -d)/t.db
continuum --db "$DB" init
continuum --db "$DB" start run_1 --goal "ship it"
continuum --db "$DB" complete run_1            # terminal: COMPLETED

continuum --db "$DB" resume run_1; echo "exit=$?"
continuum --db "$DB" --json resume run_1 | python -c "import json,sys; print(json.load(sys.stdin)['mode'])"
```

The same holds when the run row is set to `aborted` or `failed` directly.

### Root cause

`cmd_resume` in `src/continuum/cli/main.py:1347` never inspects the run's status. It resolves `run_id`, calls `RecoveryEngine.assess()`, and returns `exit_code_for(effective_mode)` at `src/continuum/cli/main.py:1518`. `assess()` folds the event log and finds a consistent state, so it yields `RecoveryMode.RESUME`, which maps to `ExitCode.OK`. The only terminal-aware logic in the command is the implicit-target branch (`get_active_run`, `main.py:1351`); the explicit `run_id` path has no such check, and `Storage.require_usable_run_id` (`storage/base.py:208`) only rejects blank ids.

### Impact

The documented safety contract is broken on the exact automation pattern the exit codes were designed to protect. A script that resumes by explicit id (the recommended form, since `--db` runs often name the run) gets exit `0` and `mode: resume` for a run whose work is finished or was aborted, and proceeds to start an agent against closed state. It also misleads an operator reading the verdict: "Next permitted action: continue" on a completed run is wrong, and `--repair` on a terminal run can append `RECOVERY_STARTED` to a closed event log.

### Logs or CI run

```
$ continuum --db t.db resume run_1
CONTINUUM RECOVERY
Run: run_1
Checkpoint: v0
State validation:
  [ok] goal - v1
  [ok] progress - 0 completed
Recovery decision: RESUME
  because all state verified against the environment
Next permitted action: continue
$ echo $?
0
$ continuum --db t.db --json resume run_1 | grep '"mode"'
  "mode": "resume",

# Same for runs set to status=aborted and status=failed: mode=resume, exit 0.
```

Environment: Windows 11, Python 3.11, `continuum-agent` 0.1.2 built from `main` at `cef019d`, installed from source.

### Suggested fix

In `cmd_resume`, after resolving `run_id` and loading the run, refuse or downgrade a terminal run, e.g. emit a `REQUEST_HUMAN`-equivalent verdict with an explicit "run is completed/aborted/failed; nothing to resume" rationale and a non-zero exit, and skip the `--repair` append. `RunStatus` already carries the terminal set used by `get_active_run` (`storage/sqlite.py:296`), so the same four statuses can be reused.

### Before reporting

### Affected component

### Severity

abyyxhek:
## Description

`continuum resume <run_id>` never inspected the run's status. It resolved the id, folded the event log, and returned `exit_code_for(mode)`. The folded log of a cleanly closed run is consistent, so `assess()` answered `RESUME` — the one mode mapped to exit `0`, the code `exitcodes.py` reserves for *"verified safe to resume"* — and the command printed `Next permitted action: continue` for a run whose work was finished or had been aborted. The automation the exit codes exist to protect then launched an agent onto closed state:

```
continuum resume "$RUN" && ./start-agent.sh
```

`--repair` made it worse: it appended `RECOVERY_STARTED` to a finished event log.

The implicit-target path already refuses these runs — `get_active_run()` excludes `completed`, `crashed`, `aborted` and `failed` because *"a completed run is terminal: it can never be offered for resume"* (`tests/test_cli.py`). The explicit-id path was simply more permissive. It now refuses the same set before assessment, reporting `Recovery decision: NOT_RESUMABLE` with exit 20 and `safe: false`. A closed run is reported, not escalated, so unlike a blocked run it does not ring the webhook bell.

```
$ continuum --db t.db resume run_1
CONTINUUM RECOVERY

Run: run_1

Recovery decision: NOT_RESUMABLE
  because the run is completed: a terminal run has nothing to resume.
Next permitted action: none (start a new run instead)

$ echo $?
20
```

The invariant this PR holds is: **a terminal run never exits 0, never claims `safe`, and never reports mode `resume`**. The specific presentation — mode `request_human`, exit 20 — is the closest existing band in `_MODE_CODES` and matches the house rule this command already applies when a family block must not read as permission to continue (issue #741); a dedicated `TERMINAL` code would be an equally valid follow-up but would renumber a public machine contract. The JSON payload carries `status` and `terminal: true` so a machine consumer gets the real reason rather than having to infer it.

## Related issues

Fixes #1197

## Types of changes

- Type: `bugfix`

## Breaking changes

Behaviour changes only for runs that were already finished. Previously `resume` on a completed/aborted/failed/crashed run exited `0` and reported `mode: resume`; now it exits `20` with `mode: request_human`, `safe: false` and `terminal: true`. Any script that gated on the exit code is strictly safer; a script that treated exit 0 as "fine" was the bug. The `--json` payload for these runs is a focused one (`run_id`, `goal`, `status`, `mode`, `safe`, `next_allowed_action`, `terminal`) rather than the full assessment payload, since there is no assessment to report. Active runs are unchanged.

## How has this been tested?

```
pytest tests/test_cli_resume_terminal.py --no-cov   # 10 passed
pytest tests/test_cli.py tests/test_cli_resume_terminal.py --no-cov   # 120 passed
pytest tests/test_docs_counts.py --no-cov           # 3 passed
uvx ruff@0.16.4 check src/continuum/cli/main.py tests/test_cli_resume_terminal.py
uvx ruff@0.16.4 format --check src/continuum/cli/main.py tests/test_cli_resume_terminal.py
mypy src/continuum/cli/main.py                     # Success: no issues found
uv build --wheel --no-build-isolation              # Successfully built
```

The new tests are genuine regression tests: with the fix stashed, 9 of 10 fail (each terminal status exits 0 and reports `safe: true`), and the tenth — an active run that must still exit 0 — passes either way, as intended. The issue's own CLI reproduction was also run before and after, for all four terminal statuses plus `--repair` (which now appends nothing to the log).

The suite collects 2,265 tests, 24 above the documented ~2,241 — inside the docs-count guard's TOLERANCE of 30, so no docs re-sync is needed.

## Checklist

## Additional context

The terminal-status set is defined locally in `cmd_resume`'s module rather than shared with the storage layer. `get_active_run()` builds the same tuple inline in both `sqlite.py` and `postgres.py`; extracting a shared constant would touch both backends and is a larger change than this fix needs, so the guard mirrors the documented set and cites it. The redundant `storage.get_run(run_id)` later in `cmd_resume` was folded into the row the guard already fetched.


src/continuum/cli/main.py
```diff
@@ -1344,6 +1344,50 @@ def cmd_notify_test(args: argparse.Namespace, storage: None, out: Any, err: Any)
     return ExitCode.OK
 
 
+# The statuses get_active_run deliberately excludes: a run in any of them is
+# finished and can never be offered for resume ("a completed run is terminal",
+# test_cli.py). resume's explicit-id path must refuse the same set.
+_TERMINAL_RUN_STATUSES = frozenset(
```
Cyrax321, line 1350: Verified against `get_active_run` (sqlite.py:297): the same four statuses. The mirror is faithful, and citing the guard you are matching in the comment is the right call.

src/continuum/cli/main.py
```diff
@@ -1344,6 +1344,50 @@ def cmd_notify_test(args: argparse.Namespace, storage: None, out: Any, err: Any)
...
+        f"Run: {run.run_id}\n\n"
+        "Recovery decision: NOT_RESUMABLE\n"
+        f"  because the run is {status}: a terminal run has nothing to resume.\n"
+        "Next permitted action: none (start a new run instead)\n"
+    )
+    payload: dict[str, Any] = {
+        "run_id": run.run_id,
+        "goal": run.goal,
+        "status": status,
+        "mode": RecoveryMode.REQUEST_HUMAN.value,
```
Cyrax321, line 1376: This is the decision worth making consciously. The docstring on 20 says "a person must decide", but a completed run needs no decision from anyone, and two lines down the text says `Next permitted action: none`. Non-blocking because every field a machine reads is correct; flagged so the choice is on the record.

src/continuum/cli/main.py
```diff
@@ -1344,6 +1344,50 @@ def cmd_notify_test(args: argparse.Namespace, storage: None, out: Any, err: Any)
...
+        "safe": False,
+        "next_allowed_action": "none",
```
Cyrax321, line 1378: The engine vocabulary for this field is action names (`revalidate`, `dataset_revalidation`); `"none"` is new, so consumers keying on the field will not recognise the value. It reads correctly here. If it stays, one line in `references/cli.md` naming it as the terminal-run value closes the gap.

src/continuum/cli/main.py
```diff
@@ -1356,6 +1400,16 @@ def cmd_resume(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
...
             return 2
         run_id = active.run_id
+    # A terminal run would otherwise assess as RESUME: the folded log of a
+    # cleanly closed run is consistent, and RESUME is the mode exit_code_for
+    # sends to 0 -- the code reserved for "verified safe to resume". That let
+    # `resume "$RUN" && ./start-agent.sh` launch an agent onto a run that was
+    # already completed or aborted, and let --repair append RECOVERY_STARTED
+    # to a closed log. get_active_run excludes these statuses from the
+    # implicit target for the same reason; the explicit path now matches it.
+    run = storage.get_run(run_id)
```
Cyrax321, line 1410: On the implicit path this row was already fetched as `active` and is guaranteed non-terminal, so this is a second query for the same row. Hoisting the fetch into the branch (`run = active` there, `storage.get_run` for the explicit id) removes it, which also matches the fold you already made of the later `get_run(run_id).goal`.

Cyrax321 (changes requested):
Requesting changes. The fix is correct where it counts: the guard runs before assessment, the terminal set is exactly the four statuses `get_active_run` excludes, and the contract that matters (never exit 0, never `safe: true`, never `mode: resume`) is tested across all four statuses, with `--repair` confirmed to append nothing to a closed log.

Blocking, and it has to live in the summary because the file has no hunks in this diff. `references/cli.md` opens with "Exit codes are a safety contract" and documents exit 20 as "a human must decide". As of this PR, exit 20 also means "this run is finished", the command prints a new `NOT_RESUMABLE` decision line, and the JSON payload gains a `terminal` field. None of that is in the diff. Either add a short paragraph to that section, or untick "Documentation updated" in the PR body, which currently asserts docs were updated when no doc file changed.

Non-blocking, but worth deciding before merge. Parking a finished run under `REQUEST_HUMAN` inherits exactly the conflation #1206 is splitting out for `WAIT`: "a person must decide" and "this run is over" are different answers, and the text printed below already says so. Once `terminal: true` and 20 ship together, scripts pin to 20. The body concedes a dedicated code would be equally valid; if that is the plan, one line now is cheaper than a second renumbering later.

Heads up that #1217 also fixes issue #1197. Worth coordinating so one of you does not rebase for nothing.

Cyrax321 (changes requested):
Author context: abyyxhek, external contributor. Follow-up to the prior review. The terminal-run guard direction is still correct, but the references/cli.md docs gap still blocks: exit 20 now also means the run is finished, with a new NOT_RESUMABLE decision line and a terminal JSON field, and none of that is in the diff. Add the short paragraph to the exit-code section or untick Documentation updated in the PR body. Also coordinate with #1217 on the same issue #1197 so only one lands.

abyyxhek:
Closing as superseded — issue #1197 (resume on a terminal run exits non-zero) was fixed on `main` by #1217, merged 2026-09-20. This branch is now 43+ commits behind and conflicts with the merged fix, so I'm dropping it rather than rebasing duplicate work onto a red base. Happy to reopen if #1217 turns out not to cover a case this one did.

PR #1206, opened 2026-09-16
fix(cli): give WAIT and ROLLBACK exit codes distinct from their band-mates

### LINKED ISSUE #1170: cli: recovery modes do not map to distinct exit codes, despite the documented safety contract
## Summary

`README.md:540` states: "Every other mode maps to a distinct non-zero code". The `exitcodes.py` module docstring builds its whole rationale on that promise — "Distinct codes let a script react proportionately (retry a repair, page a human on an unknown side effect)". `_MODE_CODES` does not deliver it:

| Mode | Exit code |
|:--|:--|
| `RESUME` | 0 |
| `REPAIR_AND_RESUME` | 10 |
| `REPLAN` | 10 |
| `WAIT` | 20 |
| `REQUEST_HUMAN` | 20 |
| `ROLLBACK` | 30 |
| `ABORT` | 30 |

Six non-`RESUME` modes collapse onto four codes. A shell consumer cannot do what the docstring tells it to: exit 20 is simultaneously "wait for a condition" and "page a person", and exit 30 is both "roll back" and "abort outright" — the two pairs the docstring names as needing different reactions. The primary guarantee (only a verified-safe run exits 0, so `continuum resume "$RUN" && ./start-agent.sh` cannot launch onto stale state) still holds; this issue is only about the distinctness claim on top of it.

## Affected component

CLI — `src/continuum/cli/exitcodes.py`; the claim is restated in `README.md:540` and in `references/cli.md`s exit-code table.

## Severity

Minor. Automation receives a correct but coarser signal than the one documented, so scripts that try to branch on 20 vs 30 cannot.

## System information

- Python: 3.13.5

```bash
python -c "
from continuum.cli.exitcodes import _MODE_CODES
print({m.name: _MODE_CODES[m] for m in RecoveryMode})
"
# {RESUME: 0, REPAIR_AND_RESUME: 10, REPLAN: 10, WAIT: 20, REQUEST_HUMAN: 20, ROLLBACK: 30, ABORT: 30}
```

```bash
git grep -n "distinct" README.md src/continuum/cli/exitcodes.py references/cli.md
# README.md:540: ... Every other mode maps to a distinct non-zero code ...
# src/continuum/cli/exitcodes.py:13: Distinct codes let a script react proportionately ...
```

## What a fix looks like

1. **Make the codes distinct** — e.g. `WAIT` 25 and `ROLLBACK` 35, keeping the fall-through-to-`UNSAFE` default that makes an unclassified mode fail closed. This delivers the docstrings promise but renumbers a public machine contract: check `references/cli.md`s exit-code table, the dashboard/`serve` surfaces, and any tests pinning 20/30 before renumbering (`#1119` already tracked the table omitting `ExitCode.ERROR`).
2. **Correct the wording** — README:540 and the docstring should say the modes group into four escalating bands (ok / repair / human / unsafe), which is what the code has always done.

Option 1 matches the documented intent; option 2 matches the shipped behaviour.

Cyrax321:
Closes #1170. Takes option 1 from the issue: make the codes distinct rather than rewording the promise away.

`_MODE_CODES` collapsed two pairs onto one code each: `WAIT` and `REQUEST_HUMAN` both exited 20, `ROLLBACK` and `ABORT` both exited 30. Those were exactly the pairs the module docstring names when it says distinct codes let a script react proportionately. A shell consumer could not tell "the run went quiet, poll a clock" from "a person must decide", nor "roll back to a checkpoint" from "give up", from the exit code alone. The primary guarantee is untouched: only a verified-safe run exits 0, and an unclassified mode still falls through to `UNSAFE`.

The change, in bands of ten where the band names the reaction and the offset names the mode:

```
RESUME            0
REPAIR_AND_RESUME 10   } repair band (shared: same reaction)
REPLAN            10   }
REQUEST_HUMAN     20   } human band
WAIT              25   }
ABORT             30   } unsafe band
ROLLBACK          35   }
```

`REQUEST_HUMAN` keeps 20 and `ABORT` keeps 30, so any script that range-checks a band (`20 <= code < 30`) is unaffected. Only the two split-out modes moved. `REPAIR_AND_RESUME` and `REPLAN` still share 10 deliberately: they demand the same reaction, and the issue only flagged the human/unsafe bands. README:540 is reworded to say the codes are grouped into bands with distinct values for the modes a script must tell apart, since "distinct" would otherwise still be false for the repair pair.

One implementation note beyond the table: `cmd_liveness` returned a hard-coded `20` for a liveness breach rather than calling `exit_code_for`. That literal would have silently kept returning 20 for WAIT while the table said 25, so it now routes through `exit_code_for(RecoveryMode.WAIT)` and cannot drift from the table again.

The affected surfaces are updated in step: the `references/cli.md` exit-code table (with a paragraph explaining the band structure), the liveness-watch guide, the `recovery/health.py` docstring, and README:540. The `docs/assets/crash-recovery` transcript and image are deliberately untouched: they show the `REQUEST_HUMAN` refusal path, which still exits 20.

Compatibility: this renumbers a public machine contract. Scripts matching the exact values 20 or 30 to mean "any human-ish" or "any unsafe" mode will now see 25 or 35 for WAIT and ROLLBACK. Those scripts should read the mode from `--json` if they need the mode rather than the reaction. Since WAIT is only two commits old (#1137) and ROLLBACK was not reachable before the recent recovery-engine work, real-world exposure should be minimal.

Verified: full suite in a clean worktree at `cef019d` (Python 3.13.5), 2258 tests collected, 2232 passing, 26 skipped, 0 failed, including three new tests pinning that the split pairs stay distinct and that every mode stays within its reaction band. `ruff check`, `ruff format --check`, and `mypy` are clean on the changed files. A CHANGELOG entry is under Unreleased > Changed.


src/continuum/cli/exitcodes.py
```diff
@@ -46,17 +55,23 @@ class ExitCode:
...
+    ROLLBACK = 35
+    """Not safe as-is; resuming requires rolling back to a prior checkpoint."""
+
 
 _MODE_CODES: dict[RecoveryMode, int] = {
     RecoveryMode.RESUME: ExitCode.OK,
     RecoveryMode.REPAIR_AND_RESUME: ExitCode.REQUIRES_REPAIR,
     RecoveryMode.REPLAN: ExitCode.REQUIRES_REPAIR,
-    RecoveryMode.WAIT: ExitCode.REQUIRES_HUMAN,
+    RecoveryMode.WAIT: ExitCode.WAIT,
```
Cyrax321, line 72: Blocking. This dict is the source of truth, and `docs/guides/enforcement.md:21-26` copies it into a table that still reads `WAIT, REQUEST_HUMAN | 20 | A person must decide` and `ROLLBACK, ABORT | 30 | Not safe at all`. Both rows are now false in the guide a reader reaches from the README line this PR reworded. Anchoring the comment here rather than on enforcement.md because that file has no hunks in this diff, so an inline comment there would be rejected.

docs/guides/liveness-watch.md
```diff
@@ -76,7 +76,7 @@ Liveness: ok, silence 22.8s within threshold 3600s (phase otherwise).
 ```
 
 That sample is a healthy run, which is why it reads `ok` and still exits 0.
-Breached runs report the breach in the same block and exit 20 instead, so a
+Breached runs report the breach in the same block and exit 25 instead, so a
```
Cyrax321, line 79: Blocking. This is the only breach-to-code reference in the file that moved to 25. Lines 42, 65, 89, 103 and 106 still say 20 for the same condition, and line 65 in the section above now directly contradicts this line.

CHANGELOG.md
```diff
@@ -8,6 +8,29 @@ All notable changes to this project are documented here. The format follows
...
+  verified-safe run exits 0, and an unclassified mode still falls through to
+  `UNSAFE`.
+
+  This is a renumbering of a public machine contract. Scripts that range-check a
+  band (`20 <= code < 30`) keep working unchanged; scripts that matched the exact
+  values 20 or 30 to mean any non-human or any unsafe mode will now see 25 or 35
+  for the two split-out modes and should read the mode from `--json` instead. The
+  affected surfaces are updated in step: the `references/cli.md` exit-code table,
+  the liveness-watch guide (a breach exits 25), the `recovery/health.py`
+  docstring, and the literal `20` that `cmd_liveness` returned, which now routes
```
Cyrax321, line 29: Non-blocking. There is no `cmd_liveness` in the tree; the hunk is in `cmd_watch`. The changelog is permanent, so a reader who goes looking for that name finds nothing.

src/continuum/cli/main.py
```diff
@@ -1175,10 +1175,12 @@ def cmd_watch(args: argparse.Namespace, storage: Storage, out: Any, err: Any) ->
...
         palette=getattr(args, "_palette", None),
     )
-    # Exit code: breached maps to WAIT which is REQUIRES_HUMAN (20), otherwise OK
+    # Exit code: a breach proposes WAIT, so route it through exit_code_for
+    # rather than hard-coding the value (#1170: WAIT is now distinct from
+    # REQUEST_HUMAN and the two must not drift apart here).
     if breached:
-        return 20
-    return 0
+        return exit_code_for(RecoveryMode.WAIT)
```
Cyrax321, line 1182: Routing this through `exit_code_for` instead of a literal is the highest-value line in the PR: the breach code now lives in exactly one place.

Cyrax321 (commented):
Requesting changes. The renumbering is sound and band-preserving, and routing `cmd_watch` through `exit_code_for` removes a genuine second source of truth for the breach code. What blocks the merge is that two documentation files still quote the old map, and both now state something false about the exact contract this PR changes. Specifics are on the lines below.

Non-blocking, noted here because neither file has hunks in this diff: the CHANGELOG attributes the hard-coded 20 to a function named `cmd_liveness`, which does not exist in the tree (the hunk is in `cmd_watch`), and `tests/test_rewind_dual_state.py:143` pins the tuple `(ExitCode.OK, 10, 20, 30)`, which no longer admits WAIT (25) or ROLLBACK (35) and will fail the day `validate` proposes either mode after a rewind.

Cyrax321 (commented):

Ref: #1206 / Issue #1170
Surface: CLI exit codes, src/continuum/cli/exitcodes.py

Bug: _MODE_CODES collapsed WAIT onto REQUEST_HUMAN's 20 and ROLLBACK onto ABORT's 30, so a script could not distinguish poll-later from human-decide, or rollback from give up, from the exit code alone, contradicting the module's own stated purpose for distinct codes.

Fix verified at head 656d177d: the bands are sound. REQUEST_HUMAN keeps 20 and ABORT keeps 30, so scripts range-checking a band are unaffected; only WAIT (25) and ROLLBACK (35) moved, and the unmapped-mode fallthrough still lands on UNSAFE rather than OK, which preserves the primary guarantee that only a verified-safe run exits 0. The README reword honestly narrows the claim to modes a script must tell apart, since REPAIR_AND_RESUME and REPLAN deliberately share 10.

Proof: ran tests/test_liveness_events.py and tests/test_cli.py at the head commit: 122 passed, 6 failed. All 6 failures are attest tests raising RuntimeError for the missing optional cryptography package in my local environment, unrelated to this change; the same suite is green in CI on the head commit, which does install the extra.

Cyrax321 (commented):
Maintainer re-review at current head. The WAIT and ROLLBACK exit-code renumbering direction is still sound and band-preserving. Prior review flagged two doc files still quoting the old map plus a CHANGELOG function-name error and a rewind tuple pin; confirm all are resolved on the rebased head with docs and tests green. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review. Rebase out of dirty state first.

PR #1209, opened 2026-09-16
docs: remove non-existent monitored_commands from embed-codex guide

### LINKED ISSUE #1161: docs: embed-codex guide offers a monitored_commands key in .continuum/config that no code reads

wized2:
## Summary
Closes #1161.

`docs/guides/embed-codex.md` suggested `monitored_commands` in `.continuum/config`, which no code reads. Point only at the real `checkpoint_node` adapter path.


docs/guides/embed-codex.md
```diff
@@ -108,7 +108,7 @@ PY
 # or wrap apply_patch in a Bash echo so the hook still sees it
 ```
 
-If you cannot route through Bash, the durability alternative is `continuum observe` via the adapter's `checkpoint_node` or via `monitored_commands` in `.continuum/config` (not covered here). The constraint is Codex's, not CONTINUUM's, so this doc states it explicitly rather than hiding it.
+If you cannot route through Bash, the durability alternative is `continuum observe` via the LangChain/LangGraph adapter's `checkpoint_node` (there is no `.continuum/config` `monitored_commands` key). The constraint is Codex's, not CONTINUUM's, so this doc states it explicitly rather than hiding it.
```
anya-research, line 111: This line redoes work covered by open #1162. Both PRs close #1161 and rewrite the same sentence that offered `monitored_commands` in `.continuum/config`. Keeping both open splits discussion and risks conflicting merges. The older #1162 also names the adapter files and adds a docs guard plus changelog.

anya-research (changes requested):
Thank you for contributing to Continuum, @wized2. This removes the `monitored_commands` fallback from the embed-codex guide for #1161, and there are some changes needed before this can merge.

This duplicates open #1162, which fixes the same line for the same issue with a fuller rewrite plus a regression guard. The overlap is exact on `docs/guides/embed-codex.md:111`, so one of the two should close in favor of the other.

The doc direction here is correct. No `monitored_commands` key exists under `src`, and `checkpoint_node` is real on both adapters.

- Diff is one line in `docs/guides/embed-codex.md:111`
- `grep monitored_commands` finds no hits under `src`, only the denial sentence in the guide
- `checkpoint_node` exists at `src/continuum/adapters/langgraph.py:263` and `src/continuum/adapters/langchain.py:248`
- `git diff --check` is clean
- CI Test matrix is SUCCESS across ubuntu, macos and windows on Python 3.11 through 3.13, plus Markdown lint SUCCESS and Detect docs-only change SUCCESS
- anya-review is FAILURE on empty `GH_TOKEN` in the workflow, unrelated to the doc line

wized2:
Removes the non-existent `monitored_commands` key from embed-codex guide so operators don't configure a dead field.

Cyrax321 (changes requested):
Author context: wized2, external contributor. This duplicates the older #1162 on the same issue #1161 and rewrites the same monitored_commands sentence. Keeping both open splits discussion and risks conflicting merges. Please close as superseded by #1162, which additionally names the adapter files and adds a docs guard plus changelog.

PR #1211, opened 2026-09-16
docs: resume may write webhook notification events

### LINKED ISSUE #1175: cli: `continuum resume` is documented read-only, but a blocked run appends NOTIFICATION_FAILED to the event log

wized2:
## Summary
Closes #1175.

Plain `resume` is still non-mutating for run state/checkpoints, but when webhooks are configured it can append `NOTIFICATION_SENT` / `NOTIFICATION_FAILED` to the event log. Documented that exception in README and `references/cli.md`.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that read-only CLI commands do not mutate run state.
  - Documented that `resume` may append webhook notification events when webhooks are configured, while the recovery decision itself remains non-mutating.


anya-research (approved):
Thank you for contributing to Continuum, @wized2. This corrects the read-only claim for plain `resume` for #1175 to note it may append webhook notification events while the recovery decision stays non-mutating.

The wording matches the implementation. `src/continuum/recovery/webhooks.py:297` and `:304` append `NOTIFICATION_SENT` or `NOTIFICATION_FAILED` with dedup surviving restarts, and `src/continuum/cli/main.py:1487` calls the blocked-run notify before the `--repair` branch at `:1490`, so the write happens on a plain invocation.

- `git diff origin/main` shows 2 files with 5 insertions and 3 deletions, limited to `README.md` and `references/cli.md`
- `git diff --check` is clean
- `gh issue view 1175` is open and describes exactly this blocked-run notification write
- `gh pr list` shows no other open PR covering #1175, with neighboring docs PRs on unrelated counts and modes
- `cmd_watch` at `src/continuum/cli/main.py:1063` already discloses its own conditional breach writes, so keeping this change focused on `resume` is accurate

CI is green on the required gates: all nine Test jobs report SUCCESS plus Markdown lint SUCCESS and docs-only detection SUCCESS, with Lint and Type-check skipped as docs-only. The `anya-review` workflow reports FAILURE on a missing GH_TOKEN in its own automation setup, unrelated to this docs content.

Appreciate the tight scope that narrows the guarantee to run state plus the explicit notification exception.

Cyrax321 (approved):
Reviewed and approved. Good to merge.

PR #1213, opened 2026-09-16
fix: request_human also notifies requires_review webhook endpoints

### LINKED ISSUE #1180: webhooks: the documented `requires_review` event filter can never fire

wized2:
## Summary
Closes #1180.

Production `notify_blocked` always receives a `RecoveryMode` string (`request_human`). Endpoints that only listed `requires_review` never matched.

`WebhookRegistry.for_event("request_human")` now also returns endpoints subscribed to `requires_review`. Added a regression test.

## Summary by CodeRabbit

- **Bug Fixes**
  - Webhook subscribers configured for review-required events now also receive notifications when human intervention is requested.
  - Prevents missed webhook deliveries for this related event scenario.

- **Tests**
  - Added coverage confirming the notification is sent successfully and delivered exactly once.


src/continuum/recovery/webhooks.py
```diff
@@ -93,8 +93,17 @@ class WebhookRegistry:
...
+        """Endpoints whose event filter subscribes to ``event``.
+
+        ``request_human`` also matches endpoints that only listed
+        ``requires_review``: production notify always passes a RecoveryMode
+        string, and ``requires_review`` is not a RecoveryMode member, so
+        without this alias those endpoints would never fire (#1180).
+        """
+        aliases = {event}
+        if event == EVENT_REQUEST_HUMAN:
+            aliases.add(EVENT_REQUIRES_REVIEW)
```
anya-research, line 105: The alias adds EVENT_REQUIRES_REVIEW to every request_human lookup. A requires_review only endpoint then fires for ordinary human gates with no StateStatus.REQUIRES_REVIEW component. The filter condition lives in the validation report, which for_event never sees, so registry level aliasing cannot express the documented rule.

```suggestion
return [e for e in self.endpoints if event in e.events]
```

anya-research (changes requested):
Thank you for contributing to Continuum, @wized2. This makes `request_human` also notify `requires_review` endpoints for #1180, and there are some changes needed before this can merge.

PR #1203 is open on the same issue and overlaps `tests/test_webhooks.py`. That PR wires a second `notify_blocked` call in `cmd_resume` only when the validation report carries `REQUIRES_REVIEW`, plus docs. Only one direction should land, and the two approaches conflict.

The `requires_review` filter is documented as opt in for self certified runs awaiting confirmation (`src/continuum/recovery/webhooks.py:53`, `docs/guides/webhooks.md:31`). The production caller in `src/continuum/cli/main.py:1487` always passes `RecoveryMode.REQUEST_HUMAN`, with no report check.

- CI Test matrix SUCCESS across ubuntu, macos, windows on 3.11 through 3.13 plus Postgres backend, and codecov patch SUCCESS
- CI Lint and Type-check FAILURE, which guards formatting and types, needs a fix first with one file reported unformatted
- `git diff --check` is clean on the head diff

1. `src/continuum/recovery/webhooks.py:105` unconditional alias notifies plain human gates that have nothing to review.

wized2:
Still the right fix for #1180 — production only passes `request_human`. Happy to add a second commit if you want the delivery `event` field to show `requires_review` when that was the subscription match.

wized2:
Closing in favor of the more precise direction in #1203: aliasing every `request_human` to `requires_review` would notify plain human gates incorrectly. The validation-report filter belongs at the call site, not in the event alias map.

PR #1215, opened 2026-09-16
fix(briefing): read the full history so the agent summary survives compaction (#1128)

### LINKED ISSUE #1128: briefing loses the agent summary after compaction: the curated section vanishes and --raw-summary claims none was recorded
## Summary

The agent's reasoning summary disappears from `continuum briefing` on a compacted run, in both of its rendering paths, because each reads the live tail only:

**The curated briefing (the default).** `curate_briefing` (`src/continuum/recovery/briefing_curation.py:129`) includes the agent-summary section only when `_latest_agent_summary` returns a record, and that function scans `storage.read_events(run_id)` for `REASONING_SUMMARY` events (`briefing_curation.py:77`). After `continuum compact` moves the pre-anchor prefix into `events_archive`, the scan is empty, the function returns `None` (`:79-80`), and the section is silently dropped — it does not even appear in the output's `omitted` list, which is only populated when pins fail to hold (`:228-237`). The resumed agent inherits a briefing with no trace that a summary ever existed, which is exactly the context-rehydration loss #235 was about.

**The `--raw-summary` diagnostic (#742).** `cli/main.py:2413` filters `storage.read_events(run_id)` the same way, so the explicit "verbatim, for an operator debugging the curation" path prints `No reasoning summary recorded for {run_id}` on a run whose summary is sitting in the archive — the diagnostic exists to debug cases like this and cannot see them.

Verified end-to-end on a run with one `REASONING_SUMMARY` event (the payload shape `mcp/server.py:833` writes):

```
BEFORE compact: _latest_agent_summary -> ({'status': ..., 'next': ...}, True)   # section included
AFTER  compact: _latest_agent_summary -> None                                   # section vanishes
raw-summary sees live: 0 | sees all: 1
```

Same class as #1050 / #1072 / #1126 — a live-tail read behind a command that must see full history — but a distinct pair of sites.

## Affected component

continuum.recovery (briefing curation); continuum.cli (cmd_briefing --raw-summary)

## Severity

Major functionality broken, workaround exists. The briefing is the primary resume artifact for the next agent session, and compaction — the operation designed for exactly these long, summary-bearing runs (#239) — silently deletes its most context-rich section. Workaround: don't compact runs that need their summary.

## System information

- Python: 3.13.5

```python
import tempfile, os
from types import SimpleNamespace
from continuum.recovery.briefing_curation import _latest_agent_summary

store = SQLiteStorage(os.path.join(tempfile.mkdtemp(), "sum.db"))
store.append_event("run_1", EventType.REASONING_SUMMARY,
    {"summary": {"status": "writing the report", "next": "send for review"}},
    source=Origin.EXTERNAL_AGENT)

decision = SimpleNamespace(validation=SimpleNamespace(report=SimpleNamespace(statuses=[])))
print(_latest_agent_summary(store, "run_1", decision))   # ({...}, True)

store.compact_run("run_1")

print(_latest_agent_summary(store, "run_1", decision))   # None — summary gone
print(len([e for e in store.read_events("run_1")
           if e.type is EventType.REASONING_SUMMARY]))    # 0 (what both readers scan)
print(len([e for e in store.read_all_events("run_1")
           if e.type is EventType.REASONING_SUMMARY]))    # 1 (the fix)
```

## What a fix looks like

Two one-word changes: `read_all_events(run_id)` at `briefing_curation.py:77` and at `cli/main.py:2413`. `read_all_events` (`storage/base.py:135`) folds the archived prefix back in by sequence, so "newest summary" stays correct across compaction. A regression test in the style of the #1050/#1072 ones — record a summary, compact, assert the curated briefing still carries the agent-summary section and `--raw-summary` still prints it — pins both sites.

Adhi1-2:
## Summary

Fixes #1128. The agent's reasoning summary vanished from `continuum briefing`
after compaction, in **both** of its rendering paths, because each scanned the
live tail only.

- **The curated briefing (default).** `curate_briefing` includes the
  agent-summary section only when `_latest_agent_summary` finds a record, and
  that function scanned `storage.read_events(run_id)`. After `compact` moves
  the pre-anchor prefix into `events_archive`, the scan is empty, the function
  returns `None`, and the section is silently dropped. It does not even appear
  in the output's `omitted` list, which is only populated when pins fail to
  hold. The resumed agent inherits a briefing with no trace that a summary
  ever existed.
- **The `--raw-summary` diagnostic.** `cmd_briefing` filtered the same live
  tail, so the explicit verbatim path printed "No reasoning summary recorded"
  on a run whose summary was sitting in the archive. The diagnostic exists to
  debug cases exactly like this and could not see them.

Compaction is the operation designed for long, summary-bearing runs, so the
two live-tail reads failed precisely when they mattered most.

## The fix

Two read sites changed to `read_all_events(run_id)`, which folds the archived
prefix back in, ordered by sequence, so "newest summary" stays correct across
compaction:

- `src/continuum/recovery/briefing_curation.py:77` (the curated section)
- `src/continuum/cli/main.py:2413` (the `--raw-summary` diagnostic)

No behavioural change for un-compacted runs: `read_all_events` returns the same
sequence as `read_events` when the archive is empty.

## Tests

Two regression tests in `tests/test_briefing_curation.py`, following the style
of the existing compaction tests:

- `test_curated_briefing_keeps_the_agent_summary_after_compaction` records a
  summary, compacts, asserts the summary really is archived (live tail empty,
  full history non-empty), then asserts the curated briefing still renders the
  section and its contents.
- `test_raw_summary_diagnostic_survives_compaction` asserts `--raw-summary`
  still prints the verbatim payload after compaction.

Both were verified to **fail on pre-fix source** for the right reasons (the
curated one loses the section; the raw one emits prose instead of JSON), then
pass after the fix. The full suite stays green (2232 passed, 25 skipped), and
`ruff` and `mypy` are clean.

## Scope

This is the same class as #1050 (assess), #1072 (watch), #1126 (pinning) and
#1186 (webhook dedup): a live-tail read behind a surface that must see full
history. It does not touch #1169 (the projection fold reporting 11 of its own
event types as not understood), which is a separate issue in the same
neighbourhood.


Cyrax321 (dismissed):
Thank you for contributing to Continuum, @Adhi1-2. This fixes the vanishing agent summary in #1128 at both rendering paths by switching the two live tail reads to read_all_events, which returns the full history sorted by sequence, so newest stays correct after compaction moves the prefix into the archive.

I reviewed both read sites and the new curation tests, and I verified the helper contract on the storage base: full history including the archived prefix, sorted by sequence. The full CI matrix plus lint is green with no bot findings. No blocking issues found. LGTM, marking this as ready to pull. Thank you again for catching both paths, including the diagnostic that existed to debug exactly this case.

PR #1218, opened 2026-09-16
fix(tui): family_lines fetches the run once instead of twice

### LINKED ISSUE #1157: tui: family_lines calls storage.get_run twice and discards the first result
## Summary

`family_lines` in `src/continuum/tui/model.py:305-308` calls `storage.get_run(run_id)` twice and discards the first result:

```python
def family_lines(storage: Storage, run_id: str) -> list[str]:
    """The `tree` view: the parent verdict and every child's, read-only."""
    storage.get_run(run_id)
    run = storage.get_run(run_id)
```

The first call is the run-existence guard — it raises `RunNotFound` before the view renders a header for a run that isn't there. The second fetches the record actually used. Both hit storage for the same row.

Wasteful rather than wrong, but worth fixing because `get_run` on the SQLite and Postgres backends is a round trip, and this view is one the operator re-renders while watching a run tree. The neighbouring views (`checkpoint_rows`, `action_rows`, `event_rows`, `budget_rows`) each call it exactly once for the same guard purpose, so this one is the outlier.

## Affected component

CLI

## Severity

Minor inconvenience or cosmetic issue — one extra storage round trip per render of the tree view.

## System information

- Python: 3.13.5

```bash
sed -n '305,312p' src/continuum/tui/model.py
```

## What a fix looks like

Delete the first call and keep the assignment, which preserves the guard — `run = storage.get_run(run_id)` raises on a missing run exactly as the standalone call did:

```python
def family_lines(storage: Storage, run_id: str) -> list[str]:
    """The `tree` view: the parent verdict and every child's, read-only."""
    run = storage.get_run(run_id)
    lines = [f"{run_id}  [{run.status.value}]  {run.goal[:60]}"]
```

The existing `RunNotFound` test for the tree view will confirm the guard still fires. A fixer might also check `overview_lines` (`:141`), which relies on `CheckpointManager.restore` to raise rather than calling `get_run` itself — different by design, so no change needed there.

Parthipashok04:
## Description

`family_lines` (the TUI `tree` view) called `storage.get_run(run_id)` twice and discarded the first result:

```python
def family_lines(storage: Storage, run_id: str) -> list[str]:
    """The `tree` view: the parent verdict and every child's, read-only."""
    storage.get_run(run_id)        # existence guard, result discarded
    run = storage.get_run(run_id)  # the record actually rendered
```

The first call was the run-existence guard: it raises `RunNotFound` before the view renders a header for a run that isn't there. The second fetched the record the header actually uses. Both hit storage for the same row, and `get_run` is a round trip on the SQLite and Postgres backends. This view is one an operator re-renders while watching a run tree, so the duplicate is per render.

The fix deletes the standalone call and keeps the assignment, which preserves the guard: `run = storage.get_run(run_id)` raises on a missing run exactly as the standalone call did. No behaviour changes beyond the spared query. The neighbouring views (`checkpoint_rows`, `action_rows`, `event_rows`, `budget_rows`) each call `get_run` exactly once for the same guard purpose, so this removes the outlier rather than introducing a new pattern.

## Related issues

Fixes #1157.

## Types of changes

- Type: `performance`

## Breaking changes

No. The view renders an identical header for an existing run and raises the same `RunNotFound` for a missing one; only the number of storage round trips changed.

## How has this been tested?

- `pytest tests/test_tui.py` (33 passed), including the new `test_family_lines_refuse_a_missing_run`, which asserts the guard still raises `RunNotFound` for a missing run. There was no such test before, and it is what proves the deleted line was redundant rather than load-bearing.
- `pytest tests/test_tui.py tests/test_family.py` (48 passed) to cover the family roll-up paths that read the same run.
- `ruff check src/continuum/tui/model.py tests/test_tui.py`, `ruff format --check` on the same files, and `mypy src/continuum/tui/model.py`: all clean.
- Run on Windows 11 against the project venv (Python from `.venv`).

## Checklist

Tests added or updated for the changed behaviour. Documentation updated where the change affects user-facing behaviour. CI passes on the latest commit. Security-relevant changes state known limitations explicitly in this description (none: the change is read-only and touches no enforcement path).

## Additional context

`cmd_tree` in `src/continuum/cli/main.py` has the same shape (a standalone guard call plus a second `get_run` in the header format). It is a separate implementation from `family_lines` with its own header shape, so it is left alone here rather than folded into this fix.


CHANGELOG.md
```diff
@@ -8,6 +8,19 @@ All notable changes to this project are documented here. The format follows
 
 ### Changed
 
+- **The TUI `tree` view fetches the run once instead of twice (#1157).**
+  `family_lines` in `src/continuum/tui/model.py` called
+  `storage.get_run(run_id)` twice and discarded the first result: the first
+  call was the run-existence guard, the second fetched the record the header
+  actually renders. Both hit storage for the same row, and on the SQLite and
+  Postgres backends that is a round trip on a view an operator re-renders
+  while watching a run tree. The assignment now does both jobs — `run =
```
Cyrax321: Blocking: U+2014 is barred by AGENTS.md in docs. Please use a comma or parentheses and preserve meaning.

```suggestion
  while watching a run tree. The assignment now does both jobs, `run =
```

Cyrax321 (changes requested):
Thank you for the cleanup, @Parthipashok04. The direction is correct and the guard behavior is preserved.

Ref: #1218 / Closes #1157
Surface: TUI tree view (src/continuum/tui/model.py:family_lines)

Author context: Parthipashok04 is an external contributor. Issue reporter Adhi1-2 authored the dashboard commit that introduced the duplicate. No durable negative signal.

Bug: family_lines called storage.get_run(run_id) twice and discarded the first result. Wasteful rather than wrong, one extra round trip per tree render.

Cause: src/continuum/tui/model.py:307-308 on main. Line 307 was a standalone existence guard, line 308 refetched the same row for the header. Confidence: clear. The assignment raises RunNotFound identically, so the standalone call was redundant.

Provenance: introduced by ad3080a (TUI dashboard, Adhi1-2), carried forward unchanged. Current PR author differs from introducer. Confidence: clear via blame and log -S.

Best fix: yes. Delete the standalone call and keep run = storage.get_run(run_id). This matches the neighbouring views (checkpoint_rows, action_rows, event_rows, budget_rows), which each call get_run exactly once for the same guard purpose.

Refactor: no. Leaving cmd_tree in cli/main.py alone is correct. It is a separate implementation with its own header shape, so folding it in would widen scope for no safety gain.

Proof checked in this session:
- Live against local src: main costs 2 get_run calls for an existing run, PR costs 1 call with an identical header. Missing run raises RunNotFound in both shapes.
- Read the full diff (3 files, 22 insertions, 1 deletion), the neighbouring single call views, children_of use, and the new regression test.
- New test test_family_lines_refuse_a_missing_run is at the right seam and proves the deleted line was redundant rather than load bearing.
- CI on f481be9: all Test legs SUCCESS (ubuntu, macos, windows, Postgres), Lint and Type-check SUCCESS, Build and Install SUCCESS, codecov/patch SUCCESS, Gitar SUCCESS. No CodeRabbit or bot inline findings. Ruff check and format clean on the touched Python files. No open milestones.

1. CHANGELOG.md:17 uses U+2014, barred by AGENTS.md in docs. See inline suggestion. Once that line uses a comma or parentheses, this is ready.

Non-blocking notes:
- tests/test_tui.py:158 says The first get_run, which reads stale now that one call does both jobs. Consider rewording to get_run is the guard. No behavior impact.
- The PR body footer and the commit trailer carry AI attribution. AGENTS.md bars AI trailers in commits and code. Please drop those lines on rebase so the commit reads as maintainer written. Not blocking the logic.
- The PR body uses What / The fix / Testing instead of the repo PR template headings. Content covers the needed info, so not blocking, but please map to the template on the next PR.

I am marking REQUEST_CHANGES only for the convention fix. The code change itself is correct and needs no further logic change.

Parthipashok04:
Thanks for the review, @Cyrax321. All three items are addressed in 7337781 (force-pushed, CI re-running):

- **Blocking, em dash:** replaced the U+2014 in the CHANGELOG entry with a colon. Scanned the whole diff for U+2014 afterwards; the only remaining one in CHANGELOG.md is line 125, which is pre-existing text outside this diff.
- **Test docstring:** reworded to say `get_run` is the guard as well as the header's record, since one call now does both.
- **Attribution:** dropped the co-author trailer from the commit, and the PR body now uses the repo template headings (Description / Related issues / Types of changes / Breaking changes / How has this been tested? / Checklist / Additional context) with no AI footer.

The code change itself is untouched: still one `get_run` in `family_lines`, same header for an existing run, same `RunNotFound` for a missing one.

anya-research (approved):
Thank you for contributing to Continuum, @Parthipashok04. This trims `family_lines` to a single `storage.get_run` for #1157, keeping the missing run guard intact.

The change is a one line deletion with matching coverage around it.

- `src/continuum/tui/model.py:307` is now `run = storage.get_run(run_id)` only, with the header line unchanged
- `checkpoint_rows`, `action_rows`, `event_rows`, and `budget_rows` each call `get_run` exactly once for the same guard purpose, so the outlier is gone
- `tests/test_tui.py:157` adds `test_family_lines_refuse_a_missing_run`, which expects `RunNotFound` for a missing run, and its docstring states one call does both jobs
- `CHANGELOG.md` describes the spared round trip with no barred dashes in the diff, and the commit carries no AI trailer
- CI is green across the matrix plus lint and type-check, with all Test legs SUCCESS (ubuntu, macos, windows, Postgres), Lint and Type-check SUCCESS, and codecov/patch SUCCESS

Issue #1157 is still open and no other open PR touches `family_lines`, so there is no overlap to reconcile. Leaving `cmd_tree` alone keeps the scope where it belongs.

Thank you again for proving the guard while removing the extra round trip.

PR #1219, opened 2026-09-16
fix(cli,mcp): read the full history so a compacted run can still record plan and progress (#1133)

### LINKED ISSUE #1133: record-plan and the MCP write path (record_progress, record_plan) refuse every update on a compacted run: the pre-flight fold reads the live tail only
## Summary

After `continuum compact`, every structured write to the run is refused, because the pre-flight fold that gates these writes projects the live tail only:

**The MCP write path.** `_append_projectable` (`src/continuum/mcp/server.py:309`) commits only what the fold accepts (the #364 fix), but `_project_candidate` folds `ctx.storage.read_events(run_id)` — the live tail. On a compacted run that tail starts at the `EVENT_LOG_ANCHORED` marker; `RUN_STARTED` is in the archive, so the fold raises before the candidate is ever considered. Both tools that go through the helper fail:

- `continuum_record_plan` (`server.py:915`) → `run 'run_1' has no goal: the log never recorded RUN_STARTED`
- `continuum_record_progress` (`server.py:724`) → `event event_…: TASK_UPDATED before the run was started`

**The CLI.** `cmd_record_plan` (`src/continuum/cli/main.py:451-466`) runs the identical pre-flight: `history = list(storage.read_events(...))`, then `project(run_id, [*history, candidate])` — so `continuum record-plan` answers `error: plan would leave run unprojectable and was not recorded: run 'run_1' has no goal: the log never recorded RUN_STARTED` on every compacted run. (The post-write display at `:467` projects the same truncated tail.)

Verified end-to-end: a run with a plan that is projectable over `read_all_events` is refused by both folds over `read_events`:

```
compact: {'archived': 1}
CLI  record-plan pre-flight  (live tail): REFUSED -> run 'run_1' has no goal: the log never recorded RUN_STARTED
CLI  record-plan pre-flight  (full log):  projectable, 1 plan unit(s)
MCP  record_progress pre-flight (live):   REFUSED -> event event_0b8e…: TASK_UPDATED before the run was started
```

The run is not unprojectable — `read_all_events` folds it fine, and the recovery surfaces that use checkpoint versions (`resume`, `inspect`) keep working. Only the write path is blind to the archive. The irony writes itself: compaction exists for month-long runs (#239), and the plan/progress records exist so exactly those runs can resume with structure (#312) — after compaction, the agent can no longer record either.

## Affected component

continuum.mcp (write path / _append_projectable); continuum.cli (cmd_record_plan)

## Severity

Major functionality broken, workaround exists. A compacted run cannot record plan or progress updates through any surface. Workaround: none at write time (the events are legal); the run has to be re-recorded elsewhere or the plan kept outside CONTINUUM.

## System information

- Python: 3.13.5

```python
import tempfile, os
from continuum.events import EventType, Origin, Event

store = SQLiteStorage(os.path.join(tempfile.mkdtemp(), "plan.db"))
store.create_run(Run(run_id="run_1", goal="ship the report"))
store.append_event("run_1", EventType.RUN_STARTED, {"goal": "ship the report", "total": 3})
print("compact:", store.compact_run("run_1"))

# the fold both write paths run (mcp/server.py _project_candidate; cli/main.py:466)
payload = {"plan_id": "p1", "units": [
    {"id": "u1", "title": "draft", "status": "pending", "depends_on": []},
]}
history = list(store.read_events("run_1"))
head = history[-1].sequence if history else 0
candidate = Event(run_id="run_1", sequence=head + 1, type=EventType.PLAN_UPSERT,
                  payload=payload, source=Origin.HUMAN)
try:
    project("run_1", [*history, candidate])
except ValueError as exc:
    print("live tail refuses:", exc)          # what record_plan reports
print("full log accepts:",
      len(project("run_1", [*store.read_all_events("run_1"), candidate]).plan), "unit(s)")
```

## What a fix looks like

Fold the full log at both pre-flight sites: `read_all_events(run_id)` in `_project_candidate` (`mcp/server.py`) and in `cmd_record_plan` (`cli/main.py:451`, plus the display fold at `:467`). `read_all_events` (`storage/base.py:135`) merges the archived prefix back in by sequence, so `head` and the fold both stay correct across compaction, and `expected_sequence` still guards the concurrent-write race. A regression test in the style of the #1053/#1072 ones — compact a run, then record a plan unit and a progress tick through the MCP path and assert both commit — pins it.

Adhi1-2:
## Summary

Fixes #1133. A compacted run could not record a plan unit or a progress tick. Both the MCP tools (`record_plan`, `record_progress`) and the `record-plan` CLI command gate the write on a pre-flight fold that projected the **live tail only**. After `compact`, the archived prefix holds `RUN_STARTED` and the earlier plan units, so the fold concluded the run had no goal and refused a write the full log would have accepted. Through every surface, a compacted run was frozen out of its own structure.

## The fix

Three pre-flight reads switched from `read_events` (live tail) to `read_all_events` (archived prefix + live tail, merged in sequence order):

- `mcp/server.py` `_project_candidate`, shared by `record_plan` and `record_progress`
- `mcp/server.py` `ensure_run`, the first-event check
- `cli/main.py` `cmd_record_plan`

`head` is still the true tip, because the merge is sequence-sorted, so `expected_sequence` still guards the concurrent-write race.

## A second, quieter defect

Writing the regression test surfaced something the issue did not report. `ensure_run` read the live tail's first event to decide whether to backfill `RUN_STARTED`. After compaction that read is empty, so it appended a **second** `RUN_STARTED` after the anchor, rewriting the run's history to make the write succeed rather than refusing. The pre-fix test "passed" for the wrong reason, having fabricated a new start for the run.

It now reads the first event of the full merged history and raises `MalformedRunLog` when the log does not begin with `RUN_STARTED`, matching the documented contract instead of silently rewriting it.

## Tests

- `test_record_plan_and_progress_survive_compaction` (test_mcp_server.py): seeds a run, compacts, asserts the live tail lost `RUN_STARTED` while the full history keeps it, records a plan unit and a progress tick, and asserts exactly **one** `RUN_STARTED` in the resulting log. Fails pre-fix with `assert 3 == 1`.
- `test_cli_record_plan_survives_compaction` (test_plan_upsert.py): drives `record-plan` end to end on a compacted run. Fails pre-fix with "has no goal".

Full suite: 2232 passed, 25 skipped. `ruff check`, `ruff format --check`, and `mypy src/continuum` all clean.

## Scope

Deliberately limited to the plan/progress write path named in #1133. Other live-tail reads in `mcp/server.py` (`_declare_model`, `_declare_dependencies`, `continuum_checkpoint`) are separate surfaces; #1169 covers the checkpoint one and is not touched here.


Cyrax321 (dismissed):
Thank you for contributing to Continuum, @Adhi1-2. This unfreezes compacted runs for plan and progress writes in #1133 by switching all three pre-flight folds to read_all_events, and it fixes the second defect the regression test surfaced: ensure_run no longer backfills a duplicate RUN_STARTED after the anchor when the live tail starts there.

I reviewed the shared _project_candidate change covering both MCP tools, the ensure_run first event check, the CLI record plan path, and the regression tests, and I checked CI: the full matrix across all three platforms plus Postgres plus lint is green with no bot findings. The sequence sorted merge keeps head as the true tip so the concurrent write guard is unaffected. No blocking issues found. LGTM, marking this as ready to pull. Thank you again for catching the backfill defect the issue had not reported.

PR #1220, opened 2026-09-16
fix(recovery): detach the runaway worker so the run_with_limits timeout bounds the caller (#1147)

### LINKED ISSUE #1147: recovery(limits): run_with_limits joins the runaway it just timed out, so the timeout bounds nothing

Adhi1-2:
## Summary

Fixes #1147. `run_with_limits(fn, timeout=t)` did not bound anything. It was written as a `with ThreadPoolExecutor(...)` block, and leaving the block runs `shutdown(wait=True)`, which **joins the worker**. `future.result(timeout=...)` raised `TimeoutError` on time, but the caller only saw the `RecoveryTimeoutError` once the probe had run to completion.

The reporter's reproduction: a 6s probe under a 1s timeout returns after 6.0s with the message `recovery operation timed out after 1.0s`. Nothing downstream can distinguish a probe that was stopped from one that completed. A probe written to hang forever (`lock.acquire()` with no timeout, an infinite loop) blocks the caller forever **through the very API that exists to bound it** — the one `docs/threat_model.md` points operators at for the threat model's named "unbounded resource exhaustion by adversarial probes."

## The fix

The executor is managed by hand and detached with `shutdown(wait=False)` in a `finally`, so both the success and the timeout path detach:

```python
executor = concurrent.futures.ThreadPoolExecutor(max_workers=1)
future = executor.submit(fn, *args, **kwargs)
try:
    return future.result(timeout=timeout)
except concurrent.futures.TimeoutError as exc:
    raise RecoveryTimeoutError(...) from exc
finally:
    executor.shutdown(wait=False)
```

The issue's proposed snippet put `shutdown(wait=False)` in both the `except` and the `finally`; the `finally` alone covers both paths with one call, so that is what shipped.

Python cannot kill a thread, so the probe still runs to completion. What changed is that the caller is no longer held hostage to it, which is the only thing the docstring advertises: *"The caller can decide how to handle the timeout."* The module docstring and `docs/threat_model.md` now state that bound explicitly instead of implying the probe is terminated.

## Tests

The existing `test_runaway_op_is_terminated_not_hung` cannot catch this — it only asserts the exception type, which passes today (correctly typed, just late). Two tests time the gap instead:

- `test_timeout_bounds_when_the_caller_regains_control` — a 2.0s probe under a 0.1s timeout must raise in under 1.0s. Fails pre-fix with `caller waited 2.00s for a 0.1s timeout`.
- `test_fast_call_then_slow_call_does_not_join_the_first` — pins that the `finally` detaches too, not just the `except`; a `wait=True` shutdown would still join anything the pool picks up on a second call in the same interpreter.
- `test_timed_call_returns_its_result_when_it_beats_the_deadline` — covers the success path through the new detached executor.

Full suite green (exit 0, ~2258 items). `ruff check`, `ruff format --check` on touched files, and `mypy src/continuum` all clean.

## Scope

`run_with_limits` has no in-tree callers — it is a public export in `continuum.recovery` plus the two doc references. The reconciler path is unaffected (`subprocess.run(..., timeout=...)` at `reconcilers.py:159` terminates correctly).

## Summary by CodeRabbit

* **Bug Fixes**
  * Time-limited operations now return control promptly when the deadline is reached, instead of waiting for an unfinished operation to complete.
  * Operations that finish before the deadline continue to return their results normally.
  * Subsequent time-limited operations are no longer delayed by earlier unfinished operations.

* **Documentation**
  * Clarified that timeouts limit when control returns, but cannot stop an underlying operation already in progress.


Cyrax321 (commented):
Thank you for contributing to Continuum, @Adhi1-2. I compared this against main and the core fix here already landed via Parthipashok04 in commit a92ac528: run_with_limits creates the executor bare and detaches with shutdown(wait=False), with the same TimeoutError to RecoveryTimeoutError translation. The merged version additionally passes cancel_futures=True and declares __all__, so main is slightly ahead of this branch rather than behind it.

The behavioral pins are also present on main: test_timeout_bounds_when_caller_gets_control_back asserts a 3 second probe returns in under 1 second, and test_success_path_does_not_join_worker covers the fast path. Your three tests assert the same properties, so there is no coverage to port.

Please close this PR as superseded. If you see a case in your branch that main does not pin, name the exact test here and it can be added separately. Thank you again for taking this on.

Adhi1-2:
Closing as a duplicate: issue #1147 was closed by merged PR #1151 "fix(recovery): run_with_limits detaches the runaway worker instead of joining it" by @Parthipashok04, which lands the same change this branch carries.

Verified `gh issue view 1147` is CLOSED and `gh pr view 1151` is MERGED, so this branch has no remaining value and is just queue noise. Deleting the branch to keep the fork tidy.

PR #1222, opened 2026-09-16
fix(gate): wire the consumed-authority check through is_authority_consumed (#1154)

### LINKED ISSUE #1154: gate: the authority block in decide() recomputes seq twice and its dict fallback is unreachable dead code
## Summary

The authority-resurrection block in `decide()` (`src/continuum/gate.py:276-298`) contains a sequence of defensive-looking branches that cannot run, plus an exported helper that the block inlines instead of calling:

- **`is_authority_consumed`** (`gate.py:104`) is in `__all__` and documented, but `decide()` never calls it — it inlines `if isinstance(_value, str) and _value in consumed_authorities` at line 278. The helper is dead, and the invariant it was presumably written to centralize (the check lives in one place) is unenforced.
- The `seq` extraction at lines 280-284 computes `getattr(ev, "sequence", "?")` **or** `ev.get("sequence", "?")`, and is then immediately overwritten by the block at 286-288, which does `seq = ev.sequence` unconditionally when `hasattr(ev, "payload")`. The first computation is discarded every time.
- The `else` at line 289 (`payload = ev.get("payload", {})`) is unreachable. `collect_consumed_authorities` (`gate.py:79-101`) stores the **`Event` object itself** as the map value (`consumed[aid] = ev`), and `Event` always has both `.sequence` and `.payload` (`events.py:200-223`). There is no path by which the dict fallback runs.

None of this changes behaviour today — the dead branches happen to compute the same values the live ones do. The cost is that the block reads as though it handles a dict-valued map it can never receive, which matters in a security-sensitive function: a reader cannot tell the defensive code from the load-bearing code, and the next change to `collect_consumed_authorities` will silently change which branch runs.

## Affected component

Other (gate / authority provenance)

## Severity

Minor inconvenience or cosmetic issue — no incorrect behaviour; unreachable branches and an unused exported helper in the authority check.

## System information

- Python: 3.13.5

```python
import inspect
from continuum.gate import decide, collect_consumed_authorities, is_authority_consumed

print("decide() calls the helper?  ", "is_authority_consumed(" in inspect.getsource(decide))
print("decide() inlines the check? ", "in consumed_authorities" in inspect.getsource(decide))
print("map stores the Event?       ", "consumed[aid] = ev" in inspect.getsource(collect_consumed_authorities))

ev = Event(run_id="run_1", sequence=1, type=EventType.ACTION_RECORDED)
print("Event has .sequence/.payload?", hasattr(ev, "sequence") and hasattr(ev, "payload"))
# True — so the dict-fallback branch in decide() can never execute.
```

```bash
git grep -n 'is_authority_consumed' -- src/ tests/
# src/continuum/gate.py:55    <- __all__ entry
# src/continuum/gate.py:104   <- definition
# (no caller anywhere)
```

## What a fix looks like

Two reasonable directions, and the choice is a maintainer's:

1. **Wire the helper.** Replace the inlined membership test at line 278 with `is_authority_consumed(_value, consumed_authorities)`, so the check has one definition and the exported name is honest. Then delete the discarded first `seq` computation and collapse the `hasattr` ladder into a direct `ev.sequence` / `ev.payload` read, asserting or documenting that the map holds `Event` objects.
2. **Remove the helper** if the inlined form is preferred, and delete it from `__all__` rather than leaving an unused export.

Option 1 is preferable — the block already reaches for a defensive style, and the helper is that style's natural home. Whichever is chosen, the extracted `payload` / `seq` logic should match what `collect_consumed_authorities` actually stores, and a test asserting `decide()` denies on a consumed authority would pin the wiring.

Related in the same area, not duplicates: #1074 (the check scans top-level values only, so a nested spent authority passes — a real bypass, not dead code) and #1066 (the block fail-opens on a transient read error).

Cyrax321:
Closes #1154.

`is_authority_consumed` was exported and documented but had no caller: `decide()` and the gateway's `match_route()` each tested membership inline. Both call sites also carried a `hasattr` ladder whose dict fallback could never run, because `collect_consumed_authorities` stores the `Event` itself as the map value and an `Event` always has `.sequence` and `.payload`.

This PR takes the issue's option 1. Both call sites now route through the helper, and both read `ev.sequence` and `ev.payload` directly, with the map-holds-Events contract documented at the helper's one definition. The discarded first `seq` computation and the unreachable `else` are gone.

`gateway.py` carried the same inlined ladder in its mirror of the check, so it is fixed the same way rather than left to drift against `gate.py`.

Behaviour is unchanged. The dead branches happened to compute the same values the live ones did, so no message, verdict, or sequence number moves. What goes away is that a security-sensitive block no longer reads as though it handles a dict-valued map it can never receive, so the next change to `collect_consumed_authorities` cannot silently select a different branch.

Tests: two new cases in `tests/test_gate_resurrection.py` pin the helper's contract (membership, the None and empty-map cases, and that the map value is the `Event` with its payload) and that a spent authority in an unrelated argument field still denies. Existing resurrection, probe, gateway, and archived-history tests pass unchanged. `ruff check`, `ruff format --check`, and `mypy` are clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved authority-consumption checks across tool inputs, including values in fields beyond the primary authority field.
  - Actions referencing an already-consumed authority are now consistently denied with clearer information about the consuming event.

- **Tests**
  - Added coverage for consumed-authority detection, empty or missing authority data, and authorities appearing in unrelated input fields.


Cyrax321 (commented):

Ref: #1222 / Issue #1154
Surface: gate and gateway enforcement seams, consumed-authority check

Bug: is_authority_consumed was exported and documented but called nowhere; decide() and the gateway's match_route() each tested membership inline and carried a hasattr ladder whose dict fallback could never run, because collect_consumed_authorities keys to the AUTHORITY_CONSUMED Event itself, and an Event always has .sequence and .payload.

Fix verified at head 5100db0: both call sites route through the helper, the map-holds-Events contract is documented at its single definition, the gateway mirror is fixed the same way so the two enforcement seams cannot drift, and the unreachable else plus the discarded first seq computation are gone. Behaviour is unchanged, as the PR argues: the dead branches computed the same values the live ones produce, so no message, verdict, or sequence moves. This is the right shape for the fix, deleting the misleading branch instead of keeping a dead compatibility path in a security-sensitive block.

Proof: ran tests/test_gate_resurrection.py and tests/test_gateway.py at head: 38 passed. CI green on the head commit.

Note: adjacent to #1190 (consumed-authority escalation in the engine). They touch different seams of the same subsystem and neither conflicts textually today, but merge them close together and re-run the gate tests after the second lands.

Cyrax321 (commented):
Maintainer re-review at current head. The #1154 consumed-authority wiring is still correct: both enforcement seams route through is_authority_consumed, the dead hasattr fallback is gone, behavior unchanged. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review with no blocking issues. Needs an independent approval to merge. Merge state is dirty, so rebase first.

PR #1223, opened 2026-09-16
fix: TUI and dashboard complete now share the CLI's full verb (#1153)

### LINKED ISSUE #1153: the TUI and dashboard complete_run skip the resume.json cleanup and the REVIEW_CONFIRMED event the CLI verb performs
## Summary

There are three places a human can close a run as completed, and only one of them performs the full verb:

| Surface | `REVIEW_CONFIRMED` | `RUN_COMPLETED` | run row flipped | `.continuum/resume.json` cleared |
|---|---|---|---|---|
| `continuum complete` (`cli/main.py:1770`) | yes | yes | yes | **yes** (`:1813-1820`) |
| TUI `complete_run` (`tui/model.py:395`) | yes | yes | yes | **no** |
| dashboard `complete_run` (`dashboard/hitl.py:82`) | **no** | yes | yes | **no** |

Two separate gaps, both with real consequences:

**1. `resume.json` is left pointing at a completed run.** `_write_resume_json` (`checkpoint/manager.py:386-408`) rewrites `.continuum/resume.json` on every checkpoint, and the CLI's instant-resume fast path (`cli/main.py:2346-2372`, `:2423-2429`) *prefers* whatever `run_id` that file names. The CLI deletes the file when the run it names is the one being completed. The TUI and dashboard do not. So an operator who closes a run from the dashboard and starts a fresh session is resumed straight back into the run they just finished — the exact "finished work hijacks every fresh session" condition `cmd_complete`'s own docstring says the verb exists to prevent.

**2. The dashboard path skips `REVIEW_CONFIRMED`.** This is the event that clears the `self_certified` marker on goal and progress (`recovery/engine.py:258-294`) — the whole point of closing a run *from a human* rather than letting the adapter do it. Its docstring claims "the same pairing as `continuum complete`", which is not true as written. A dashboard-closed externally-driven run can therefore stay self-certified in a way a CLI-closed one does not.

## Affected component

Other (dashboard / TUI HITL surface)

## Severity

Major functionality broken, workaround exists — run `continuum complete <run_id>` from the CLI after closing from the dashboard or TUI, which performs the cleanup both omit.

## System information

- Python: 3.13.5

```bash
git grep -n 'resume.json' -- src/continuum/
# cli/main.py:1814   <- the only place the file is ever deleted
# checkpoint/manager.py:389  <- written on every checkpoint

git grep -n 'REVIEW_CONFIRMED' -- src/continuum/dashboard/hitl.py
# (only the confirm_state verb at :75; complete_run at :82 has none)
```

The behavioural repro needs a session: checkpoint a run so `resume.json` names it, complete it from the dashboard or TUI, then start a new session and observe the resume target.

## What a fix looks like

Extract the tail of `cmd_complete` — the `REVIEW_CONFIRMED` append plus the `resume.json` conditional delete — into a shared helper in `dashboard/hitl.py` or a new `continuum/runs.py`, and have all three call sites use it. The delete must stay conditional on `data.get("run_id") == run_id` exactly as the CLI has it, so closing one run never clobbers another run's resume pointer.

Two smaller decisions a fixer will hit: the dashboard path needs to decide whether the optional `summary` rides along in the `RUN_COMPLETED` payload (it already does there, so only the missing event needs adding), and the bare `except Exception: pass` around the delete should probably stay — a resume file the process cannot read should not block completing a run, which is the current behaviour and correct.

Cyrax321:
## Description

Three surfaces can close a run as completed from a human, `continuum complete`, the TUI's complete verb, and the dashboard HITL button, and only the CLI performed the whole verb. The CLI appends `REVIEW_CONFIRMED` before `RUN_COMPLETED` (both `Origin.HUMAN`), flips the run row to `COMPLETED`, and clears `.continuum/resume.json` when the file names that run. The TUI skipped the file clear, and the dashboard skipped both the clear and the confirmation.

Both omissions had real consequences. `CheckpointManager` rewrites `.continuum/resume.json` on every checkpoint and the SessionStart fast path prefers whatever run it names, so a run closed from the dashboard or the TUI left the file pointing at finished work and the next session resumed straight back into the run just closed, the exact hijack `cmd_complete` exists to prevent. Separately, `REVIEW_CONFIRMED` is the event that clears the `self_certified` marker on goal and progress (`recovery/engine.py`), so a dashboard-closed externally-driven run stayed certified by the agent that did the work in a way a CLI-closed one does not.

The tail of `cmd_complete` is extracted into a new `continuum/runs.py` `close_run()`, and all three surfaces call it with their own `closed_by` label for the `RUN_COMPLETED` payload. The helper is deliberately only the write path: each surface keeps its own already-completed handling and output formatting, so the CLI's idempotency short-circuit and the TUI's message are unchanged. The resume delete stays conditional on the file naming the run being closed, so closing one run never clobbers another run's pointer, and the bare `except Exception` around the read stays: a resume file the process cannot parse must not block completing a run.

`_RESUME_JSON` in `checkpoint/manager.py` becomes public `RESUME_JSON`, since the writer and the clearer now share the one literal and a private name would lie about that.

## Related issues

Fixes #1153

## Types of changes

- Type: `bugfix`

## Breaking changes

No. Each surface lands the same events and row state it did before, plus the two gaps the issue names; the event payloads are unchanged in shape (`closed_by` keeps distinguishing the surface). The one observable difference is that dashboard and TUI completes now also clear the resume file and land the confirmation, which is the point of the issue.

## How has this been tested

New `tests/test_complete_run_parity.py` drives all three surfaces against the same seeded run and pins the shared invariant with `pytest.mark.parametrize` over the CLI, TUI, and dashboard: the last two events are `REVIEW_CONFIRMED` then `RUN_COMPLETED` with `Origin.HUMAN` and `components` `["goal", "progress"]`, the run row is `COMPLETED`, the resume file is gone when it named the closed run, a file naming a different run is left untouched, and `RecoveryEngine.assess` reports goal and progress as `valid` afterwards. It also covers the CLI's already-completed short-circuit (one `RUN_COMPLETED`, no duplicates) and a corrupt resume file not blocking a dashboard close.

`tests/test_dashboard_hitl.py` gains an HTTP-level test that the complete button lands the confirmation before the completion event, so the new event is covered through the endpoint and not just the module.

I verified the tests are load-bearing: stashing only the `src/` changes makes six of them fail (tui and dashboard resume clearing, dashboard self-certification, the unreadable-file case, and the new HTTP assertion), and they pass again once restored.

Commands run, from a worktree on Python 3.13 with `PYTHONPATH` pointed at the worktree `src` (the environment has an editable install pointing at the main checkout):

```
PYTHONPATH=$PWD/src python -m pytest tests/test_complete_run_parity.py tests/test_dashboard_hitl.py tests/test_dashboard.py tests/test_tui.py tests/test_cli.py tests/test_cli_precompact.py tests/test_mcp_instant_resume.py tests/test_checkpoint_manager.py tests/test_docs_counts.py -q -p no:cacheprovider
ruff check src/ tests/
```

All 309 tests pass, lint and format are clean.

## Checklist

Tests added or updated for the changed behaviour. Documentation updated where the change affects user-facing behaviour (CHANGELOG Unreleased/Fixed). CI passes on the latest commit. Security-relevant changes state known limitations explicitly in this description: no new mutation surface is introduced, the dashboard endpoint keeps its existing `CONTINUUM_DASHBOARD_TOKEN` fail-closed gate, and the best-effort resume read remains deliberately non-blocking.

## Additional context

The alternative of putting the helper in `dashboard/hitl.py` was rejected because it would make the CLI and TUI import the dashboard module; `continuum/runs.py` is a neutral home for a tail shared by all three. The dashboard's `confirm_run` remains a separate button with its own `via: dashboard` payload, so this adds no duplicate confirmation for runs that were already confirmed by hand. The documented test counts in README, CONTRIBUTING_ONBOARDING, and CHANGELOG are not re-synced: this adds 12 tests, inside the guard's tolerance of 30, and the three files still agree with each other.


Cyrax321 (commented):
Maintainer review at current head. Two blocking defects verified from the bot threads. 1. Unresolved merge conflict markers are committed into CHANGELOG.md (HEAD, separator, origin/main lines), corrupting the rendered doc. Remove the marker lines and keep both entries. 2. close_run appends REVIEW_CONFIRMED and RUN_COMPLETED before calling get_run, contradicting its own docstring that a missing run raises before anything is written; reorder to fetch and validate first. Also blocking: Lint and Type-check FAILURE. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review. Fix all three and confirm green.

PR #1216, opened 2026-09-16
docs: README narrative test count is ~2,241 (not 1380)

wized2:
## Summary
The closing narrative still said **1380 tests**; collection on main is **~2,241**. Align with the rest of the README.

## Summary by CodeRabbit

- **Documentation**
  - Updated the README to report the current approximate test count: 2,241 tests.

PR #1237, opened 2026-09-17
test: catch documented test counts in narrative text (#1226)

### LINKED ISSUE #1226: docs: README's Origin section says 1380 tests while the same file's guarded figure is 2,241, and the count guard cannot see the claim
## Summary

`README.md:564` closes the Origin narrative with "Validated with real kills and 1380 tests". The same file's Empirical Verification section (`README.md:235`) and Quick Start (`README.md:102`) state ~2,241 collected, the figure `tests/test_docs_counts.py` pins. The translated `README.es.md:516` carries the same 1380 claim.

The guard does not catch the contradiction because all three of its regexes require a leading tilde (`~N collected`, `roughly N tests collected`, `~N tests`), and "and 1380 tests" has none:

```python
>>> [m for rx in _COLLECTED_RES for m in rx.findall("Validated with real kills and 1380 tests")]
[]
```

so `documented_total` reads only 2,241 from README.md and the assertion that the file states exactly one figure passes. Any un-tilde'd count is therefore unchecked, which is how a figure from an early milestone survived ~860 tests of growth unnoticed.

## Affected component

Documentation; Tests.

## Severity

Minor inconvenience or cosmetic issue. A reader who compares the two sections sees a 6x discrepancy and cannot tell which is authoritative.

## System information

- Python: 3.13.5

```bash
grep -n 'tests' README.md | grep -E '1380|2,241'
python -c "from tests.test_docs_counts import documented_total; from pathlib import Path; print(documented_total(Path('README.md')))"
```

## What a fix looks like

Two parts, both small:

1. Replace 1380 with the current collected total in `README.md:564` and `README.es.md:516` (or rephrase to avoid stating a count in the narrative at all, since the guarded figure two sections up already carries it).
2. Widen `_COLLECTED_RES` in `tests/test_docs_counts.py` so a bare "N tests" is also matched, so the next stale un-tilde'd claim fails CI instead of aging silently. Note the existing `~N tests` alternative deliberately requires the tilde to avoid matching unrelated figures, so the new pattern should be anchored to a verb context ("with N tests", "validated ... N tests") rather than matching any digit followed by the word tests anywhere in the file.

Ava-91:
## Description

Updates the documentation test-count guard so it detects bare test counts used in narrative text, such as `with 2,241 tests` and `validated ... 2,241 tests`.

Also synchronizes the current documented collected-test count to ~2,311 across the English and translated READMEs, onboarding/testing/install references, and the current CHANGELOG entry.

The count matcher is intentionally limited to narrative patterns rather than matching arbitrary numbers followed by `tests`.

## Related issues

Fixes #1226

## Types of changes

* Type: `tests`
* Type: `docs`

## Breaking changes

No.

## How has this been tested?

* Updated `tests/test_docs_counts.py` to cover narrative English and Spanish test-count patterns.
* Added regression coverage for the documented narrative count forms.
* Synchronized the current documented collected-test count to ~2,311 across the guarded documentation.
* Ran `python -m pytest tests/test_docs_counts.py -k "not slow" -q` successfully.

## Checklist

* [x] Tests added or updated for the changed behaviour.
* [x] Documentation updated where the change affects user-facing content.
* [ ] CI passes on the latest commit.
* [x] No security-relevant changes.


tests/test_docs_counts.py
```diff
@@ -33,6 +33,7 @@
     re.compile(r"~([\d,]+)`?\s+collected"),
     re.compile(r"roughly\s+([\d,]+)\s+tests\s+collected"),
     re.compile(r"~([\d,]+)\s+tests\b"),
+    re.compile(r"\b(?:with|validated(?:\s+\w+){0,3})\s+([\d,]+)\s+tests\b", re.IGNORECASE),
```
Cyrax321: Blocking: this pattern does not match the strings from #1226, so the guard still misses the reported class.

I checked live against this exact pattern:
- "Validated with real kills and 1380 tests" -> no match
- "Validado con muertes reales y 1380 tests" -> no match
- "Validated with real kills and ~2,241 tests" is caught only by the pre-existing ~N tests pattern, not by this new one.

Two reasons: the anchor vocabulary is English only (with, validated) while one target file is Spanish (Validado, con, y), and the number group needs whitespace directly before digits while the narrative uses a tilde (~2,241). The anchor also sits more than three words from the number in the real sentence.

What exactly to do:
1. Revise this pattern so it matches the exact sentences in #1226, including the tilde form and the Spanish verbs, or add a second Spanish anchored pattern.
2. Add regression asserts in this file for "and 1380 tests", "and ~2,241 tests", and the Spanish equivalents, showing documented_total now sees both figures.
3. Run PYTHONPATH=src pytest tests/test_docs_counts.py -q and confirm.

I am marking this blocking because the core of the issue (fail CI on the next stale bare claim) is not yet delivered.

README.es.md
```diff
@@ -513,7 +513,7 @@ CONTINUUM se sitúa en la intersección de ejecución durable, seguimiento idemp
...
 
-En tres semanas construí CONTINUUM desde un invariante, cada hecho lleva su origen. El resultado es un registro encadenado con `verify()`, un libro mayor con deduplicación por clave estable, una puerta y un gateway que bloquean efectos no reclamados, y un motor de recuperación que sella un contrato. Cinco costuras exponen el mismo registro a Claude Code, LangGraph, LangChain, OpenAI, HTTP y OpenTelemetry. Validado con muertes reales y 1380 tests, imprime `0 duplicados` donde la reproducción ingenua imprime `50`.
+En tres semanas construí CONTINUUM desde un invariante, cada hecho lleva su origen. El resultado es un registro encadenado con `verify()`, un libro mayor con deduplicación por clave estable, una puerta y un gateway que bloquean efectos no reclamados, y un motor de recuperación que sella un contrato. Cinco costuras exponen el mismo registro a Claude Code, LangGraph, LangChain, OpenAI, HTTP y OpenTelemetry. Validado con muertes reales y ~2,241 tests, imprime `0 duplicados` donde la reproducción ingenua imprime `50`.
```
Cyrax321: Blocking: this line leaves the Spanish file internally inconsistent.

Line 516 now reads ~2,241 tests, but lines 99, 232, and 398 still use the 2,195 family. A Spanish reader now sees two totals in one document, and README.es.md is outside COUNTED_FILES so CI cannot flag it.

What exactly to do, pick one:
A. Sync the whole Spanish file in this PR: update every Spanish count figure to the same live total the English files are synced to, with consistent phrasing.
B. Or revert this line and handle the full Spanish sync separately, including the open question of whether to add README.es.md to COUNTED_FILES (that needs Spanish aware patterns for recogidos and cerca de, not just this line).

Please do not merge the half synced state.

Cyrax321 (changes requested):
Thank you for contributing to Continuum. I appreciate the quick pickup of #1226 and the clean two commit split between guard and docs.

This PR adds one narrative pattern to _COLLECTED_RES in tests/test_docs_counts.py and updates the Spanish Origin narrative from 1380 to ~2,241 tests. I reviewed the diff, the linked issue, the COUNTED_FILES guard, both README files on the PR branch, and the existing bot feedback, and I ran the new pattern live against the exact strings from the issue.

Two items block merge. First, the new pattern matches none of the reported bare strings, so the guard still cannot catch the next stale claim; exact failing cases and steps are inline on tests/test_docs_counts.py:36. Second, README.es.md now contradicts itself (2,195 vs 2,241); exact steps are inline on README.es.md:516. The 9 failing Test jobs (live 2273 vs documented 2241) are pre-existing drift, not caused by this PR, but they still gate merge, so the three counted files need a re-sync and the checklist CI claim needs correcting. The English README.md half of #1226 needs no action here since the PR base already contains the #1216 fix.

Thanks again for taking this on. Once the pattern actually catches the issue strings with regression coverage and the Spanish file is fully synced or explicitly deferred, this will be ready to approve.

Ava-91:
Thanks for the detailed review! I’ve addressed the blocking points:

* Updated the narrative-count matcher to cover the exact English and Spanish forms, including `~`-prefixed counts and the longer `Validated ... and N tests` form.
* Added regression coverage for all four reported English/Spanish cases through `documented_total()`.
* Reverted the partial Spanish README count update for now rather than leaving `README.es.md` internally inconsistent. The full Spanish documentation sync can be handled separately, including the question of adding it to `COUNTED_FILES`.
* I also left the existing English README changes in `main` untouched as noted in the review.

The fixes are pushed to the existing PR branch. Thanks again for catching these issues!

tests/test_docs_counts.py
```diff
@@ -22,6 +22,9 @@
     ROOT / "docs" / "CONTRIBUTING_ONBOARDING.md",
     ROOT / "CHANGELOG.md",
 )
+SPANISH_COUNTED_FILES = (
+    ROOT / "README.es.md",
+)
```
Cyrax321, line 36: Blocking: ruff format check fails here (CI Lint and Type-check is red). Please run ruff format so this collapses to a single line.

tests/test_docs_counts.py
```diff
@@ -33,6 +36,15 @@
...
     re.compile(r"~([\d,]+)\s+tests\b"),
+    re.compile(r"\bwith\s+~?([\d,]+)\s+tests\b", re.IGNORECASE),
+    re.compile(
+        r"\bvalidated(?:\s+\w+){0,5}\s+and\s+~?([\d,]+)\s+tests\b",
+        re.IGNORECASE,
+    ),
+    re.compile(
+        r"\bvalidado(?:\s+\w+){0,5}\s+y\s+~?([\d,]+)\s+tests\b",
+        re.IGNORECASE,
+    ),
```
Cyrax321, line 64: Blocking: the new Spanish guard does not see the two canonical Spanish collected forms in README.es.md (the recogidos line 99 form and the cerca de line 232 form). I checked documented_total against the PR head file and only the tilde tests lines match, so test_spanish_documented_counts_agree cannot fail on drift in those lines since SPANISH_COUNTED_FILES has one file. Please either add matchers for those Spanish collected forms with a cross-file assertion, or remove the Spanish guard from this PR and leave it to #1256.

Cyrax321 (changes requested):
Thank you for the follow ups. The narrative patterns now catch all four #1226 strings (verified live) and README.es.md is fully synced to 2,241, which resolves both prior blockers. The tmp_path and with-form coverage fixes are good.

One required item remains before approval, left inline:

- tests/test_docs_counts.py:27: ruff format check fails (CI Lint and Type-check is red). Please run ruff format.
- tests/test_docs_counts.py:47: the new Spanish guard does not see the recogidos and cerca de forms, and the single file agree check cannot fail. Please either add those matchers with a cross-file assertion, or remove the Spanish guard from this PR and leave it to #1256.

Please also correct the CI passes checklist until CI is green. Once that lands and Lint is green, this is ready to approve. This will close #1226 and supersede #1238.

Cyrax321 (changes requested):
Hi Ava, thank you for the rebase in a9838a9. The conflict is resolved correctly, the PR is now test only, Lint is green, and your three added patterns do not break consistency. I checked all 10 stating files with your guard and they agree. The tmp_path fix and the narrowed guard from 5224b73 are good.

The 9 red Test jobs are not caused by your patterns. They all fail on pre existing drift:

suite collects 2311 tests but docs say ~2278

Diff 33 is over tolerance 30. Main moved to 2311 after #1256 synced docs to 2278. To make this green, please sync docs to 2311 in this same PR.

On your branch run:

git checkout fix/1226-test-count-docs
git pull origin fix/1226-test-count-docs

Replace 2,278 with 2,311 in exactly these 9 files, every occurrence:

- README.md (lines 102, 235, 428, 564)
- README.es.md (lines 99, 232, 398, 516)
- README.ja.md
- README.ko.md
- README.pt-BR.md
- README.zh-CN.md
- docs/CONTRIBUTING_ONBOARDING.md (line 31)
- references/testing.md (lines 17, 21)
- references/install.md (line 100)

For CHANGELOG.md update only the line containing ~2,278 collected around line 888. Do not touch the historical note around line 53 that says README.md stated 2,278, that is history.

Do not change ~2,253 passed or ~25 skipped. The guard ignores those. Your synthetic 2,241 cases in test_documented_narrative_count_forms are fine as is, they test the matcher, not the live total.

Verify before pushing:

git status --short
git diff --stat
grep -rn "2,278" README.md README.es.md README.ja.md README.ko.md README.pt-BR.md README.zh-CN.md docs/CONTRIBUTING_ONBOARDING.md references/testing.md references/install.md || echo "no stale 2278, good"
ruff format --check tests/test_docs_counts.py
python -m pytest tests/test_docs_counts.py -k "not slow" -q

The pytest command must pass. Then:

git add README.md README.es.md README.ja.md README.ko.md README.pt-BR.md README.zh-CN.md docs/CONTRIBUTING_ONBOARDING.md references/testing.md references/install.md CHANGELOG.md
git commit -m "docs: sync documented test counts to 2,311 (#1226)"
git push origin fix/1226-test-count-docs

Also update the PR description: remove the Spanish README claim since this is now guard plus re sync, and uncheck CI passes until CI is green. Then request review again. Once CI is green this is ready to approve and will close #1226.

Cyrax321 (approved):
Thank you for contributing to Continuum, @Ava-91. I really appreciate the careful follow through across the rebase, the narrowed guard, the cerca de coverage, and the docs re sync to 2311.

I reviewed the final diff, the guard patterns, the regression coverage, and the required CI results. LGTM, no blocking issues remaining. Marking this as ready to pull.

Thanks again for improving the project.

PR #1238, opened 2026-09-17
docs: README.es Origin section ~2,241 tests (not 1380)

### LINKED ISSUE #1226: docs: README's Origin section says 1380 tests while the same file's guarded figure is 2,241, and the count guard cannot see the claim

wized2:
## Summary
Closes #1226 for the Spanish README.

`README.md` Origin line already uses ~2,241; `README.es.md` still said **1380 tests**. Align both.

## Summary by CodeRabbit

- **Documentation**
  - Updated the Spanish README to reflect the current approximate number of tests: 2,241 instead of 1,380.


README.es.md
```diff
@@ -513,7 +513,7 @@ CONTINUUM se sitúa en la intersección de ejecución durable, seguimiento idemp
...
 
-En tres semanas construí CONTINUUM desde un invariante, cada hecho lleva su origen. El resultado es un registro encadenado con `verify()`, un libro mayor con deduplicación por clave estable, una puerta y un gateway que bloquean efectos no reclamados, y un motor de recuperación que sella un contrato. Cinco costuras exponen el mismo registro a Claude Code, LangGraph, LangChain, OpenAI, HTTP y OpenTelemetry. Validado con muertes reales y 1380 tests, imprime `0 duplicados` donde la reproducción ingenua imprime `50`.
+En tres semanas construí CONTINUUM desde un invariante, cada hecho lleva su origen. El resultado es un registro encadenado con `verify()`, un libro mayor con deduplicación por clave estable, una puerta y un gateway que bloquean efectos no reclamados, y un motor de recuperación que sella un contrato. Cinco costuras exponen el mismo registro a Claude Code, LangGraph, LangChain, OpenAI, HTTP y OpenTelemetry. Validado con muertes reales y ~2,241 tests, imprime `0 duplicados` donde la reproducción ingenua imprime `50`.
```
anya-research, line 516: Line 516 now states ~2,241 tests while lines 232 and 398 in the same file still state 2,195 tests. A reader sees two different totals in one document. The count guard cannot catch this because COUNTED_FILES omits README.es.md.

anya-research (changes requested):
Thank you for contributing to Continuum, @wized2. This updates the Spanish Origin line for #1226 from 1380 to ~2,241 tests, and there are some changes needed before this can merge.

This overlaps open PR #1237, which contains the same `README.es.md` hunk plus a guard change in `tests/test_docs_counts.py`. One of the two should land and the other should close, to keep a single fix for #1226.

The corrected line matches the English Origin line, which states ~2,241 tests. The Spanish file is now internally split, with ~2,195 claims remaining at two earlier spots while the Origin line states ~2,241.

- English README states ~2,241 tests in the Origin narrative
- Spanish README states ~2,241 at the Origin line and ~2,195 at two earlier spots
- The count guard covers `README.md`, `CONTRIBUTING_ONBOARDING.md` and `CHANGELOG.md`, not `README.es.md`
- CI is green across the 9 job test matrix plus Markdown lint and title-check, with no whitespace errors

1. `README.es.md` at line `516` leaves the file self contradictory and unguarded.

Cyrax321 (changes requested):
This PR was correct when it was opened: at your base commit `README.es.md` read "1380 tests" while `README.md` already said 2,241, so the Spanish narrative was genuinely stale. The fix is the right idea.

It is now superseded. Main has since been re-synced, and `README.es.md` on main already reads "~2,452 tests", matching `README.md`. Merging this branch would move the Spanish figure back to 2,241 while the English file stays at 2,452, and `tests/test_docs_counts.py::test_documented_counts_agree` fails on any disagreement between the counted files. The branch is also conflicting against main for the same reason.

Please rebase onto current main. If you do, the diff will most likely collapse to nothing, because the line you came here to fix is already fixed. If it does, the clean thing is to close the PR rather than carry an empty branch. If rebase leaves a real difference, keep the Spanish figure identical to `README.md`'s, whatever it is at that point, and regenerate rather than hardcoding, so the next count bump does not re-break it.

README.es.md
Cyrax321, line 516: Blocking: writing ~2,241 here fails CI and regresses a line that is already correct. tests/test_docs_counts.py treats every translated README as a guarded doc (OPTIONAL_FILES globs README.*.md) and parses exactly this sentence shape: the literal string "Validado con muertes reales y ~2,241 tests" is asserted as a detectable form at tests/test_docs_counts.py:125. With this change README.es.md would state 2241 while README.md, docs/CONTRIBUTING_ONBOARDING.md, CHANGELOG.md and references/testing.md all state ~2,452, and test_documented_counts_agree asserts that every stated collected total is identical, so the guard goes red on merge. Separately, main"s copy of this paragraph already reads ~2,452 tests, so even ignoring the guard this reverts the Spanish README from the current figure to an outdated one.

Cyrax321 (changes requested):
This updates the Spanish Origin paragraph from 1380 to ~2,241 tests, but main has already moved past both figures: on origin/main, README.es.md line 516 already reads "Validado con muertes reales y ~2,452 tests" and README.md states ~2,452 in the same paragraph, so the fix this PR carries has already landed (by a later docs-metrics sync) and the branch now conflicts with main. The premise in the PR description, that README.md"s Origin line already uses ~2,241, is stale; main says ~2,452 there. As written, landing this would also reintroduce drift and fail the docs-count guard, details inline. Requesting changes; the actionable path is to close this PR, or retarget it at ~2,452 (or whatever pytest --collect-only currently reports) if any straggler remains after a fresh sweep.

Issue #1101, opened 2026-09-15
the build-your-own-dashboard guide reads decision.human_steps, which does not exist on RecoveryDecision: every copy-paste example raises AttributeError


Adhi1-2:
## Summary

`docs/guides/bring-your-own-dashboard.md` reads `decision.human_steps` in three places — `:34`, `:196`, and `:207` (the last two inside the "drop in and adapt" Flask app and the `/api/run/<run_id>/resume` handler). But `RecoveryDecision` (`src/continuum/recovery/engine.py:90`) is a `@dataclass(frozen=True, slots=True)` whose fields are `run_id, mode, contract, plan, state, validation, environment_diff, restored, rationale, safe, next_allowed_action, uncertain_actions, permits, impacted_files, informed_retry, tail_evidence` — there is no `human_steps` field and no `__getattr__` fallback, so every copy-paste example in the guide raises `AttributeError` at runtime.

The real API is `continuum.recovery.guidance.human_steps_for(decision, run_id=...)` (`src/continuum/recovery/guidance.py:91`), which is exactly how the CLI (`src/continuum/cli/main.py:903-911`) and the sidecar (`src/continuum/serve/server.py:639`) build the field.

Corroborating drift in the same family: `docs/recipes/control-plane.md:20` tells a dashboard author to read `contract.human_steps` — but `human_steps` is a **sibling** of `contract` in the `--json resume` payload (`cli/main.py:1446`), not a child of it. `RecoveryContract` has no `human_steps` field either.

## Affected component

Documentation (guides and recipes)

## Severity

Minor inconvenience — but every code sample in the guide is broken as written, so a contributor's first run fails.

## System information

- OS: macOS 26.0.1 (Darwin 25.0.0)
- Python: 3.13.5
- continuum-agent: from source, commit `579376d` on `upstream/main`
- Installed via: `uv pip install -e ".[dev]"`

## Command or code to reproduce

```bash
python - <<'EOF'
from continuum.recovery.engine import RecoveryDecision
print('human_steps' in RecoveryDecision.__dataclass_fields__)   # False
EOF

git grep -n 'decision.human_steps' -- docs/
# docs/guides/bring-your-own-dashboard.md:34
# docs/guides/bring-your-own-dashboard.md:196
# docs/guides/bring-your-own-dashboard.md:207
```

## What a fix looks like

[...]

PR #1239, opened 2026-09-17
docs: human_steps is computed via human_steps_for, not on RecoveryContract

### LINKED ISSUE #1234: docs: human_steps is a computed top-level key, not an attribute of RecoveryDecision or RecoveryContract, and five doc sites teach the broken access
## Summary

`human_steps` is not a field of either model. `RecoveryDecision` (`src/continuum/recovery/engine.py:101`) carries `run_id`, `mode`, `contract`, `plan`, `validation`, `restored`, `uncertain_actions`, `rationale`, `impacted_files`, `tail_evidence` and `informed_retry`. `RecoveryContract` has `post_checkpoint_observations` but no `human_steps`. The steps are computed per call by `human_steps_for` in `src/continuum/recovery/guidance.py:91`, and the CLI (`cli/main.py:1030`, `:1454`, `:2466`, `:2513`) and MCP server (`mcp/server.py:1091`) splice the result into their output as a top-level key.

Five doc sites still teach readers to reach it through the object or the contract, which raises `AttributeError`:

- `docs/guides/bring-your-own-dashboard.md:34` — `print(decision.human_steps)` on a `RecoveryEngine(store).assess(...)` result
- `docs/guides/bring-your-own-dashboard.md:196` — `human_steps=decision.human_steps` in a Flask `render_template_string` call
- `docs/guides/bring-your-own-dashboard.md:207` — `"human_steps": decision.human_steps` in an API payload builder
- `docs/recipes/control-plane.md:20` — lists `contract.human_steps` among "key fields for a dashboard", so a control plane doing `c["contract"]["human_steps"]` gets a `KeyError` while `c["human_steps"]` works
- `docs/release-notes.md:28` — describes `RecoveryContract` as carrying `evidence` / `reason` / `next_allowed_action` / `human_steps`

#1159 and #1101 were closed against the first three of these, but the sites are still in the file, so the fix landed somewhere else or was partial. The JSON blocks in the same files (`bring-your-own-dashboard.md:55`, `embed-claude-code.md:89`, `embed-codex.md:90`, `api/serve-http.md:291`) are correct, because they show the top-level key of the serialised output, not an attribute of the engine object.

## Affected component

Documentation; Core (state, checkpoints, recovery).

## Severity

Major functionality broken, workaround exists, for a reader who copy-pastes the guide: the dashboard recipe and the control-plane recipe both fail on their first hit of the field.

## System information

- Python: 3.13.5

```bash
python -c "
import sys; sys.path.insert(0,'src')
from continuum.recovery.engine import RecoveryDecision
print('human_steps' in RecoveryDecision.__annotations__)
"
grep -n 'human_steps' docs/guides/bring-your-own-dashboard.md docs/recipes/control-plane.md docs/release-notes.md
```

## What a fix looks like

In `bring-your-own-dashboard.md`, the three accesses should go through the same helper the CLI and MCP server use (`from continuum.recovery.guidance import human_steps_for; human_steps_for(decision, run_id)`) so the sample keeps working when the guidance rules change, rather than switching to a hardcoded list. In `control-plane.md:20`, move `human_steps` out of the `contract.*` bullet to the top-level field list where `mode`, `safe` and `informed_retry` already sit. In `release-notes.md:28`, drop `human_steps` from the list of `RecoveryContract` fields or reword it to say the verdict carries executable human steps alongside the sealed contract. Worth adding a note in one of the guides that the key is computed per call from the plan and reconciler registry rather than stored, since that is why it is not on the model.

wized2:
## Summary
Closes #1234.

Clarify in release notes and control-plane recipe that `human_steps` is produced by `human_steps_for(...)` / CLI JSON top-level key, not `RecoveryContract.human_steps`.


Cyrax321 (approved):

This PR corrects the last two sites from #1234: control-plane.md now places human_steps as a top level JSON key from human_steps_for instead of contract.human_steps, and release-notes.md no longer lists human_steps as a RecoveryContract field. I verified against the code (no human_steps on RecoveryContract or RecoveryDecision; human_steps_for in recovery/guidance.py is the producer) and grepped the docs on the PR branch: no decision.human_steps or contract.human_steps accesses remain, and the three dashboard guide sites are already on human_steps_for via #1101. CI is green.

One non blocking follow-up worth a separate issue: README.md:167 and README.es.md:164 still describe the sealed contract as carrying human_steps, which teaches the same model though it breaks no code. Rewording those two table cells the same way would finish the cleanup.

No blocking issues. Ready to approve. Thank you again.

PR #1240, opened 2026-09-17
fix(gateway): reject a colon that shifts the memory key's segments (#1149)

### LINKED ISSUE #1149: gateway: a colon in store_id shifts the memory key segments and defeats the bound-tenant check
## Summary

The memory key template is `mem:{store_id}:{tenant}:{record_key}`, and `match_route` (`src/continuum/gateway.py:227-244`) enforces the bound tenant by splitting the *rendered* key on `:` and comparing `parts[2]` to the bound tenant. Nothing validates that `store_id` and `tenant` themselves contain no colon, so a caller who controls one field can shift the segments until `parts[2]` lands on whatever value the check wants to see, while the real `tenant` field is something else entirely.

The confusing part: the values are still interpolated into the rendered key by name, so the key is well-formed and looks honest. Only the positional check is fooled.

## Affected component

Other / not sure (gateway, memory-store identity)

## Severity

Major functionality broken, workaround exists — the guarantee the bound-tenant check exists to make (one tenant cannot reach another tenant's records through a shared store) does not hold for store ids containing a colon.

## System information

- Python: 3.13.5

```python
from continuum.gate import render_key

honest = render_key("mem:{store_id}:{tenant}:{record_key}",
                    {"store_id": "pg1", "tenant": "acme", "record_key": "inv-42"})
attack = render_key("mem:{store_id}:{tenant}:{record_key}",
                    {"store_id": "pg1:acme", "tenant": "other", "record_key": "inv-42"})
print(honest)  # mem:pg1:acme:inv-42
print(attack)  # mem:pg1:acme:other:inv-42

def tenant_of(rendered):
    parts = rendered.split(":")
    return parts[2] if len(parts) >= 4 else None

print(tenant_of(honest))  # acme
print(tenant_of(attack))  # acme   <- passes the bound_tenant == "acme" check
```

The attacker's real `tenant` is `other`. `match_route` extracts `parts[2]` = `acme` and compares it against the bound tenant `acme`, so the call is allowed and the write lands under a different tenant than the one the route was bound to.

## What a fix looks like

Pick one and apply it at the boundary, not at the read site:

1. **Validate at write time.** Reject any placeholder value containing ":" in `render_key` / `normalize_key_value` before interpolation, so a key can never carry ambiguous segments. This is the fix that makes the positional check sound, because the count of colons is then fixed.
2. **Stop splitting positionally.** Have `render_key` return the rendered key *and* the resolved field values, and let `match_route` compare the structured `tenant` field rather than a segment cut out of the flattened string. This removes the whole class of confusion, not just the colon case.
3. At minimum, extend `load_gate_config` to reject a template whose non-`mem:` fields can produce a colon, and to require that memory templates name exactly three placeholders so the segment positions are provable.

Option 2 is the durable one; option 1 is the smaller one. Note this is the same confusion family as #1051 (route prefix parsed but never enforced) and #1074 (authority check scans top-level values only) — each is a different instance of the gateway trusting a flattened string it later re-parses.

abyyxhek:
## What

`match_route` enforces the bound tenant by splitting the *rendered* memory key on `:` and comparing `parts[2]`. Nothing checked that `store_id` or `tenant` themselves carried no colon, so a caller who controls one field could shift the segments until `parts[2]` read as the bound tenant while the real `tenant` field was someone else's. The key stayed well-formed and looked honest — only the positional check was fooled.

Closes #1149.

## The attack

```python
honest = render_key("mem:{store_id}:{tenant}:{record_key}",
                    {"store_id": "pg1", "tenant": "acme", "record_key": "inv-42"})
attack = render_key("mem:{store_id}:{tenant}:{record_key}",
                    {"store_id": "pg1:acme", "tenant": "other", "record_key": "inv-42"})
# honest -> mem:pg1:acme:inv-42        parts[2] == "acme"
# attack -> mem:pg1:acme:other:inv-42  parts[2] == "acme"  <- passes bound_tenant == "acme"
```

The attacker's real tenant is `other`. `match_route` extracts `parts[2]` = `acme`, compares it against the bound tenant `acme`, allows the call, and the write lands under a different tenant than the route was bound to.

## The fix

Reject any placeholder value containing `:` at the render boundary, but **only for memory templates**, where the segments are colon-delimited. This is option 1 from the issue, scoped so a non-memory template that legitimately embeds a colon in a value keeps working.

Both `render_key` implementations are covered — the hook's (`continuum.gate`) and the proxy's (`continuum.gateway`) — because the two seams must derive the same key or a call claimed through one looks unclaimed at the other. `match_route`'s existing `GatewayConfigError` handler turns the rejection into a fail-closed denial, so the check is enforced at the write boundary rather than the read site.

## Verification

- The issue's reproduction now raises instead of emitting an ambiguous key; the honest key still renders.
- New test `test_gateway_denies_a_colon_that_shifts_the_tenant_segment` in `tests/test_memory_forensic.py` covers both the render-boundary rejection and the end-to-end `match_route` denial.
- `tests/test_memory_forensic.py`, `tests/test_cli_gate.py`, `tests/test_gateway.py`: 61 passed.
- `ruff check` / `ruff format --check` clean on all touched files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Memory key templates now reject string values containing colons when rendering keys.
  - Requests that could bypass tenant validation by shifting key segments are denied with an appropriate configuration or routing error.
  - Added regression coverage for key-rendering failures and route denial behavior.


Cyrax321 (changes requested):
Thank you for contributing to Continuum, @abyyxhek. This closes the tenant shifting hole in #1149 at both render seams: gate.render_key and gateway.render_key now reject a colon in any placeholder before the terminal segment of a memory template, so a crafted store_id can no longer push parts[2] onto the bound tenant while the real tenant differs. Guarding the formatted value rather than only strings, and leaving the terminal record key free for values like doc:section:1, are both the right calls.

I reviewed both seams and the forensic tests, and the CI on this head is fully green across the matrix plus lint. Two items block merge, both from main moving underneath this branch, and the branch now conflicts:

1. Please drop .github/workflows/anya-review.yml from this PR entirely. The fork stand down for missing review tokens just landed via #1266, so that hunk now duplicates and conflicts with main. Take the main side for that file on rebase.
2. Resolve the remaining conflicts (doc counts from the older base) and confirm Lint and the Test matrix are green on the rebased head before re-requesting review.

Once conflicts are resolved and CI is green this will be ready for a final review. Thanks again for fixing both seams instead of one.

Cyrax321 (changes requested):
Author context: abyyxhek, external contributor. Follow-up to the prior review. The #1149 colon-segment fix at both render seams is still correct, but the rebase blockers still stand: drop .github/workflows/anya-review.yml and take the main side, resolve the remaining doc-count conflicts, and confirm Lint plus the Test matrix green on the rebased head.

src/continuum/gateway.py
```diff
@@ -159,14 +159,39 @@ def render_key(template: str, body: dict[str, Any]) -> str:
...
         raise GatewayConfigError(f"key template {template!r} needs body field(s) {missing}")
-    return template.format(**{f: normalize_key_value(body[f]) for f in fields})
+    values = {f: normalize_key_value(body[f]) for f in fields}
+    if is_memory_template(template):
+        # Only the placeholders before the terminal segment can shift the
+        # positions, and the check reads the formatted value rather than only
+        # ``str`` ones: ``str.format`` renders a non-string such as
+        # ``["x:acme:"]`` as ``['x:acme:']``, which carries the same shifting
+        # colon without ever being a string.
+        guarded = set(fields[:-1])
```
Cyrax321, line 191: Non-blocking. This guard is now part of the identity rule for memory keys, and it exists twice: here and verbatim in gate.py's render_key. Both modules already argue that the two seams must derive the same key or a call claimed at one looks unclaimed at the other, and the same drift argument applies to this check: a colon accepted at one seam but rejected at the other reopens #1149 asymmetrically, with no test that would necessarily catch the skew. Consider a single helper in continuum.gate, which already owns is_memory_template and normalize_key_value, that takes the exception class to raise so the rule can only ever change in one place.

Cyrax321 (commented):
Rejecting colons in non-terminal memory placeholders closes the #1149 segment shift at both render seams, and I verified the fail-closed mapping holds on each: gate.decide catches GateConfigError and denies, match_route catches GatewayConfigError and denies, so the rejection can never surface as an unhandled raise. The terminal-segment exemption is sound: parts[2] cannot be moved by a colon in the last placeholder, and cmd_tombstone re-parses the remainder as ":".join(parts[3:]) (cli/main.py), so keys like doc:section:1 keep working; I also confirmed the #1319 reviewer's conclusion, since this PR touches only render_key and not match_route's route selection, the prefix scoping in #1319 composes with this change rather than interacting. Nothing blocks merge.

Two non-blocking notes. The guard is duplicated verbatim across gate.py and gateway.py (inline). And the new restriction that store_id and tenant must not contain ':' is not documented in docs/guides/memory_governance.md or docs/multi_agent_isolation.md, both of which describe the mem:{store_id}:{tenant}:{record_key} convention; a one-line note there would save an operator from learning the rule through a denial message.

PR #1241, opened 2026-09-17
fix(serve): validate record_progress against the fold before committing (#1221)

### LINKED ISSUE #1221: serve: record_progress commits the event before checking the fold, so one bad update poisons the run permanently (the #364 ordering the MCP server already enforces)
### Summary

`continuum serve`'s `record_progress` commits the event to the log *before* it checks whether the log still projects. The MCP server was hardened against exactly this ordering (`_append_projectable`, the #364 fix), so the two transports that expose the same tool now disagree: MCP refuses a bad update and records nothing; the sidecar records it, then notices.

A caller can therefore permanently poison a run's event log with a single progress update that omits `total`.

### Expected behaviour

The write path rejects what the read path rejects, *before* committing — the invariant `_append_projectable` exists to enforce (`src/continuum/mcp/server.py:309`). A refused call writes nothing.

### Actual behaviour

The event is appended unconditionally (`src/continuum/serve/server.py:690`), then folded with `on_unprojectable="degrade"` (`:695`). The response even admits the break — it carries a `projection_failed_at` block — but the event is already durable by then, and no later event can correct it, because the fold validates every intermediate state.

The pre-write guards in `_h_record_progress` (`:681-684`) only catch the case where `total` arrives in the *same* payload:

```python
if total is not None and completed + failed > total:
    raise BadParams(...)
```

They are blind to the `total` already on record, which is the ordinary case — an agent reports `total` once and then sends `completed`/`failed` updates. `total` is only known to this handler when the caller passes it; the fold is the only check that sees the accumulated figures. That is precisely the gap #364 closed on the MCP side.

### Steps to reproduce

```python
import tempfile, os
from continuum.serve.server import SidecarServer

srv = SidecarServer(storage=SQLiteStorage(os.path.join(tempfile.mkdtemp(), "t.db")))

srv.dispatch("record_progress", {"run_id": "r1", "completed": 0, "total": 5, "goal": "g"})
# second call omits total; completed(4) + failed(4) > the recorded total of 5
r = srv.dispatch("record_progress", {"run_id": "r1", "completed": 4, "failed": 4})
print("acknowledged the break:", "projection_failed_at" in r)   # True

print("event is durable:", srv.storage.read_events("r1")[-1].payload)
# {'completed': 4, 'failed': 4}

project("r1", srv.storage.read_events("r1"))   # raises ValidationError
```

### Root cause

`_h_record_progress` appends first and validates after, instead of validating the candidate state before committing. The MCP server's `_project_candidate`/`_append_projectable` pair does the opposite and states why in its docstring: an event the fold refuses, once durable, leaves every projecting surface for that run dead permanently, while the action tools keep working — so the run can still authorise real side effects while recovery is unable to say whether continuing is safe.

The sidecar never got that fix. The same bad payload over MCP is refused with nothing written:

```
MCP  _append_projectable -> ValueError: TASK_UPDATED {'completed': 4, 'failed': 4}
     would leave run 'r1' unprojectable and was not recorded: ...
     (log length after: 3, unchanged)
```

### Impact

The run is unrecoverable from that point on, and the damage is silent at first:

- `checkpoint` on the poisoned run dies outright (`ValidationError`).
- `validate` and `resume` degrade to `mode=request_human`, `safe=False`, still reporting the **pre-poison** figures (`completed=0, total=5`) with no indication the log is broken.
- The strict fold the CLI uses for `status`, `inspect`, `replay`, `show-contract` and `briefing` raises on the same log.
- `intercept_action` and the other action tools still work, so the run can keep authorising external side effects while recovery cannot verify anything about it — the inversion the #364 fix exists to prevent.

The one existing test for this (`tests/test_serve.py:52` `test_record_progress_rejects_over_total`) covers only the same-call case, so the regression is invisible to the suite.

Severity is major rather than blocking: it affects one run, not the database, and it needs a caller to report counters that contradict a `total` it did not restate. There is no workaround for the affected run — the event cannot be removed without editing the log.

### Evidence

```
first (total=5): 0 5 src_seq 3
second response: {'run_id': 'r1', 'completed': 0, 'pending': 5, 'failed': 0, 'total': 5,
  'source_sequence': 3, 'projection_failed_at': {'sequence': 4, 'type': 'TASK_UPDATED',
  'reason': 'Value value, completed + pending + failed exceeds total'}}
log length: 4  last: TASK_UPDATED {'completed': 4, 'failed': 4}

checkpoint: DEAD -> ValidationError: completed + pending + failed exceeds total
resume mode: request_human  safe: False
resume progress: {'completed': 0, 'pending': 5, 'failed': 0, 'total': 5}   # stale
CLI status/inspect fold: DEAD
```

Full test suite at `b81db85`: 0 failures (baseline clean; this is a code-behaviour finding, not a failing test).

### Suggested solution

Route the sidecar's write path through the same validate-then-commit sequence the MCP server uses, rather than its own append-then-project one. `_append_projectable` is already transport-independent — it takes a storage handle and writes with `expected_sequence` — so `_h_record_progress` can call it (or a shared helper lifted out of `continuum.mcp.server`) and drop the redundant same-call `total` check, since the candidate fold subsumes it. This also gains the optimistic-concurrency guard the sidecar currently lacks, and would close the last surface where the two transports can disagree about what a legal update is.

### System information

- OS: Windows 11 Home (10.0.26200)
- Python: 3.11.15
- continuum-agent: 0.1.2, from source at `b81db85` on `main`
- Installed via: editable install (`uv pip install -e .`)

### Affected component

MCP server — specifically `continuum.serve`, the sidecar that mirrors the MCP tool surface over stdio/HTTP.

### Severity

Major functionality broken, workaround exists (for the affected run there is no workaround; classified major rather than blocking because the database and other runs are unaffected).

### Regression check

- [ ] This worked in an earlier version or commit — No: the sidecar has appended before projecting since the handler was written; the MCP side was fixed in #364 and this handler predates or missed that hardening.

abyyxhek:
## What

`continuum serve`'s `record_progress` appended the event to the log *before* it checked whether the log still projects. The MCP server was hardened against exactly this ordering (`_append_projectable`, the #364 fix), so the two transports that expose the same tool disagreed: MCP refused a bad update and recorded nothing; the sidecar recorded it, then noticed.

Closes #1221.

## The cost

A caller could permanently poison a run's event log with one progress update. Because the fold validates every intermediate state, no later event could correct it, and every projecting surface for that run stayed dead — `status`, `resume`, `briefing`, `replay` — while the action tools kept working. That inversion (side effects still authorised, recovery unable to vouch for them) is what makes an unprojectable log worse than a rejected call.

The pre-write guards in `_h_record_progress` only catch the case where `total` arrives in the *same* payload. They are blind to the `total` already on record, which is the ordinary case — an agent reports `total` once and then sends `completed`/`failed` updates.

## The fix

Ports the MCP's `_append_projectable` shape to the sidecar as `SidecarServer._append_projectable`:

1. Fold the log with the candidate event **in memory**.
2. Refuse with `bad_params` if the fold rejects it — nothing is recorded.
3. Only then commit, passing `expected_sequence` so a second writer that advances the run between validation and append triggers a re-validation rather than composing two individually-legal payloads into a log neither would have been allowed to produce. Bounded retries, so a permanently busy run answers instead of spinning.

The candidate is folded in memory rather than written-and-rolled-back: no transaction spans the append, and a failed rollback would leave behind precisely the state this prevents.

The existing pre-write guards are kept — they are blind to the recorded `total`, but they give the common single-call typo its own clear message. Since a refused call now writes nothing, the response no longer carries the `projection_failed_at` degrade block, which existed only to describe a break the caller could no longer fix.

## Verification

The issue's reproduction, updated:

```
first ok: completed=0 total=5 pending=5
second refused: BadParams | TASK_UPDATED {'completed': 4, 'failed': 4} would leave run 'r1' unprojectable...
event count: 2          # RUN_STARTED + the first update only
run still projects: YES
legal update accepted: completed=3 failed=1 total=5 pending=1 seq=3
```

- New tests in `tests/test_serve.py`: the refused update records nothing and leaves the run projectable; an accepted update reports the folded figures.
- `tests/test_serve.py`, `tests/test_serve_http.py`, `tests/test_mcp_sidecar_provenance.py`: 91 passed.
- `ruff check` / `ruff format --check` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Progress updates that would make a run inconsistent are now rejected before being saved.
  - Valid progress updates return accurate completion, failure, pending, total, and source-sequence values.
  - Concurrent progress changes are handled more reliably, reducing the risk of conflicting updates.


Cyrax321 (changes requested):
Thank you for contributing to Continuum, @abyyxhek. This ports the MCP _append_projectable shape to the sidecar for #1221 so record_progress folds the candidate in memory and refuses unprojectable updates before anything is durable, with bounded revalidation on write races instead of failing a valid update that lost a race. The design matches the hardened MCP path including the poisoned log analysis.

I reviewed the new helper, the handler wiring, and the new serve tests. The fix itself looks right. Two items block merge, both from main moving underneath this branch:

1. This branch conflicts with main. Part of it is the .github/workflows/anya-review.yml hunk: the fork stand down for missing review tokens just landed via #1266, so please drop that file from this PR entirely and take the main side on rebase. The remainder is stale doc counts from the older base.
2. There is no fresh CI signal on a merged head. After rebasing, please confirm Lint and the Test matrix are green and re-request review.

Once conflicts are resolved and CI is green this will be ready for a final review. Thanks again for closing the transport gap.

Cyrax321 (changes requested):
Author context: abyyxhek, external contributor. Follow-up to the prior review. The #1221 sidecar validation direction is still correct, but the rebase blockers still stand: drop .github/workflows/anya-review.yml from this PR and take the main side (the fork stand down landed via #1266), resolve the remaining conflicts, and confirm Lint plus the Test matrix green on the rebased head.

src/continuum/serve/server.py
```diff
@@ -282,6 +283,89 @@ def _ensure_run(self, run_id: str, goal: str | None = None) -> Run:
     def _ledger(self, run_id: str) -> ActionLedger:
         return ActionLedger(self.storage, run_id, source=AGENT_SOURCE)
 
+    def _append_projectable(
```
Cyrax321, line 286: Non-blocking: this duplicates ContinuumMCP._append_projectable nearly verbatim, including the retry loop and both error strings. Issue #1221 exists precisely because #364 fixed the ordering on one transport and the other drifted, so a shared helper (taking storage plus source, living somewhere both transports import) would keep the next fix from landing on only one side again. Fine to land as-is for parity now, but worth a follow-up before a third consumer copies it.

src/continuum/serve/server.py
```diff
@@ -282,6 +283,89 @@ def _ensure_run(self, run_id: str, goal: str | None = None) -> Run:
...
+        would have been allowed to produce (for example ``total=50`` landing
+        between the read and the write of a ``completed=75`` that omits
+        ``total``). Losing that race says nothing about whether this update is
+        valid, so it re-validates against the new head rather than failing;
+        bounded, so a permanently busy run answers instead of spinning.
+        """
+        from continuum.storage import ConcurrentWriteError
+
+        for remaining in range(attempts - 1, -1, -1):
+            history = list(self.storage.read_events(run_id))
```
Cyrax321, line 325: Non-blocking: read_events returns only the live tail, so on a compacted run this fold raises ProjectionError ("the log never recorded RUN_STARTED") no matter how legal the candidate is, and the refusal message names the caller's payload as the cause. CheckpointManager.project_current defaults to read_all_events for exactly this reason (issue #648). Today _ensure_run raises MalformedRunLog first, so this call site is masked, but the helper is shaped transport-generic and MCP's twin already surfaces the misleading message on compacted runs. Folding read_all_events here would make record_progress actually work on compacted runs instead of merely failing more honestly.

Cyrax321 (commented):
This ports the MCP transport's validate-then-commit ordering (#364) to the sidecar's record_progress, closing the #1221 hole where a refused update was already durable. The read-modify-write gap between the in-memory fold and the append is closed properly with expected_sequence plus bounded re-validation, the error translation to bad_params/400 is correct, and the new tests cover refusal, acceptance, and both race outcomes. Nothing here blocks merge; three non-blocking notes inline.

src/continuum/serve/server.py
```diff
@@ -282,6 +283,89 @@ def _ensure_run(self, run_id: str, goal: str | None = None) -> Run:
...
+                    run_id,
+                    event_type,
+                    payload,
+                    expected_sequence=head,
+                    source=AGENT_SOURCE,
+                )
+            except ConcurrentWriteError:
+                if remaining:
+                    continue
+                raise BadParams(
```
Cyrax321, line 357: Non-blocking: losing the optimistic-concurrency race three times is transient server-side contention, not a malformed request, but BadParams maps to bad_params on the wire and HTTP 400 in serve-http, so a client reading the status will conclude its payload was wrong rather than retry. The message does say to check for a competing owner, which helps, but a distinct SidecarError code (or 409 in the HTTP mapping) would let clients distinguish retryable contention from a genuine bad_params. MCP has the same classification, so this is a parity note rather than a new problem.

PR #1242, opened 2026-09-17
docs(adapters): correct the three constructor signatures the framework adapters reject (#1230)

### LINKED ISSUE #1230: docs: adapters.md documents an engine= keyword that the LangGraph, OpenAI and LangChain adapters reject
## Summary

`docs/api/adapters.md` teaches one constructor shape for every adapter:

```
`continuum.adapters.GenericAgentAdapter(storage, *, engine=None)`      :16
`continuum.adapters.LangGraphAgentAdapter(storage, *, engine=None)`    :54
`continuum.adapters.OpenAIAgentAdapter(storage, *, engine=None)`       :67
`continuum.adapters.LangChainAgentAdapter(storage, *, engine=None)`    :80
`continuum.adapters.BrowserAdapter(storage, *, engine=None)`           :87
`continuum.adapters.ContainerAdapter(storage, image, *, engine=None)` :117
`continuum.adapters.KubernetesAdapter(storage, *, namespace="default", engine=None)`  :148
`continuum.adapters.FilesystemSandboxAdapter(storage, sandbox_dir, *, engine=None)`    :179
`continuum.adapters.python_inproc.PythonInProcAdapter(storage, workdir, *, engine=None)` :211
```

The Generic, Browser, Container, Kubernetes, Filesystem and PythonInProc entries are correct: those constructors do accept `engine`. The three framework adapters do not. Their `__init__` signatures:

- `src/continuum/adapters/langgraph.py:125` — `(storage, *, graph=None, state_to_semantic=None, auto_file=None, auto_total=None)`
- `src/continuum/adapters/openai.py:146` — same minus `graph`
- `src/continuum/adapters/langchain.py:121` — same minus `graph`

So `LangGraphAgentAdapter(storage, engine=my_engine)` raises `TypeError: __init__() got an unexpected keyword argument 'engine'`, while the same file's Generic entry suggests it should work. The three adapters still reach an engine, but they build it themselves via `super().__init__(storage, auto_file=..., auto_total=...)` and read `self.engine` afterwards, so an operator cannot inject one.

The same three entries also omit `graph` and `state_to_semantic`, which are the primary customization points on the LangGraph adapter and the only way to point it at a compiled graph.

## Affected component

Adapters (LangGraph, OpenAI Agents, LangChain); Documentation.

## Severity

Major functionality broken, workaround exists, for a reader following the docs: the documented call fails and the error names a keyword the doc just told them to pass.

## System information

- Python: 3.13.5

```bash
python -c "
import inspect, sys; sys.path.insert(0, 'src')
from continuum.adapters import langgraph
print(inspect.signature(langgraph.LangGraphAgentAdapter.__init__))
"
```

## What a fix looks like

Correct the three entries to the real signatures, including `graph` and `state_to_semantic` for LangGraph, and add a sentence that these three build their own `RecoveryEngine` internally so an injected engine is not among their options (that is the fact of the code today; whether it should change is a separate design question). No example in `examples/` passes `engine=` to any of the three, so this is docs-only.

abyyxhek:
## What

`docs/api/adapters.md` taught one constructor shape per adapter. Three of them documented an `engine=` keyword the classes do not accept, so copying the documented line raised `TypeError` naming a keyword the page had just told the reader to pass.

Closes #1230.

## The three wrong lines

| Documented | Real signature |
|---|---|
| `LangGraphAgentAdapter(storage, *, engine=None)` | `(storage, *, graph=None, state_to_semantic=None, auto_file=None, auto_total=None)` |
| `OpenAIAgentAdapter(storage, *, engine=None)` | `(storage, *, state_to_semantic=None, auto_file=None, auto_total=None)` |
| `LangChainAgentAdapter(storage, *, engine=None)` | `(storage, *, state_to_semantic=None, auto_file=None, auto_total=None)` |

The three adapters still reach an engine — they build one themselves via `super().__init__(storage, auto_file=..., auto_total=...)` and read `self.engine` afterwards — so an operator cannot inject one. The page now says so instead of offering a keyword that fails.

The same three entries also omitted `graph` and `state_to_semantic`, the primary customization points on the LangGraph adapter and the only way to point it at a compiled graph.

`GenericAgentAdapter`'s line was not wrong but was incomplete; its real keyword-only parameters are listed now. The remaining five entries (Browser, Container, Kubernetes, FilesystemSandbox, PythonInProc) already matched the classes and are unchanged.

## Guard

Adds `tests/test_adapter_docs.py` so the drift cannot recur silently: it parses each documented constructor line, resolves the class, and compares the positional and keyword-only parameter lists against `inspect.signature`. A second test fails if an exported adapter loses its constructor line entirely. Same shape as `test_mcp_docs`, which audits the MCP tool table for the same reason — the page `__all__`-exports these classes, so the signature is part of the promised surface rather than an implementation detail.

I verified the guard has teeth by reverting one line to the old `engine=` form and confirming it fails.

## Verification

- `tests/test_adapter_docs.py`: 10 passed.
- `tests/test_mcp_docs.py`, `tests/test_cli_docs.py`: pass.

## Note on a pre-existing red gate

`tests/test_docs_counts.py::test_documented_count_matches_suite` fails on `main` **before** this change: the suite collects 2275 tests while the docs say ~2241 (off by 34, tolerance 30). My new tests add 10, taking it to 2285. I did not re-sync the count because #1226 has that work in flight (PRs #1237, #1238), and changing the same figures there would conflict.

## Summary by CodeRabbit

- **Documentation**
  - Updated adapter documentation to describe automatic progress tracking during capture and resume.
  - Documented optional progress file and total settings for supported adapters.
  - Clarified recovery engine configuration and state translation options for Generic Agent, LangGraph, OpenAI, and LangChain adapters.
  - Updated constructor references, including adapter-specific graph and state translation options.

- **Tests**
  - Added checks to ensure documented adapter constructors and supported adapter coverage stay aligned with the available interfaces.


Cyrax321 (changes requested):
Thank you for contributing to Continuum, @abyyxhek. This corrects the three constructor lines in docs/api/adapters.md for #1230 that documented an engine keyword the classes do not accept, so a reader copying the documented line hit a TypeError.

I verified all three real signatures directly: LangGraphAgentAdapter takes graph, state_to_semantic, auto_file, and auto_total; the OpenAI and LangChain adapters take state_to_semantic, auto_file, and auto_total; none accepts engine. The corrected page matches the code, and Lint plus Markdown lint pass. One item blocks merge and it is unrelated to the fix:

1. Stale documented test counts (tests/test_docs_counts.py). All 9 Test jobs fail: the suite collects 2,320 but the docs state about 2,278. Please run pytest --collect-only -q, take the live total, and re-sync every stating file (README.md, the translated READMEs, docs/CONTRIBUTING_ONBOARDING.md, the current CHANGELOG entry, references/testing.md, references/install.md), leaving historical CHANGELOG notes untouched. Then run python -m pytest tests/test_docs_counts.py -k "not slow" -q locally to confirm green before pushing.

Once CI is green this will be ready for a final review. Thanks again for the exact signature table.

abyyxhek:
Thanks for the review, @Cyrax321. Pushed `ba5ac80` with the two test findings addressed, and one clarification on the counts.

**1. CodeRabbit nitpick — added.** `test_the_page_yields_parsable_constructor_lines` now asserts the page parses into cases, so a reformat that matches no `SIGNATURE_LINE` fails the suite instead of reporting the parametrized guard as *skipped*.

**2. gitar finding — real, fixed.** `test_the_page_documents_every_public_adapter` enumerated with `vars(package)`, but `continuum.adapters` resolves Browser/Container/Kubernetes/LangChain/LangGraph/OpenAI through PEP 562 `__getattr__` and caches them in `globals()` only after first access. `vars()` saw just the two eagerly imported adapters, so the guard excused six of the eight. It now enumerates `__all__` and resolves each name with `getattr`, the same path a reader's `from continuum.adapters import BrowserAdapter` takes — all eight adapters are covered and all eight have a constructor line.

**3. Stale documented counts — already clear on this head.** This branch is rebased on `8ef2474`, whose README states 2,311. `pytest --collect-only` here collects 2,329, and `test_documented_count_matches_suite` (tolerance 30) passes, as do the agreement and required-files tests. The 2,278 figure you saw was the older base; no file needs re-syncing now.

**Remaining red Test job is pre-existing on `main` and untouched by this PR.** `tests/test_adapters_openai.py::test_wrap_function_tool_invocation_binds_args_and_intercepts` fails because openai-agents 0.22.3 reads `ToolContext._function_tool_arguments`, which the `FakeTC` in that file never defined. This PR changes only `docs/api/adapters.md` and `tests/test_adapter_docs.py`; the failure reproduces on `main` and is covered by #1286 / #1295.

Cyrax321 (changes requested):
Author context: abyyxhek, external contributor. Follow-up to the prior review. The three #1230 constructor corrections are still verified against the real signatures, but the stale docs-count blocker still stands. Re-run pytest --collect-only on a rebased tree and resync every stating file, then confirm the Test matrix green.

docs/api/adapters.md
```diff
@@ -13,11 +13,17 @@ adapter = GenericAgentAdapter(SQLiteStorage("continuum.db"))
...
-`continuum.adapters.GenericAgentAdapter(storage, *, engine=None)`
+`continuum.adapters.GenericAgentAdapter(storage, *, engine=None, auto_file=None, auto_total=None)`
 
 The concrete adapter for standard Python agent loops. Construct it with a
 `Storage` implementation; it owns a `CheckpointManager` and a `RecoveryEngine`.
 
+`engine` injects a pre-built `RecoveryEngine`; omit it and the adapter
+constructs one from the storage. `auto_file` and `auto_total` together turn on
+automatic progress recording: when both are set, the adapter records file
+progress on every `capture_state` and `resume` rather than asking the loop to
```
Cyrax321, line 24: Blocking: `resume` does not record file progress. `GenericAgentAdapter.resume` in src/continuum/adapters/generic.py is a short delegate to `self.engine.assess(...)`, and `assess` performs no progress recording anywhere in its body. The auto path fires in exactly two places: `capture_state` (guarded by `self.auto_file is not None and self.auto_total is not None`) and `_auto_progress`, which runs at the end of a fresh `intercept_action`. Telling the reader progress refreshes on resume sets the expectation that resuming a crashed loop re-derives file progress, and it never does.

Same sentence: the backticked `record_progress` is not a callable on any documented surface. The public names are the MCP tool `continuum_record_progress` and `continuum.hooks.record_file_progress`, so a reader searching for `record_progress` finds nothing. Suggested rewrite: when both are set, the adapter records file progress on every `capture_state` and after every intercepted action, rather than asking the loop to record progress itself.

Cyrax321 (changes requested):
Corrects the three framework adapter constructor lines in docs/api/adapters.md and adds a signature-guard test. All four rewritten signatures (GenericAgentAdapter, LangGraphAgentAdapter, OpenAIAgentAdapter, LangChainAgentAdapter) were verified against the classes on main and match exactly, and CI is green.

One blocking factual error in the new GenericAgentAdapter prose stands between this and merge: auto progress does not record on resume, and the named fallback callable does not exist. Details inline.

Non-blocking: sibling PR #1331 fixes the same three signature lines for #1230 (reviewed separately, also verified against main), so whichever lands second will need a rebase or close.

abyyxhek:
Closing as a duplicate — #1230 is also covered by #1331 (both edit docs/api/adapters.md), and I'm keeping the single non-duplicate PR for the issue rather than asking the maintainer to arbitrate between two. Mine additionally added tests/test_adapter_docs.py as a guard; if that guard is wanted, it can be lifted from this branch or I can reopen on request.

PR #1243, opened 2026-09-17
docs: name the three environment variables no document mentioned (#1229)

### LINKED ISSUE #1229: docs: CONTINUUM_MCP_SLIM silently deletes 9 of the 12 MCP tools and no document mentions it
## Summary

`src/continuum/mcp/server.py:1560` shrinks the server's tool surface when the variable is set:

```python
if os.environ.get("CONTINUUM_MCP_SLIM") == "1":
    keep = {"continuum_resume", "continuum_validate", "continuum_list_actions"}
    for name in list(server._tool_manager._tools.keys()):
        if name not in keep:
            del server._tool_manager._tools[name]
```

Because this runs inside `build_server`, every caller path that constructs a server honours it, not just the CLI entry point. The `## Environment variables` section of `docs/api/mcp.md:102` documents `CONTINUUM_MCP_ALLOW`, `CONTINUUM_MCP_MUTATING_CLIENTS` and `CONTINUUM_DB`, and does not mention it. In a deny-by-default project where the documented tool count (12, `README.md:362`) is part of the security posture, a variable that quietly removes 9 mutating tools belongs in that list, with a note on what a slim server can and cannot do.

Two smaller gaps of the same kind, worth folding into the same fix:

- `CONTINUUM_SIGNER_KEY` (`src/continuum/cli/main.py:3586`, error message at `:3588`, and the `--key` help at `:4208`) is the attestation key path override, documented only inside `--help` text. `references/attestation.md` is the natural home.
- `CONTINUUM_WEBHOOK_SECRET` (`src/continuum/recovery/notify.py:24`, `SECRET_ENV_VAR`) is the HMAC secret for webhook signatures. `docs/guides/webhooks.md:81` describes the signature scheme but never names the variable that supplies the key.

## Affected component

MCP server; Documentation.

## Severity

Minor inconvenience or cosmetic issue. Behaviour is correct; the gap is that an operator cannot discover the switch from the docs.

## System information

- Python: 3.13.5

```bash
grep -rn 'CONTINUUM_MCP_SLIM\|CONTINUUM_SIGNER_KEY\|CONTINUUM_WEBHOOK_SECRET' docs/ references/ README.md   # no hits
grep -rn 'CONTINUUM_MCP_SLIM' src/    # src/continuum/mcp/server.py:1560
```

## What a fix looks like

Add a `CONTINUUM_MCP_SLIM` entry to the `## Environment variables` list in `docs/api/mcp.md` stating that it reduces the surface to the three read-only-ish resume, validate and list_actions tools, and that it is off by default. Then document `CONTINUUM_SIGNER_KEY` in `references/attestation.md` and `CONTINUUM_WEBHOOK_SECRET` in `docs/guides/webhooks.md`. If the slim mode is intended as an internal or unsupported escape hatch rather than a supported configuration, the alternative fix is to say so explicitly at `server.py:1560` in a comment, so a reader does not assume it is load-bearing for a documented deployment shape.

abyyxhek:
## What

Three environment variables that change what a deployment exposes or verifies were discoverable only by reading source. In a deny-by-default project where the documented tool count (12) is part of the security posture, that is a real gap.

Closes #1229.

## The three variables

**`CONTINUUM_MCP_SLIM`** — checked inside `build_server`, so every caller path that constructs a server honours it, not just the CLI entry point. Set to `1` it removes nine of the twelve tools — every mutating one — leaving `continuum_resume`, `continuum_validate` and `continuum_list_actions`. I verified this against the live server:

```
slim tools: 3 ['continuum_list_actions', 'continuum_resume', 'continuum_validate']
```

Documented in the `## Environment variables` section of `docs/api/mcp.md`, with the note that it is not a security boundary on its own: it shrinks what a caller may ask for, but the allowlist is still what decides who may ask.

**`CONTINUUM_SIGNER_KEY`** — the attestation signing key when `--key` is absent; without either, the command refuses with "no signing key" rather than looking for a default. **`CONTINUUM_SIGNER`** — the signer name embedded in the document; nothing derives authority from it, so a verifier that cares about identity must still check `public_key`. Both were documented only inside `--help` text; added to `references/attestation.md` as a new `### Environment variables` subsection.

**`CONTINUUM_WEBHOOK_SECRET`** — the HMAC secret webhook signatures are keyed on. `docs/guides/webhooks.md` described the signature scheme in detail, including the constant-time comparison a receiver should use, without ever naming the variable that supplies the key. Added it, with the precedence (per-endpoint `secret` in the registry wins; the env var is the default for endpoints that declare none) confirmed against `src/continuum/recovery/webhooks.py:145` and `notify.py:59`.

## Verification

- `tests/test_mcp_docs.py`, `tests/test_cli_docs.py`: pass.
- Slim surface verified against the live server (3 read tools, as above).
- Documentation-only change; no behaviour touched.


anya-research (approved):
Thank you for contributing to Continuum, @abyyxhek. This names the three environment variables no document mentioned for #1229.

The three additions state the implementation exactly. `CONTINUUM_MCP_SLIM` in `docs/api/mcp.md` matches `src/continuum/mcp/server.py:1560`, twelve tools with the nine mutating ones removed and `continuum_resume`, `continuum_validate` and `continuum_list_actions` kept, checked inside `build_server` with the allowlist caveat intact. The attestation subsection matches `src/continuum/cli/main.py:3586`, flag wins over env, and refusal with "no signing key" when both are absent. The webhook precedence matches `src/continuum/recovery/notify.py:59` with per-endpoint `secret` winning and the env var as default, and the absent header behavior matches the unsigned path.

- Docs-only change, 36 additions across 3 files, `git diff --check` clean
- No duplicate open PR covers this, #1229 stays open, nearest docs PR touches different files
- CI is green across the matrix plus Markdown lint, title check, and docs-only detection

No blocking issues found. This is ready to merge. The care with precedence and boundary notes makes these docs trustworthy.

Issue #1118, opened 2026-09-15
after #1064 (benchmarks/), scripts/ and demo-run/ are the last Python directories outside every lint gate, carrying 9 findings and formatting drift


Adhi1-2:
## Summary

Follow-up to #1064 (benchmarks/ excluded from every lint gate): the repo's lint scope is `src/`, `tests/`, and `examples/` everywhere —

- `.github/workflows/ci.yml:136` — `ruff check src/ tests/ examples/`
- `.github/workflows/ci.yml:139` — `ruff format --check src/ tests/ examples/`
- `.pre-commit-config.yaml:27` and `:29` — `files: ^(src|tests|examples)/`

— which leaves exactly two Python directories ungated: `scripts/` (3 files) and `demo-run/` (2 files). Running the repo-pinned `ruff==0.16.5` against them turns up 9 errors plus formatting drift:

```
demo-run/generate_crash_visual.py:1:1:   EXE001  Shebang is present but file is not executable
demo-run/generate_crash_visual.py:53:12: PLW1510 `subprocess.run` without explicit `check` argument
demo-run/generate_crash_visual.py:62:12: PLW1510 `subprocess.run` without explicit `check` argument
demo-run/generate_crash_visual.py:106:14: PLW1510 `subprocess.run` without explicit `check` argument
demo-run/generate_crash_visual.py:116:14: PLW1510 `subprocess.run` without explicit `check` argument
demo-run/worker.py:1:1:                  I001  [*] Import block is un-sorted or un-formatted
demo-run/worker.py:13:8:                 BLE001 Do not catch blind exception: `Exception`
scripts/generate_crash_visual.py:1:1:    EXE001 Shebang is present but file is not executable
scripts/mcp_smoke.py:1:1:                EXE001 Shebang is present but file is not executable
```

and `ruff format --check scripts demo-run` reports 3 of the 4 files would be reformatted.

These files are user-facing entry points, not scratch: `scripts/generate_crash_visual.py` exists so both documented invocations in `README.md` and `docs/release-notes.md` work, and `scripts/mcp_smoke.py` is the live MCP smoke test. With #1064 covering `benchmarks/`, gating these two completes the sweep — every Python file in the repo would sit behind a lint gate.

## Affected component

CI; Other (lint scope)

## Severity

Minor inconvenience or cosmetic issue.

## System information

- OS: macOS 26.0.1 (Darwin 25.0.0)
- Python: 3.13.5
[...]

Issue #1064, opened 2026-09-15
benchmarks/ is imported by tests but excluded from every lint gate, carrying 9 findings


Cyrax321:
## Summary

Three directories are outside every lint gate: `benchmarks/`, `_bugaudit/`, and `demo-run/`. CI runs `ruff check src/ tests/ examples/` and the pre-commit hooks scope to `^(src|tests|examples)/`, both by design and both documented as such. The gap is cosmetic for `_bugaudit/` and `demo-run/`, but `benchmarks/` is imported by the test suite, so the code is shipping under test while carrying nine lint findings that would fail CI in any of the three linted trees.

## Affected component

CI and tooling (lint coverage), benchmarks

## Severity

Minor inconvenience or cosmetic issue for the library. No behaviour is wrong; the risk is that the excluded trees drift further from the style the rest of the repo enforces, and that a lint-level bug in a benchmark path goes unflagged.

## Evidence

`.github/workflows/ci.yml:136` runs `ruff check src/ tests/ examples/` and `:139` runs `ruff format --check src/ tests/ examples/`. `.pre-commit-config.yaml:21-27` scopes both hooks to `^(src|tests|examples)/`, and `:13-15` states that the three excluded directories "are not ruff-clean and are not checked by CI". Tests reach into `benchmarks/` at `tests/test_risk_bench.py:7`, `tests/test_benchmark_counters.py:120`, `tests/test_dropped_constraint_reconstruction.py:178-179`, `tests/test_bench_unsafe_edit.py:169-170`, and `tests/test_horizon_benchmark.py`.

## Command or code to reproduce

```bash
uv run ruff check benchmarks/ _bugaudit/ demo-run/
```

Output at HEAD `14f8a84`, 9 errors, 6 fixable with `--fix`:

```
_bugaudit/check_env.py:1:1: I001 [*] Import block is un-sorted or un-formatted
_bugaudit/repro_import_check.py:1:1: I001 [*] Import block is un-sorted or un-formatted
benchmarks/fault_injection/runner.py:355:11: SIM115 Use a context manager for opening files
benchmarks/fault_injection/runner.py:361:9: I001 [*] Import block is un-sorted or un-formatted
benchmarks/fault_injection/runner.py:410:17: SIM102 Use a single `if` statement instead of nested `if` statements
benchmarks/fault_injection/runner.py:487:13: I001 [*] Import block is un-sorted or un-formatted
[...]

PR #1245, opened 2026-09-17
ci: bring _bugaudit/ inside the ruff lint gates (#1232)

### LINKED ISSUE #1232: ci: _bugaudit/ is the last tracked Python directory outside every lint gate, with 2 findings and formatting drift
## Summary

Follow-up to #1064 (benchmarks/) and #1118 (scripts/, demo-run/): both of those issues describe their target as completing the sweep, but a third tracked directory is still ungated. `_bugaudit/` ships two Python files on `main` and sits outside the ruff scope in CI and pre-commit:

- `.github/workflows/ci.yml:136` — `ruff check src/ tests/ examples/`
- `.github/workflows/ci.yml:139` — `ruff format --check src/ tests/ examples/`
- `.pre-commit-config.yaml:27` and `:29` — `files: ^(src|tests|examples)/`

`.pre-commit-config.yaml:13` documents the exclusion in plain terms: "demo-run/ and `_bugaudit/` are not ruff-clean and are not checked by CI". The first half of that sentence is what #1118 fixes, which makes the second half stale rather than descriptive. Against the repo-pinned `ruff==0.16.5`:

```
_bugaudit/check_env.py:1:1      I001  [*] Import block is un-sorted or un-formatted
_bugaudit/repro_import_check.py:1:1  I001  [*] Import block is un-sorted or un-formatted
Found 2 errors. [*] 2 fixable with the `--fix` option.
2 files would be reformatted
```

## Affected component

CI; Other (lint scope).

## Severity

## System information

- Python: 3.13.5
- Installed via: `uv pip install -e ".[dev]"` (ruff 0.16.5, the `pyproject.toml:38` pin)

```bash
git ls-tree -r main --name-only | grep '\.py$' | grep -vE '^(src|tests|examples|benchmarks|scripts|demo-run)/'   # _bugaudit/check_env.py, _bugaudit/repro_import_check.py
ruff check _bugaudit            # 2 errors (verified with ruff 0.16.5)
ruff format --check _bugaudit   # 2 files would be reformatted
```

## What a fix looks like

The maintainers should pick one of two directions, and either is fine; the issue is that the current state is neither:

1. Complete the sweep the same way #1118 does it: extend `ci.yml:136`, `ci.yml:139` and both `.pre-commit-config.yaml` `files:` regexes to include `_bugaudit/`, run `ruff check --fix` and `ruff format` on the two files (both findings are auto-fixable `I001`), and delete the now-false sentence at `.pre-commit-config.yaml:13`.
2. Decide `_bugaudit/` is scratch and stop tracking it, adding it to `.gitignore` alongside `.worktrees/`. Note `.gitignore` currently has no `_bugaudit` entry, so the directory is tracked and would keep drifting either way.

Option 1 is the smaller change and is consistent with #1118's stated goal that every Python file in the repo sit behind a lint gate.

abyyxhek:
## What

The sweep started by #1064 (benchmarks/) and #1118 (scripts/, demo-run/) left one tracked Python directory ungated. `_bugaudit/` ships two files on `main` and sat outside the ruff scope in both CI and pre-commit.

Closes #1232.

## Changes

Extended both gates to `_bugaudit/`:

- `.github/workflows/ci.yml` — `ruff check` and `ruff format --check` now include `_bugaudit/`
- `.pre-commit-config.yaml` — both `files:` regexes now `^(src|tests|examples|_bugaudit)/`

Fixed the two findings, which are auto-fixable `I001` import sorting plus the formatting drift that follows them. Both files are scratch checks that print the interpreter version and where `continuum` imports from, so the diff is blank lines around the import statements:

```diff
 import sys
+
 print(sys.version)
 try:
     import continuum
+
     print("continuum imported from:", continuum.__file__)
```

The pre-commit config documented the exclusion in plain terms — "benchmarks/, demo-run/ and _bugaudit/ are not ruff-clean and are not checked by CI" — which was stale prose rather than a decision. The comment now names only the two directories that genuinely remain outside the sweep (`benchmarks/`, `demo-run/`), so a reader can tell an intentional exclusion from a gap.

I chose option 1 from the issue over removing `_bugaudit/` from the tree: it is the smaller change and is consistent with #1118's stated goal that every Python file in the repo sit behind a lint gate.

## Verification

The gated scope is clean and the two gates agree exactly:

```
$ ruff check src/ tests/ examples/ _bugaudit/        → All checks passed!
$ ruff format --check src/ tests/ examples/ _bugaudit/ → 320 files already formatted
```

Confirmed the pre-commit regex actually matches the `_bugaudit/` files (an underscore-prefixed directory is easy to get wrong in `^(src|tests|examples|_bugaudit)/`), and that the CI and pre-commit scope lists are identical.


Cyrax321 (changes requested):
Thank you for contributing to Continuum, @abyyxhek. This closes the last gap in the lint sweep for #1232 by bringing _bugaudit inside both the CI and pre-commit ruff gates, fixing the two import sort findings, and correcting the stale exclusion comment so it names only the directories that genuinely remain outside.

I reviewed the workflow and pre-commit changes plus the two file cleanups. Lint and Type-check passes, which is the gate this PR is about. One item blocks merge and it is unrelated to the change itself:

1. Stale documented test counts (tests/test_docs_counts.py). All 9 Test jobs fail: the suite collects 2,310 but the docs state about 2,278. Please run pytest --collect-only -q, take the live total, and re-sync every stating file (README.md, the translated READMEs, docs/CONTRIBUTING_ONBOARDING.md, the current CHANGELOG entry, references/testing.md, references/install.md), leaving historical CHANGELOG notes untouched. Then run python -m pytest tests/test_docs_counts.py -k "not slow" -q locally to confirm green before pushing.

Once CI is green this will be ready for a final review. Thanks again for finishing the sweep.

Cyrax321 (changes requested):
Author context: abyyxhek, external contributor. Follow-up to the prior review. The #1232 lint sweep direction is still correct, but the stale docs-count blocker still stands with dirty merge state. Re-run pytest --collect-only on a rebased tree and resync every stating file, leaving historical notes untouched, then confirm the Test matrix green.

.github/workflows/ci.yml
```diff
@@ -145,10 +145,10 @@ jobs:
         run: pip install -e ".[dev]"
 
       - name: Ruff - lint
-        run: ruff check src/ tests/ examples/
+        run: ruff check src/ tests/ examples/ _bugaudit/
```
Cyrax321, line 148: Non-blocking. Approved, still-open PR #1312 adds benchmarks/ on this exact line, so the two PRs cannot both merge textually: whichever lands second must rebase and combine the scopes into `ruff check src/ tests/ examples/ benchmarks/ _bugaudit/`, with the same combining needed on the format line below and on both `files:` regexes in .pre-commit-config.yaml. Worth agreeing the merge order now so the conflict is expected rather than a surprise. The fixed files themselves do not overlap: #1312 auto-fixes benchmarks/fault_injection/runner.py, this PR auto-fixes the two _bugaudit/ files.

.pre-commit-config.yaml
```diff
@@ -9,10 +9,10 @@
 # file in CI, which is the failure this config exists to prevent: bump both
 # together.
 #
-# files scopes the hooks to the three directories CI lints. benchmarks/,
-# demo-run/ and _bugaudit/ are not ruff-clean and are not checked by CI, so
-# leaving them in would fail `pre-commit run --all-files` on an untouched tree
-# and rewrite files no gate asks about.
+# files scopes the hooks to the directories CI lints. benchmarks/ and demo-run/
+# are not ruff-clean and are not checked by CI, so leaving them in would fail
```
Cyrax321, line 13: Non-blocking. Accurate at HEAD today, but this sentence goes stale the moment #1312 merges, because benchmarks/ will then be checked by CI while the comment still names it alongside demo-run/ as not checked. A reader would then be unable to tell an intentional exclusion from a gap, which is the exact failure this rewrite set out to fix. Fold a rewording that names only demo-run/ into whichever of the two rebases comes second.

Cyrax321 (approved):
Approving. I verified the widened gate at head fac8a68 in a scratch tree with the pinned ruff==0.16.5: `ruff check _bugaudit/` and `ruff format --check _bugaudit/` are both clean, and replicating the scope assertions of tests/test_precommit_config.py against the head copies of ci.yml and .pre-commit-config.yaml shows the CI job and both hooks now agree exactly on src/, tests/, examples/, _bugaudit/, with benchmarks/ still excluded on both sides. The docs-count blocker from my earlier reviews is resolved: the full Test matrix, which runs test_docs_counts, is green at this head and the tree reports mergeable.

Two non-blocking coordination notes are inline; both concern the approved but still unmerged #1312, which edits the identical lines to add benchmarks/. One further note cannot be anchored to a diff line: thirteen docs quote `ruff check src/ tests/ examples/` as the gate CI enforces (README.md:103, the five translated READMEs, CONTRIBUTING.md:100, docs/CONTRIBUTING_ONBOARDING.md:37, docs/research.md:57, docs/release-checklist.md:21, docs/guides/memory_governance.md:185, references/testing.md:12, references/install.md:106). #1312 updates all of them when adding benchmarks/, while this PR updates none, so once both land the verification guides will still omit _bugaudit/ even though CI lints it. Non-blocking, but please fold those quoted invocations into whichever rebase comes second, so the guides do not drift from CI the way the ruff pins drifted in #840.

PR #1246, opened 2026-09-17
test: extend the __all__ guard from 5 modules to every module that declares one (#1228)

### LINKED ISSUE #1228: tests: the __all__ guard covers 5 of the 98 modules that declare one, so 93 modules' export lists are unchecked
## Summary

`tests/test_module_all_exports.py` asserts two invariants: every name in a module's `__all__` resolves to a real attribute, and star-imports of those modules resolve at runtime. It enumerates the modules by hand:

```python
modules = [action_ledger, recovery_contract, gate, pinning, recovery_gate]
```

98 modules declare an `__all__` today, so 93 of them get neither check. A refactor that renames a function and forgets `__all__`, or lists a name that no longer exists, goes green on 93 modules and red on 5. This is exactly the shape of drift the #903 / #913 / #1093 / #1111 series has been finding by hand after the fact.

I ran the stronger check against all 98 modules and it passes cleanly right now, so this is a guard extension, not a bug hunt:

```python
import importlib, pkgutil, continuum
bad = []
for m in pkgutil.walk_packages(continuum.__path__, "continuum."):
    mod = importlib.import_module(m.name)
    a = getattr(mod, "__all__", None)
    if not a: continue
    for name in a:
        if not hasattr(mod, name):
            bad.append(f"{m.name}.{name}")
assert not bad  # currently passes: 0 failures
```

## Affected component

Tests.

## Severity

## System information

- Python: 3.13.5

```bash
python -c "import pkgutil, importlib, continuum; print(sum(1 for m in pkgutil.walk_packages(continuum.__path__, 'continuum.') if hasattr(importlib.import_module(m.name), '__all__')))"
pytest tests/test_module_all_exports.py -q
```

## What a fix looks like

Replace the hand-rolled five-module list with a walk over the installed package that collects every module carrying `__all__`, and keep the existing per-module assertions (the five named-symbol tests) as documentation of intent. The `test_star_import_execution` subprocess should keep importing only the five it names today, or grow to import all 98; importing all of them is the stronger option but pulls optional dependencies, so it is worth confirming locally that the subprocess stays exit 0 before choosing that. Skip modules whose import fails on a missing optional extra rather than letting the guard error out, and skip `continuum.__main__` since importing it has a side effect.

abyyxhek:
## What

The `__all__` guard asserted its two invariants for five modules enumerated by hand. 98 modules declare an `__all__` today, so 93 of them got no check at all.

Closes #1228.

```python
modules = [action_ledger, recovery_contract, gate, pinning, recovery_gate]   # 5 of 98
```

A refactor that renames a function and forgets `__all__`, or lists a name that no longer exists, went green on 93 modules and red on 5 — the shape of drift the #903 / #913 / #1093 / #1111 series kept finding by hand after the fact.

## Changes

Replaces the hand-rolled list with a `pkgutil` walk that collects every importable module carrying `__all__`:

- `continuum.__main__` is skipped — importing it runs the CLI.
- Modules behind an optional extra (`otel`, `postgres`) are skipped when the extra is absent: a module that cannot be imported cannot be checked, and a missing dependency is not an export defect.
- Any *other* import failure still propagates, so a genuinely broken module fails rather than being silently exempted.

The five named tests stay, on purpose: they document which symbols a reader is meant to rely on, which a sweep over every module cannot express.

`test_star_import_execution` now imports every module the walk found rather than five, so the `from module import *` the guard exists to validate runs against the whole surface. The subprocess still exits 0 — verified locally against all 98.

Two tests added:
- `test_all_entries_are_strings` — a non-string entry turns every star-import of that module into a `TypeError`.
- `test_every_module_is_covered` — a walk that returns nothing (say, `pkgutil` stopped recursing) would make the assertions above vacuously pass, which is worse than not having them.

## Verification

```
$ pytest tests/test_module_all_exports.py      → 9 passed
```

Verified the guard has teeth by adding a nonexistent name to `pinning`'s `__all__`, which fails both `test_all_symbols_exist_on_their_module` and `test_star_import_execution`. Removed the poison and confirmed green again.

## Note

`tests/test_docs_counts.py::test_documented_count_matches_suite` fails on `main` both before and after this change (the suite collects 2275 vs a documented ~2241; tolerance 30). My new tests add 4 to that gap. Re-syncing the count is #1226's work, in flight as PRs #1237/#1238.


abyyxhek:
Closing this — upstream landed a different implementation for the same guard in baf769d / 0b97ebb / c2ee540 (an enumerated module list with a generic export invariant, from the #1227 work), so this branch conflicts with `main` as it stands.

Rather than rebase over that and replace the direction the maintainer chose, I'm stepping back. The complementary idea is here if it's useful: this branch's `pkgutil` walk auto-discovers modules, so it covers any module added later without someone having to remember to extend the hand-list. Happy to rework it on top of the landed version if that's wanted — the two are not mutually exclusive, the sweep could run *in addition* to the named list.

PR #1247, opened 2026-09-17
fix(bench): publish the fault-injection suite's real detection rate (#1060)

### LINKED ISSUE #1060: benchmarks/run.py renders the fault-injection detection rate as 0 in the published tables
## Summary

`benchmarks/run.py` renders the fault-injection detection rate as a literal `0` in the README bench table and in `references/bench.md`. It reads `detection_rate` and `unsafe_resume_rate` off `BenchmarkReport.summary()`, a method that never returns either key. The suite's real detection rate is `1.0` with an unsafe-resume rate of `0.0`, and the correct figures are already written to `benchmarks/out/fault_injection_report.json` by the same run, so the published number is the only place the truth is lost.

## Affected component

Benchmarks (published results in README.md and references/bench.md)

## Severity

Minor inconvenience or cosmetic issue for the library. It matters for the project's published claims: a reader checking whether the fault-injection suite actually detects anything is told it detects nothing, on a suite that currently detects everything.

## Evidence

`src/continuum/benchmark/phase6/metrics.py` defines `summary()` to return `total`, `passed`, `failed` and `by_outcome` only. `benchmarks/run.py:105` calls `fault_report.summary()` and `:108` reads `f_summary.get('detection_rate', 0)` and `f_summary.get('unsafe_resume_rate', 0)`, both of which hit the default. The regeneration paths then carry those zeros into the README and bench.md tables. The emitter in `benchmarks/fault_injection/emitter.py` computes the real aggregates into its own envelope, which is why the JSON file is right and the published table is wrong.

At HEAD `14f8a84` the horizon half of the table was regenerated after the #1044 merge and is current, while the fault-injection line beneath it still reads:

```
Fault-injection: 7 scenarios, detection 0, unsafe 0
```

in both `README.md:253` and `references/bench.md:116`.

```python
from benchmarks.fault_injection.runner import run_benchmark_suite
rep = run_benchmark_suite()
print('summary() keys          :', sorted(rep.summary().keys()))
print('run.py renders detection :', rep.summary().get('detection_rate', 0))
```

Output at HEAD `14f8a84`:

```
summary() keys          : ['by_outcome', 'failed', 'passed', 'total']
run.py renders detection : 0
```

For the published real numbers, the suite's own summary and the written report both give:

```
{'total': 6, 'detected': 6, 'detection_rate': 1.0, 'unsafe_resume': 0, 'unsafe_resume_rate': 0.0, 'false_positive_rate': 0.0}
benchmarks/out/fault_injection_report.json: detection_rate 1.0, unsafe_resume_rate 0.0
```

## Root cause

Introduced in `16432e7` ("bench: horizon-scale scenarios with simulated years and a recovery-decision judge", Refs #398), which added the horizon report and the shared bench-table renderer. The renderer was written against the fault-injection summary dict that `run_fault_injection_suite` returns, then pointed at `BenchmarkReport.summary()`, which does not carry those keys.

## Expected behaviour

The published table should show the suite's actual detection and unsafe-resume rates, and should never print `0` for a metric the harness measured.

## Suggested fix shape

Either extend `BenchmarkReport.summary()` to carry the metrics the table needs, or have the table read them from the results the way the emitter already does. Prefer averaging across results over trusting `results[0]`, which is the separate order-dependence defect filed separately; fixing this by copying the emitter's `results[0]` read would import that fragility.

## System information

- Python: 3.14.5

abyyxhek:
## Description

`benchmarks/run.py` rendered the published fault-injection line from `BenchmarkReport.summary()`. That method counts outcomes only — `total`, `passed`, `failed`, `by_outcome` — and never returns `detection_rate` or `unsafe_resume_rate`. Both reads therefore hit their `0` default, so the bench table reported a suite that detects every fault it injects as detecting nothing:

```
Fault-injection: 7 scenarios, detection 0, unsafe 0
```

in both `README.md` and `references/bench.md`, while the JSON report the same run writes (`benchmarks/out/fault_injection_report.json`) carries the real figures.

The renderer now averages the rates the scenarios actually carry — the same convention the horizon columns in the same function already use — rather than reading them off the shared envelope's outcome counts. `summary()` is intentionally generic across suites (it is shared with horizon), so the suite-specific rates belong on the results, which is where the runner already puts them.

The clean-control scenario is excluded from the average rather than counted as a zero: it carries only its own `false_positive_rate` and never measured detection, so dividing by `len(results)` would average the suite against a scenario that has no opinion.

## Related issues

Fixes #1060

## Types of changes

- Type: `bugfix`

## Breaking changes

No. The change is confined to the benchmark renderer and the published bench-table figures; no library, CLI or wire behaviour changes.

## How has this been tested?

Reproduction before the fix (issue's own snippet, confirmed at HEAD `18e81d0`):

```
summary() keys          : ['by_outcome', 'failed', 'passed', 'total']
run.py renders detection : 0
```

After the fix, the same call renders:

```
Fault-injection: 7 scenarios, detection 1.0, unsafe 0.0
```

and both `README.md` and `references/bench.md` were updated to match, since the regeneration path (`benchmarks/run.py`, gated by `tests/test_horizon_benchmark.py`) rewrites exactly that line.

Two regression tests were added to `tests/test_fault_injection.py`, both verified to **fail before** the fix and pass after:

- `test_published_detection_rate_reads_the_results_not_a_summary_default` — fast guard that `summary()` carries no rate key, that the control scenario stays out of the average, that `_mean_rate` reports the suite's real rates, and that an empty scenario set answers `0` instead of raising.
- `test_published_bench_line_carries_the_real_detection_rate` (slow) — renders the actual table line through `_bench_table_lines` and asserts it agrees with the suite's own summary figures.

Commands run, Windows 11 / Python 3.13, from the repo root with `PYTHONPATH=src`:

```
ruff check benchmarks/run.py tests/test_fault_injection.py   # clean
ruff format --check benchmarks/run.py tests/test_fault_injection.py  # clean
mypy src/continuum                                           # no issues, 127 files
pytest tests/test_fault_injection.py tests/test_docs_counts.py  # 13 passed
pytest tests/                                                # 2242 passed, 38 skipped
```

The collected total moved 2278 -> 2280, so the documented figures in `README.md`, `docs/CONTRIBUTING_ONBOARDING.md` and `CHANGELOG.md` were re-synced to keep `tests/test_docs_counts.py` green (the guard's own remedy for test-adding PRs). Note this PR's re-sync also absorbs a small pre-existing drift: at the pre-rebase base `18e81d0` the suite already collected 2273 against a documented 2241, so that guard was already red on main before this branch.

## Compatibility considerations

None. `summary()` is unchanged; the fix is local to `benchmarks/run.py`'s table rendering.


Cyrax321 (dismissed):
Thank you for contributing to Continuum, @abyyxhek. This fixes the rendering side of the fault injection number for #1060: benchmarks/run.py read detection and unsafe resume rates off the shared outcome envelope, which never carries them, so the bench table printed 0 for a suite that detects every fault. Averaging the rates the scenarios actually carry matches the horizon convention in the same function, and excluding the clean control rather than counting it as zero is the honest denominator.

I reviewed the renderer change, the bench table updates, and the new tests, and I checked CI alongside its companion #1250: the full matrix across all three platforms plus Postgres plus lint is green. The two PRs stay independent as described, one for the emitted reports and one for the rendered table. No blocking issues found. LGTM, marking this as ready to pull. Thank you again for pairing the two halves cleanly.

Cyrax321 (approved):

Ref: #1247 / Issue #1060
Surface: benchmarks/run.py fault-injection table rendering

Bug: the published bench table read detection_rate and unsafe_resume_rate off BenchmarkReport.summary(), which counts outcomes only and never carries those keys, so both reads hit their 0 default and the README and references/bench.md advertised a suite that detects nothing while the JSON report from the same run carried the real figures. Verified the mismatch class is real: summary() is the shared envelope across suites.

Fix: the renderer averages the rates the scenarios actually carry, matching the convention the horizon columns in the same function already use. The clean-control exclusion is handled correctly and for the right reason: the control carries only its own false_positive_rate, so skipping results without the key keeps a scenario with no opinion out of the detection average instead of diluting it as a fake zero.

Proof:
- Read _mean_rate and the rewritten fault-injection line; the fallback when no result carries the key is 0, which is the honest output for an empty suite rather than a crash.
- Ran tests/test_fault_injection.py at head efcf1908: 11 passed, including the new coverage pinning the rendered line and the control exclusion.

Issue #365, opened 2026-08-25
Identity-token fallback collapses same-basename resources in different directories, silently skipping a side effect


Cyrax321:
## What happens

Two genuinely different resources that share a filename are treated as the same action, so the second side effect is never performed and the caller is handed the first one's receipt.

## Reproduction (MCP only, no explicit `key`)

```
intercept_action(type="tenant_notify", arguments={"path": "/tenants/acme/report.csv"})
  -> proceed=true, action_key=69b1c5f0...
complete_action(action_key=69b1c5f0..., external_id="notify-acme-001")
  -> completed

intercept_action(type="tenant_notify", arguments={"path": "/tenants/globex/report.csv"})
  -> proceed=false
     status=completed
     external_id="notify-acme-001"
     guidance="Already performed. Reuse the previous result; do not repeat it."
```

Globex is never notified. The response returns acme's `external_id` as if it were globex's, and the returned `action_key` is acme's, so a caller that follows the two-phase protocol would settle acme's record a second time.

## Why

The exact argument-hash lookup misses, because the paths differ. That engages `_identity_match`, whose comparison runs on `leaf_tokens`, and `leaf_tokens` deliberately drops any token containing a separator:

```python
return frozenset(t for t in tokens if "/" not in t and "\\" not in t)
```

Both claims reduce to `{report.csv, report}`, containment holds in both directions, exactly one completed candidate matches, and the claim is answered as a duplicate.

`leaf_tokens` documents this outcome ("two same-type actions on same-named files in different directories are treated as one"), so the behaviour is known. What makes it a defect rather than a documented tradeoff is that it produces precisely the failure `_identity_match` says it exists to prevent: "Under mere intersection the second side effect would be silently swallowed, the exact failure this ledger exists to prevent." Containment plus leaf reduction swallows it too.

## Why the default is the dangerous one

Passing an explicit `key` avoids the collapse, verified:

```
intercept_action(type=..., key="notify:/tenants/acme/report.csv")   -> proceed=true
intercept_action(type=..., key="notify:/tenants/globex/report.csv") -> proceed=true, distinct key
```

But the tool description frames `key` as being for operations that are legitimately repeated ("sending the same reminder twice is two sends, not one"), so nothing tells a caller that omitting it risks silently dropping distinct work. Per-directory files with conventional names (`report.csv`, `invoice.pdf`, `index.json`, `config.yaml`) are a common fan-out shape, so the exposure is not exotic.

## Suggested fix

[...]

Issue #36, opened 2026-08-13
identity_tokens drops purely-numeric resource identifiers, so the cross-session fallback fails to dedupe numeric ids


Cyrax321:
- [x] I searched existing issues for a duplicate
- [x] I am reporting this against the correct repository
- [x] I am using a supported Python version (3.11 or later)
- [x] I can reproduce this on the latest commit on `main`

## System information
OS: macOS 15.x (darwin)\nPython: 3.13.5\ncontinuum-agent: 0.1.0 (editable install from source)\nSource commit: a539948 (== origin/main)

## Command or code to reproduce
``_is_strong_token`` returns False for any token where ``token.isdigit()`` is True. So a resource identified only by a numeric id (database row id, account number, numeric invoice id) is dropped from `identity_tokens`. The ledger's defensive fallback (`_identity_match`) then finds no shared token and opens a brand-new action slot even when the logical operation is identical and only the argument field name drifted between sessions.

```python
from continuum.storage.sqlite import SQLiteStorage
from continuum import models, EventType, ActionLedger
from continuum.actions.idempotency import identity_tokens, _is_strong_token
from continuum.models import Run, RunStatus

print('4821 strong?', _is_strong_token('4821'), '| tokens:', identity_tokens(external_id='4821'))

store = SQLiteStorage(':memory:')
store.create_run(Run(run_id='r1', goal='g', status=RunStatus.RUNNING))
store.append_event('r1', EventType.RUN_STARTED, {'goal':'g'})
led = ActionLedger(store, 'r1')
o1 = led.claim('db.update', {'row_id': 4821})
o2 = led.claim('db.update', {'id': 4821})  # same op, renamed arg (session drift)
print('two slots?', str(o1.key) != str(o2.key))
```

## Current behaviour
Two separate action records are created for the same logical operation, so the side effect can be performed twice. The end-to-end suite only ever uses ids like `INV-001` (contains a letter, so not purely numeric and not dropped), which masks the defect.

## Expected behaviour
Numeric resource identifiers are legitimate identities; a row id of 4821 identifies one row just as well as `INV-001` identifies one invoice. The fallback should treat them as strong tokens (or, at minimum, not discard them), so argument-drift between sessions still deduplicates.

## Proof
Reproduction prints:
```
4821 strong? False | tokens: frozenset()
two slots? True
```

Cyrax321:
[...]

PR #1248, opened 2026-09-17
test(idempotency): cover idempotency token helpers directly (#1233)

### LINKED ISSUE #1233: tests: the idempotency token helpers that carry the #365 double-effect fix have no direct test
## Summary

The ledger's defensive fallback, the path that matches two attempts describing the same resource differently, rests on five pure functions in `src/continuum/actions/idempotency.py`. No test file names any of them:

- `identity_tokens` (`:293`) — distinctive resource tokens from an argument mapping
- `leaf_tokens` (`:342`) — the tokens that name a resource rather than locate it
- `location_tokens` (`:362`) — the exact complement, what `leaf_tokens` sets aside
- `same_location` (`:383`) — whether two path-like tokens can be one file rendered differently
- `locations_agree` (`:405`) — whether two token sets locate the same resource, or locate nothing

Their docstrings carry the load-bearing invariants. `leaf_tokens` explains that discarding the container is what lets `_identity_match` demand containment instead of a shared token, and that doing it unconditionally "silently swallowed the second side effect (issue #365)". `locations_agree` explains that a side carrying no path-like token makes no location claim and cannot contradict one. `same_location` is deliberately not a filesystem check so the answer is identical on every machine.

`tests/test_action_ledger.py` and its neighbours exercise the ledger end to end, so these helpers are covered indirectly, but the property-level invariants (containment not equality, complementarity of leaves and locations, a missing path not contradicting a present one) are only asserted if a reader infers them from prose. #1116 made the same argument for `RewindError` and `LedgerLockError`.

## Affected component

Core (state, checkpoints, recovery); Tests.

## Severity

## System information

- Python: 3.13.5

```bash
grep -rl 'identity_tokens\|leaf_tokens\|location_tokens\|locations_agree\|same_location' tests/   # no output
```

## What a fix looks like

A `tests/test_idempotency_tokens.py` covering the documented properties directly, no storage needed: `leaf_tokens` and `location_tokens` partition the full token set between them; `/data/invoices/INV-5.pdf` and `invoices/INV-5.pdf` agree on location while `/tenants/acme/report.csv` and `/tenants/globex/report.csv` do not (the #365 case); a basename shared with no directory in common is not the same file; an integer argument renders as a token; an empty `location_tokens` on one side never blocks a match. A `hypothesis` strategy over path fragments would fit this module's style, and the drift cases in the docstrings (`/data/invoices/INV-5.pdf` vs `invoices/INV-5.pdf`) are the cases a regression would break, so they should be named tests rather than incidental coverage.

soorajsuru2004-sudo:
## Description

Adds `tests/test_idempotency_tokens.py` to directly cover the five pure functions in `src/continuum/actions/idempotency.py` that underpin the ledger's fallback and the issue #365 double-effect fix:

- `leaf_tokens` and `location_tokens`: verifies disjointness and partition over arbitrary token sets with Hypothesis.
- `identity_tokens`: verifies string token extraction, path derivations (basename and stem), integer scalars as tokens (issue #36), boolean exclusion, nested structures, external id extraction, volatile parameter stripping, and token filtering.
- `same_location`: verifies suffix drift comparison, cross-platform separator handling, symmetry, and rejection of distinct containers sharing only a basename (issue #365).
- `locations_agree`: verifies missing location tokens never block a match, matches on reconcilable pairs, and refuses mismatched directories.

## Related issues

Fixes #1233

## Types of changes

- Type: `tests`

## Breaking changes

No

## How has this been tested?

```
pytest tests/test_idempotency_tokens.py -v
pytest -q
ruff check tests/test_idempotency_tokens.py
ruff format --check tests/test_idempotency_tokens.py
```

Environment: Windows, Python 3.12.10, pytest 9.1.1, hypothesis 6.168.0.

## Checklist

## Additional context

Tests the documented invariants directly without requiring storage or ledger fixtures.

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive coverage for idempotency token extraction and classification.
  - Added tests for nested inputs, external identifiers, volatile arguments, weak tokens, and empty values.
  - Added coverage for location comparison, including path drift, distinct containers, empty paths, suffix drift, symmetry, and missing or conflicting location claims.


anya-research (approved):
Thank you for contributing to Continuum, @soorajsuru2004-sudo. This adds `tests/test_idempotency_tokens.py` covering the five pure helpers for #1233, and it overlaps directly with open #1249 which adds the same test path for the same issue plus count bumps in three docs files, so only one of the two can land.

The contracts in the new tests match `src/continuum/actions/idempotency.py` exactly. Partition holds by separator presence, integer scalars become tokens with booleans excluded, volatile keys are stripped before collection, weak and short tokens are filtered, and `same_location` accepts trailing suffix drift while rejecting same basenames in distinct containers with `locations_agree` passing through empty sides.

CI is green across the matrix plus lint and type-check. Codecov patch is green. The single FAILURE is `anya-review`, which guards prior review deduplication and fails on a missing `GH_TOKEN` in the workflow, unrelated to this change.

- Test matrix on ubuntu, macos, and windows across Python 3.11 to 3.13 is SUCCESS
- Lint and type-check is SUCCESS, Codecov patch is SUCCESS
- `leaf_tokens` plus `location_tokens` partition logic matches the separator predicate in implementation
- `identity_tokens`, `same_location`, and `locations_agree` cases match the documented #36 and #365 behavior

No blocking issues found. This is ready to merge once maintainers pick between this and #1249. Your drift and symmetry cases are thorough and clearly tied to the regression they guard.

PR #1249, opened 2026-09-17
test(actions): direct tests for the idempotency identity-token helpers (#1233)

### LINKED ISSUE #1233: tests: the idempotency token helpers that carry the #365 double-effect fix have no direct test

abyyxhek:
## Description

Adds direct tests for the five public helpers in `continuum.actions.idempotency` that carry the ledger's defensive fallback: `identity_tokens`, `leaf_tokens`, `location_tokens`, `same_location` and `locations_agree`. These are the functions that recognise an already-recorded action when two attempts describe the same resource differently (a renamed argument field, a path re-rendered between sessions).

`tests/test_action_ledger.py` exercises the fallback end to end, so the helpers were covered indirectly, but the invariants their docstrings state were asserted only in prose. This PR asserts them directly, so a change to one half of the pair is caught by the suite rather than by a reader noticing a comment no longer matches the code.

The `#365` case is the load-bearing one and gets a named test: `/tenants/acme/report.csv` and `/tenants/globex/report.csv` share every leaf, so leaf comparison alone matched them and globex was never notified. `location_tokens` is what distinguishes them, and the ledger requires `locations_agree` before accepting a leaf-level match. To confirm the new tests actually catch that regression, I temporarily reverted `location_tokens` to the pre-#365 behaviour (returning an empty set); four tests failed, including the dedicated `#365` test. The source was then restored and the file re-verified clean.

`tests/test_idempotency_tokens.py` (25 tests) covers: the leaf/location partition and a hypothesis property over generated arguments; the path-drift case (`/data/invoices/INV-5.pdf` vs `invoices/INV-5.pdf`); the `#365` disagreement; suffix-not-equality location comparison; separator-agnostic and empty-path handling; integer and `external_id` tokens; weak/stopword dropping; volatile-argument exclusion; and symmetry properties for both comparisons. Generated paths use a portable ascii vocabulary so the suite means the same thing on posix and Windows.

The documented test counts are re-synced to `~2,303 collected` / `~2,278 passed`, the 25 new tests, in `README.md`, `docs/CONTRIBUTING_ONBOARDING.md` and `CHANGELOG.md`. `tests/test_docs_counts.py` guards those three files against drift past a tolerance of 30, so a test-count change without a re-sync turns CI red.

No source or behaviour change.

## Related issues

Fixes #1233

## Types of changes

- Type: `tests`

## Breaking changes

No.

## How has this been tested?

Branched from `upstream/main` at `baf769d` (rebased mid-work when upstream advanced five commits; the branch is one commit ahead and contains no merges). Windows 11, Python 3.11.15, editable `.[dev]` install.

```
pytest tests/test_idempotency_tokens.py tests/test_docs_counts.py -q   # 28 passed
pytest -q                                                              # full suite, exit 0
pytest --collect-only -q                                               # 2,303 collected
ruff check src/ tests/ examples/                                       # All checks passed!
ruff format --check src/ tests/ examples/                              # 319 files already formatted
mypy src/continuum                                                     # no issues found in 127 source files
git diff --check                                                       # clean
```

Regression check: with `location_tokens` temporarily reverted to the pre-#365 behaviour, `test_leaf_and_location_tokens_partition_the_set`, `test_partition_holds_for_generated_arguments`, `test_leaf_tokens_reach_a_path_through_its_basename` and `test_same_named_files_in_different_directories_disagree` fail. Source restored before commit.

The full suite was run three times during development (before the rebase, and twice after), each exiting 0 with no failures.

## Checklist

## Additional context

The count re-sync is included because the guard compares a live `pytest --collect-only` against the three documented figures with a tolerance of 30; the 25 new tests would otherwise drift past it. The passed/skipped split is kept internally consistent (`~2,278 passed`, `~25 skipped`). The `161 test files` figure in the README is not re-synced here: it is already stale on `main` and is tracked separately by #1068.

## Summary by CodeRabbit

* **Documentation**
  * Updated reported test totals across the README, changelog, and contributor onboarding guide: approximately 2,303 tests collected, including 2,278 passed and 25 skipped.
  * Corrected the project overview’s test count to match the latest suite size.

* **Tests**
  * Added comprehensive coverage for idempotency token and location-comparison behavior, including nested inputs, path variations, excluded values, and edge cases.


Cyrax321 (approved):
Thank you for contributing to Continuum, @abyyxhek. This adds direct tests for the five public idempotency helpers behind the ledger fallback for #1233, asserting the docstring invariants that were previously covered only indirectly through the ledger end to end tests.

I reviewed the new test module and the count re-sync, and I checked CI: the full matrix across all three platforms plus Postgres plus lint is green, with no bot findings. The mutation check is the strongest evidence here: reverting location_tokens to its pre #365 behavior fails four tests including the dedicated tenant path case, so the suite genuinely pins the regression. No blocking issues found. LGTM, marking this as ready to pull. Thank you again for verifying the tests catch what they claim to.

abyyxhek:
Closing as superseded — the maintainer landed the same coverage in #1248 (b97256d, merged), which is already on main. Thanks for the review nudge; no rebase will make this branch add anything beyond what main has now.

PR #1250, opened 2026-09-17
fix(bench): take the fault-injection suite rate from the results, not results[0] (#1061)

### LINKED ISSUE #1061: fault-injection emitter takes the suite-level rate from results[0], so it is order-dependent
## Summary

`emit_fault_injection_report` takes the suite-level detection rate from `report.results[0]` on a comment asserting every result carries the same aggregates. The premise is false: the clean control scenario is appended with only `false_positive_rate` and has no `detection_rate` key at all. The published number therefore depends on result order, and a control placed first reports a detection rate of `0` for a suite that detected every fault.

## Affected component

Benchmarks (fault-injection emitter and report)

## Severity

Minor inconvenience or cosmetic issue for the library. Today the control happens to be appended last, so the published figure is accidentally correct; the defect is a latent wrong number waiting for a reorder or an added scenario.

## Evidence

`benchmarks/fault_injection/runner.py:606-649` builds the report: the per-fault results each carry `detection_rate`, `unsafe_resume_rate`, and `false_positive_rate` copied from the suite summary, while the control appended at `:641-649` carries only `metrics={"false_positive_rate": ...}`. `benchmarks/fault_injection/emitter.py:69-74` reads `report.results[0].metrics` and comments "Use the first result's aggregates as the suite-level (they are all same)". The comment is the bug: it is not true, and the fallback default of `0` turns the ordering into the published rate.

```
results[0]               : fault_fabricated_progress -> 1.0
control has detection_rate? False
published (control last) : 1.0
published (control first): 0
```

```python
import sys, json, os, tempfile
sys.path.insert(0, '.')
from benchmarks.fault_injection.runner import run_benchmark_suite
from benchmarks.fault_injection.emitter import emit_fault_injection_report

rep = run_benchmark_suite()
print('results[0]               :', rep.results[0].scenario, '->', rep.results[0].metrics.get('detection_rate'))
print('control has detection_rate?', 'detection_rate' in rep.results[-1].metrics)

out = os.path.join(tempfile.mkdtemp(), 'fi')
env = json.loads(emit_fault_injection_report(rep, out)[0].read_text())
print('published (control last) :', env['summary']['detection_rate'])

rep.results.insert(0, rep.results.pop())   # move the control to the front
out2 = os.path.join(tempfile.mkdtemp(), 'fi')
env2 = json.loads(emit_fault_injection_report(rep, out2)[0].read_text())
print('published (control first):', env2['summary']['detection_rate'])
```

Output at HEAD `14f8a84`:

```
results[0]               : fault_fabricated_progress -> 1.0
control has detection_rate? False
published (control last) : 1.0
published (control first): 0
```

## Root cause

Introduced in `7f94ddb` ("bench: fault-injection chaos suite with detection-rate and unsafe-resume metrics"). The adapter copied the suite-level rate onto each fault result and then read it back from the first element, which was correct only while every result was a fault result; the control was appended afterwards without the same keys.

## Expected behaviour

The suite-level rate should not depend on which result happens to be first. It should be computed once from the results that carry the metric, or carried on the report itself.

## Suggested fix shape

Replace the `results[0]` read with the average the loop already computes: `round(sum(detection_rates) / len(detection_rates), 3)` when `detection_rates` is non-empty, falling back to `0` only when no result carries the key. The `detection_rates`, `unsafe_rates`, and `fp_rates` lists are already collected at `emitter.py:56-68` and are order-independent, so the same change covers `unsafe_resume_rate` and `false_positive_rate`.

## System information

- Python: 3.14.5

abyyxhek:
## Description

`emit_fault_injection_report` read the suite-level `detection_rate`, `unsafe_resume_rate` and `false_positive_rate` off `report.results[0]`, on a comment asserting that every result carries the same aggregates. The premise is false: each fault scenario carries the suite's own aggregate, but the clean control scenario is appended with only its `false_positive_rate`. So the published figure depended on result order — today the control happens to be appended last and the number is accidentally right; a control placed first publishes a detection rate of `0` for a suite that detects every fault:

```
results[0]               : fault_fabricated_progress -> 1.0
control has detection_rate? False
published (control last) : 1.0
published (control first): 0
```

The collection loop had already gathered `detection_rates`, `unsafe_rates`, `fp_rates` and `prop_distances` — and then only used the last one. Those collected lists are now the source of the summary, averaged over the scenarios that actually reported each metric. Scenarios that never measured a rate are skipped rather than counted as a zero, which is what keeps the control out of the detection average. `propagation_distance` was already averaged this way; it now goes through the same helper.

Companion to #1247 (the rendering side of the same published number): that PR fixes the rate the README/bench table shows, this one fixes the rate the emitted JSON and Markdown reports carry. They are independent defects with independent root causes.

## Related issues

Fixes #1061

## Types of changes

- Type: `bugfix`

## Breaking changes

No. The report envelope is unchanged in shape; only the values are now order-independent. A report whose scenarios carry no rate for a metric still answers `0` rather than dropping the documented key.

## How has this been tested?

The issue's own reproduction, confirmed at HEAD `baf769d` both before and after the fix (control-last and control-first orderings). One regression test added to `tests/test_fault_injection.py`, verified to **fail before** the fix and pass after:

- `test_emitter_rate_is_order_independent` — asserts the control carries no `detection_rate`, that the baseline published rate is `1.0`, then rotates every result through the front of the list and re-emits, asserting the published rate never moves. This fails on the old code at the first rotation (`assert 0 == 1.0`).

Commands run, Windows 11 / Python 3.13, from the repo root with `PYTHONPATH=src`:

```
ruff check benchmarks/fault_injection/emitter.py tests/test_fault_injection.py   # clean
ruff format --check benchmarks/fault_injection/emitter.py tests/test_fault_injection.py  # clean
mypy src/continuum        # no issues, 127 files
pytest tests/             # 2241 passed, 38 skipped
```

The collected total moves 2278 -> 2279, which stays inside the `test_docs_counts.py` tolerance of 30, so no documented figure needed re-syncing.

## Compatibility considerations

None — the emitted report files (`benchmarks/out/fault_injection_report.json|md`) now carry order-independent summary values; nothing else reads this emitter's internals.


Cyrax321 (approved):
Thank you for contributing to Continuum, @abyyxhek. This fixes the order dependent aggregate in emit_fault_injection_report for #1061: suite level rates are now the mean over the scenarios that reported each metric instead of a copy of results[0], so a control placed first no longer publishes a detection rate of 0 for a suite that detected every fault.

I reviewed the emitter change, the shared mean helper with its documented empty case, and the new tests, and I checked CI: the full matrix across all three platforms plus Postgres plus lint is green. The companion rendering side in #1247 stays independent of this change. No blocking issues found. LGTM, marking this as ready to pull. Thank you again for the precise reproduction in the description.

Issue #1235, opened 2026-09-17
tests: is_quiet_window and render_trajectory_report are the trajectory report's trigger and renderer, and neither has a direct test


Cyrax321:
## Summary

`tests/test_trajectory_report.py` covers the build and record paths well: report generation per compaction window, the min-authority non-amplification rule, zero overhead when quiet never occurs, bounded size, digest auditability, and archive determinism. Two exported functions in `src/continuum/analysis/trajectory_report.py` are never named in any test file:

- `is_quiet_window` (`:48`) — decides whether a stretch of events counts as quiet, which is the trigger that makes the rest of the module run at all. It is exercised indirectly by `test_health_idle_trigger_generates_for_quiet_and_not_for_busy` and `test_zero_overhead_when_quiet_never_occurs`, but its own boundary is not asserted.
- `render_trajectory_report` (`:8`) — renders a recorded report for briefing consumption. The module's `__all__` exports it and `docs/guides/embed-claude-code.md:63` names it as a briefing field, but no test renders a report and checks the output shape.

Between them they are the "when does this fire" and "what does a human see" ends of the feature, with the middle (build, record, digest) fully covered.

## Affected component

Tests.

## Severity

Minor inconvenience or cosmetic issue.

## System information

- OS: macOS 26.0.1 (Darwin 25.0.0)
- Python: 3.13.5
- continuum-agent: from source, commit `bbfdd74` on `main`
- Installed via: `uv pip install -e ".[dev]"`

## Command or code to reproduce

```bash
grep -rn 'is_quiet_window\|render_trajectory_report' tests/    # no output
grep -n 'def is_quiet_window\|def render_trajectory_report' src/continuum/analysis/trajectory_report.py
```

## What a fix looks like

Add cases to the existing `tests/test_trajectory_report.py` rather than a new file, reusing its storage fixtures. For `is_quiet_window`: a window of only non-side-effect events is quiet, a single `TOOL_COMPLETED` breaks it, an empty event list is quiet (the "never occurs" case that already drives the zero-overhead test), and the window boundary is inclusive on the side it checks. For `render_trajectory_report`: render a report built by `build_trajectory_report` and assert the rendered text names the top stall and scarcity rate the report records, and that a report with no lessons renders something honest rather than an empty string. Both are pure or storage-light, so they fit the module's existing fast style and do not need the `@pytest.mark.slow` marker.

PR #1251, opened 2026-09-17
test: direct tests for is_quiet_window and render_trajectory_report (…

abyyxhek:
…#1235)

tests/test_trajectory_report.py covered the build and record paths well -- generation per compaction window, the min-authority non-amplification rule, zero overhead when quiet never occurs, bounded size, digest auditability, archive determinism. The two exported functions at the ends of that pipeline were never named in any test file:

  is_quiet_window     decides whether a stretch of events counts as quiet, the
                      trigger that makes the rest of the module run at all
  render_trajectory_report  renders a recorded report for briefing consumption

Both are exercised indirectly, but the property-level boundaries were only inferrable from prose: which events break a quiet window, what a human sees.

Adds cases to the existing file rather than a new one, reusing its storage fixtures. For is_quiet_window: empty is quiet, a non-progressing event is quiet, WORK_COMPLETED with count>0 breaks it, a failed completion does not, TASK_UPDATED with completed>0 breaks it, DECISION_CREATED breaks it, a window stays busy from the first breaker wherever it sits, and a completed counter that cannot be read as an integer is treated as busy rather than assumed zero.

For render_trajectory_report: a report built by build_trajectory_report renders its id, window, scar_rate and every stall site by name, and a report with no lessons still renders an honest header rather than an empty string a reader could mistake for missing data.

Verified both directions: poisoning is_quiet_window to always return True fails four tests, and stripping the renderer to a bare header fails two.

## Types of changes

## Checklist

## Summary by CodeRabbit

- **Tests**
  - Added coverage for quiet-window detection, including empty windows, non-progressing events, breaker events, and unreadable counters.
  - Added coverage for trajectory report rendering, including stall-site and scar-rate labels, empty-window output, and derived-origin labeling.


Cyrax321 (commented):
Thank you for contributing to Continuum, @abyyxhek. I reviewed this alongside the two sibling PRs covering the same #1235 ground, #1254 and #1252. Your new is_quiet_window and render_trajectory_report cases pass. Two things to be direct about.

First, #1252 by Adhi1-2 covers the same two functions with eleven cases reusing the existing fixtures, includes the docs re-sync, and is fully green across the matrix, so I have approved it as the one to land. Landing all three would triple cover the same surface and they all append to the same test file, so they cannot all merge cleanly.

Second, this PR as it stands cannot merge anyway: its Test jobs fail on stale documented test counts (the suite collects about 2,317 while the branch docs state about 2,278). Your added tests themselves pass.

Please close this PR as superseded by #1252. If you believe your cases pin an edge #1252 lacks, name the exact test here so it can be ported over rather than losing the coverage. Thank you again for taking this on.

Cyrax321 (approved):

Ref: #1251 / Issue #1235
Surface: tests/test_trajectory_report.py, is_quiet_window and render_trajectory_report

The gap is real: the two exported endpoints of the trajectory-report pipeline were only exercised indirectly, and their property-level boundaries (which events break a quiet window, what the rendered output must contain) were inferrable only from prose. The PR adds cases to the existing file, reusing its fixtures, which is the right organization for coverage of the same module.

Proof:
- Ran tests/test_trajectory_report.py at head 275cb13: 15 passed.

Merge-order note: #1254 by @soorajsuru2004-sudo covers the same two functions for the same issue. Keep whichever lands first and close or rebase the other; the property lists should be reconciled so nothing the other PR covered is lost.

PR #1252, opened 2026-09-17
test(analysis): cover the quiet-window trigger and the report renderer (#1235)

Adhi1-2:
## Description

`is_quiet_window` and `render_trajectory_report` are the two ends of the trajectory report feature — the first decides when a report is built at all, the second is what a human reads. The middle (build, record, digest) is well covered; neither of these two had a direct test (`grep -rn 'is_quiet_window\|render_trajectory_report' tests/` was empty).

This PR adds cases to the existing `tests/test_trajectory_report.py`, reusing its storage fixtures, rather than a new file.

**`is_quiet_window`** — the quiet boundary, per event type and payload:
- An empty window is quiet (the case that already drives the zero-overhead test).
- `RUN_STARTED`, `TOOL_FAILED`, `STATE_CHECKPOINTED`, `LIVENESS_SILENCE_DETECTED`, `ACTION_RECORDED`, `TRAJECTORY_REPORT` alone leave the window quiet.
- `WORK_COMPLETED`: `count > 0` breaks quiet; `count == 0` and `failed: true` do not; a missing count defaults to one unit of work, and an *unreadable* count is treated as progress rather than silently ignored.
- `TASK_UPDATED`: `completed > 0` breaks quiet; `completed == 0` and a missing `completed` do not; an unreadable `completed` is treated as progress.
- `DECISION_CREATED` breaks quiet regardless of payload.
- A progress event anywhere in the window breaks quiet, not only at the tail.
- The window bound is half-open, `(start, end]`: starting a window *at* a progress event excludes it (a report is still built), ending one *at* it includes it (nothing is built).

**`render_trajectory_report`** — the renderer echoes what the report recorded:
- The rendered text names the `report_id`, the `window_start->window_end` range, every recorded stall site and failure type, and the scar rate at two decimals.
- A report with no stall sites and no failure types still renders its header and metrics rather than an empty list, and omits the `stall_sites:` / `top failures:` lines.
- The derived-origin label splits the way the code splits it: `external_agent`/`llm` render `unverified (derived)`, everything else renders `derived from <origin>`.
- End to end: a report built by `build_trajectory_report` from a real storage renders the stall site and scar rate it actually computed.

No production code is touched.

### One observation, not changed here

The issue's "what a fix looks like" suggests asserting that a single `TOOL_COMPLETED` breaks quiet. It does not, and never did: `is_quiet_window` at `src/continuum/analysis/trajectory_report.py:48` checks only `WORK_COMPLETED`, `TASK_UPDATED` and `DECISION_CREATED`, and that file is byte-identical between the issue's commit (`bbfdd74`) and `main`. I tested the behaviour the code has rather than the prose.

Whether that is a gap is a maintainer's call: `checkpoint/policy.py:62` classifies `TOOL_COMPLETED` as a `SIDE_EFFECT_EVENT` ("events that mean the outside world changed"), so a compaction window containing only successful tool calls currently counts as quiet and produces a trajectory report. That may be intended — the feature's notion of progress is completed work units and decisions, not individual tool calls — but if it is not, the tests above name the boundary the fix would move. I left the behaviour alone since this issue is scoped to tests.

## Related issues

Addresses #1235

## Types of changes

- Type: `tests`

## Breaking changes

No.

## How has this been tested

All commands run in an isolated worktree branched from `upstream/main` (commit `700228e`), Python 3.13.5, editable install:

```
pytest tests/test_trajectory_report.py -q        # 35 passed (was 8)
pytest -q                                        # full suite, exit 0
ruff check src/ tests/                           # clean, under ruff 0.16.5 (the pyproject pin)
ruff format --check src/ tests/                  # clean, under ruff 0.16.5
mypy src/continuum                               # Success: no issues found in 127 source files
```

The new tests were mutation-checked to confirm they are not vacuous: disabling `WORK_COMPLETED` handling in `is_quiet_window` fails 6 of them, and dropping the `stall_sites`/`top failures` lines in `render_trajectory_report` fails 2. Both mutations were reverted; no production file is modified.

## Checklist

Tests added or updated for the changed behaviour. Documentation updated where the change affects user-facing behaviour — not applicable, no behaviour changed. CI passes on the latest commit. Security-relevant changes state known limitations explicitly — not applicable.

## Additional context

The second commit re-syncs the documented test count, and it lands for exactly the reason the convention survey predicted: this PR's 27 new tests push the docs-count guard (`tests/test_docs_counts.py`, `TOLERANCE=30`) past tolerance — CI collects 2333 while README/CHANGELOG/CONTRIBUTING_ONBOARDING still documented ~2,278, a delta of 55. The count was already drifting (28 over) before this PR. Every documented site now states ~2,333 collected / ~2,280 passed, the figures the CI guard computes; a minimal local env collects 2305, still inside tolerance. No behaviour changed, and no CHANGELOG entry is added beyond that number bump in the existing "README refresh" entry.

## Summary by CodeRabbit

- **Tests**
  - Expanded validation for trajectory reports, including report identity, time windows, metrics, stalls, failures, lessons, and derived origins.
  - Added coverage for quiet-window behavior across empty windows, progress events, decision events, payload edge cases, and sequence boundaries.
  - Improved confidence in reports generated from stored data.


Cyrax321 (dismissed):
Thank you for contributing to Continuum, @Adhi1-2. This covers both untested ends of the trajectory report pipeline for #1235: the is_quiet_window trigger boundary per event type and payload, and the render_trajectory_report output including honest empty lessons and derived origin labels, reusing the existing storage fixtures.

I reviewed the eleven new cases and the docs re-sync, and I checked CI: the full matrix across Python 3.11 through 3.13 on all three platforms plus Postgres plus lint is green. No blocking issues found. LGTM, marking this as ready to pull. Thank you again for the thorough edge coverage.

PR #1254, opened 2026-09-17
test(analysis): cover is_quiet_window and render_trajectory_report directly (#1235)

### LINKED ISSUE #1235: tests: is_quiet_window and render_trajectory_report are the trajectory report's trigger and renderer, and neither has a direct test

soorajsuru2004-sudo:
Closes #1235.

`is_quiet_window` and `render_trajectory_report` are the two ends of the trajectory report feature — when it fires, and what a human sees — and neither was referenced by name in any test. The middle (build, record, digest) was already covered; this adds the edges.

## What changed

Only `tests/test_trajectory_report.py`. No source or behaviour change.

### `is_quiet_window` — the trigger

Direct tests over constructed events, no storage:

- An empty window is quiet.
- A window of recorded facts only (tool calls, tool completions, failed and unresolved actions) is quiet — these say work was *attempted*, not that it landed.
- `WORK_COMPLETED` breaks quietness; a zero count or a failed batch does not, and an unparseable count is treated as one unit of work rather than silently read as zero.
- `TASK_UPDATED` breaks quietness on a positive completion count; a missing or zero completion does not, and an unparseable completion is treated as progress rather than crashing the predicate.
- `DECISION_CREATED` always breaks quietness regardless of payload.
- The window boundary is start-exclusive and end-inclusive, asserted through `maybe_generate_trajectory_report`, so a busy event sitting exactly on `window_start` cannot suppress a report it does not belong to.

### `render_trajectory_report` — the renderer

- A report built by `build_trajectory_report` renders lines naming the report id, the window, the top stall site and the scar rate.
- A report with no lessons still renders an honest header and `scar_rate 0.00` rather than an empty string — the briefing layer concatenates these lines, so vanishing would read as silence.
- A report derived from `external_agent` or `llm` events is labelled `unverified (derived)`; a deterministic derivation is labelled with its origin instead.

## Verification

`pytest tests/test_trajectory_report.py` — 18 passed. `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for quiet-window classification and boundary handling.
  - Added tests for trajectory report rendering, including empty lessons and derivation-origin labels.


Cyrax321 (commented):
Thank you for contributing to Continuum, @soorajsuru2004-sudo. I reviewed this alongside the two sibling PRs covering the same #1235 ground, #1252 and #1251. Your new is_quiet_window and render_trajectory_report cases pass. Two things to be direct about.

First, #1252 by Adhi1-2 covers the same two functions with eleven cases reusing the existing fixtures, includes the docs re-sync, and is fully green across the matrix, so I have approved it as the one to land. Landing all three would triple cover the same surface and they all append to the same test file, so they cannot all merge cleanly.

Second, this PR as it stands cannot merge anyway: its Test jobs fail on stale documented test counts (the suite collects about 2,316 while the branch docs state about 2,278). Your added tests themselves pass.

Please close this PR as superseded by #1252. If you believe your cases pin an edge #1252 lacks, name the exact test here so it can be ported over rather than losing the coverage. Thank you again for taking this on.

Cyrax321 (changes requested):
Author context: soorajsuru2004-sudo, external contributor. Follow-up to the prior review. #1252 remains the approved coverage for the same #1235 surface and is fully green, so this stays superseded. Please close as superseded by #1252, or name the exact edge your cases pin that #1252 lacks so it can be ported rather than merged as a third copy.

tests/test_trajectory_report.py
```diff
@@ -364,3 +366,212 @@ def test_health_idle_trigger_generates_for_quiet_and_not_for_busy() -> None:
...
+
+# --- the trigger: is_quiet_window (issue #1235) ------------------------------ #
+
+
+def _event(type: EventType, payload: dict[str, object] | None = None, sequence: int = 1) -> Event:
+    """Build a bare, unsealed event for a pure predicate test (no storage)."""
+    return Event(run_id="run_1", sequence=sequence, type=type, payload=payload or {})
+
+
+def test_is_quiet_window_empty_window_is_quiet() -> None:
```
Cyrax321, line 379: Blocking: this test and the four that follow it duplicate coverage that is already merged on main. tests/test_trajectory_report.py on main now has test_is_quiet_window_empty_and_non_progressing_events_are_quiet (empty window, count 0, failed batch), test_is_quiet_window_progress_and_decisions_break_it (WORK_COMPLETED, TASK_UPDATED, DECISION_CREATED all break quietness), and test_is_quiet_window_rejects_an_unreadable_completed_counter (unparseable 'completed' reads as busy), asserting the same behaviours with the same payloads. Merging this would leave two near-identical test clusters for one predicate in one file, which is why the branch is already in conflict with main. Please rebase and drop the cases main already pins down.

tests/test_trajectory_report.py
```diff
@@ -364,3 +366,212 @@ def test_health_idle_trigger_generates_for_quiet_and_not_for_busy() -> None:
...
+            [
+                _event(EventType.TOOL_FAILED, {"tool_name": "edit"}, sequence=1),
+                _event(EventType.DECISION_CREATED, {"summary": "retry"}, sequence=2),
+            ]
+        )
+        is False
+    )
+
+
+def test_quiet_window_boundaries_are_start_exclusive_end_inclusive() -> None:
```
Cyrax321, line 449: Non-blocking: this is the part of the PR worth keeping. Main's merged wave covers the predicate and the renderer but does not assert the (window_start, window_end] selector semantics, so the claim that a busy event sitting exactly on window_start cannot suppress a report is genuinely new. When you rebase, keep this test and the unresolved/failed-actions case, and fold the WORK_COMPLETED absent/unparseable count cases into main's existing work-completed coverage.

tests/test_trajectory_report.py
```diff
@@ -364,3 +366,212 @@ def test_health_idle_trigger_generates_for_quiet_and_not_for_busy() -> None:
...
+
+    for i in range(2):
+        _add_failed_action(storage, run_id, "test.stall", f"stall_{i}")
+    ledger = ActionLedger(storage, run_id)
+    ledger.claim("test.scar", {"y": 1}, key="scar_0")
+    end = storage.last_sequence(run_id)
+    return build_trajectory_report(storage, run_id, 0, end)
+
+
+def test_render_trajectory_report_names_top_stall_and_scar_rate() -> None:
```
Cyrax321, line 513: Blocking: same duplication on the renderer side. Main already has test_render_trajectory_report_names_the_stall_sites_and_scar_rate, test_render_trajectory_report_is_honest_when_there_are_no_lessons, and test_render_trajectory_report_labels_a_derived_origin, which assert the same header fields, the same 'scar_rate 0.00' honesty check, and the same derived-origin labels this block adds. Retitling the same assertions does not add coverage; it adds a second copy to maintain whenever the renderer format changes.

tests/test_trajectory_report.py
```diff
@@ -364,3 +366,212 @@ def test_health_idle_trigger_generates_for_quiet_and_not_for_busy() -> None:
         assert direct is None
     finally:
         storage.close()
+
+
+# --- the trigger: is_quiet_window (issue #1235) ------------------------------ #
+
+
+def _event(type: EventType, payload: dict[str, object] | None = None, sequence: int = 1) -> Event:
```
Cyrax321, line 374: Non-blocking: after a rebase this helper collides with _ev, which the file on main already defines for exactly this purpose (bare, unsealed events with a caller-chosen sequence). Reusing _ev keeps one construction path for predicate fixtures and avoids two helpers that differ only in name.

Cyrax321 (changes requested):
The intent here is right: is_quiet_window and render_trajectory_report are the two ends of the trajectory report feature and they deserve direct tests. The problem is timing: since this branch was cut, main has merged its own wave of coverage for issue #1235 into this same file, so eight of the ten tests added here shadow tests that already exist on main with equivalent assertions, and the branch now conflicts with that file (mergeStateStatus DIRTY). On top of that the whole Test matrix is red, and one of the failures is this PR's to fix: tests/test_docs_counts.py fails with 'suite collects 2316 tests but docs say ~2278', because the ten new test functions push the suite past the documented-count tolerance without re-syncing README.md, docs/CONTRIBUTING_ONBOARDING.md, and CHANGELOG.md. The test_adapters_openai failure is pre-existing drift on the stale base and is not caused by this diff. Blocking merge until the branch is rebased and reduced to the real delta. The genuinely new coverage worth keeping after a rebase is the window-boundary test, the unresolved-and-failed-actions window, and the WORK_COMPLETED cases with an absent or unparseable count; the rest should be dropped in favour of what main already has.

PR #1255, opened 2026-09-17
fix(storage): guard the Postgres compaction bound the SQLite one already enforces (#1078)

### LINKED ISSUE #1078: compact_run's through_sequence anchor guard exists on SQLite but is absent on Postgres, so a Postgres run can archive and delete its own anchor marker

Adhi1-2:
## Description

`PostgresStorage.compact_run` was otherwise a line-for-line twin of the SQLite one — same forced anchor checkpoint, same single-transaction marker-plus-move, same `needs_fresh_anchor` recomputation, same `through < 1` refusal — but it had kept every safety check except the `through_sequence` anchor guard (issue #705). So an explicit `through_sequence` at or above the anchor marker's sequence ran the `DELETE … sequence <= %s` over the marker and every live row after it. The live log no longer retained its anchor, the next append minted a fresh genesis with `prev_hash = None`, and the live chain forked away from the archive — precisely the corruption the guard exists to prevent.

Reachability matches what #705 had on SQLite: the CLI's `compact` never passes `through_sequence`, so only a direct caller of the storage API can hit it. That is still a real hole for a library whose storage is a public interface, and for any future CLI flag that exposes the bound.

The fix lifts the guard into `Storage._validate_compaction_bound` on the shared base, so both backends call one definition and the two cannot drift apart again — the option the issue floated as the more robust of the two. The SQLite path now calls the shared helper instead of its inline copy; the error text is unchanged.

## Related issues

Fixes #1078

## Types of changes

- Type: `bugfix`

## Breaking changes

No. The SQLite behaviour and error message are byte-identical; the Postgres backend gains a rejection it was always documented (and is now docstring'd) to have.

## How has this been tested?

- New regression test `tests/test_storage_postgres.py::test_pg_compact_rejects_through_sequence_that_would_eat_the_anchor`, modelled on the SQLite suite's `test_compact_run_rejects_through_sequence_that_would_eat_the_anchor`. **Verified it fails against pre-fix code** (`DID NOT RAISE ValueError`) and passes with the fix — reproduced locally with a live Postgres 16 server.
- Full suite: `2272 passed, 6 skipped, 0 failed`, run against both backends (`CONTINUUM_POSTGRES_DSN` set so the pg suites actually execute):
  - `pytest tests/test_storage_postgres.py tests/test_provenance_postgres.py` — 20 passed
  - `pytest tests/test_compaction.py tests/test_compact_twice.py tests/test_cli_precompact.py tests/test_provenance_compaction.py` — 50 passed
  - `ruff check` / `ruff format --check` (ruff 0.16.5, the pinned version) and `mypy src/continuum/storage/{base,postgres,sqlite}.py` — clean
- One unrelated local failure, `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend`, is **pre-existing and environmental**: it expects `open_storage("postgresql://localhost/continuum")` to raise an error mentioning `psycopg`, which only happens when the driver is absent. My dev machine has both psycopg and a live server on `localhost:5432`, so the call connects and reports `database "continuum" does not exist` instead. I confirmed it fails identically on pristine `upstream/main`. CI is unaffected: the main test job installs no `postgres` extra, and the `test-postgres` job runs only `test_storage_postgres.py` and `test_action_index.py` as the `continuum` user against `continuum_test`.

Environment: macOS 26 (Darwin 25.0.0), Python 3.13.5, Postgres 16, editable install.


Cyrax321 (approved):
Thank you for contributing to Continuum, @Adhi1-2. This lifts the through_sequence anchor guard into Storage._validate_compaction_bound so both compaction backends share one definition for #1078, closing the Postgres hole where an explicit bound at or above the anchor marker archived the marker itself and forked the live chain away from the archive.

I reviewed the shared helper, both backend call sites, and the Postgres regression tests. SQLite behavior and error text are unchanged, and the full CI matrix plus lint is green. No blocking issues found. LGTM, marking this as ready to pull. Thank you again for choosing the drift proof shape the issue recommended.

PR #1257, opened 2026-09-17
fix: pin-marker surface no longer contradicts itself (#1099)

### LINKED ISSUE #1099: pin-marker surface contradicts itself: docs name pin_markers_for_state as the emitter, but context.py imports the private _pin_marker and the helper has no caller
## Summary

The constraint-pin marker surface contradicts itself three ways:

1. **The docs name a dead function as the emitter.** `docs/concepts/constraint-pinning.md:77` and `docs/guides/constraint-pinning.md:123` both say the `[pin:<constraint_id>:<sha256[:8]>]` markers are "Emitted by `pin_markers_for_state(state)`" — but `pin_markers_for_state` (`src/continuum/state/semantic.py:1024`) has no caller outside its own module (tests aside).
2. **The real emitter uses the private twin instead.** The code that actually builds the marker section, `checkpoint/context.py:_pins_section`, reaches across modules for the private helper: `from continuum.state.semantic import _pin_marker` (`src/continuum/checkpoint/context.py:141`), then `marker = _pin_marker(pin)` at :145 — a hand-rolled inline of the public one-liner `pin_markers_for_state` already is.
3. **The pin-accounting cluster is invisible on its own module.** `state/semantic.py`'s `__all__` (:55-61) lists five projection names and omits the entire pin cluster — `account_pins_in_context` (:938), `pin_markers_for_state` (:1024), `check_pin_accounting` (:1029), `constraint_pins_payload` (:1057) — while `checkpoint/context.py:34` imports `account_pins_in_context` by name and three test modules (`tests/test_pin_surface.py:20`, `tests/test_pin_reconstruction.py:17-18`, `tests/test_dropped_constraint_reconstruction.py:35-37`) import the rest.

## Affected component

Documentation; Other (state / constraint-pinning surface)

## Severity

Minor inconvenience or cosmetic issue — the mechanism works, but the docs describe a wiring that is not the one that exists, and a public helper and its private duplicate diverge in risk.

## System information

- Python: 3.13.5

```bash
git grep -w pin_markers_for_state -- src/
# src/continuum/state/semantic.py:1024:    def pin_markers_for_state... (definition only)

git grep -n '_pin_marker' -- src/continuum/checkpoint/context.py
# :141  from continuum.state.semantic import _pin_marker   <- private cross-module import
# :145  marker = _pin_marker(pin)
```

## What a fix looks like

Make `_pins_section` call `pin_markers_for_state` (deleting the private cross-module import — the function is exactly `[_pin_marker(pin) for pin in state.pins.values()]`), add the four pin helpers to `state/semantic.py`'s `__all__`, and the two doc sentences become true as written. Extending `tests/test_module_all_exports.py` to cover `state/semantic.py` prevents the `__all__` half from regressing.

soorajsuru2004-sudo:
## Description

The constraint-pin marker surface contradicted itself three ways (issue #1099):

1. The docs named a dead function as the emitter. `docs/concepts/constraint-pinning.md` and `docs/guides/constraint-pinning.md` both say the `[pin:<constraint_id>:<sha256[:8]>]` markers are "Emitted by `pin_markers_for_state(state)`", but that helper had no caller outside its own module (tests aside).
2. The real emitter used the private twin. `checkpoint/context.py:_pins_section` reached across modules for `from continuum.state.semantic import _pin_marker`, a hand-rolled inline of the public one-liner `pin_markers_for_state` already is.
3. The pin-accounting cluster was invisible on its own module. `state/semantic.py`'s `__all__` listed five projection names and omitted all four pin helpers, while `checkpoint/context.py` and three test modules import them by name.

This PR fixes all three, so the two doc sentences are now true as written.

**Approach.** `_pins_section` now builds its markers through the public `pin_markers_for_state` and the private cross-module import is gone. The pins are still sorted by `constraint_id` for deterministic output, and the markers returned by the emitter are keyed by `constraint_id` rather than positionally zipped, so the pairing is self-evidently correct regardless of dict order. `zip(..., strict=True)` makes a length mismatch between the pin set and the emitter's output fail loudly instead of silently mis-pairing. The four pin helpers are added to `state/semantic.py`'s `__all__`.

Rendered output is unchanged: verified byte-identical over five pins inserted in deliberately non-alphabetical order, which exercises the sort. No behaviour change, no gating change.

## Related issues

Fixes #1099

## Types of changes

- Type: `refactor`

## Breaking changes

No. `__all__` gains four names that were already importable by name, and the rendered recovery context is byte-identical.

## How has this been tested?

```
.venv/Scripts/python.exe -m pytest tests/test_pin_reconstruction.py tests/test_pin_surface.py tests/test_module_all_exports.py tests/test_dropped_constraint_reconstruction.py --no-cov
# 35 passed in 1.78s

.venv/Scripts/ruff.exe check src/ tests/      # on the changed files: clean
.venv/Scripts/ruff.exe format --check src/ tests/   # on the changed files: clean
```

New regression coverage:

- `tests/test_pin_reconstruction.py::test_rendered_section_spells_pins_the_way_the_emitter_does` - the rendered ACTIVE CONSTRAINTS section contains every marker the public emitter produces, and accounting reads all of them as `present`. This fails if the section hand-rolls its own spelling again.
- `tests/test_pin_reconstruction.py::test_context_module_does_not_reach_for_the_private_marker` - pins the private `_pin_marker` import out of `checkpoint/context.py` by name.
- `tests/test_module_all_exports.py::test_state_semantic_exports_pin_helpers` - all four pin helpers stay in `state/semantic.py`'s `__all__`. Adding the module to `test_all_symbols_exist_on_modules` also pins `__all__` against the module's own definitions, so a dropped or stray export fails the build.

Environment: Windows 11, Python 3.12, editable install from source, branch `fix/1099-pin-marker-surface` off `upstream/main` at b97256d.

Note on `mypy`: I could not run it locally. `mypy src/continuum` fails to start in this environment with `ImportError: DLL load failed while importing main: An Application Control policy has blocked this file`, so the strict type-check is unverified here and relies on CI. The change is annotation-neutral (no new signatures, `zip` with a literal `strict=True`).

## Checklist

Tests added or updated for the changed behaviour. Documentation updated where the change affects user-facing behaviour (the two doc sentences were already correct and now match the code; no doc edit needed). CI passes on the latest commit. Security-relevant changes state known limitations explicitly in this description.

## Additional context

Alternatives considered: making `pin_markers_for_state` sort internally and positionally zipping it against sorted pins was rejected, because two independently sorted sequences only correspond by an invariant a reader must prove; keying the pairing by `constraint_id` makes the correspondence self-evident.


Cyrax321 (changes requested):
Thank you for contributing to Continuum, @soorajsuru2004-sudo. This resolves the three way pin marker contradiction in #1099 by routing _pins_section through the public pin_markers_for_state emitter, removing the private cross module import, and exporting the four pin helpers from state/semantic.py so the docs sentences are true as written.

I reviewed the emitter change, the __all__ additions, and the reconstruction tests. I also verified the pairing the description relies on: pin_markers_for_state maps state.pins.values() in order, and _pins_section zips that output against the same dict, so the constraint_id keying is correct and zip strict fails loudly on length drift. The design looks right. Two items block merge, both in CI:

1. Stale documented test counts (tests/test_docs_counts.py). The suite collects 2,309 but the docs state about 2,278, diff 31 over tolerance 30. Please run pytest --collect-only -q, take the live total, and re-sync every stating file (README.md, the translated READMEs, docs/CONTRIBUTING_ONBOARDING.md, the current CHANGELOG entry, references/testing.md, references/install.md), leaving historical CHANGELOG notes untouched. Then run python -m pytest tests/test_docs_counts.py -k "not slow" -q locally to confirm green before pushing.

2. One Windows only failure in test_horizon_benchmark.py where benchmarks/run.py exceeded its 60 second timeout. Your change does not touch that path, so this reads as a slow runner flake. Please retry that job, and if it persists, report back so we can look together.

Once CI is green this will be ready for a final review. Thanks again for taking this on.

Cyrax321 (changes requested):
Author context: soorajsuru2004-sudo, external contributor. Follow-up to the prior review. The #1099 pin-marker direction is still correct, but the blockers still stand: stale documented test counts (branch pins older figures against a suite now near 2,452) plus dirty merge state. Re-run pytest --collect-only on a rebased tree, resync every stating file, and confirm the Test matrix green including a retry of the Windows horizon flake.

tests/test_module_all_exports.py
```diff
@@ -125,6 +126,7 @@ def test_all_symbols_exist_on_modules() -> None:
         recovery_notify,
         security_provenance,
         security_revalidation,
+        state_semantic,
```
Cyrax321, line 129: Blocking. This hunk no longer has anywhere to land: main rewrote this file in #1300 (#1228) and replaced the hand-rolled module lists with a `pkgutil.walk_packages` sweep in `_modules_with_all()`, so `test_all_symbols_exist_on_modules` and the star-import block now cover `continuum.state.semantic` automatically once the helpers are in `__all__`. The branch also conflicts in `CHANGELOG.md`, which gained new entries at the same `### Fixed` anchor, and the head CI run is red against the stale base (the docs-count guard reports 2309 collected vs ~2278 documented, plus an unrelated adapters/openai failure that is green on current main). Please rebase onto main; after the walk, only `test_state_semantic_exports_pin_helpers` below needs to carry over, optionally plus an `assert callable(pin_markers_for_state)` line in the rewritten star-import block.

tests/test_pin_reconstruction.py
```diff
@@ -213,3 +213,40 @@ def test_grace_window_default_advisory() -> None:
...
+        accounting = account_pins_in_context(state, rendered)
+        assert all(info["status"] == "present" for info in accounting.values())
+    finally:
+        storage.close()
+
+
+def test_context_module_does_not_reach_for_the_private_marker() -> None:
+    """``checkpoint/context.py`` must not import the private ``_pin_marker`` (issue #1099).
+
+    The private twin is an inline duplicate of the public one-liner. Reaching
```
Cyrax321, line 243: Non-blocking. The framing here is off: `_pin_marker(pin)` in `state/semantic.py` is not an inline duplicate of `pin_markers_for_state`; it is the primitive both the public emitter (semantic.py:1052) and `account_pins_in_context` (semantic.py:1005) are built on, so the spelling never actually diverged. The problem this PR fixes was the cross-module reach for a private name, and stating it as a duplicated one-liner teaches the next reader a wrong model of the module (the PR body and CHANGELOG entry repeat it). Suggest rewording to say the section bypassed the public emitter for a private helper.

tests/test_pin_reconstruction.py
```diff
@@ -213,3 +213,40 @@ def test_grace_window_default_advisory() -> None:
...
+    across modules for it is what let the two surfaces diverge in the first
+    place, so the import is pinned out by name.
+    """
+    from pathlib import Path
+
+    import continuum.checkpoint.context as context_mod
+
+    source = Path(context_mod.__file__).read_text(encoding="utf-8")
+    assert "_pin_marker" not in source
```
Cyrax321, line 252: Non-blocking. This is a whole-file substring assertion, so it fails on prose as well as code: a future comment in `checkpoint/context.py` that merely mentions `_pin_marker` (for example, explaining why the private import was removed) trips the guard without any import existing. Since the intent is to pin out the import and its calls, matching import/call shapes (e.g. `"_pin_marker("`, `"import _pin_marker"`) or walking the AST would pin behavior rather than wording.

Cyrax321 (changes requested):
This reroutes `_pins_section` through the public `pin_markers_for_state`, publishes the four pin helpers in `state/semantic.py`'s `__all__`, and pins both with regression tests; rendered output is correctly unchanged and the two doc sentences become true as written. What blocks the merge is mechanical, not conceptual: the branch conflicts with main (mergeStateStatus DIRTY) because #1300 rewrote `tests/test_module_all_exports.py` around a package walk and `CHANGELOG.md` gained entries at the same anchor, and the head CI run is red against that stale base. Rebase and re-shape the export-guard hunks per the inline note and this should be straightforward to land.

PR #1258, opened 2026-09-17
feat: wire the recovery anchor into `resume --repair` (#1097)

### LINKED ISSUE #1097: CheckpointManager.checkpoint_on_recovery is never called: RECOVERY anchors and the last_recovery_anchor lookup are dead surface
## Summary

`CheckpointManager.checkpoint_on_recovery` (`src/continuum/checkpoint/manager.py:336`) takes an unconditional `RECOVERY`-trigger anchor immediately after a non-RESUME decision, and `last_recovery_anchor` (:315) is the lookup that answers "where would I roll back to?". Neither has a caller in `src/` — six and three references respectively, all in tests. The project's own docs state this outright, `docs/ARCHITECTURE_EVOLUTION.md` §15.5:

> The auto-checkpoint hook is not yet invoked automatically by the adapter loop; callers must call `checkpoint_on_recovery` after a non-RESUME decision. Wiring is a small, well-isolated change left for the next step.

The knock-on makes the dead surface larger: `manager.py:349` (inside `checkpoint_on_recovery`) is the sole producer of `CheckpointTrigger.RECOVERY`, so the `keep_anchors` guard at `manager.py:373` and the `is_anchor` branch at `src/continuum/recovery/cleanup.py:49` protect a set that is always empty in-repo. (`prune` itself also has no `src/` caller.)

## Affected component

Other (checkpoint manager / recovery wiring)

## Severity

Minor inconvenience or cosmetic issue — no wrong behavior today, but a Phase 4 feature (lease-protected recovery anchors) that cannot occur through any product path.

## System information

- Python: 3.13.5

```bash
git grep -w checkpoint_on_recovery -- src/
# src/continuum/checkpoint/manager.py:336:    def checkpoint_on_recovery(   <- definition only

git grep -w last_recovery_anchor -- src/
# src/continuum/checkpoint/manager.py:315:    def last_recovery_anchor(     <- definition only
```

## What a fix looks like

**Option A: wire it.** One call in the post-decision path after a non-RESUME verdict (the engine caller or `cmd_resume`), which is exactly what `ARCHITECTURE_EVOLUTION.md` §15.5 describes as "a small, well-isolated change". `prune` and the anchor-preservation guards then have something real to protect.

**Option B: remove it.** Delete the hook, the lookup, the `RECOVERY` trigger, and the now-pointless anchor guards, until the wiring is actually wanted.

soorajsuru2004-sudo:
## Summary

Closes #1097 (Option A: wire it).

`CheckpointManager.checkpoint_on_recovery` and `last_recovery_anchor` had no
caller in `src/` — only tests — so `CheckpointTrigger.RECOVERY`, the
`keep_anchors` guard in `prune`, and the `is_anchor` branch in
`cleanup_ephemeral_artifacts` all protected a set that could never be populated
through any product path. `ARCHITECTURE_EVOLUTION.md` §15.5 called the wiring
"a small, well-isolated change left for the next step"; this is that change.

## What changed

- **Producer:** `continuum resume <run> --repair` now calls
  `checkpoint_on_recovery` after a non-RESUME verdict, **before** the
  `RECOVERY_STARTED` event so the anchor pins the state the verdict judged
  rather than the state after the repair bookkeeping landed in the log.
- **Consumer:** `continuum restore <run> --reason ... --to-recovery-anchor`
  resolves `last_recovery_anchor` and rolls back to it. It errors when no anchor
  exists and refuses a `--to`/`--anchor` given alongside — a restore discards
  `(anchor, head]`, so the target must be unambiguous.

## Why the write lives where it does

The write stays in the **acting caller**, never in `RecoveryEngine.assess`,
which keeps its read-only promise (judging is separate from acting). A plain
`resume` therefore stays read-only and records nothing — it is an inquiry a
caller may repeat while polling, and `test_resume_without_repair_is_still_read_only`
still passes unchanged.

The anchor deliberately records **no environment**. Stamping it with the current
environment would let a later `resume` diff that environment against itself and
read `RESUME` over still-stale state. Recording none forces dependency
revalidation instead — verified below.

## Correctness notes

- The anchor is written after the verdict is computed and emitted, so it cannot
  change the mode, the payload, or the exit code of the invocation that created
  it.
- An anchor failure is reported on stderr and swallowed: the verdict and its
  exit code are already final, and the caller still has the repair plan to act
  on.
- After the anchor becomes the latest checkpoint, a subsequent assessment gets
  an empty environment diff and flags dependencies as *unknown* (→ human
  revalidation), which is strictly more cautious than the pre-existing
  `REPAIR_AND_RESUME` — it never weakens a verdict.

## Verification

- `pytest --no-cov -q` — full suite green (2,312 collected; ~30 skips are
  environmental: Postgres without `CONTINUUM_TEST_POSTGRES_DSN`, adapter tests
  without optional deps).
- `ruff check src/ tests/ examples/` and `ruff format --check` — clean.
- `test_docs_counts.py` — re-synced the documented counts in README.md,
  `docs/CONTRIBUTING_ONBOARDING.md` and `CHANGELOG.md` (2,278 → 2,312
  collected); the pre-existing drift of 28 meant the guard was already one test
  away from failing.
- Manual end-to-end smoke test of both commands against a real SQLite store.
- `mypy src/continuum` **could not be run locally** — an Application Control
  policy blocks the mypy DLL in this Windows environment. The change is
  annotation-clean (`RecoveryDecision`, `StateCheckpoint`, `int | None`
  unchanged for `approve_restore`), but CI is the real check here.

6 new tests in `tests/test_cli.py`.


Cyrax321 (changes requested):
Thank you for contributing to Continuum, @soorajsuru2004-sudo. This wires the recovery anchor into resume --repair for #1097 (Option A) exactly where the architecture note said it belongs: the write lives in the acting CLI caller before RECOVERY_STARTED, RecoveryEngine.assess stays read-only, plain resume stays read-only, and restore gains a mutually exclusive --to-recovery-anchor target with clear errors for missing anchors and ambiguous targets.

I reviewed the anchor helper, the resume and restore call sites, the parser flag, and the new CLI tests. The design looks right, including best effort anchoring that can never change the verdict or exit code. Two items block merge, both procedural:

1. This branch conflicts with main (README.md among others). Please merge main or rebase and resolve the conflicts, keeping the current main wording where it has moved on.
2. There is no CI signal on this PR yet: no Lint or Test jobs have run. Once the conflicts are resolved, CI will run. Please confirm Lint and the Test matrix are green, and re-request review.

Once conflicts are resolved and CI is green this will be ready for a final review. Thanks again for taking this on.

Cyrax321 (changes requested):
Author context: soorajsuru2004-sudo, external contributor. Follow-up to the prior review. The #1097 design direction is still correct, but the procedural blockers still stand: dirty merge state with conflicts against main, and no green CI signal required before merge. Rebase, resolve conflicts keeping current main wording, confirm Lint and the full Test matrix green, then re-request review.

CHANGELOG.md
```diff
@@ -8,6 +8,21 @@ All notable changes to this project are documented here. The format follows
 
 ### Changed
 
+- **Recovery anchors are now produced by a product path (#1097).**
+  `CheckpointManager.checkpoint_on_recovery` and `last_recovery_anchor` had no
+  caller in `src/` — only tests — so the `RECOVERY` trigger, the `keep_anchors`
```
Cyrax321, line 13: Non-blocking: this line introduces two new em dashes, which runs against the AGENTS.md rule "No em dashes (U+2014) anywhere in commit messages, code comments, or docs." The touched line in references/cli.md also carries one (pre-existing text on a line this PR rewrites). CHANGELOG already has a couple of older violations, so this is about not drifting further from a stated rule with new content; parentheses or a sentence split keeps the same meaning.

src/continuum/cli/main.py
```diff
@@ -1493,6 +1519,19 @@ def cmd_resume(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
...
+    # pin the pre-failure state as a rollback point. This is the wiring
+    # ARCHITECTURE_EVOLUTION.md 15.5 calls out as a small, well-isolated
+    # change: the write lives here in the acting caller, never in
+    # `RecoveryEngine.assess`, which stays read-only (judging is separate from
+    # acting). The anchor precedes the RECOVERY_STARTED event below so it
+    # captures the state the verdict judged, not the state after this repair's
+    # bookkeeping landed in the log. Without --repair `resume` stays read-only
+    # (test_resume_without_repair_is_still_read_only), since a plain resume is
+    # an inquiry a caller may repeat while polling.
+    if args.repair and decision.mode is not RecoveryMode.RESUME:
```
Cyrax321, line 1532: Non-blocking: the comment above says the anchor "captures the state the verdict judged", which holds at the projection level but not quite at the sequence level on one path. When the verdict is REQUEST_HUMAN and `.continuum/webhooks.json` is configured, `_notify_blocked_run` runs a few lines earlier and appends NOTIFICATION_SENT / NOTIFICATION_FAILED before this line executes, so the pinned `source_sequence` sits after events the verdict never judged. The projector excludes those event types, so a restore to this anchor returns an identical state and keeps the notification rows, which makes the gap harmless today; flagging it only so the ordering guarantee is not later consumed at sequence granularity, for example by something that diffs `(assessed head, anchor sequence)`.

Cyrax321 (commented):
This wires `checkpoint_on_recovery` into `resume --repair` after a non-RESUME verdict and adds `restore --to-recovery-anchor` as the consumer. The placement is right: the anchor is taken after the verdict and before `RECOVERY_STARTED`, a failure to record it cannot reach back into the exit code, and the deliberately environment-free anchor does force revalidation, since the validator degrades dependencies to UNKNOWN when there is no snapshot to compare against under the default strict path. The six tests cover producer, consumer, refusal, and failure paths, and the docs-count resync matches the guard.

Nothing blocks merge; two non-blocking notes are inline. Two heads-ups beyond the code: the CHANGELOG "### Changed" section races with open PR #1290, so whichever lands second needs a rebase, and now that every `--repair` on a non-RESUME verdict writes an anchor while `prune` keeps every RECOVERY anchor, repeated repair polls will accumulate unprunable checkpoints, which is fine today but worth tracking.

PR #1261, opened 2026-09-17
fix(storage): fold the action index on wall-clock time across the archive boundary (#1054)

### LINKED ISSUE #1054: action index fold ranks an archived completion below another run's later live failure, re-opening a completed side effect
## Summary

The action index folds with a global last-write-per-key rule, but its ordering across runs is wrong after compaction. `src/continuum/storage/sqlite.py:661-669` gives archived rows negative order positions and live rows their rowid, on the stated reasoning that "everything in `events_archive` predates every live row of its run" (`sqlite.py:643-651`). Within one run that holds. The fold is not per run: it iterates `[*archived, *rows]` across every run in the store, and a key claimed with `scoped_to_run=False` lives in one store-wide namespace. So an archived `COMPLETED` in run A ranks below a *later* live `FAILED` in run B, and last-write-per-key selects the failure.

`rebuild_action_index` then overwrites the correct incremental row with the wrong folded one. A third run claiming that key sees `fresh: True` against a completed side effect and re-fires it. The incremental index is right and the rebuild makes it wrong, which is the inversion that matters: `rebuild_action_index` exists as the repair path, and here it introduces the corruption it is meant to remove. `PostgresStorage` folds identically over `ctid` and has the same defect.

## Affected component

continuum.storage (both engines: `_canonical_index_rows`, `rebuild_action_index`, `action_index_drift`)

## Severity

Major functionality broken, workaround exists. It re-opens a completed external side effect, the exact failure the idempotency ledger exists to prevent, and it does it through the repair command. It needs an unscoped key, two runs, and a compaction, so the trigger is narrow but genuinely unsafe. The workaround is to scope keys to a run, or never rebuild the index after compaction.

## Evidence

At HEAD `14f8a84`. Run B fails an unscoped key, run A later completes the same key, then run A is compacted and the index rebuilt:

```
BEFORE compact action_index: {..., 'run_id': 'rA', 'status': 'completed', 'updated_seq': 6}
compact_run('rA') -> {'archived': 3}
AFTER  compact action_index (incremental, still correct): {..., 'run_id': 'rA', 'status': 'completed'}
rebuild_action_index() -> corrections: 1
AFTER  rebuild action_index: {..., 'run_id': 'rB', 'status': 'failed', 'updated_seq': 4}
re-claim after rebuild -> fresh: True
  => the completed side effect is re-fired
```

The row flips from `rA/completed` to `rB/failed`. The incremental index stayed correct after compaction; only the rebuild corrupted it, and it reported the flip as one "correction". The existing test `tests/test_compaction.py::test_action_index_covers_the_archive_after_rebuild` passes because it only checks that the archived completion survives, never that a later live failure must not outrank it.

```python
import tempfile, os

db = os.path.join(tempfile.mkdtemp(), "a6.db")
store = SQLiteStorage(db)
for rid in ("rB", "rA"):
    store.create_run(Run(run_id=rid, goal="g"))
    store.append_event(rid, EventType.RUN_STARTED, {"goal": "g", "total": 1})

outB = ActionLedger(store, "rB").claim("refund", {"invoice": "inv-1"},
                                       scoped_to_run=False, key="shared-k")
ActionLedger(store, "rB").fail(outB.key, "gateway 500")          # earlier, FAILED
outA = ActionLedger(store, "rA").claim("refund", {"invoice": "inv-1"},
                                       scoped_to_run=False, key="shared-k")
ActionLedger(store, "rA").complete(outA.key, external_id="ext-1", result={"ok": True})

def row():
    with store._read() as conn:
        r = conn.execute("SELECT run_id, status FROM action_index WHERE key=?",
                         (outA.key,)).fetchone()
    return dict(r) if r else None

print("before:", row())                                          # rA completed
store.compact_run("rA")
print("after compact:", row())                                   # rA completed
store.rebuild_action_index()
print("after rebuild:", row())                                   # rB failed

again = ActionLedger(store, "rB").claim("refund", {"invoice": "inv-1"},
                                        scoped_to_run=False, key="shared-k")
print("re-claim fresh:", again.fresh)                            # True -> side effect re-fires
```

## Root cause

`3f7c480` (2026-08-24, "fix(storage): fold archived actions into the action index projection"). Its own commit message states the ordering rule and its scope assumption: "archived rows take negative positions below every possible live value, oldest first", which is only valid intra-run. It also notes "drift appears once per compaction until the index is rebuilt", which is the assumption this case inverts.

## Expected behaviour

Global last-write-per-key must reflect true wall-clock order across runs, so an archived completion always outranks an earlier live failure of the same unscoped key, and `rebuild_action_index` never downgrades a correct incremental row.

## Suggested fix shape

Order by real time rather than by table. Give archived rows their original `sequence` (or their original live rowid, carried into `events_archive` on insert) and fold on that, instead of the current archive-negative / live-positive split. At minimum, scope the ordering decision per run when resolving a key contested across runs. Add a regression test with the failed-then-completed cross-run pattern above, asserting the row stays `completed` after rebuild and the re-claim is not fresh.

## System information

- Python: 3.14.5

Adhi1-2:
## What

Fixes #1054 — the action-index fold ranked an archived `COMPLETED` below another run's *earlier* live `FAILED`, and `rebuild_action_index` (the repair path) then wrote that failure over the still-correct incremental row. The next claim of the unscoped key saw `fresh: True` against a completed side effect and re-fired it.

## Why it happens

`_canonical_index_rows` folds the log into `{key: (entry, order_seq)}` with a store-wide last-write-per-key rule. For a compacted store it iterated `[*archived, *rows]` — the whole `events_archive` first, then every live row — on the stated reasoning that "everything in `events_archive` predates every live row of its run." That holds *within one run*. The fold is not per run: an unscoped key (`scoped_to_run=False`) lives in one store-wide namespace, so the archive's rows compete with live rows from *other* runs, which number their rows independently. Once run A completes a key that run B had failed earlier, and run A then compacts, the archive-first ordering selects run B's stale failure:

```
BEFORE compact   action_index: {'run_id': 'rA', 'status': 'completed', ...}
compact_run('rA') -> {'archived': 3}
AFTER  compact   action_index: {'run_id': 'rA', 'status': 'completed'}   # incremental, still correct
rebuild_action_index() -> corrections: 1
AFTER  rebuild   action_index: {'run_id': 'rB', 'status': 'failed', ...}
re-claim after rebuild -> fresh: True   # completed side effect re-fired
```

The inversion is the part that matters: the incremental index was right after compaction, and the rebuild — the command that exists to remove drift — introduced the corruption and reported the flip as one "correction."

## The fix

Merge the two streams on `(timestamp, run_id, sequence)` so every row lands on one wall-clock timeline; last-write-per-key then means true order, and within a single run the archive still precedes the live tail, so an uncompacted store folds identically to before. Timestamps are ISO-8601 text with microsecond resolution on both engines, so the sort key is a plain string compare.

`order_seq` deliberately stays **table-relative** (live rows keep their rowid / ctid-stream index, archived rows keep a negative position below every live one). The `action_index` table holds one row per key, so that value never has to rank a live row against an archived one — it only has to agree between `_canonical_index_rows` and `action_index_drift`. Keeping it table-relative means an uncompacted store produces byte-identical folds to before this change.

Both engines are fixed: `SQLiteStorage` (`rowid`) and `PostgresStorage` (`ctid`).

The fold also degrades on an unparsable payload rather than raising. It reads raw rows with no chain check first (the CLI refuses to repair a store whose chain is broken, but the storage-level fold gets there first), so one corrupt entry has to skip rather than abort `rebuild_action_index` — the repair path for *every* key in the store. SQLite already skipped; Postgres raised. Both skip now, and both are covered.

## Tests

Two regression tests, both reproducing the issue's rA/rB repro directly:

- `tests/test_compaction.py::test_rebuild_keeps_an_archived_completion_above_an_earlier_live_failure` — run B fails an unscoped key, run A later completes it, run A compacts, then rebuild. Asserts the index row stays `rA/completed` through compact **and** rebuild, `action_index_drift() == 0`, `foreign_action` sees the completion, and a third run's claim of the same key returns `fresh is False` with the recorded `external_id`.
- `tests/test_storage_postgres.py::test_pg_rebuild_keeps_an_archived_completion_above_an_earlier_live_failure` — the same shape over the Postgres engine.

Plus the two covering the payload-degradation branch, one per engine:

- `tests/test_action_index.py::test_rebuild_skips_an_unparsable_payload` — corrupt the ACTION payloads, assert the fold is empty, the rebuild drops the stale row instead of raising, and drift is then zero.
- `tests/test_storage_postgres.py::test_pg_rebuild_skips_an_unparsable_payload` — the same over Postgres, scoped to one run because that suite shares one database across files.

The existing `test_action_index_covers_the_archive_after_rebuild` is untouched and still passes; per the issue it only checked that an archived completion *survives*, never that a later one must not be demoted.

### Before / after

Both fold-order tests fail on `upstream/main` with exactly the issue's evidence, then pass with this change:

```
sqlite:   assert {'run_id': 'rB', 'status': 'failed'} == {'run_id': 'rA', 'status': 'completed'}
postgres: assert {'run_id': 'pg_1054_b', 'status': 'failed'} == {'run_id': 'pg_1054_a', 'status': 'completed'}
```

## Verification

- `pytest` full suite (SQLite + a real Postgres 16 via `CONTINUUM_TEST_POSTGRES_DSN`): **2274 passed, 6 skipped, 0 failed**
- Re-ran after the two follow-up commits: full suite green (0 failures), `--cov` confirms the fold's payload-guard branch is now hit
- `ruff check` / `ruff format --check` (0.16.5): clean
- `mypy` on both changed modules: clean

One pre-existing, environment-only failure I did not touch: `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend` expects the error to mention `psycopg`, but this machine has psycopg installed and a live server, so it connects for real and reports `database "continuum" does not exist`. It fails identically on unmodified `upstream/main` and is unrelated to this change; CI's main job has no `postgres` extra and the Postgres job does not run `test_storage.py`.

The Postgres suite shares one database across files, so it needs a clean `continuum_test` before each run — CI gets that from a fresh service container.

## CI

`.github/workflows/ci.yml`: the Postgres job now collects and uploads coverage. The only matrix job that uploads installs `.[dev]` with no `postgres` extra, so `tests/test_storage_postgres.py` skips there and every line of `continuum.storage.postgres` — plus this PR's new Postgres test — reported as uncovered in the codecov patch, no matter how well the Postgres job actually exercised them. That job ran with no coverage collection at all. Codecov merges the two uploads per line, so the second engine now counts toward the patch target instead of being invisible to it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed action history ordering when archived and live events share an unscoped key.
  * Rebuilding the action index now preserves the most recent completed action across compaction.
  * Prevented completed side effects from being incorrectly treated as new and run again.
  * Ensured action lookups and subsequent claims consistently resolve to the completed action across supported storage backends.


Adhi1-2:
The `codecov/patch` check reports 50% of the diff hit against an 83.31% target. I dug into it locally and it is structural rather than a coverage regression in this PR.

The only coverage upload comes from the Linux Python 3.12 `Test` job, which installs `.[dev]` — without `psycopg`, so `tests/test_storage_postgres.py` skips wholesale and every changed line in `src/continuum/storage/postgres.py` is unreachable in the uploaded `coverage.xml`. The `Test (Postgres backend)` job *does* install `.[dev,postgres]` and runs the Postgres contract suite (including this PR's `test_pg_rebuild_keeps_an_archived_completion_above_an_earlier_live_failure`), but it has no `codecov-action` step, so its execution never reaches the report.

Roughly half of this PR's source diff is the Postgres twin of the SQLite fix, which is why the diff reads as 50%: the SQLite half is covered, the Postgres half is covered on a job that does not upload.

Two ways to close the gap, if you want it closed:
- add a `codecov-action` upload (with a `flags: postgres` entry, or relying on codecov's per-line merge) to the `test-postgres` job, or
- give the py3.12/Linux `Test` job a Postgres service and the `postgres` extra so the one uploading report sees the second engine.

Both are `.github/workflows/ci.yml` changes, so I have not made them here.

Cyrax321 (dismissed):
Thank you for contributing to Continuum, @Adhi1-2. This fixes the cross run fold inversion in #1054 by merging the archive and live streams on wall clock time instead of table order, in both the SQLite and Postgres backends, with deterministic tie breaks and a skip for unparsable payloads so one corrupt row cannot abort the repair path. The Postgres coverage upload is a sensible supporting change so that engine counts in the patch report.

I reviewed both _canonical_index_rows rewrites, the CI workflow change, the regression coverage, and the full CI matrix plus lint, all green. One observation, non blocking: cross run ordering now hinges on client timestamps, so skewed clocks could theoretically misorder two rows from different runs. Ties already break deterministically and the incremental path is unchanged, so this is strictly safer than the archive first order it replaces.

No blocking issues found. LGTM, marking this as ready to pull. Thank you again for the precise root cause writeup.

Cyrax321 (dismissed):

Ref: #1261 / Issue #1054
Surface: storage action index fold, sqlite.py and postgres.py rebuild path

Bug: the archive-first fold assumed everything in events_archive predates every live row, which only holds within one run. An unscoped key lives in one store-wide namespace, so after run A compacts, run B's earlier live failure of the same key outranked run A's later archived completion, and rebuild_action_index, the repair path, was what wrote the wrong row: the next claim saw fresh: True against a completed side effect and re-fired it. The failure mode is severe and silent, exactly the class #1054 describes.

Fix reviewed: the fold now merges both streams on one wall-clock timeline with a deterministic (timestamp, run_id, sequence) tie-break, so replay is stable across rebuilds. The order_seq bookkeeping stays table-relative (negative positions for archived rows, live indices for live rows), which is correct because the stored value only needs to agree with action_index_drift, never to rank across tables. Both engines carry the same fix, so the backends cannot drift apart.

Proof:
- Read the full _canonical_index_rows rewrite on both backends and traced the interleaved timeline case; last-write-per-key now falls out of real order rather than by accident.
- Regression tests cover the cross-run compaction scenario on both SQLite and Postgres, and the Postgres CI job is green.
- The only failing check is codecov/patch, which is advisory patch coverage, not a required test; every functional CI job passes at head 335b690e.

Note: this touches the same action-index subsystem as #1324, #1325, and #1323 in the 09-20 batch. Merge this first (it fixes the fold ordering the others depend on), then rebase the batch.

Cyrax321 (approved):
Author context: Adhi1-2, external contributor. Re-review at current head. The defect is real and severe: the archive-first fold assumed archived means older, which holds within one run but not store-wide, so after run A compacts, run B's earlier live failure of the same unscoped key outranked run A's later archived completion, and the repair path wrote the wrong row. Merging on (timestamp, run_id, sequence) with a deterministic tie-break puts last-write-per-key on real order, and keeping order_seq table-relative is correct since it only has to agree with action_index_drift. The wall-clock skew note is a fair observation but strictly safer than the order it replaces, and unparsable payloads skip rather than abort the repair. Both prior maintainer approvals were dismissed by rebases, so this fresh approval is for the current head. Merge state is dirty, so a rebase is still required; the codecov/patch gap is the known Postgres uploading-leg gap fixed by #1330.

src/continuum/storage/postgres.py
```diff
@@ -733,30 +733,57 @@ def action_index_drift(self) -> int:
...
-        for i, row in enumerate([*archived, *rows]):
-            payload = (
-                row["payload"] if isinstance(row["payload"], dict) else json.loads(row["payload"])
-            )
-            entry = index_entry_from_payload(EventType(row["type"]), payload)
-            if entry is not None:
-                canonical[entry[0]] = (entry, i if i >= offset else i - offset)
+        archived_seen = 0
+        for _ts, _run_id, _seq, live_idx, event_type, payload in timeline:
+            parsed = payload if isinstance(payload, dict) else json.loads(payload)
```
Cyrax321, line 780: Blocking. The body says both engines now degrade on an unparsable payload (`SQLite already skipped; Postgres raised. Both skip now, and both are covered`), but this `json.loads` is unguarded: the `payload` column is `TEXT NOT NULL`, so psycopg always returns a str here and this call always runs. WHY it matters: one corrupt ACTION payload in either events table raises out of `_canonical_index_rows`, which aborts `rebuild_action_index` and makes `action_index_drift` raise too, so the repair path dies for every key in the store while `continuum verify --index` crashes instead of reporting; the SQLite fold at sqlite.py:690 already guards this branch, so the fix's claimed parity is not in the diff. Please add the same `except json.JSONDecodeError: continue` plus the two cited covering tests, or narrow the body so the remaining Postgres gap is documented rather than asserted fixed.

src/continuum/storage/sqlite.py
```diff
@@ -645,34 +645,60 @@ def rebuild_action_index(self) -> int:
...
+            rows = [
+                (r["timestamp"], r["run_id"], r["sequence"], r["rid"], r["type"], r["payload"])
+                for r in conn.execute(
+                    "SELECT timestamp, run_id, sequence, rowid AS rid, type, payload FROM events"
+                )
+            ]
+        # Both streams arrive in their own insertion order; the merge is what
+        # interleaves them. Ties on the same instant break on (run_id,
+        # sequence) so the fold stays deterministic replay after replay.
+        timeline = sorted([*archived, *rows], key=lambda row: row[:3])
```
Cyrax321, line 684: Non-blocking. The fold winner is now chosen by wall-clock timestamp, but incremental maintenance still upserts by insertion order (`rowid` here, `nextval('action_index_ord_seq')` on Postgres). WHY it matters: if the clock steps backward between two writes of the same unscoped key (NTP correction, host suspend/resume), a rebuild ranks the genuinely later write below the earlier one and re-exposes a completed action as fresh, the same failure shape as #1054 under a different trigger, and `action_index_drift` will then never return 0 because the two orderings permanently disagree. Same-instant ties across runs also break on lexicographic `run_id`, which is deterministic but not causal. None of this is worse than the pre-fix fold and no better cross-table ordering is recorded today, so this should not block; please add one sentence to the docstring stating the monotonic-clock assumption so the next reader does not have to reconstruct it.

Cyrax321 (changes requested):
This PR rewrites `_canonical_index_rows` on both engines to merge the archived and live event streams on `(timestamp, run_id, sequence)` instead of appending the archive ahead of the live table, so an unscoped key's last write wins by wall clock across the archive boundary. That is the right shape for #1054, and the two fold-order regression tests repro the rA/rB inversion end to end on both engines.

Two things block merge. First, the Postgres fold still raises on an unparsable payload: `json.loads(payload)` at postgres.py:777 is unguarded on a TEXT column, while the SQLite fold skips, yet the body states both engines skip now and cites two tests (`test_rebuild_skips_an_unparsable_payload`, `test_pg_rebuild_skips_an_unparsable_payload`) that are not in this diff. Second, and related, the body describes a ci.yml coverage-upload change for the Postgres job that is also absent; `codecov/patch` fails on this PR for exactly the reason that missing change was meant to address, so either the commits were dropped in a rebase or the body predates them.

One non-blocking note inline: the fold now picks its winner by wall clock while incremental index maintenance still orders by insertion, so a backward clock step can make a rebuild demote a genuinely later write and drift will then never settle. Not worse than the unfixed behavior and there is no better cross-table order recorded today, but the assumption deserves one sentence in the docstring.

Adhi1-2:
Superseded by **#1337**.

This PR fixes the fold's *ordering* — the wall-clock merge that stops `rebuild_action_index` from demoting a completed row and re-firing the side effect — and that part is correct and worth keeping. I compared the two branches directly and found one thing this one leaves unfixed: `updated_seq` stays table-relative (live `rowid` / `ctid` index, archived rows a negative position), and `compact_run` never rewrites the index row. So after compaction the stored number is a live rowid while the fold recomputes a negative archived position, and `action_index_drift()` reports a false positive on a store nothing tampered with — that is issue #1322.

#1337 fixes both halves by making the number a property of the event (epoch microseconds of its own `timestamp`) rather than of its row, so the incremental writer and the fold agree whether the row is live or archived. It contains the same wall-clock merge this PR introduces, plus the regression tests for the repair path, and it stacks on #1323.

I'm leaving this open until #1337 merges so #1054 always has a PR attached — closing it is the maintainer's call. If you'd rather have one PR, #1337 is the broader fix and this one can close.

Co-Authored-By: Claude <noreply@anthropic.com>

Adhi1-2:
Closing as superseded by #1337, which also resolves #1054 but follows the action-event-count fold that shipped in main (#1321/#1322). This PR's wall-clock fold conflicts with that design, so #1337 is the single live fix for #1054.

Issue #1098, opened 2026-09-15
Follow-up to #392: recovery/derived.py is never imported, so both derived-artifact producers inline the stamping and the invariant check has no consumer


Adhi1-2:
## Summary

Follow-up to #392 (non-amplification invariant for derived artifacts). The module built for that invariant, `src/continuum/recovery/derived.py`, is never imported by anything in `src/`. It offers three things — `stamp_derived` (:14, stamp every derived artifact with the origin of its source events), `derived_label` (:22, the human rendering), and `is_derived_unverified` (:36, the invariant check) — and the two actual producers of derived artifacts each bypass it and inline the stamping:

- `src/continuum/recovery/summary.py:140` → `derived_provenance_for_events(events)`, then `"derived_origin": derived_origin.value` at :151
- `src/continuum/analysis/trajectory_report.py:192` → the same pattern

So `stamp_derived` is duplicated logic, and `derived_label` and `is_derived_unverified` have no consumer at all: nothing ever displays or enforces the non-amplification invariant the module exists for. Reference counts are 5/4/5, all in one test file.

## Affected component

Other (recovery provenance surface)

## Severity

Minor inconvenience or cosmetic issue.

## System information

- OS: macOS 26.0.1 (Darwin 25.0.0)
- Python: 3.13.5
- continuum-agent: from source, commit `579376d` on `upstream/main`
- Installed via: `uv pip install -e ".[dev]"`

## Command or code to reproduce

```bash
git grep -n 'recovery.derived\|recovery import derived' -- src/
# (no output)

git grep -w stamp_derived -- src/
# src/continuum/recovery/derived.py only

git grep -n 'derived_provenance_for_events' -- src/
# the two producers inline it instead of calling stamp_derived
```

## What a fix looks like

**Option A: wire it.** Route both producers through `stamp_derived` (deleting the duplicated inline stamping), use `derived_label` wherever a derived artifact is rendered (briefing output), and `is_derived_unverified` wherever derived artifacts are surfaced — making the #392 invariant observable.

**Option B: remove it.** Delete the self-contained 45-line module and its test file; recoverable from git history if the invariant enforcement is wanted later.

Adhi1-2:
Status check against `upstream/main` at `e150b47` (2026-09-23): **still open.**

[...]

Issue #392, opened 2026-08-26
feat(provenance): non-amplification invariant for derived artifacts


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

### What problem does this solve?

The project is about to add derived artifacts: attempt lessons (#313), trajectory reports, informed-retry summaries. Derivation has an authority hazard: an artifact built from events asserted by external agents can be projected back as if it were independently verified, laundering the weakest source's trust level into a stronger-looking claim. Nothing currently enforces that a derived object's authority is capped by its sources.

### Proposed solution

One invariant, enforced everywhere derivation happens:

> A derived artifact's effective authority equals the minimum authority of its source events, forever.

- Every derivation site stamps the minimum origin/trust of its inputs onto the output (new field on lesson/report/summary payloads).
- The projector marks derived components with the degraded origin; the validator treats them exactly like EXTERNAL_AGENT facts (REQUIRES_REVIEW where applicable).
- Briefing renders them labelled; they can never clear a review gate on their own.

Land this BEFORE #313 implementation so lessons are born compliant instead of retrofitted.

Scope: `provenance_map.py`, derivation helpers in `recovery/summary.py`, projection + validator touch points, briefing renderer.

Acceptance criteria:

- A lesson sourced only from EXTERNAL_AGENT events appears labelled unverified and cannot clear REQUIRES_REVIEW.
- Mixing in one trusted source does not upgrade the whole artifact (minimum wins).
- Existing artifacts without the new field degrade to unverified rather than erroring.

Falsifiable test: fabricate progress via MCP, derive a lesson from it, assert the lesson carries EXTERNAL_AGENT-derived marking and a resume built around it stays request_human.

Alternatives considered: case-by-case handling inside each consumer. Rejected: N consumers, N chances to forget; the invariant belongs at derivation time.

### Is this a breaking change?

No

### Would you like to work on this?

No

Blocks: #313 implementation, sleep-time consolidation reports. Sequential-chain head.

Cyrax321:
Claiming #392

PR #1262, opened 2026-09-17
fix(provenance): wire the derived-provenance module into its producers (#1098)

soorajsuru2004-sudo:
## What

`src/continuum/recovery/derived.py` was written to enforce the non-amplification invariant (#392) — a derived artifact is never more trustworthy than its least-verified source — but nothing in `src/` imported it, so the invariant was neither displayed nor observable anywhere (issue #1098). This wires it in (the issue's **Option A**) and deletes the duplicated inline logic.

## Changes

- **Both producers stamp through the shared helper.** `build_informed_retry` and `build_trajectory_report` each inlined their own `derived_provenance_for_events` lookup; both now go through `stamp_derived`, so the origin a derived artifact carries can no longer diverge between the two surfaces for the same events. The private `_derived_label` copy in `summary.py` is gone.
- **Both renderers render through `derived_label`.** The trajectory renderer's hand-rolled label printed `derived from ` (empty string) for a report that was never stamped; it now reads `unverified (derived from unverified sources)`, matching the informed-retry block.
- **The invariant is now observable where artifacts are surfaced.** `curate_briefing` calls `is_derived_unverified` and tiers a section built from agent-asserted events (trajectory reports, informed-retry block) as unverified agent material rather than system-derived, with `unverified provenance` in the section title.

## Behaviour

Output for a run whose sources are all trusted is byte-identical — `derived_label` and `stamp_derived` reproduce the previous label and origin exactly. Only artifacts derived from unverified events are now labelled and tiered differently.

## Tests

Adds `tests/test_derived_provenance_wiring.py` (6 tests) pinning the wiring so the module cannot silently go unused again: a spy on the shared helper at each producer, the shared label at each renderer (including the unstamped fallback), and the briefing tiering in both directions.

Targeted suites pass locally: `test_derived_provenance_wiring`, `test_provenance_non_amplification`, `test_informed_retry`, `test_trajectory_report`, `test_briefing_curation`, `test_module_all_exports`, `test_docs_counts`, `test_cli`, `test_dashboard_hitl`, `test_recovery_engine`. `ruff check` is clean on all touched files. `mypy` could not be run here — an Application Control policy blocks its DLL — but the changes are type-trivial (same-signature helper swaps).

The branch is rebased onto current `main`; the only conflicts were stale documented test-count strings, resolved in favour of upstream's newer figures, so this PR touches only the four files above.


Cyrax321 (changes requested):
Thank you for contributing to Continuum, @soorajsuru2004-sudo. This wires the shared derived provenance helper into both producers for #1098 (Option A) and removes the duplicated inline logic, so the origin a derived artifact carries can no longer diverge between the two surfaces.

I reviewed the producer changes in src/continuum/analysis/trajectory_report.py and src/continuum/recovery/summary.py, the tiering changes in src/continuum/recovery/briefing_curation.py, the six new wiring tests in tests/test_derived_provenance_wiring.py, and the CI results. The wiring itself looks right: trusted source output stays byte identical, unverified derivations are now labelled and tiered as agent material, and every functional test passes. One item blocks merge:

1. Stale documented test counts (tests/test_docs_counts.py). All 9 Test jobs fail: the branch docs state about 2,284 collected while the suite collects 2,316, and the stating files disagree among themselves. Please run pytest --collect-only -q, take the live total, and re-sync every stating file (README.md, the translated READMEs, docs/CONTRIBUTING_ONBOARDING.md, the current CHANGELOG entry, references/testing.md, references/install.md), leaving historical CHANGELOG notes untouched. Then run python -m pytest tests/test_docs_counts.py -k "not slow" -q locally to confirm green before pushing.

Once CI is green this will be ready for a final review. Thanks again for taking this on.

soorajsuru2004-sudo:
Thanks for the review. The requested count re-sync is pushed (75ca8d9).

**Test counts.** `pytest --collect-only` on this branch reports **2,324 collected**, so every stating file now reads 2,324 instead of relying on the 30-test tolerance: `README.md`, all five translated READMEs, `docs/CONTRIBUTING_ONBOARDING.md`, `references/install.md`, `references/testing.md`, and the current `[Unreleased]` CHANGELOG entry. Historical CHANGELOG notes are untouched, and the files agree with each other exactly rather than only within tolerance. `python -m pytest tests/test_docs_counts.py` (all 5 tests, including the slow live-collection check) is green locally, as are the PR's own suites (`test_derived_provenance_wiring`, `test_provenance_non_amplification`, `test_informed_retry`, `test_trajectory_report`) and `ruff check` on the touched files.

**About the failing Test jobs.** The remaining red check is not caused by this PR. All 8 failed jobs fail on a single test, `tests/test_adapters_openai.py::test_wrap_function_tool_invocation_binds_args_and_intercepts`, with:

```
assert "An error occurred while running the tool. ... 'FakeTC' object has no attribute '_function_tool_arguments'" == {'endpoint': 'https://x', 'method': 'POST'}
```

`openai-agents` is declared as `>=0.2` with no upper bound, so CI resolves a newer release than the one the test double was written against, and that release no longer carries `_function_tool_arguments` on `FunctionTool`. The test passes locally against `openai-agents` 0.22.2. The same test fails identically on unrelated branches that share none of this PR's changes, e.g. run 35362893856 on `fix/1111-module-all-omissions`, so it is a repo-wide dependency-resolution issue rather than a regression here. This PR's diff is four files (`summary.py`, `trajectory_report.py`, `briefing_curation.py`, and the new wiring test), none of which touch the adapters. I left it alone to keep this PR focused; happy to raise a separate issue or pin the SDK in a separate PR if that is preferred.

Ready for another look when you have a moment.

Cyrax321 (changes requested):
Author context: soorajsuru2004-sudo, external contributor. Follow-up to the prior review. The wiring direction for #1098 is still correct, but the stale docs-count blocker still stands and has grown: this checkout collects around 2,452 while the branch still pins older figures, and merge state is dirty. Re-run pytest --collect-only on a rebased tree and resync every stating file (READMEs, CONTRIBUTING_ONBOARDING, current CHANGELOG entry, references/testing.md, references/install.md), leaving historical notes untouched. Note this duplicates #1310 by Parthipashok04 on the same issue #1098; only one of the two should land. Confirm CI green after the rebase.

src/continuum/analysis/trajectory_report.py
```diff
@@ -215,7 +214,12 @@ def build_trajectory_report(
         stall_sites=stalls,
         top_failure_action_types=top,
         created_at=created,
-        derived_origin=derived_origin.value,
+    )
+    # A report is a derived artifact of its window, stamped through the shared
+    # non-amplification helper (#392) so its origin can never diverge from the
+    # one the informed-retry block would carry for the same events.
+    candidate = TrajectoryReport.model_validate(
```
Cyrax321, line 221: Non-blocking. Constructing the model, dumping it to JSON, stamping, and validating it back re-serializes every field (including the created_at datetime round trip) when stamp_derived only needs a plain dict. Build the dict literal, stamp it, and validate once; the dict-first shape in #1310 shows the simpler form. Same outcome, one validation instead of two, and the while-loop below then dumps a model you are already holding.

src/continuum/recovery/briefing_curation.py
```diff
@@ -196,14 +197,27 @@ def curate_briefing(storage: Storage, run_id: str, decision: RecoveryDecision) -
...
+        trajectory_unverified = any(
+            is_derived_unverified(report.model_dump(mode="json")) for report in reports
+        )
         sections.append(
             {
-                "title": "trajectory reports (sleep-time, system-derived)",
-                "provenance": _PROVENANCE_SYSTEM,
-                "reason": "distilled from archived history between sessions (issue #393)",
+                "title": "trajectory reports (sleep-time, system-derived)"
+                + (", unverified provenance" if trajectory_unverified else ""),
```
Cyrax321, line 213: Non-blocking. When trajectory_unverified is true the rendered title reads 'trajectory reports (sleep-time, system-derived), unverified provenance': the section still claims system-derived authority in the same breath as the caveat, and the title is what the briefing surfaces. Swap the qualifier rather than append to it, e.g. 'derived from unverified sources' as #1310 does, so the title stops contradicting itself.

Cyrax321 (changes requested):
The wiring itself is correct: both producers now stamp through stamp_derived with the same events argument their inline lookups used, both renderers go through derived_label, and the briefing tiering fails closed on a missing origin. What blocks merge is CI: all nine Test jobs fail, but the sole failure is tests/test_adapters_openai.py::test_wrap_function_tool_invocation_binds_args_and_intercepts, which this diff does not touch; the branch's merge base predates main's #1033 fix that builds FakeTC through ToolContext.__init__, so the stale test errors under openai-agents 0.22.2+. Rebasing onto current main should turn the matrix green, and the docs-count hunks will conflict there and should be re-synced (this PR also bumps collected to ~2,324 while leaving '~2,253 passed' unchanged, which is now internally inconsistent). The label wording and the briefing tier flip are user-visible behavior changes, so this warrants a CHANGELOG entry; the PR edits CHANGELOG counts but adds none.

Note that open PR #1310 implements the same Option A wiring for issue #1098, and it is the cleaner of the two: it builds the stamped dict directly instead of the dump/revalidate round trip, swaps the contradictory section title, removes the dead payload['derived_origin'] line in record_trajectory_report, and ships a CHANGELOG entry plus CLI end-to-end tests.

yuki-fuyutsuki:
**Code Review**

This PR correctly wires the previously dead `src/continuum/recovery/derived.py:14` helpers into both producers and both renderers, deleting the duplicated `_derived_label` inline copies and making `curate_briefing` surface `#392`'s non-amplification invariant as a provenance tier rather than only a label string. The consolidation is the right direction and the six wiring tests pin the integration against silent drift. Output for fully-trusted runs stays byte-identical, while unverified derivations now fail closed. Two non-blocking correctness/readability issues remain.

**medium**
`src/continuum/recovery/briefing_curation.py:213`
When `trajectory_unverified` is true the title becomes `trajectory reports (sleep-time, system-derived), unverified provenance`. The prefix still asserts system-derived authority in the same string that downgrades provenance to `agent`, which is the user-visible trust signal the briefing exists to make unambiguous. An operator scanning titles can reasonably treat the section as trusted despite the caveat. The same contradiction applies to the retry block at line 237. Swap the qualifier instead of appending it.

```suggestion
                "title": "trajectory reports (sleep-time, derived from unverified sources)"
                if trajectory_unverified
                else "trajectory reports (sleep-time, system-derived)",
```

**low**
`src/continuum/analysis/trajectory_report.py:221`
`build_trajectory_report` constructs a `TrajectoryReport`, dumps it to JSON, stamps the dict, then validates it back, re-serializing every field including the `created_at` datetime round-trip. The result is functionally correct but does two validations and an extra allocation where `stamp_derived` only needs a plain dict. Building the dict literal first, stamping, and validating once preserves the exact same origin and avoids the datetime serialization fragility and the hidden double-validate cost, as the dict-first shape in `#1310` demonstrates. The existing `while` loop should then operate on the already-validated model.

```suggestion
    candidate_dict = {
        "report_id": report_id,
        "window_start": window_start,
        "window_end": window_end,
        "compaction_seq": window_end,
        "attempts": attempts,
        "scar_rate": scar,
        "stall_sites": stalls,
        "top_failure_action_types": top,
        "created_at": created,
    }
    # A report is a derived artifact of its window, stamped through the shared
    # non-amplification helper (#392) so its origin can never diverge from the
    # one the informed-retry block would carry for the same events.
    candidate = TrajectoryReport.model_validate(stamp_derived(candidate_dict, events))
```


[github run](/Cyrax321/CONTINUUM/actions/runs/35773008481)

Cyrax321 (changes requested):
Hi @soorajsuru2004-sudo,

This PR is currently blocked by failing CI tests and has become redundant due to a competing implementation.

1.  **CI Failures**: All test jobs are failing due to `test_adapters_openai.py`, which is unrelated to your changes but indicates the branch is out of sync with `main`. Please **rebase** onto the latest `main` to resolve this.
2.  **Redundancy**: As noted in the review, PR **#1310** implements the same fix for issue #1098 with a cleaner approach (dict-first construction, corrected titles, and CLI tests).

Given that #1310 is further along and addresses the review nits more comprehensively, please consider **closing this PR** in favor of #1310. If you believe your implementation offers distinct value, please rebase, resolve the conflicts, and address the remaining review comments regarding title wording and model validation efficiency.

Issue #1055, opened 2026-09-15
gateway buffers the entire upstream response with no cap while the threat model says reads are capped


Cyrax321:
## Summary

The gateway caps the request body at 10 MB and refuses 413 before reading it, but buffers the entire upstream response with no limit at all. `_handle` calls `resp.read()` into memory, then writes the buffer back to the client. A hostile or misbehaving upstream can therefore hold the agent's own proxy hostage by sending a large body, and `docs/threat_model.md` at line 32 claims this class is fended off. The documented cap covers only the inbound half of the proxy.

## Affected component

src/continuum/gateway.py: response read, request cap (`MAX_BODY_BYTES`), drain bound. Documentation claim at docs/threat_model.md.

## Severity

Major functionality broken, denial of service. The gateway is a local proxy bound to 127.0.0.1 that an agent's outbound traffic is pointed at, so an oversized upstream reply degrades the agent's own process and memory, not a remote tenant's.

## Evidence

At HEAD `14f8a84`, a live gateway in front of a TLS upstream returning a 128 MiB JSON body for a claimed invoice post:

```
response bytes: 134217719
peak traced allocation: 320 MiB
peak RSS: 54 -> 494 MiB
```

The response is over twelve times the request cap the same file enforces, and nothing in the code path limits it.

## Command or code to reproduce

Builds on the gateway harness from the route-prefix issue. Run with `uv pip install -e ".[dev]"`:

```python
import http.client, json, os, resource, tempfile, tracemalloc
from pathlib import Path
# ... same gateway + TLS upstream setup as the prefix repro ...
RESPONSE = 128 * 1024 * 1024

upstream(lambda path: (200, '{"blob":"' + "x" * (RESPONSE - 20) + '"}'))
server = gateway(db, work / "gateway.json")

before = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
tracemalloc.start()
conn = http.client.HTTPConnection(f"127.0.0.1:{server.port}", timeout=60)
conn.request("POST", "/v1/invoices", body=json.dumps({"id": "I-1"}), headers={"Host": HOST})
resp = conn.getresponse()
n = len(resp.read())
conn.close()
_, peak = tracemalloc.get_traced_memory()
tracemalloc.stop()
after = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss

print(f"response bytes: {n}")
print(f"peak traced allocation: {peak / 1024 / 1024:.0f} MiB")
[...]

PR #1263, opened 2026-09-17
fix(gateway): cap the upstream reply, not just the request body (#1055)

Adhi1-2:
## What

The gateway capped what a client could make it hold, but not what an upstream sent it. `MAX_BODY_BYTES` refused an oversized request body with `413` before reading it; the reply half then buffered the whole upstream response with `resp.read()` and no limit at all (#1055).

A hostile or merely broken upstream sending a large reply therefore held the agent's own proxy and memory hostage — the exact DoS `docs/threat_model.md` describes as fended off, covered on only one of the proxy's two directions.

## Fix

- **`MAX_RESPONSE_BYTES`** (`MAX_BODY_BYTES`' sibling, 10 MB) and **`_read_bounded_response`**: a reply whose declared length passes the cap is refused with `502` before it is buffered; a reply that declares no length at all — a chunked reply, which carries no length to check up front — is bounded by the running total rather than read whole. Reads in 64 KiB chunks so the working buffer stays cap-sized: the fix refuses the body without introducing a second unbounded allocation.
- The claim settles **uncertain, not completed**. An upstream that cannot answer inside the cap may still have applied the side effect, so `unknown` is the honest state and replay re-asks rather than trusting a completion that was never echoed (`fail(..., certain=False)`).
- `docs/threat_model.md`: the body-cap bullet now describes the reply half, with citations.
- Docs count re-synced (2,278 → 2,313 collected).

## Tests

Three new cases in `tests/test_gateway.py`, all failing pre-fix (the first two silently echoed 128 MiB / 8 KiB with `200` instead of refusing):

- declared `Content-Length` over the cap → `502` and `side_effect_uncertain is True`
- no declared length, streaming past the cap → `502` (the running-total path, not the declared check)
- a reply within the cap is forwarded byte-identical and the action completes

## Verification

- full suite green with a live Postgres (`CONTINUUM_TEST_POSTGRES_DSN`): no failures, no errors
- `ruff check`, `ruff format --check` (0.16.5), `mypy` — all clean


Cyrax321 (dismissed):
Thank you for contributing to Continuum, @Adhi1-2. This caps the upstream reply half of the gateway proxy with bounded 64 KiB chunked reads, a declared length fast path to 502, and honest uncertain settlement so replay re-asks instead of trusting an un echoed completion. I reviewed the helper, the call site, the threat model update, and the three regression tests covering declared over cap, undeclared streaming past the cap, and verbatim forwarding within the cap. Full CI matrix plus lint is green. No blocking issues found. LGTM, marking this as ready to pull. Thank you again for closing a real DoS gap.

Issue #390, opened 2026-08-26
feat(budgets): bind retry budgets to authorization instances so fresh keys cannot reset them


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

### What problem does this solve?

Run-level retry budgets (#240, shipped) cap attempts per action type. The counter resets whenever a caller arrives with a fresh idempotency key, so one logical authorization can be executed repeatedly under freshly minted keys: each attempt looks like a brand-new operation to the budget registry, and the cap never binds.

### Proposed solution

Bind budget consumption to the authorization instance, not only the action type:

- Introduce a durable monotonic counter keyed by `(run_id, action_type, authorization)` where authorization is derived from the stable key when present and from identity tokens otherwise.
- Confirmation events consume the same budget: an authorized operation gets an execution budget, and every claim slot and every confirmed completion draws it down.
- Fresh keys for the same authorization do not reset the counter; the registry refuses claims once the authorization's budget is exhausted, with an error that names both the type cap and the authorization id.
- Backward compatible: runs and configs without authorization data behave exactly as today.

Scope: `.continuum/budgets.json` schema gains optional authorization-bound entries; `budgets.py`, `actions/idempotency.py`, gate claim path.

Acceptance criteria:

- Same send attempted under N distinct fresh keys exhausts the authorization budget on the Nth attempt regardless of key freshness.
- Distinct authorizations keep independent budgets.
- Unset authorization fields leave behaviour byte-identical to today.

Falsifiable test: loop intercept_action over generated unique keys for one logical resource, assert refusal at cap with the authorization id named, assert today's code passes the loop (demonstrating the hole).

Alternatives considered: key-normalization hardening alone (already shipped). Necessary but insufficient: a determined reissuer mints genuinely new keys.

### Is this a breaking change?

No

### Would you like to work on this?

No

Parallel-safe with the edit-precondition and constraint-pinning tracks.

Cyrax321:
Verified on origin/main@58d222f. Authorization-bound budgets track already on main via PRs #411-414 merged and #415 verified closed via PR #483, with deterministic identity, drawdown and refusal surfaces. All 5 children 411-415 CLOSED. Closing parent. Refs #390

PR #1264, opened 2026-09-17
fix(budget): padded argument token can no longer move the authorization bucket (#1052)

### LINKED ISSUE #1052: authorization-bound retry budget is bypassable by padding one throwaway argument token
## Summary

The authorization-bound retry budget derives its bucket from the argument tokens alone, which means padding a request with one throwaway token moves it to a fresh bucket at full `max_attempts`. `_budget_authorization_id` deliberately ignores the explicit idempotency key so that freshly minted keys cannot reset the cap. That closes the fresh-key bypass from #390 and opens the inverse one: keep the key fixed, keep the real arguments fixed, vary a noise field, and the cap never binds. A budgets.json cap of 2 that correctly refuses a third attempt on identical arguments stays open indefinitely when each retry carries a new `trace_id`.

## Affected component

src/continuum/actions/ledger.py: `_budget_authorization_id`, called from `claim`. src/continuum/actions/idempotency.py: `resolve_authorization_id` and the token derivation.

## Severity

Major functionality broken, workaround exists. Declaring the noise field `volatile` at every claim site would exclude it, but the caller supplies the arguments and the harness cannot know which fields are incidental, so this is not enforceable from configuration alone. The cap that exists to stop an agent loop hammering a failing upstream never binds.

## Evidence

At HEAD `14f8a84` with a cap of 2 and a fixed idempotency key:

```
baseline: same arguments every time, fixed idempotency key
  attempt 1: CLAIMED then failed
  attempt 2: CLAIMED then failed
  attempt 3: REFUSED -> budget exhausted for 'send_invoice' / '3c532d13...' (2 of 2 used, 0 remaining)
  attempt 4: REFUSED -> budget exhausted for 'send_invoice' / '3c532d13...' (2 of 2 used, 0 remaining)

attack: fixed key, one throwaway argument token varied per attempt
  attempt 1: CLAIMED then failed
  attempt 2: CLAIMED then failed
  attempt 3: CLAIMED then failed
  attempt 4: CLAIMED then failed
  attempt 5: CLAIMED then failed
  attempt 6: CLAIMED then failed

buckets created for action send_invoice: 7
  3c532d13...: counter=2 of max_attempts=2
  4d9d62c9...: counter=1 of max_attempts=2
  ...five more, each counter=1 of max_attempts=2
```

The cap is real and enforced in the baseline. Six padded attempts created six fresh buckets that each started at their full allowance.

```python
import json, os, tempfile
from pathlib import Path
from continuum.actions.ledger import LedgerError

work = Path(tempfile.mkdtemp())
home = work / "home"
(home / ".continuum").mkdir(parents=True)
(home / ".continuum" / "budgets.json").write_text(json.dumps({"default_max_attempts": 2}))
os.chdir(home)
os.environ["HOME"] = str(home)
os.environ.pop("CONTINUUM_BUDGETS_PATH", None)

db = str(work / "run.db")

def attempt(n, pad):
    args = {"invoice_id": "INV-401"}
    if pad is not None:
        args["trace_id"] = pad
        try:
            outcome = ledger.claim("send_invoice", args, key="invoice:INV-401")
            ledger.fail(outcome.key, "upstream 503", certain=True)
            return f"attempt {n}: CLAIMED then failed"
        except LedgerError as exc:
            return f"attempt {n}: REFUSED -> {exc}"

print("baseline: fixed key, identical arguments")
for i in range(1, 5):
    print("  " + attempt(i, None))
print("attack: fixed key, one throwaway token varied")
for i in range(1, 7):
    print("  " + attempt(i, f"pad-{i:04d}"))
```

## Root cause

Introduced by `921a272` (2026-08-28, "feat(budgets): bind budgets to authorization via drawdown (Refs #413)"). The docstring in `_budget_authorization_id` records the reasoning: passing the explicit key "would give each fresh key its own bucket and defeat the cap-amplification fix", so the helper derives from resource tokens alone. Both halves are true and together they are the bug. The bucket follows the arguments, and the arguments are caller-controlled noise plus the real resource. `docs/api/security.md` documents the guarantee this defeats, that one logical operation draws from one bucket even when callers mint fresh keys.

## Expected behaviour

An operator-set cap of 2 binds the effect being retried. Two failed sends of invoice INV-401 exhaust the budget and a third is refused, whether or not each retry pads a trace id, a request id, or any other field the operation's identity does not depend on. Fresh-key minting must still share the bucket, so the fix cannot simply bind to the key.

## Suggested fix shape

Bind the bucket to the identity the ledger already dedups on. When `claim` receives an explicit key, the effect is named by that key, so the bucket is keyed off the action type plus that key. When the key is derived from arguments, restrict token extraction to the fields the rendered key depends on, the key-template fields in the gateway path and the corresponding derivation in the ledger path, so noise can neither move the bucket nor mint a fresh one. Keep the current unrestricted token set only as the fallback when no key is available. Note the fix is not to declare `trace_id` volatile at call sites: that moves the trust decision to every caller, and a caller that wants around the cap simply forgets to declare it.

## System information

- Python: 3.14.5

Parthipashok04:
## Summary

Fixes #1052. The authorization-bound retry budget derived its bucket from every
argument token, and arguments are caller-controlled noise plus the real
resource. Holding the idempotency key fixed while varying one throwaway field
(a `trace_id`, a request id) moved every retry into a fresh bucket at its full
allowance, so an operator-set cap never bound.

## Root cause

`ActionLedger.claim` called `_budget_authorization_id(action_type, None, arguments, volatile)`.
The helper deliberately ignores an explicit key — that is what keeps fresh-key
minting sharing one bucket (#390, #413) — and derives from resource tokens
alone. But the token set is computed from *every* argument, so it follows the
noise as much as the resource. The two halves of the reasoning in
`_budget_authorization_id` are both true, and together they are the bug.

Reproduced against the issue's script at `main` (cap of 2, fixed key):

```
baseline: attempt 1 CLAIMED, attempt 2 CLAIMED, attempt 3 REFUSED
attack:   attempt 1..6 all CLAIMED  (7 buckets, each counter=1)
```

## The fix

The bucket now follows the identity the ledger has *already decided this claim
is*. `claim` resolves `existing` — the record the attempt defers to, whether
because the caller held the idempotency key fixed or because the token fallback
recognised a prior attempt — and derives the bucket from that record's stored
arguments. Those name the operation; the incoming arguments name the operation
plus whatever the caller padded this time.

After the fix, the attack is byte-for-byte the baseline:

```
attack: attempt 1 CLAIMED, attempt 2 CLAIMED, attempt 3 REFUSED  (1 bucket, counter=2)
```

This is deliberately *not* binding the bucket to the explicit key: that is the
#390 amplification hole, and `tests/test_bench_fresh_key_reissuance.py` still
pins fresh-key minting to one bucket. It is also not declaring `trace_id`
`volatile` at call sites — that moves the trust decision to every caller, and a
caller that wants around the cap simply forgets to declare the field.

### Pinned on the record (review follow-up)

Deriving per call still left two openings, both flagged in review:

- The token derivation reads the caller's `volatile` sequence to decide which
  arguments to drop. A retry that flipped its declaration re-derived a fresh
  bucket and started its count over.
- `complete`/`reconcile` derive with no `volatile` sequence, so a claim that
  *did* declare one settled against a different bucket than the attempt it
  settled.

The resolved id is now stored on the record: `Action.budget_authorization_id`.
The claim resolves it once and every later attempt or settlement reads it back,
so neither the caller's declaration nor the settlement path can move the bucket
after the fact. Settlements fall back to deriving from the stored arguments only
when the claim left no id. `claim` copies the id through the FAILED and
COMPENSATED re-claim paths as well.

Tested by `test_changing_the_volatile_declaration_cannot_move_the_bucket` and
`test_settlement_uses_the_bucket_the_claim_pinned`.

### Documented residual

A caller that mints both a fresh key *and* fresh noise per attempt presents no
identity the ledger can see, and stays on the token fallback as before. Nothing
in the arguments distinguishes resource from noise at that point, and the issue
accepts this rather than moving the trust decision to callers.

## Changed

- `src/continuum/actions/ledger.py` — derive the claim bucket from the deferred-to
  record's arguments, and resolve it once per record; `complete`/`reconcile`
  settle against that stored id.
- `src/continuum/models.py` — `Action.budget_authorization_id`, the bucket an
  attempt draws from, round-tripped with the rest of the record.
- `tests/test_budget_drawdown.py` — four new tests: padding cannot move the
  bucket, the volatile declaration cannot move it, settlement uses the bucket
  the claim pinned (each confirmed to fail on unpatched `ledger.py`), and the
  fresh-key-plus-fresh-noise case is pinned as the documented residual.
- `docs/api/security.md`, `CHANGELOG.md` — the guarantee and its residual.
- README.md, the five translated READMEs, `docs/CONTRIBUTING_ONBOARDING.md`,
  `references/install.md`, `references/testing.md` — re-synced documented suite
  count; the docs-count guard (#630) fails a gap over 30.

## Verification

- Full suite: **2276 passed, 38 skipped, 2314 collected** (exit 0).
- `ruff check src/ tests/ examples/` and `ruff format --check` clean;
  `mypy src/continuum` clean (127 files).
- `tests/test_docs_counts.py` green against the re-synced figures.
- The new regression tests were confirmed to fail on unpatched `ledger.py` and
  pass with the fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Retries now remain tied to the original budget bucket even when callers change throwaway request or tracing details.
  - Confirmations and retries consistently share the same budget, preventing retries from resetting usage limits.
  - Attempts using both a new idempotency key and new request details continue to use token-based fallback behavior.

- **Documentation**
  - Added guidance explaining budget-bucket behavior, retry handling, and the remaining token-fallback scenario.
  - Updated documented test-suite counts across project guides and translated READMEs.


src/continuum/actions/ledger.py
```diff
@@ -974,7 +987,22 @@ def claim(
...
+        # that record's arguments. Deriving it from the incoming arguments
+        # instead would make the bucket follow caller-controlled noise: one
+        # throwaway ``trace_id`` moved every retry into a fresh bucket at its
+        # full allowance, and the cap never bound. The stored arguments are
+        # also what the settlement paths below derive from, so a retry and its
+        # confirmation now draw from one bucket by construction.
+        budget_arguments = existing.arguments if existing is not None else arguments
+        budget_auth_id = self._budget_authorization_id(
+            action_type, None, budget_arguments, volatile
+        )
```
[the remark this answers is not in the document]
Parthipashok04: Fixed in 840b6c0, along the lines of the suggested change.

The resolved bucket is now stored on `Action.budget_authorization_id` and round-trips with the rest of the record. `claim\ resolves it once — still anchored to the record it defers to, per the earlier commit — and every later attempt or settlement reads it back. `complete` and `reconcile` settle against the stored id, falling back to deriving from the stored arguments only when the claim left none. The FAILED and COMPENSATED re-claim paths copy the id through as well, so a retry cannot pick a fresh bucket by flipping its `volatile` sequence.

One deliberate deviation: I did not gate on `model_fields_set` to distinguish an explicitly persisted `None` from a legacy record lacking the field. `None` is only ever resolved when the arguments carry no distinctive tokens (or no budget registry exists), in which case re-deriving is deterministic and yields `None` again — the `volatile` sequence cannot change that, since there are no tokens to drop. So both cases behave identically, and the field stays a plain defaulted attribute rather than one the reader has to reason about in two modes. Happy to add the `model_fields_set` gate if you'd rather have the persisted `None` win when a budget registry appears mid-run.

Regression tests added: `test_changing_the_volatile_declaration_cannot_move_the_bucket` (fixed key, `volatile=["trace_id"]` → `[]` → `["trace_id"]`, cap 2, third attempt refused) and `test_settlement_uses_the_bucket_the_claim_pinned` (claim declares `volatile`, `complete` settles into the same single bucket). Both fail on the pre-fix code.

Cyrax321 (dismissed):
Thank you for contributing to Continuum, @Parthipashok04. This closes the authorization bucket bypass in #1052 by pinning the bucket onto the action record at claim and reusing it at settlement, with a documented fallback for legacy records. I reviewed the claim, complete, and reconcile paths, the new model field, and the regression coverage, and I checked the full CI matrix plus lint. The #390 fresh key behavior is preserved and the CodeRabbit persistence point is addressed at head. No blocking issues found. LGTM, marking this as ready to pull. Thank you again for the thorough writeup.

src/continuum/actions/ledger.py
```diff
@@ -977,7 +1003,33 @@ def claim(
...
+        # re-derived, because token derivation also reads the caller's
+        # ``volatile`` declaration, and a retry that changes either the
+        # arguments or that declaration would compute a different bucket from
+        # the same stored record. Pinning it once means a retry, its
+        # confirmation and its reconciliation all draw from one counter by
+        # construction, whatever the caller sends later. Records written
+        # before the field existed carry None and are re-derived here, which
+        # pins them at the first claim that sees them.
+        budget_arguments = existing.arguments if existing is not None else arguments
+        budget_auth_id = existing.budget_authorization_id if existing is not None else None
```
Cyrax321, line 1028: Pinning the resolved id on the record, and reading it back on every later claim, is what makes a retry that flips its `volatile` declaration a no-op instead of a bucket move. Not binding the bucket to the explicit key is the right call here: that is the #390 amplification hole, and `tests/test_bench_fresh_key_reissuance.py` still pins fresh-key minting to one bucket. Non-blocking, no change requested.

src/continuum/actions/ledger.py
```diff
@@ -447,6 +460,19 @@ def _budget_authorization_id(
...
+
+        The claim that opened the slot pinned its bucket onto the record
+        (issue #1052), so the settlement reads it back and the two share one
+        counter whatever the caller sends between them. A record written before
+        the field existed carries ``None`` and is re-derived from its stored
+        arguments with no volatile declaration, which is the pre-#1052 rule.
+        """
+        if action.budget_authorization_id is not None:
+            return action.budget_authorization_id
+        return self._budget_authorization_id(action.action_type, None, dict(action.arguments), ())
```
Cyrax321, line 474: Non-blocking, and only for records written before this field existed: the fallback re-derives with no `volatile` declaration, so if such a record's original claim declared volatile fields and this retry declares different ones, the retry consumes a bucket that differs from the one the original claim consumed. Those two counts then sit in separate buckets and the cap for that authorization is one unit weaker, once. It self-corrects on the first claim that pins the id. Storing the claim's `volatile` sequence on the record would let the fallback reproduce the original derivation exactly, but the blast radius is pre-existing records, so leaving it is defensible.

src/continuum/actions/ledger.py
```diff
@@ -439,6 +439,19 @@ def _budget_authorization_id(
...
         resource tokens alone. Ledger anchoring is not used here so a
         prior failed attempt does not make a retry look unbound.
+
+        ``arguments`` are caller-supplied, which is the other half of the same
+        problem: a token derived from every argument follows the noise as much
+        as the resource, so padding one throwaway field (a ``trace_id``, a
+        request id) moved each retry into a fresh bucket at its full allowance
+        and the cap never bound (issue #1052). The caller therefore passes the
+        arguments of the record the claim defers to when there is one: those
+        name the operation the ledger has already decided this attempt *is*, and
```
Cyrax321, line 449: Non-blocking: this sentence says the settlement paths derive from those arguments, but after the second commit they read the pinned id and fall back to deriving only when the claim left none. The comment in `claim` and the `_settlement_authorization_id` docstring are accurate; this one and the `security.md` section carry the earlier mechanism.

docs/api/security.md
```diff
@@ -154,6 +154,25 @@ wins and ignores arguments; otherwise the id derives from distinctive resource
...
+
+The token derivation reads every argument, and arguments are caller-controlled
+noise plus the real resource, so the bucket is only as stable as the arguments
+it was computed from. `ActionLedger.claim` therefore derives the bucket from the
+record the claim *defers to* when one exists -- the caller held the idempotency
+key fixed, or the token fallback recognised a prior attempt -- rather than from
+the incoming arguments. A retry that pads a throwaway `trace_id` or request id
+is the same operation the ledger already recorded, so it draws from that
+record's bucket and cannot reset the cap by padding (issue #1052). Settlements
+derive from the same stored arguments, so a retry and its confirmation share one
```
Cyrax321, line 167: Non-blocking: this describes the derive-from-stored-arguments rule from the first commit, but the shipped mechanism pins the id on `Action.budget_authorization_id` and `complete`/`reconcile` read it back. The guarantee is stronger than the wording implies: the bucket cannot move even if the caller changes its `volatile` declaration between the attempt and its confirmation. Naming the pinned field also makes it discoverable. The CHANGELOG bullet and the `_budget_authorization_id` docstring carry the same older wording, so all three could be updated together.

Cyrax321 (approved):
Approving. I verified the fix rather than only reading it: I reverted the two changed hunks in `ledger.py` inside a throwaway worktree at this HEAD, and three of the four new tests fail there, the padding test, the volatile-flip test, and the settlement-pinned-bucket test, while the residual-case test still passes as it must. Back at HEAD the full suite exits 0, `ruff check` and `ruff format` are clean, mypy reports no issues across 128 files, and the docs-count guard passes against a live collection, so the 2,402 figure in the READMEs matches this tree. The new `Action.budget_authorization_id` round-trips because `_record` embeds `action.model_dump()` in the event payload and `fold_action_events` validates it back, so the pin survives a replay and a process restart.

Non-blocking, and outside this diff: the `on_unknown` resolver branch in `claim` records an `ACTION_RECONCILED` outcome without a settlement drawdown, where a caller that let the `UnknownSideEffect` propagate and reconciled later would have been charged one. CodeRabbit flagged this as major. I do not think the suggested patch closes it, because it skips resolutions to `COMPLETED`, which is exactly the case `complete()` draws a settlement for, so the main gap would remain. It is a real accounting asymmetry and worth a follow-up issue, but it predates this PR and does not affect the #1052 fix.

Nothing blocks. The doc passages noted inline still describe the derive-from-stored-arguments mechanism from the first commit; the second commit replaced that with a pinned id read back at settlement, and the code comments in `claim` and `_settlement_authorization_id` are the accurate ones.

PR #1265, opened 2026-09-17
docs(cli): clarify global --json flag placement across READMEs and references

### LINKED ISSUE #1144: docs: README and references/cli.md say every command accepts --json, but the trailing position is rejected with exit 2

Voyagerroc-Lab:
Resolves #1144

### Summary
Clarifies in README.md, 
eferences/cli.md, and translated READMEs (s, ja, ko, pt-BR, zh-CN) that --json (like --db) is a global flag that must precede the subcommand (e.g. continuum --json <command>), aligning with docs/api/cli.md:100-103.

## Summary by CodeRabbit

- **Documentation**
  - Clarified CLI usage across translated guides and reference documentation.
  - Documented that the global `--json` option must appear before the command: `continuum --json <command>`.
  - Clarified that placing `--json` after the command is not supported.


Cyrax321 (changes requested):
Thank you for contributing to Continuum. I appreciate you picking up #1144 and covering the translated READMEs.

There is currently a conflict in README.md and references/cli.md because main already carries a newer version of the same English fix with additional context. Please resolve the conflict by keeping the main side for those two files and keeping your translation updates.

Once the conflict is resolved and CI is green, this will be ready to approve.

Cyrax321 (changes requested):
Author context: Voyagerroc-Lab, first-time contributor. Thank you for picking up #1144 and covering the translated READMEs. The translation updates are the valuable part of this PR. The prior review already flagged the conflict in README.md and references/cli.md, where main carries a newer version of the same English fix, and that conflict still stands with dirty merge state. Resolve by keeping the main side for those two English files and keeping your translation updates. Note this overlaps with #1296 on the same issue; the translation coverage here is what distinguishes it. Once rebased with the English conflict resolved and CI green, this is approvable.

README.md
```diff
@@ -495,7 +495,7 @@ continuum tree <parent_run_id>                    # multi-agent hierarchy view
 
 Optional registries live beside your code and are data, not code: `.continuum/gate.json` (side-effect tools + stable-key templates), `.continuum/reconcilers.json` (probes that check external systems), `.continuum/gateway.json` (upstream routes).
 
-Every command accepts `--json`, and read-only commands never write, so they are safe against a live database while an agent is mid-run. Exit codes are a safety contract (only a verified-safe run exits 0). Full command list, exit-code table, and state-diff output in [references/cli.md](references/cli.md).
+Every command accepts `--json` (placed before the command: `continuum --json <command>`, not `continuum <command> --json`), and read-only commands never write, so they are safe against a live database while an agent is mid-run. Exit codes are a safety contract (only a verified-safe run exits 0). Full command list, exit-code table, and state-diff output in [references/cli.md](references/cli.md).
```
Cyrax321, line 498: Blocking. This hunk (and its twin in references/cli.md) conflicts with main and is superseded: #1165 already merged this clarification, and this exact sentence on main is now "Most commands accept global `--json` **before** the subcommand (e.g. `continuum --json resume RUN`)", which is more precise on two counts: "Most" rather than "Every", and a webhook caveat qualifying the read-only claim. Re-applied on top of main, this wording would regress both. Please rebase and drop README.md and references/cli.md so only the five translation hunks remain, since those are the part still missing from main.

README.es.md
```diff
@@ -466,7 +466,7 @@ continuum tree <parent_run_id>                   # vista de jerarquía multiagen
 
 Los registros opcionales viven junto a tu código y son datos, no código: `.continuum/gate.json` (herramientas de efecto secundario + plantillas de clave estable), `.continuum/reconcilers.json` (sondas que comprueban sistemas externos), `.continuum/gateway.json` (rutas ascendentes).
 
-Cada comando acepta `--json`, y los comandos de solo lectura nunca escriben, por lo que son seguros contra una base de datos viva mientras un agente está a mitad de ejecución. Los códigos de salida son un contrato de seguridad (solo una ejecución verificada segura sale con 0). Lista completa de comandos, tabla de códigos de salida y salida de diff de estado en [references/cli.md](references/cli.md).
+Cada comando acepta `--json` (colocado antes del comando: `continuum --json <comando>`, no `continuum <comando> --json`), y los comandos de solo lectura nunca escriben, por lo que son seguros contra una base de datos viva mientras un agente está a mitad de ejecución. Los códigos de salida son un contrato de seguridad (solo una ejecución verificada segura sale con 0). Lista completa de comandos, tabla de códigos de salida y salida de diff de estado en [references/cli.md](references/cli.md).
```
Cyrax321, line 469: Non-blocking. The placement rule checks out (verified against `build_parser()` on main: `--json` lives only on the top-level parser and trailing it after the subcommand exits 2), so the translations are the right files to keep. Once rebased, consider aligning the sentence with main's canonical phrasing, "Most commands" rather than "Every command" plus the webhook caveat on read-only commands, so the translations describe the same CLI the English docs do rather than the pre-#1165 wording.

Cyrax321 (changes requested):
This PR adds the correct placement rule for the global `--json` flag (before the subcommand) to the two English canonical docs and five translated READMEs. I verified the claim against the parser on main: `build_parser()` registers `--json` only on the top-level parser and `continuum <command> --json` exits 2 with `unrecognized arguments`, so the guidance itself is factually right.

The problem is mergeability and duplication: the README.md and references/cli.md hunks re-do what #1165 already merged on 2026-09-16 (main now reads "Most commands accept global `--json` **before** the subcommand" at README.md:500 and references/cli.md:55), and GitHub reports this PR as conflicting, so it cannot merge as submitted. Requesting changes: rebase onto current main, drop the two English hunks, and keep the five translation updates, which are still needed since the translated READMEs still carry the placement-less sentence.

On the overlap question: open PR #1296 asserts the same before-placement rule, not the opposite, and both are consistent with the parser, so there is no factual contradiction between the two PRs. They do edit the same sentences in the same files plus more in #1296's case, so only one of them can land; worth coordinating which one proceeds.

PR #1269, opened 2026-09-18
feat(contract): publish a compatibility corpus and conformance suite

### LINKED ISSUE #764: feat(contract): publish a compatibility corpus and conformance suite for recovery contracts
## Affected component

`docs/research/cross_agent_portability.md` defines a portable `RecoveryContract` subset, but also documents an unresolved forward-compatibility problem: an older verifier cannot validate a newly sealed contract when it contains an unknown hash-covered optional field. Current tests cover specific historical fields such as `evidence` and `reason`, rather than a general compatibility contract. This makes cross-framework and cross-version verification depend on implementation details rather than a published, testable interchange promise.

## Proposed solution

Publish a versioned recovery-contract compatibility corpus containing canonical JSON fixtures and expected verification outcomes for current, legacy, forward-extension, malformed, and tampered contracts. Add a conformance test helper that any verifier implementation can run. Define canonical serialization, the fields covered by each contract version, unknown-field handling, and the supported upgrade path before changing verification behavior.

Keep the corpus focused on recovery contracts. It must not duplicate the broader semantic-state interchange implementation or weaken the existing integrity guarantee by silently ignoring unknown hash-covered data.

## Acceptance criteria

- A versioned, checked-in corpus includes valid current contracts, valid legacy contracts, forward-extension cases, malformed payloads, and tampering cases.
- Each fixture states its contract version, canonical serialization bytes or digest input, and expected verification result.
- A reusable conformance test helper verifies fixtures independently of the production `verify_contract` implementation.
- The compatibility policy explicitly defines how an old verifier handles newer fields and how a new verifier handles older contracts.
- Unknown or unsupported contract versions and unknown hash-covered fields fail closed with actionable diagnostics unless an explicitly documented compatible path exists.
- Existing legacy `evidence` and `reason` behavior remains covered by regression tests.
- Documentation explains how external verifiers consume the corpus and how future fields or versions are added.

## Alternatives considered

Continue adding one-off legacy fallbacks whenever a field changes. Rejected because it does not establish an auditable compatibility guarantee and will not scale to independent framework verifiers. Ignoring unknown fields during verification is also rejected unless their treatment is cryptographically specified, because it could accept a tampered or semantically different contract.

## Is this a breaking change?

No

Yes, with guidance

Cyrax321:
Closes #764.

The integrity hash covered whatever the model defined at seal time, so a contract sealed by one build could only be verified by a build with the same field set. The legacy `evidence`/`reason` path was a comment in `verify_contract`, not a rule an external verifier could implement against.

## What this adds

**Published version table.** `CONTRACT_VERSIONS` pins the field set each version covers: v0 is the pre-Phase-1 contract (no `evidence`/`reason`), v1 is current. A verifier recognizes a version or fails closed.

**Checked-in corpus** at `src/continuum/recovery/corpus/`, one JSON per case across every category the policy depends on — current, legacy, forward-extension, malformed, tampering — each stating its version, canonical digest input, and expected outcome.

**Independent checker.** `continuum.recovery.conformance` recomputes each digest from the published rules and never calls `verify_contract`, so the spec and the production verifier remain two things that can disagree. `tests/test_contract_conformance.py` also runs the production verifier over the same corpus and asserts agreement, since a second implementation can be wrong in the same direction as the first.

**Diagnostics.** `verify_contract_detailed` reports which version accepted and why. An unknown version fails closed naming the versions this build knows, instead of hashing whatever fields happen to be present.

**Spec.** `docs/contract_compatibility.md` is what an external verifier implements against, including how to add a field or a version.

## Compatibility policy

New verifier reading an old contract: try the declared version, then each older one in descending order. Old verifier reading a new contract: fail closed with a diagnostic. This replaces the previous "ignore unknown optional fields" guidance, which could not distinguish a field the verifier does not understand from one the producer removed.

## Notes

- The digest is now sha256 over the canonical bytes directly. Hashing an already-canonical string through `stable_hash` re-encodes it, which is not reproducible from the published input alone.
- `RecoveryContract` gains `contract_version`, excluded from the digest because it selects the payload rather than being part of it. Absent on the wire, it means the pre-versioning shape.
- `test_legacy_sealed_contract_verifies` now excludes `contract_version` from the legacy hash for the same reason.
- The one full-suite failure is a pre-existing Hypothesis health-check flake in `test_idempotency_tokens.py` under load; it passes in isolation on both this branch and clean `origin/main`.

## Summary by CodeRabbit

- **New Features**
  - Recovery contracts now support versioned integrity verification, including compatibility with legacy contracts.
  - New contracts record the contract version used for verification.
  - Detailed verification can identify the accepted version and provide failure reasons.
  - Unknown or malformed contract versions are rejected with actionable diagnostics.

- **Documentation**
  - Added the recovery-contract compatibility specification, including canonical digest rules and guidance for external verifiers.

- **Tests**
  - Added conformance coverage for valid, legacy, malformed, forward-compatible, and tampered contracts.


Cyrax321 (commented):
Maintainer review at current head. The compatibility corpus and conformance suite direction is sound, but the gitar-bot Phase-1 legacy finding looks real and blocking: contracts sealed before contract_version existed default to 0, and the order construction only tries versions below declared, so the legacy digest path may never match the stored hash. Verify legacy sealed contracts still verify on this head and fix the version fallback if not. Also blocking: full Test matrix red on all nine legs plus dirty merge state. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review. Rebase, fix the legacy path, confirm green.

PR #1271, opened 2026-09-18
feat(plugins): run registered domain validation rules during recovery assessment

### LINKED ISSUE #761: feat(plugins): run registered domain validation rules during recovery assessment
## Affected component

CONTINUUM exposes a `ValidationRule` plugin seam for domain-specific staleness, but it is not a consumer in the recovery path. `references/integration-architecture.md` calls out cases where environment diffs are insufficient, such as a decision becoming invalid because a cited policy is revoked. `docs/ARCHITECTURE_EVOLUTION.md` also identifies the plugin seams as declared but not load-bearing. Without an execution path, downstream integrations must fork the validator or cannot contribute evidence-backed domain rules at all.

## Proposed solution

Thread an explicitly supplied or registered collection of `ValidationRule` implementations through recovery assessment. Rules should receive the projected semantic state and the current environment, return structured component-validation entries, and be merged with built-in validation using the existing most-cautious-wins semantics. Provide one small built-in example rule and a conformance test helper that third-party rule authors can run.

Rules must be read-only, deterministic for identical inputs, namespaced in diagnostics, and unable to override an existing unsafe verdict with a less cautious result. Failed rule execution must be represented as an explicit unknown or requires-review result, not silently ignored.

## Acceptance criteria

- `RecoveryEngine.assess` accepts an explicit, documented rule collection or registry-backed equivalent without changing default behavior when none is configured.
- Returned rule findings are provenance-labeled, namespaced by rule identifier, and appear in the sealed contract and JSON/text diagnostic surfaces.
- A rule can increase caution but cannot reduce the severity of built-in or another rule's finding.
- A malformed, crashing, duplicate-named, or nondeterministic-output rule fails closed with a diagnosable result.
- Rules cannot mutate storage, emit events, or change repair plans directly during assessment.
- A built-in example and reusable conformance tests cover valid, stale, unknown, error, ordering, and severity-composition cases.
- Documentation explains the extension trust boundary and shows how to register a rule without making claims about automatic policy learning.

## Alternatives considered

Let integrations patch `StateValidator` directly. Rejected because it forks core recovery behavior and makes provenance, ordering, and upgrade compatibility difficult to audit. Automatically loading arbitrary third-party code is also out of scope; registration remains explicit.

## Is this a breaking change?

No

Yes, with guidance

Cyrax321:
Closes #761. The `ValidationRule` seam had no consumer, so an integration that needed staleness the environment diff cannot express (a decision void because the policy it cites was revoked, not because any dependency moved) had to patch `StateValidator` and fork core recovery behaviour. This wires the seam into `RecoveryEngine.assess`, adds a built-in example rule and a conformance helper, and documents the trust boundary.

Rules are given the projected state and the current environment and nothing else, so a rule cannot reach storage, an event writer, or the repair plan. Findings merge by maximum caution per component, commutatively, so registration order does not change the report and a rule can raise a status but can never lower one a built-in check or another rule already produced. No rules configured, or a rule that finds nothing, leaves the report and the sealed contract byte-identical.

Two judgement calls worth flagging for review:

- A duplicate rule name fails closed rather than being deduplicated. Silently picking one would attribute the other's findings to the wrong rule, and an auditor reading "rule X failed" has to know which X.
- Rules are not filtered by an assessment `scope`. A domain rule that knows a decision is unauthorized is not made wrong by the caller asking about one dependency, so it speaks. This is the one place a localized recovery can pick up a finding from outside its subtree, and it errs toward caution.

Acceptance criteria, each covered by a test in `tests/test_validation_rules.py`:

- `assess` accepts an explicit rule collection or a `Registry`, with no behaviour change when none is configured (the baseline-vs-with-rules test compares report bytes and the contract hash).
- Findings are provenance-labelled (`entry.rule`) and namespaced `component:id [rule:name]` in the contract and in the text and JSON renderings.
- Most-cautious-wins merge: a rule can escalate, cannot relax, and two rules resolve the same either way around.
- A crashing, malformed, nameless, or duplicate-named rule becomes a `validation_rule` entry at `requires_review` that withholds resume and names the rule and what it did.
- The seam passes no storage, event, or plan handle; a hostile rule left the event log and checkpoint version untouched.
- The built-in `RevokedApprovalRule` plus `check_validation_rule` cover valid, stale, unknown, error, ordering, and severity-composition cases, including a nondeterministic rule and one that mutates the state it is handed.
- `docs/guides/validation-rules.md` documents the boundary and registration, and states plainly that the seam does not learn policy.

The conformance helper deliberately does not check that a rule's verdict is *true*, only that it is deterministic, read-only, and well-formed. Reviewing the logic is still a human job.

Documented test counts are re-synced to 2,338 collected across README, the five translated READMEs, CHANGELOG, CONTRIBUTING_ONBOARDING, and the two `references/` docs. That number is the CI-collectable total: `tests/test_otel.py` (8 tests) skips under `.[dev]` because opentelemetry is not in the dev extras.


Cyrax321 (commented):

Ref: #1271 / Issue #761
Surface: ValidationRule plugin seam, RecoveryEngine.assess, plugins registry

What this adds: the previously consumer-less ValidationRule seam is wired into assess, with a built-in example rule, a conformance helper, and a documented trust boundary. The motivating case (a decision void because cited policy was revoked, expressible neither by environment diff nor by built-in checks) is the right gap for this seam.

Design reviewed at head a919c656:
- Rules receive the projected state and the current environment only, so a rule cannot reach storage, the event writer, or the repair plan. The isolation is enforced by the signature, not by convention.
- Findings merge by maximum caution per component, commutatively, so registration order cannot change the report, and a rule can raise a status but never lower one already produced. That asymmetry is the safety property that matters.
- No rules configured, or a rule finding nothing, leaves report and sealed contract byte-identical, so the default path carries no behavioral risk.

Proof:
- Ran tests/test_validation_rules.py at the head commit: 36 passed.

Cyrax321 (commented):
Maintainer re-review at current head. The validation-rules plugin seam is still sound with the raise-only asymmetry. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review with no blocking issues at last review. Needs an independent approval to merge. Rebase out of dirty state and confirm CI green.

PR #1272, opened 2026-09-18
feat(recovery): scope human-gate attempt budgets to dependency ownership (#744)

### LINKED ISSUE #744: feat(recovery): scope human-gate attempt budgets to dependency ownership
## Affected component

The human-gate minimization study in `docs/research/human_gate_minimization.md` recommends per-dependency budgets so a noisy dependency does not consume a run-wide recovery-attempt allowance. The current recovery ledger records attempts and human-required outcomes, but its budget semantics do not isolate unrelated dependency scopes. A repeatedly failing integration can therefore block a safe repair path for another dependency.

## Proposed solution

Extend recovery-attempt accounting with an optional, explicit dependency scope derived from the structured plan or action metadata. Evaluate a scoped budget when ownership is known, retain the current global behavior when it is not, and record the scope in the evidence used to explain an escalation. The feature must be fail-closed: malformed or ambiguous ownership must never grant extra retry capacity.

Likely seams include `src/continuum/recovery/ledger.py`, recovery planner and contract rendering, `PlanStep.depends_on`/action dependency metadata, and recovery-ledger tests.

## Acceptance criteria

- The data model defines an optional dependency-scope key with validation and backward-compatible defaults.
- Attempts attributable to dependency A do not exhaust dependency B's scoped allowance.
- Runs without reliable ownership retain existing global budget behavior.
- Unknown, conflicting, or malformed ownership cannot bypass a global or scoped limit.
- Recovery contracts and CLI/JSON output identify the budget scope and remaining/exhausted allowance without leaking unnecessary sensitive values.
- Tests cover independent scopes, shared scopes, missing ownership fallback, compaction/replay persistence, and concurrent or repeated assessment behavior.
- Existing max-severity recovery ordering and human gates for tamper, lease contention, and explicit human verdicts remain unchanged.

## Alternatives considered

Increase the global attempt limit. Rejected because it masks a single noisy dependency and increases the chance of repeated unsafe work across the entire run.

## Is this a breaking change?

No

Yes, with guidance

Cyrax321:
Closes #744.

The recovery ledger counted recovery attempts run-wide, so one repeatedly failing integration could spend the allowance a *different* dependency needed to repair. This makes an attempt charge the budget of the dependency that owns it, which is what `docs/research/human_gate_minimization.md` asked for.

**What changes**

`RecoveryLedger` carries an optional dependency scope on each entry. `record_attempt`, `attempts`, `requires_human` and the new `budget` accept any ownership signal the system already has: an action's `dep_scope`, a dependency finding's component id, or the resource set a scoped assessment was confined to. A new `resolve_scope` reduces those candidates to one canonical key.

Escalation markers are anchored and scoped, so they survive compaction and stop at their own dependency: exhausting `dataset` leaves `other`'s repair path open. `RepairStep` records the scope the plan derived, and `build_contract` emits an evidence line naming the budget scope with its remaining or exhausted allowance. `RecoveryEngine` takes an optional ledger and reads that budget read-only; without one, every decision is byte-identical to before.

**Fail-closed, by construction**

Unknown, conflicting or malformed ownership all resolve to the run-wide bucket, which can only escalate sooner and can never grant capacity the run did not have. A scoped limit is additionally capped at the run-wide ceiling, so a per-dependency allowance configured above it buys nothing. A scope in a hand-edited ledger record that is not a non-empty string is discarded on read rather than honoured, so a controlled file cannot manufacture a private budget.

An ownership-less query still reads a *known* escalation as its own: `requires_human(run_id)` stays true once any dependency has escalated, because an unknown-ownership caller cannot assume the escalation belongs to someone else.

**Backward compatibility**

The scope key is omitted from a scope-less entry's sealed content, so a ledger file written before this change still verifies after the upgrade, and an auditor holding a pre-upgrade copy still reconciles it. Existing callers that pass no scope get identical behaviour, including the marker and count semantics. Contracts gain only an additive evidence line; the sealed payload gains no new field, so `verify_contract` keeps working for everything already on disk.

**Testing**

2282 passed, 26 skipped (2308 collected); ruff and mypy strict are clean.

`tests/test_recovery_budget_scope.py` (52 tests) covers scope isolation, case-folding, missing ownership, malformed and conflicting ownership, the run-wide ceiling, compaction, file round trips, legacy pre-scope records, repeated reads, and the contract evidence surface including the no-sensitive-values property. The phase-6 suite gains `scoped_attempt_budget` as the counterpart to `recovery_lease_exhaustion`: the same three attempts, this time attributable, gating only the dependency that earned them.

One thing worth a reviewer's judgement: `RecoveryEngine` reads the budget from an injected ledger but nothing in the CLI constructs one yet, because `Storage` is an abstract backend with a Postgres implementation and there is no storage-agnostic home for ledger files today. The budget line therefore surfaces wherever a caller wires a ledger in; persisting recovery attempts from `continuum resume` is a separate change and is not assumed here.


Cyrax321 (commented):

Ref: #1272 / Issue #744
Surface: RecoveryLedger attempt budgets, human-gate escalation, recovery scoping

What this adds: recovery attempt budgets charge the dependency that owns the attempt instead of the run as a whole, so one repeatedly failing integration can no longer spend the allowance a different dependency needs. That is the behavior docs/research/human_gate_minimization.md asked for, and the old run-wide counter made cross-dependency interference possible in exactly the way the issue describes.

Design reviewed at head 97e27c63:
- resolve_scope reduces the three existing ownership signals (action dep_scope, finding component id, scoped assessment resource set) to one canonical key rather than inventing a fourth.
- Escalation markers are anchored and scoped, so they survive compaction and stop at their own dependency: exhausting one leaves another's repair path open. Both properties are pinned by tests rather than asserted in prose.
- RepairStep records the scope so the plan cannot silently widen after escalation.

Proof:
- Ran tests/test_recovery_budget_scope.py and tests/test_budget_drawdown.py at the head commit: 61 passed.

Cyrax321 (commented):
Maintainer re-review at current head. The scoped-budget design is still sound. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review with no blocking issues at last review. Needs an independent approval to merge. Rebase out of dirty state and confirm CI green.

PR #1273, opened 2026-09-18
feat(environment): configured, discoverable providers for resume (#762)

### LINKED ISSUE #762: feat(environment): configure discoverable providers for automatic resume-time revalidation
## Affected component

`EnvironmentProvider` implementations exist for files, git, values, callables, and static inputs, but callers must pass them manually to snapshot capture and assessment. `docs/ARCHITECTURE_EVOLUTION.md` identifies the missing discovery path: the validator does not automatically apply registered providers. As a result, an integration can define a reliable world-observer but a normal `resume` path may validate only what the caller remembered to supply.

## Proposed solution

Add an explicit provider configuration and registry-resolution path for resume-time revalidation. A run records provider descriptors and normalized resource scopes at checkpoint time. On resume, CONTINUUM resolves only approved built-in or explicitly registered providers, captures the current environment, and feeds it to normal validation. Each provider result must carry provenance and a stable identifier so contracts explain which observers supplied evidence.

Do not auto-import arbitrary plugins or execute untrusted configuration. An unavailable, malformed, or failing configured provider must produce an explicit unknown or requires-review signal for the resources it owns, never a successful validation based on an empty capture.

## Acceptance criteria

- A versioned, validated configuration model records provider identifiers and bounded resource scope without serializing arbitrary callables or secrets.
- Checkpoint and resume paths can resolve configured built-in or explicitly registered providers consistently.
- Provider results are named and provenance-labeled in validation evidence and recovery contracts.
- Missing, disabled, malformed, or failing providers fail closed for their declared resources and never silently remove them from validation.
- Unconfigured runs preserve today's behavior exactly.
- Tests cover file and git providers, provider registration, unavailable providers, conflicting resource keys, deterministic ordering, compaction/replay persistence, and JSON diagnostics.
- Documentation defines the provider trust boundary, configuration lifecycle, and safe operational defaults.

## Alternatives considered

Continue requiring every adapter or CLI caller to pass provider objects manually. Rejected because omitted observers cause gaps that are difficult to see at resume time. Arbitrary plugin auto-discovery is also rejected because it would turn a recovery check into implicit execution of untrusted code.

## Is this a breaking change?

No

Yes, with guidance

Cyrax321:
Closes #762.

## What

`EnvironmentProvider` implementations existed for files, git, values, callables and static inputs, but a resume only applied the ones a caller remembered to pass to `assess()`. An integration could wire a reliable world-observer and still resume through a path that validated only what was supplied by hand, and nothing in the output said so.

A run now records its observers in the event log and resume applies them on its own:

- `ProviderSpec` / `ProviderConfig` in `src/continuum/environment/config.py`: a versioned, validated configuration model recording provider identifiers, a bounded resource scope, and JSON-native parameters.
- `ENVIRONMENT_PROVIDERS_CONFIGURED` event (append-only, newest record authoritative, survives compaction since resume reads the archived prefix).
- `RecoveryEngine.assess` resolves the configured providers when no `current_environment` was supplied, captures, and feeds the result to the validator that already existed. A caller-supplied environment still wins; an unconfigured run behaves exactly as before.
- `continuum providers add|remove|list|check` manages the configuration. `check` is read-only and resolves exactly as resume would, so a failing observer is visible before it gates a recovery.
- Captured resources carry `metadata["provider"]`, and validation entries read `verified unchanged (provider: git)`; the recovery contract inherits those strings as evidence.

## Trust boundary

A configuration cannot reach code, enforced in five places: a provider name is a lookup key matching `[A-Za-z0-9][A-Za-z0-9_.-]{0,63}`, never a path or import; only the four built-ins and names handed to a `ProviderRegistry` resolve, and an unknown name is reported `unavailable` rather than autoloaded; parameters must be JSON-native or refused; a parameter whose name looks like a secret (`password`, `secret`, `token`, `api_key`, ...) is rejected so it never reaches the hashed log; and `CallableProvider` is not configurable at all, it registers by name and the configuration references the name. There is no plugin auto-discovery anywhere; the issue explicitly rejects it and so does this.

## Fail-closed

A disabled, unavailable, malformed, conflicting or failing provider emits `UNKNOWN_VERSION` for every resource it declared instead of leaving it out of the snapshot, and `UNKNOWN` is what the validator already degrades on. Two specs claiming one resource key fail closed for both rather than letting one win by order. A provider reporting beyond its declared scope has those keys marked unknown too. The one non-blocking status is `undeclared`, because the surplus key is already unknown rather than trusted. `resolve_and_capture` never raises.

## Bug the feature exposed

`compact_run`'s forced anchor checkpoint was written with no `environment`, and the anchor becomes the newest checkpoint and the resume comparison point. A compacted run therefore had no snapshot to diff a resumed capture against and every resource read as unknown, which is precisely the long-lived-run case this feature targets. The anchor now inherits the previous checkpoint's environment, in both the SQLite and Postgres paths, which carry identical anchor logic. This is a pre-existing defect independent of the feature, and is called out in the CHANGELOG.

## Tests and verification

33 new tests in `tests/test_provider_config.py` covering the file and git providers, registration, unavailable / disabled / malformed / failing providers, conflicting keys, deterministic output, compaction persistence, JSON diagnostics, and the CLI. The discovery wiring and the compaction anchor fix were each A/B'd by stashing only the source change: the tests fail with the bug's signature and pass with the fix. Full suite, `ruff check`, `ruff format --check` and `mypy src/continuum` are all clean. Documented test counts (2,344 collected), the 51 to 52 event-type count, the 47-command CLI count and the module map are re-synced across the README and its translations, CHANGELOG, CONTRIBUTING_ONBOARDING, and references/{install,testing,cli}.md, and `docs/guides/environment-providers.md` documents the trust boundary, lifecycle, fail-closed table and safe defaults.

The two `hypothesis` deadline flakes in `test_adapter_result_fuzz.py` are environmental and unrelated; they did not recur on the final run.

## Summary by CodeRabbit

- **New Features**
  - Added configurable environment providers for resume-time verification, including trusted scopes, status diagnostics, fail-closed handling, and provider provenance in validation results.
  - Added CLI commands to add, list, check, and remove providers for a run.
  - Environment provider configuration now survives compaction and supports built-in or registered providers.

- **Bug Fixes**
  - Compaction checkpoints now preserve environment details from the latest checkpoint.

- **Documentation**
  - Added an environment-provider guide and updated CLI references, architecture documentation, translations, and project test statistics.


Cyrax321 (commented):

Ref: #1273 / Issue #762
Surface: environment providers, resume, src/continuum/environment/config.py, RecoveryEngine.assess

What this adds: a run records its environment observers in the event log as a versioned ProviderConfig, and resume applies configured providers on its own instead of validating only whatever a caller remembered to pass by hand. The motivating failure, an integration wiring a reliable world-observer and still resuming through a hand-fed path with nothing in the output saying so, is real and this closes it.

Design reviewed at head c79ebb39:
- ProviderSpec/ProviderConfig is versioned and validated with a bounded resource scope and JSON-native parameters, so a bad configuration is rejected at write time rather than surfacing mid-recovery.
- The ENVIRONMENT_PROVIDERS_CONFIGURED record is append-only with newest-authoritative semantics and is read through the archived prefix, so configured providers survive compaction.
- assess resolves configured providers only when no explicit current_environment was supplied, so the manual override path keeps its meaning and existing callers see no behavior change.

Proof:
- Ran tests/test_provider_config.py at the head commit: 33 passed, covering the configuration model, resolution precedence, and the compaction survival path.

Cyrax321 (commented):
Maintainer re-review at current head. The discoverable-providers design is still sound. Note: GitHub blocks approval events on self-authored PRs, so this is a comment review with no blocking issues at last review. Needs an independent approval to merge. Rebase out of dirty state and confirm CI green.

PR #1279, opened 2026-09-18
fix(core): add omitted public names to __all__ and drop private exports (#1111)

### LINKED ISSUE #1111: follow-up to #1093: a second batch of __all__ omissions (ReplayBlocked, HitlUnauthorized, SettleReport, SidecarHTTP, SubprocessClient) and one inversion in analysis/__init__.py

soorajsuru2004-sudo:
## Description

Follow-up to #1093 addressing public symbol omissions and inversion in `__all__`:
- Add `ReplayBlocked` to `continuum.replayguard.__all__`
- Add `HitlUnauthorized` to `continuum.dashboard.hitl.__all__`
- Add `SettleReport` to `continuum.reconcilers.__all__`
- Add `SidecarHTTP`, `AGENT_SOURCE`, `MAX_SIDECAR_BODY_BYTES`, and `SIDECAR_DRAIN_LIMIT_BYTES` to `continuum.serve.server.__all__`
- Add `SubprocessClient` to `continuum.serve.__all__`
- Drop underscore-private names `_normalize_dep` and `_STDLIB` from `continuum.analysis.__all__` and its import list
- Extend `tests/test_module_all_exports.py` to cover these symbols and assert `set(mod.__all__) == _locally_defined_public_names(mod)` for the defining modules.

## Related issues

Fixes #1111

## Types of changes

- Type: `bugfix`

## Breaking changes

No.

## How has this been tested?

Ran pytest and ruff on Python 3.12 (Windows):
```
pytest tests/test_module_all_exports.py -q
pytest tests/test_docs_counts.py -q
pytest tests/test_replayguard.py tests/test_serve_http.py tests/test_reconcilers.py -q
ruff check src/ tests/
```
All passed cleanly (18 passed in `test_module_all_exports.py`, 86 passed across related suites, 0 errors in ruff).

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Expanded the public API with additional exceptions, reports, client classes, sidecar interfaces, configuration constants, and dependency graph access.
  * Removed private analysis helpers and constants from the public exports.

* **Tests**
  * Added coverage verifying public exports, star imports, and exclusion of private analysis symbols across the affected modules.


Cyrax321 (changes requested):
Author context: soorajsuru2004-sudo, external contributor. The export fixes are individually correct: I verified every added name resolves on this checkout (HitlUnauthorized, SettleReport, ReplayBlocked, SubprocessClient, SidecarHTTP, AGENT_SOURCE, MAX_SIDECAR_BODY_BYTES, SIDECAR_DRAIN_LIMIT_BYTES), and dropping underscore-private names from analysis.__all__ is the right direction. Blocking: the full Test matrix is red on this branch on all nine legs, plus dirty merge state. Rebase onto current main and confirm the matrix goes green; also confirm against the landed __all__ guard work from #1300, which touched the same test file and may overlap or supersede parts of this.

tests/test_module_all_exports.py
```diff
@@ -123,6 +129,10 @@ def test_all_symbols_exist_on_modules() -> None:
         recovery_impact,
         recovery_limits,
         recovery_notify,
+        replayguard,
+        dashboard_hitl,
+        reconcilers,
+        serve_server,
```
Cyrax321, line 135: Blocking: this list no longer exists on main. Since #1300, test_all_symbols_exist_on_modules walks the package for existence only, and the exact set(mod.__all__) == _locally_defined_public_names(mod) guard moved into test_all_matches_locally_defined_public_names, driven by the separate _LEAF_MODULES_WITH_COMPLETE_ALL list. Resolving the conflict by restoring this list would silently resurrect the pre-#1300 structure and duplicate the walk-based coverage.

On rebase, add replayguard, dashboard_hitl, reconcilers, and serve_server to _LEAF_MODULES_WITH_COMPLETE_ALL instead of extending a list here. I checked each of the four modules against origin/main: after this PR's __all__ additions, the set of locally defined public names matches __all__ exactly (main currently differs only by the missing names this PR adds), so the equality guard will pass once the source changes land.

Cyrax321 (changes requested):
Re-review on head 5c5140d4. The source-module half of this PR is verified correct: every name added to __all__ (ReplayBlocked, HitlUnauthorized, SettleReport, SidecarHTTP, SubprocessClient, AGENT_SOURCE, MAX_SIDECAR_BODY_BYTES, SIDECAR_DRAIN_LIMIT_BYTES) is defined in the module that exports it, and the dropped _normalize_dep and _STDLIB are referenced nowhere outside continuum/analysis, so removing them from the package surface breaks no internal caller.

Blocking: GitHub reports this PR as CONFLICTING against main, and the conflict is real rather than trivial: #1300 (merged 2026-09-21) rewrote tests/test_module_all_exports.py in a way that removes the hand-rolled module list this PR extends. A rebase must rework the test additions rather than resolve the conflict textually; detail inline.

Blocking: the Test matrix is red on all nine legs. The single failing test, tests/test_adapters_openai.py::test_wrap_function_tool_invocation_binds_args_and_intercepts, sits in a file this PR does not touch and fails with FakeTC missing _function_tool_arguments, which reads as openai-agents SDK drift rather than anything caused by this diff, but main is green and branch protection requires the matrix, so rebase onto main and rerun until all legs pass before this can merge.

PR #1280, opened 2026-09-18
docs: fix stale EventType count (29/32 -> 51) and add drift guard (#1100)

soorajsuru2004-sudo:
Closes #1100.

## Problem

The docs consistently stated **29** event types (and `docs/assets/architecture.svg` stated **32**), while `EventType` actually holds **51**:

```python
>>> len(EventType)
51
```

The other counts in the same sentences (7 recovery modes, 6 checkpoint policies) were correct, so the event count was an isolated outlier, and nothing in the suite caught it.

## Changes

**Figures updated to 51 — 11 stale sites** (the issue listed 7; several more had drifted under different phrasings):
- `docs/index.html` (3 rendered references on the landing page)
- `docs/assets/architecture.svg` (was **32**)
- `references/architecture.md`, `references/architecture-flow.md` (mermaid node: "hash-chained, 29 types")
- `references/architecture-drawing-prompt.md` (2)
- `references/architecture-diagram.md` ("Event types (29) … and 20 more" → "(51) … and 42 more")

**Translated READMEs** — `README.es.md`, `README.pt-BR.md`, `README.ja.md`, `README.zh-CN.md` still stated **36** and **44** while `README.md` said 51. All eight figures brought to 51 (the `events` table row drops the stale "en v0.2" qualifier to match the English form).

**Lists labelled exhaustive completed** with the full enum membership in definition order:
- `references/architecture-data.md` ("Complete list"), `references/architecture-spec.md` ("The complete set"), `references/architecture-drawing-prompt.md` (the inventory that exists to prevent the image model from inventing types)

Also re-pointed two stale `events.py` line references in `architecture-data.md` (`:46` → `:47`, `:304` → `:401`).

**Guard added** — `tests/test_docs_event_counts.py`, pinning the docs to the live enum:
- every stated figure equals `len(EventType)`, including the localized forms (`N tipos de eventos`, `N 種のイベントタイプ`, `N 种事件类型`) so the translations cannot drift again;
- every "complete" list matches the enum membership exactly (set *and* length, so a duplicate or a dropped member both fail);
- the sample names in `architecture-diagram.md` are real members and its "and N more" still adds up.

Historical deltas (e.g. CHANGELOG's "11 new event types") are deliberately not pinned: the number must sit directly on the phrase, so a delta cannot match.

## Verification

- `pytest tests/test_docs_event_counts.py tests/test_docs_counts.py -m "not slow"` → 7 passed.
[...]

PR #1214, opened 2026-09-16
docs: correct README event-type count to 51 and guard it

Cyrax321:
Fixes #1171.

The README stated two different event-type counts, both wrong, and they disagreed with each other in the same file:

| Location | Was | Now | Actual |
|:--|:--|:--|:--|
| README.md:199 (feature table, "Tamper-evident log") | "36 event types" | "51 event types" | `len(EventType)` = 51 |
| README.md:418 (storage architecture, `events` table) | "44 event types in v0.2" | "51 event types" | `len(EventType)` = 51 |

The storage row was wrong twice over: the count is 51, and the released package is continuum-agent 0.1.2, so "in v0.2" cited a specific count for a version that has not shipped. That version qualifier is dropped along with the count.

I verified the adjacent claims in those tables are still correct, so they were left alone: "Seven recovery modes" matches `RecoveryMode` (7 members, `src/continuum/models.py:122`), and "Twelve tools" matches the 12 `@server.tool` registrations in `src/continuum/mcp/server.py`. The event count was the isolated outlier, exactly as the issue describes.

Nothing caught the drift because `tests/test_docs_counts.py` pinned only the pytest totals. This PR adds `test_readme_event_type_count_matches_enum`, which asserts every "N event types" figure in README.md equals `len(EventType)`, computed live from the enum rather than pinned, so a future enum addition surfaces as a failing test instead of stale prose.

Scope note: `references/` still states "29 event types" in five places. That is #1100's tracked work, so the guard is README-only for now; widening it to `references/` would fail on the current tree. Once #1100 resyncs those files, the guard can be widened by adding them to the scanned set.

Verified: `pytest tests/test_docs_counts.py` passes (4 tests), the new guard fails as intended when the README is reverted to 36, `ruff check` and `ruff format --check` are clean, and `markdownlint-cli2` reports 0 issues on README.md.

## Summary by CodeRabbit

- **Documentation**
  - Updated the README to reflect the current total of 51 hash-chained event types in relevant feature and architecture sections.

- **Tests**
  - Added documentation consistency checks to verify that README event-type counts remain synchronized with the application’s supported event types.

PR #1281, opened 2026-09-18
docs: fix stale EventType count (29/32 -> 51), cover translations, add drift guard (#1100)

### LINKED ISSUE #1100: docs and references claim 29 event types; the EventType enum has 51 (seven sites, two asserting a complete list)

soorajsuru2004-sudo:
Closes #1100.

## Problem

The docs stated four different event-type counts while `EventType` holds 51 members:

```python
>>> len(EventType)
51
```

- **29** in seven places (landing page x3, `architecture.md`, `architecture-spec.md`, `architecture-data.md`, `architecture-drawing-prompt.md` x2)
- **32** in `docs/assets/architecture.svg`
- **36** / **44** in the four translated READMEs, while `README.md` already said 51

Two of the "29" sites are labelled *"The complete set"* / *"Complete list"* and enumerated only 29 names, so a reader was told the vocabulary is 22 members smaller than it is. The neighbouring figures in the same sentences (7 recovery modes, 6 checkpoint policies) were correct, making the event count an isolated outlier — and nothing in the suite caught it.

## Coverage — every site that states the count

Auditing the tree turns up more than the issue's seven: **19 stale figures**, listed here in full so a reviewer can confirm nothing was missed.

| Site | Was | Now |
|---|---|---|
| `docs/index.html` x3 (recovery note, feature pill, feature blurb) | 29 | 51 |
| `docs/assets/architecture.svg` (Event Log card) | 32 | 51 |
| `references/architecture.md:39` | 29 | 51 |
| `references/architecture-data.md:115` — "Complete list" | 29 | 51 |
| `references/architecture-spec.md:106` — "The complete set" | 29 | 51 |
| `references/architecture-drawing-prompt.md` x2 (card subtitle, inventory) | 29 | 51 |
| `references/architecture-flow.md:31` (mermaid label: "hash-chained, N types") | 29 | 51 |
| `references/architecture-diagram.md:21` ("Event types (29) ... and 20 more") | 29 | 51 / "and 42 more" |
| `README.es.md`, `README.pt-BR.md` (feature row + storage row each) | 36 / 44 | 51 |
| `README.ja.md`, `README.zh-CN.md` (same two rows each) | 36 / 44 | 51 |

Beyond the issue's seven "29" sites and the SVG's "32", this also covers the mermaid label, the diagram row, the second drawing-prompt site, and all eight translated-README rows.

**Lists labelled exhaustive completed** with the full 51-member enum in definition order:
- `references/architecture-data.md` ("Complete list")
- `references/architecture-spec.md` ("The complete set")
- `references/architecture-drawing-prompt.md` (the inventory that exists to stop the image model inventing types)

Also re-pointed two stale `events.py` line references in `architecture-data.md` (`:46` -> `:47`, `:304` -> `:401`).

**Guard added** — `tests/test_docs_event_counts.py`, following the existing `tests/test_docs_counts.py` pattern:
- every figure a doc states equals `len(EventType)`, in English, Spanish, Portuguese, Japanese and Chinese, plus the SVG's inline text and the mermaid label
- every list labelled complete matches the enum membership exactly (set *and* length, so a dropped member and a duplicate both fail)
- `architecture-diagram.md`'s sample names are real members and its "and N more" still adds up

Historical deltas (e.g. CHANGELOG's "11 new event types") are deliberately not pinned: the number must sit directly on the phrase, so a delta cannot match.

## Verification

- `pytest tests/test_docs_event_counts.py tests/test_docs_counts.py` -> 8 passed.
- Confirmed the guard is not vacuously green: it fails on the pre-fix figures (29/32/36/44) and on a list with a dropped member.
- `ruff check tests/test_docs_event_counts.py` clean.
- Enum membership verified directly against `src/continuum/events.py` (51 members, `ast`-parsed, not hardcoded).

## Note on #1280 and #1198

PR #1280 fixed the same issue and has been **closed** in favour of this one; its nine shared doc files are content-identical to those here, and this branch additionally fixes the four translated READMEs and adds the three localized guard regexes that keep them honest. The guard added here fails without those translation fixes, so they are in scope for the issue rather than extra. The older README-only #1198 (whose changes are already on `main`) is likewise superseded.


soorajsuru2004-sudo:
Heads up for reviewers: the **Test** matrix is red on this PR, but the single failure is pre-existing and unrelated to this change.

```
FAILED tests/test_adapters_openai.py::test_wrap_function_tool_invocation_binds_args_and_intercepts
  assert "An error occurred while running the tool. ... 'FakeTC' object has no attribute '_function_tool_arguments'" == {'endpoint': 'https://x', 'method': 'POST'}
```

`openai-agents` is pinned only `>=0.2`, so CI resolves the newest release (0.22.3), which now reads `ToolContext._function_tool_arguments`. The test's `FakeTC` subclasses `ToolContext` with an `__init__` that doesn't call `super().__init__()`, so the attribute is absent and the SDK swallows the `AttributeError` into its standard tool-error string. It passes locally on 0.22.2, and I confirmed it fails identically on other branches that touch no Python.

Everything this PR adds is green: `tests/test_docs_event_counts.py` (3 tests) and `tests/test_docs_counts.py` all pass in the same CI run — the suite is not run with `-x`, so they executed and are absent from the failure summary.

The dependency drift is worth its own issue/PR (pin the SDK or have `FakeTC` call `super().__init__()`); I've left it out of this docs change to keep the scope clean.

Cyrax321 (changes requested):
Author context: soorajsuru2004-sudo, external contributor. The docs direction is correct: I confirmed len(EventType) is 51 on this checkout, so the stale 29/32/36/44 figures this resyncs are genuinely wrong, and guarding the complete lists plus the N-plus-M arithmetic is the right shape. Blocking: the full Test matrix is red on this branch, all nine OS and Python legs. The author notes the single failure is the openai-agents FakeTC mock drift, which is plausible and unrelated to docs, but a docs PR cannot merge with a red matrix. Rebase onto current main where the adapter pin or mock is fixed, confirm the matrix goes green, and re-measure the 51 figures on the rebased tree in case the enum moved.

Cyrax321 (commented):
This corrects 19 stale EventType figures to 51 and adds a guard that pins every stated count and every list labelled complete to the live enum. I verified the substance off the head commit: the enum holds 51 members, all three exhaustive lists match its membership exactly (no missing or extra names), the diagram row's nine cited names are real members and 9 plus 42 more sums to 51, and the corrected source references (events.py:47, events.py:401) are accurate. All nine Test jobs are red, but on tests/test_adapters_openai.py, a file this docs-only change cannot affect; PRs #1316 and #1318 updated after this run pass all Test jobs, so a branch update plus re-run should clear it, and branch protection will require that before merge. Non-blocking overlap note: open PR #1214 also corrects the README event-type count and adds a README-only guard; this PR's guard supersedes that guard, so the two should not both land without dropping one.

tests/test_docs_event_counts.py
```diff
@@ -0,0 +1,178 @@
...
+    bad: list[str] = []
+    stated = 0
+    for path in _SCANNED_FILES:
+        if not path.is_file():
+            continue
+        for form, figure in _documented_figures(path):
+            stated += 1
+            if figure != expected:
+                bad.append(f"{path}: {form} says {figure}, enum has {expected}")
+    assert stated, "no document states an event-type figure; the guard lost its spine"
```
Cyrax321, line 114: Non-blocking. The only assertion that the guard still has input is the global `stated` counter: if any single doc (for example one translated README, which is exactly what drifted to 36/44 here) were edited to drop its count row entirely, the guard stays green because other files still state a figure. The sibling guard `tests/test_docs_counts.py` handles this with a REQUIRED_FILES list whose docs error when they stop stating the total; applying the same treatment to at least `README.md` would close the silent-removal hole this PR exists to fix.

tests/test_docs_event_counts.py
```diff
@@ -0,0 +1,178 @@
...
+)
+
+# Every prose and markup form the docs use for the total. The number must sit
+# directly against the phrase so a historical delta ("11 new event types") or
+# an unrelated figure cannot match.
+_FIGURE_RES = {
+    "N event types": re.compile(r"(\d+)\s+event types"),
+    "event types (N)": re.compile(r"event types\s*\((\d+)\)", re.IGNORECASE),
+    "hash-chained, N types": re.compile(r"hash-chained,\s*(\d+)\s+types"),
+    "svg text": re.compile(r">\s*(\d+)\s+event types\s*<"),
```
Cyrax321, line 52: Non-blocking. The SVG figure matches both this rule and the generic `N event types` rule above it, since the `>` and `<` anchors are outside the capture. That means a future drift in architecture.svg is reported twice in the failure message under two different form labels, which makes the output noisier than it needs to be; scoping the generic rule to exclude the `>...<` shape, or dropping this one, keeps each figure to a single report.

PR #1282, opened 2026-09-18
fix(snapshot): the store's read side rejects a key that is not a digest (#1268)

### LINKED ISSUE #1268: security: the file-snapshot store joins an unvalidated digest into a filesystem path, so rewind can copy an arbitrary file into the workspace
## Summary

The content-addressed snapshot store never checks that its key is actually a content address. `snapshot_path` builds the storage location by pure path concatenation (`src/continuum/environment/file_snapshot.py:16-18`):

```python
def snapshot_path(sha256: str) -> Path:
    """Return the content-addressed snapshot path for a SHA-256 digest."""
    return _SNAPSHOT_DIR / sha256
```

A `sha256` containing `../` escapes `.continuum/file-snapshots` and lands anywhere the process can reach. `restore_file` then copies whatever that path points at over a workspace file (`file_snapshot.py:60-73`), and `checkpoint/rewind.py:197-208` feeds it digests read straight out of stored `TOOL_COMPLETED` payloads:

```python
after_digest = after_payload.get("sha256")                       # rewind.py:157
before_digest = before_payload.get("sha256") ...                 # rewind.py:159
...
snapshot = snapshot_path(before_digest)                          # rewind.py:197
if not snapshot.exists(): ...                                    # existence oracle
if not restore_file(path, before_digest): ...                    # arbitrary copy
```

The only validation anywhere on the read path is `isinstance(after_digest, str)`.

The write side of this same store was hardened in #1110 (issue #1077): `snapshot_file` now digests the bytes as it copies and refuses a supplied digest the content does not match. That closed storing content under a name it does not have. It did not touch the read side, and the read side is the worse direction: it is an arbitrary-file read into the workspace, and it is reached through `continuum rewind`, the command whose purpose is to restore trusted content.

## Reachability, stated honestly

Every `TOOL_COMPLETED` writer shipped today computes the digest itself rather than accepting one, so no CLI or MCP path currently lets an agent plant a traversal string:

- `clienthooks.observe_event_payload` hashes the file in bounded chunks (`src/continuum/clienthooks.py:156-176`)
- `continuum observe` builds its payload through that same helper, so `--payload-file` cannot inject one either
- `gateway` records a fixed payload with no digest at all (`src/continuum/gateway.py:522-530`)

So this is a defect in the sink rather than a complete exploit chain today. What makes it worth fixing is that the trust the store relies on is nowhere enforced: `snapshot_path` is public API in `__all__`, the `sha256` payload field is documented as part of the `TOOL_COMPLETED` vocabulary, and any embedding application, adapter, or future tool that records a digest it did not compute itself turns `continuum rewind` into a file-read primitive. That is the same class of gap the project closes elsewhere by validating at the boundary rather than trusting every caller, and the content-addressed contract ("the path IS the digest") is currently unenforced on the side that copies bytes into the workspace.

`snapshot_path(before_digest).exists()` at `rewind.py:197` is also a path-existence oracle for arbitrary filesystem locations, independent of the copy.

## Affected component

continuum.environment (`snapshot_path`, `restore_file`), reached from continuum.checkpoint (`rewind.py`).

## Severity

Security hardening, not an active exploit. Real defect in the sink, no shipped write path reaches it with attacker-controlled input today, so I would not call it urgent. It becomes a genuine arbitrary-read the moment any writer trusts an agent-supplied digest, and the fix is one validation at the boundary that makes that impossible regardless of future callers.

## Evidence

At `origin/main` (`ad9941a4`), against a checkout where the snapshot directory already exists (it is created by the first real snapshot):

```
$ python -c "
from continuum.environment.file_snapshot import snapshot_path, restore_file
evil = '../../../secret/keyfile.txt'
print('snapshot_path ->', snapshot_path(evil))
print('exists        ->', snapshot_path(evil).exists())
print('restore_file  ->', restore_file('workspace/leaked.txt', evil))
print('workspace now ->', open('workspace/leaked.txt').read().strip())
"
snapshot_path -> .continuum/file-snapshots/../../../secret/keyfile.txt
exists        -> True
restore_file  -> True
workspace now -> TOP_SECRET
```

`/tmp/secret/keyfile.txt` was never snapshotted and is outside the store; it was copied into the workspace verbatim. Note the intermediate `.continuum/file-snapshots` must exist for the traversal to resolve, which it always does in a store that has taken at least one snapshot.

The `#1110` write-side fix is present on the same checkout and correctly refuses a mismatched digest, which confirms the read side is the remaining gap rather than the same bug twice.

## Fix sketch

Validate in `snapshot_path` that the digest is 64 lowercase hex characters and reject anything else, so a non-digest can never become a path. A `ValueError` at that single choke point covers `snapshot_file`, `restore_file` and the existence check in `rewind.py` at once, and a caller passing a genuinely malformed digest gets a clear message naming the value instead of a silent wrong-file copy. Existing snapshots on disk are unaffected: every one was written under a real digest.

Adhi1-2:
Closes #1268.

## The defect

The content-addressed snapshot store built its storage location by pure path concatenation:

\`\`\`python
def snapshot_path(sha256: str) -> Path:
    return _SNAPSHOT_DIR / sha256
\`\`\`

A \`sha256\` containing \`../\` escaped \`.continuum/file-snapshots\` and landed anywhere the process could reach. \`restore_file\` then copied whatever that path pointed at over a workspace file, and \`snapshot_path(key).exists()\` was an existence oracle for the same range — both reached through \`continuum rewind\`, the command whose purpose is to restore trusted content.

The **write** side of this store was hardened in #1110 (issue #1077), which closed filing content under a digest it does not have. It did not touch the read side, which is the worse direction.

## Reachability, stated honestly

No shipped CLI or MCP path lets an agent plant a traversal string today — every \`TOOL_COMPLETED\` writer computes its own digest (\`clienthooks.observe_event_payload\` hashes in bounded chunks; \`continuum observe\` goes through the same helper; the gateway records no digest at all). This is a defect in the **sink**, not a complete exploit chain. It is worth closing because the trust the store relies on was enforced nowhere on the side that copies bytes into the workspace: \`snapshot_path\` is public API in \`__all__\`, the \`sha256\` payload field is documented \`TOOL_COMPLETED\` vocabulary, and any adapter or embedding application that records a digest it did not compute itself turns \`rewind\` into a file-read primitive.

## The fix

A key is only a name in this store when it is exactly 64 lowercase hex characters, enforced at the single choke point where paths are assembled — which covers the write side, the read side, and the existence oracle together:

- \`snapshot_path\` raises \`ValueError\` naming the offending value.
- \`snapshot_file\` and \`restore_file\` fail closed to their existing \`None\`/\`False\` contracts, so a caller that does not catch still cannot reach \`copyfile\`. A caller that needs to tell a bad key from a missing one can pre-check the newly exported \`is_content_digest\`.
- \`rewind\` reports a malformed recorded digest as \`unrecoverable\` with the value in the message, rather than propagating or following it. The check runs before the \`dry_run\` branch too, so \`--dry-run\` is not an existence oracle either.

## Verification

- \`pytest tests/test_file_snapshot_edges.py tests/test_rewind_dual_state.py\` — 22 passed.
- Full suite: **2297 passed, 26 skipped, 0 failures**.
- \`ruff check\` + \`ruff format --check\` clean; \`mypy\` clean on both changed modules. (Local ruff is 0.16.2, CI pins 0.16.5.)
- Docs-count guard passes; the collected total stays within the documented ±30 tolerance, so no doc re-sync was needed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved snapshot security by validating snapshot identifiers before accessing stored files.
  - Invalid or unsafe identifiers are now rejected, preventing access outside the trusted snapshot storage location.
  - Snapshot and restore operations now fail safely without reading from or writing to unintended paths.
  - Rewind operations continue processing and clearly report files that cannot be recovered instead of failing unexpectedly.


Cyrax321 (dismissed):

Ref: #1282 / Issue #1268
Surface: content-addressed snapshot store, read side, src/continuum/environment/file_snapshot.py

Bug: snapshot_path joined the raw sha256 value into the filesystem path, so a stored digest containing ../ escaped .continuum/file-snapshots; restore_file would copy whatever that path named over a workspace file and snapshot_path(key).exists() served as an existence oracle. #1110 hardened the write side and left the read side open, which is the worse direction since it ends in arbitrary file copy.

Fix reviewed: snapshot_path now validates against a strict 64-lowercase-hex regex and raises ValueError on anything else, with the contract stated in the docstring and an exported is_content_digest helper. The rewind caller catches the rejection and reports the offending path as unrecoverable rather than copying from it, which is the right failure mode: loud, non-destructive, and per-file.

Proof:
- Ran tests/test_file_snapshot_edges.py and tests/test_rewind_dual_state.py at head 78f5110e: 22 passed, including traversal attempts with ../ and non-hex keys.
- Scanned the added lines for U+2014: none.
- The one failing CI check is Test (windows-latest, Python 3.12), and the failure is tests/test_serve_http.py::test_an_oversize_drain_over_the_bound_is_413_and_closes timing out on WinError 10053, a socket drain test this PR does not touch. The rest of the Windows matrix and every other platform pass. Treat as unrelated flake; re-run or merge with the green remainder.

Adhi1-2:
Rebased onto current main (head 812274c). All 30 checks green, including the full Windows matrix.

Note for whoever merges: the rebase push dismissed the earlier approval under the dismiss-stale-approvals rule, so this needs a fresh approving review before it can merge.

Cyrax321 (approved):
Author context: Adhi1-2, external contributor. Re-review at new head 812274c after the rebase that dismissed the prior approval. The security fix is unchanged and still correct: snapshot_path validates against a strict 64-lowercase-hex regex at the single choke point, covering the write side, the read side, and the rewind existence oracle. restore_file fails closed to False consistent with its other failure paths, and rewind reports a malformed stored digest as unrecoverable per file rather than copying from it. Verified locally: tests/test_file_snapshot_edges.py plus tests/test_rewind_dual_state.py, 17 passed. All 30 CI checks green on the branch. Approving.

src/continuum/checkpoint/rewind.py
```diff
@@ -198,7 +198,13 @@ def rewind_to_checkpoint(
             except OSError as exc:
                 conflicts.append(f"{path}: delete failed: {exc}")
         else:
-            snapshot = snapshot_path(before_digest)
+            try:
+                snapshot = snapshot_path(before_digest)
+            except ValueError as exc:
+                # A stored digest that is not a digest names a path, not a key;
+                # report it rather than copying whatever it points at (#1268).
+                unrecoverable.append(f"{path}: {exc}")
```
Cyrax321, line 206: Non-blocking: this interpolates the ValueError verbatim, and snapshot_path's message embeds the full offending value as `{sha256!r}` with no length bound, while every other message in this function truncates digests to `[:12]`. A poisoned payload can carry a sha256 string of any length, so it lands unbounded in `unrecoverable` and from there in the RewindError text that interpolates `unrecoverable[:3]`. Consider bounding the value here (for example the first 12 chars plus its length) so a malformed log entry cannot balloon the error surfaced to the user, matching the convention the surrounding messages already follow.

src/continuum/environment/file_snapshot.py
```diff
@@ -3,18 +3,43 @@
...
+]
 
 MAX_SNAPSHOT_BYTES = 10 * 1024 * 1024
 
 _SNAPSHOT_DIR = Path(".continuum/file-snapshots")
 
+# A SHA-256 hex digest: 64 lowercase hex characters and nothing else. The store
+# is content-addressed, so the path *is* the digest -- anything else joined into
+# it is a path, not a key (issue #1268).
+_DIGEST_RE = re.compile(r"\A[0-9a-f]{64}\Z")
```
Cyrax321, line 26: Non-blocking: this is now the third copy of the 64-hex contract in the tree, after `models._SHA256_PATTERN` and the inline `re.fullmatch(r"[0-9a-f]{64}", ...)` in `actions/ledger.py`. Since `is_content_digest` is exported here, a follow-up could have those two sites delegate their character check to it, so the canonical definition of what counts as a digest lives in exactly one place and cannot drift between modules.

Cyrax321 (commented):
This closes the read side of the snapshot store by validating the key at the single place paths are assembled, so snapshot_file, restore_file, and the exists() oracle all fail closed on anything that is not a 64-char lowercase hex digest, and rewind reports a poisoned recorded digest as unrecoverable ahead of the dry_run branch rather than following it. The anchoring is correct: \A...\Z with match() is a full match and also rejects trailing newlines, which $ would not. I traced both in-tree callers (clienthooks computes its own digest; rewind catches the ValueError), the write side is refused by the same check, and every read path funnels through snapshot_path, so the fix is complete at the chosen altitude. CI is green.

Nothing blocks the merge. Two non-blocking notes inline.

Cyrax321:
Hi @Adhi1-2,

This PR is approved and all checks are green. The only remaining issue is a **merge conflict** in `CHANGELOG.md` caused by recent updates on `main`.

Please **rebase** onto the latest `main`, resolve the conflict, and push. Once the branch is clean, this is ready to merge.

PR #1285, opened 2026-09-18
fix(storage): PostgresStorage.rebuild_action_index reports the rows it corrected (#1267)

### LINKED ISSUE #1267: storage: PostgresStorage.rebuild_action_index always reports 0 corrected rows, so verify --repair-index lies on Postgres
## Summary

```python
fixed = int(rebuild())
```

## Affected component

## Severity

## Evidence

```
        corrections = sum(
            1
            if before.get(k) != val
        )
        return corrections

                )
        return 0
```

```
```

## Fix sketch

Adhi1-2:
Closes #1267.

## The defect

`PostgresStorage.rebuild_action_index` ended in an unconditional `return 0` (`src/continuum/storage/postgres.py:703`), while `Storage` declares the method as \"recompute the index from the log; **returns corrected rows**\" and `SQLiteStorage` honours that. It was the only return-value divergence between the two engines.

The value is not internal — `cmd_verify` surfaces it straight to the operator:

```python
fixed = int(rebuild())
index_lines.append(f\"[ok] action i...
```

So on Postgres, a repair that fixed a corrupted projection reported success-with-zero-work: a repair report that cannot be trusted to describe the repair it just ran.

## The fix

The count now mirrors the SQLite engine exactly — changed-or-added rows plus rows the rebuild removed:

\`\`\`python
with self._write():
    before = {
        r[\"key\"]: (int(r[\"updated_seq\"]), r[\"status\"])
        for r in self._connection.execute(
            \"SELECT key, updated_seq, status FROM action_index\"
        ).fetchall()
    }
    self._connection.execute(\"DELETE FROM action_index\")
    ...
corrections = sum(1 for k, val in canonical if before.get(k) != val)
return corrections
\`\`\`

The `before` snapshot is read **inside the write lock and ahead of the `DELETE`** it is measured against, so a row a concurrent writer shifted between the two is counted rather than silently absorbed — the same reasoning the SQLite engine already applies by snapshotting inside its transaction.

## Verification

- New Postgres tests in `tests/test_storage_postgres.py` (mirroring the SQLite drift test): a corrupted status counts as a correction, a spurious row counts as a correction and is removed, and a clean rebuild reports `0`.
- Full suite: **2292 passed, 28 skipped, 0 failures**. The 2 new Postgres tests skip locally (no Postgres in this environment) and run in CI via the Postgres service container.
- `ruff check` + `ruff format --check` clean; `mypy` clean on the changed module. (Local ruff 0.16.2 vs CI-pinned 0.16.5.)


Adhi1-2:
Rebased onto current `main` and folded in the point the now-closed twin (#1288) made better than this branch did: the docstring now states *why* the rebuild is global (keys live in one store-wide namespace, so a per-run rewrite could collide with another run's legitimate row of the same key).

Also fixed a defect in the new tests I had not caught: the first Postgres regression test asserted `action_index_drift() == 0` as a precondition, but this suite shares one database across tests, so an earlier test's residue made the store-wide count non-zero on a full-file run and CI would have gone red on it. The precondition now checks the specific row the corruption is about to lie about instead, and the post-rebuild `drift() == 0` assertions are safe because the rebuild is global by design.

Verified on a live Postgres server (`CONTINUUM_TEST_POSTGRES_DSN` against a freshly recreated database): the full `tests/test_storage_postgres.py` suite is green, the stale-row and spurious-row counts both match the SQLite engine's, and a clean second rebuild reports 0. ruff, ruff format, mypy and markdownlint are clean.

Adhi1-2:
Superseded by #1325, which fixes the same #1267 symptom (the unconditional `return 0`) along with the drift criterion it depended on: `action_index_drift` no longer compares `updated_seq`, so the old count #1285 computed against it is no longer the right criterion. #1325 preserves this PR's concurrency property — the `before` snapshot is still read inside the write lock, ahead of the DELETE it is measured against. The branch stays available if #1325 doesn't land.

Issue #37, opened 2026-08-13
OpenAI adapter: tool arguments misbound and idempotency bypassed because __signature__ drops ctx


Cyrax321:
- [x] I searched existing issues for a duplicate
- [x] I am reporting this against the correct repository
- [x] I am using a supported Python version (3.11 or later)
- [x] I can reproduce this on the latest commit on `main`

## System information

OS: macOS 15.x (darwin)
Python: 3.13.5
continuum-agent: 0.1.0 (editable install from source)
openai-agents: 0.20.0
Source commit: a539948 (== origin/main)

## Command or code to reproduce

```python
import asyncio, tempfile, os
from continuum.storage.sqlite import SQLiteStorage
from continuum.adapters.openai import OpenAIAgentAdapter
from continuum.actions.ledger import ActionLedger
from agents.tool_context import ToolContext

db = tempfile.mktemp(suffix=".db")
store = SQLiteStorage(db)
adapter = OpenAIAgentAdapter(store)
adapter.start_run(goal="repro", run_id="run_oa_repro")
ledger = ActionLedger(store, run_id="run_oa_repro")

ctx_types = []
@adapter.wrap_function_tool("external.api_call")
def api_call(ctx, endpoint: str, method: str = "GET") -> dict:
    ctx_types.append(type(ctx).__name__)
    return {"endpoint": endpoint, "method": method}

class FakeTC(ToolContext):
    def __init__(self):
        self.tool_name = "api_call"
        self.context = None
        self.tool_input = {"continuum_run_id": "run_oa_repro"}

async def run():
    tc = FakeTC()
    for payload in ['{"endpoint": "https://x", "method": "POST"}',
                    '{"endpoint": "https://x", "method": "POST"}']:
        r = await api_call.on_invoke_tool(tc, payload)
        print("RESULT:", r)
    print("ctx types seen by tool:", ctx_types)
    print("action ledger entries:", len(ledger.all()))

asyncio.run(run())
os.remove(db)
```

## Current behaviour

The wrapped tool runs, but its arguments are silently misbound and the action
ledger is never written:

```
RESULT: {'endpoint': 'POST', 'method': None}
RESULT: {'endpoint': 'POST', 'method': None}
ctx types seen by tool: ['str', 'str']
action ledger entries: 0
```

Root cause: `OpenAIAgentAdapter.wrap_function_tool` builds the wrapper with
[...]

PR #1286, opened 2026-09-18
test(adapters): build the FakeTC ToolContext through the SDK's own __init__

Adhi1-2:
## Summary

The `openai-agents` requirement is unpinned (`>=0.2`), so the CI resolver installs the latest release. 0.22.3 made `FunctionTool._on_invoke_tool` read `ctx._function_tool_arguments`, an attribute `ToolContext.__init__` sets. The `FakeTC` double in `tests/test_adapters_openai.py` assigned `tool_name`, `context` and `tool_input` by hand without calling `super().__init__()`, so the attribute was missing and the #37 regression test failed with:

```
'FakeTC' object has no attribute '_function_tool_arguments'
```

That broke **every job of the test matrix — `main` included** (last green main run predates the 0.22.3 release), so it is blocking the merge gate on all open PRs, not just this one. No shipped adapter behaviour changed: in real use the SDK constructs the `ToolContext` itself, so the attribute is always present.

## Change

The double now constructs through the SDK's own constructor, passing `tool_call_id` and `tool_arguments` alongside the fields the test inspects:

```python
class FakeTC(ToolContext):
    def __init__(self) -> None:
        super().__init__(
            context=ContinuumContext(continuum_run_id=run_id, goal="g"),
            tool_name="api_call",
            tool_call_id="call_oa_37",
            tool_arguments=payload,
            tool_input={"continuum_run_id": run_id},
        )
```

It tracks whatever the SDK's invocation path requires instead of freezing a private field list from one release. `payload` moved above the class so the context can carry the raw arguments string.

## Verification

Reproduced the failure under `openai-agents==0.22.3` (identical to CI) and confirmed the fix is green under both 0.22.3 and 0.22.2, so it does not narrow the supported range. Full suite green on both versions; `ruff check`, `ruff format --check` and `mypy` clean.


Cyrax321:

Adhi1-2:
Closing as superseded: this fix landed on `main` in #1303 (`test(adapters): build FakeTC through ToolContext.__init__`), which takes the same approach as this branch (construct the double through `ToolContext.__init__` instead of assigning attributes by hand). Verified `upstream/main` @ `2e32336` passes `tests/test_adapters_openai.py` with `openai-agents==0.22.3` installed, so the duplicate has no remaining value.

Keeping the tree tidy rather than leaving a competing branch for the same one-line change.

Issue #532, opened 2026-08-30
CLI continuum events shows only live tail after compaction, missing archived prefix


Cyrax321:
Follow-up to #520 which fixed the dashboard pagination hint to count archived events.

## Summary

After `continuum compact <run>`, events before the anchor move to `events_archive`. The dashboard now correctly counts both live and archived for its pagination hint (fixed in #520). However `continuum events <run>` still reads only live events via `storage.read_events(run_id)`, so the hint's directive `see continuum events <run> for full log` is misleading after compaction: the command shows only the tail, not the full log.

## Reproduction

On main at 82ec017 with the dashboard fix applied:

```bash
python - <<'PY'
from continuum.storage import SQLiteStorage
from continuum.models import Run
from continuum.events import EventType
from continuum.cli.main import main
import io

db = "/tmp/test.db"
import pathlib, os
pathlib.Path(db).unlink(missing_ok=True)
from continuum.storage import SQLiteStorage
store = SQLiteStorage(db)
store.create_run_started(Run(run_id="run_1", goal="g"))
for i in range(98):
    store.append_event("run_1", EventType.TOOL_CALLED, {"i": i})
print("before", len(store.read_events("run_1")), len(store.read_archived_events("run_1")))
store.compact_run("run_1", through_sequence=80)
print("after live", len(store.read_events("run_1")), "arch", len(store.read_archived_events("run_1")))
store.close()

# CLI events shows only live tail
out, err = io.StringIO(), io.StringIO()
main(["--db", db, "events", "run_1"], out=out, err=err)
print(out.getvalue()[:500])
print("lines", len(out.getvalue().splitlines()))
PY
```

Before compact: 99 live, 0 archived. After compact through 80: 20 live + 80 archived = 100 total (plus 2 system events). Dashboard correctly shows `Showing last 20 of 100 events`. `continuum events run_1` shows only 20 live rows, not 100.

## Expected

Either:

- `continuum events <run>` should combine live and archived (via `read_archived_events` + `read_events` sorted by sequence, or `read_all_events` helper) so the hint is truthful, with flags like `--after` and `--upto` applied over the combined stream, or
- The hint should clarify that `continuum events` shows the live tail and that archived history is accessed differently (for example via `verify` or a dedicated `--include-archived` flag).

[...]

PR #1287, opened 2026-09-18
fix(tui): the Actions tab folds the archived prefix, so a compacted run keeps showing its actions (#1182)

### LINKED ISSUE #1182: tui: the Actions tab folds the live log only, so a compacted run shows no actions while it is blocked

Adhi1-2:
Closes #1182.

## The defect

`action_rows` in `src/continuum/tui/model.py` folded `storage.read_events(run_id)` — the **live tail only**. Compaction moves the claim into `events_archive` and leaves only anchor markers live, so the Actions tab rendered **no actions at all** for a compacted run.

This is the view an operator reaches for *while a run is blocked*, and it emptied itself at exactly the moment it was needed — hiding the rows that name what is blocking the run.

Every other consumer of the same fold already reads the archive, which is what makes this an outlier rather than a design choice:

| Site | Read |
|---|---|
| `tui/model.py:action_rows` | `read_events` ← the bug |
| `dashboard/hitl.py:130` | `read_all_events` |
| `ActionLedger._replay` | `read_archived_events` + `read_events` |
| `reconcilers.py:262` | `read_all_events` |
| `cli/main.py` (gate, `continuum actions`) | `read_all_events` |

`ActionLedger._replay`'s docstring states the invariant this breaks: *"a claim settled before compaction must keep protecting afterwards, or exactly-once would quietly reset at the anchor boundary."*

## The fix

One line — read the same stream `event_rows` (one function down, #532) already reads:

\`\`\`python
\`\`\`

The two views now fold the same history and cannot diverge again. The docstring is updated to say why, so the next reader doesn't reintroduce the live-tail assumption.

## Verification

- New regression test `test_action_rows_survive_compaction`. It asserts the action event genuinely left the live tail (so it is testing the bug, not a tautology), then that the row still renders. **Confirmed it fails on the unpatched code and passes with the fix** — I reverted the source and ran it both ways.
- Full suite: **2293 passed, 26 skipped, 0 failures**.
- `ruff check` + `ruff format --check` clean; `mypy` clean on the changed module. (Local ruff 0.16.2 vs CI-pinned 0.16.5.)


Cyrax321:

Adhi1-2:
Closing as a duplicate: this fix already landed on `main` through #1107 (fix(tui): connect the dashboard to current core APIs and harden every read), which rewrote `action_rows` to fold through `ActionLedger(storage, run_id).folded()` -- the ledger's own archive-aware view, the same one the dashboard HITL buttons use. Verified against `upstream/main` at `19cc0f0`: the function is already archive-aware, so this branch would only re-express the same behaviour with `fold_action_events(storage.read_all_events(...))` instead of the ledger's own method. #1107's delegation to the ledger is the cleaner of the two, so the duplicate has nothing left to add.

Issue #1182 is still open even though the behaviour it reports is gone; I have pointed the issue at #1107 so a maintainer can close it.

PR #1288, opened 2026-09-18
fix(storage): Postgres rebuild_action_index reports the rows it corrected (#1267)

### LINKED ISSUE #1267: storage: PostgresStorage.rebuild_action_index always reports 0 corrected rows, so verify --repair-index lies on Postgres

Adhi1-2:
## Summary

`PostgresStorage.rebuild_action_index` ended in an unconditional `return 0`, while the `Storage` base class declares the method as *"recompute the index from the log; returns corrected rows"* and `SQLiteStorage` honours that. It was the only return-value divergence between the two engines.

The value is user-facing: `cmd_verify` surfaces it straight to the operator, so on Postgres `continuum verify --index --repair-index` always printed

```
[ok] action index repaired from the log (0 row(s) corrected)
```

even when the rebuild rewrote every row. An operator repairing a drifted index could not distinguish "nothing was wrong" from "I just fixed N rows" — exactly the distinction the count exists to carry. The repair itself was always correct and the index consistent afterwards; nothing safety-critical depended on the figure.

## Change

Postgres now computes the same before/canonical comparison SQLite does — reading `action_index` before the rewrite and counting keys whose stored row was missing, stale or spurious relative to the canonical fold, plus keys present in the store but absent from the fold — and returns it. Its docstring states the contract the base class already declared, where before it dropped it.

## Tests

The existing Postgres index test (`test_pg_action_index_covers_the_archive_after_rebuild`) exercised the rebuild and asserted drift before and after but never looked at the return value, which is why the regression went unnoticed. It now asserts the returned count equals the drift `action_index_drift()` reported, so a hardcoded `0` fails the suite. Run against a live Postgres instance via `CONTINUUM_TEST_POSTGRES_DSN`: the 19-test Postgres suite is green, the full suite is green, and `ruff`/`mypy` are clean.

Closes #1267.


Adhi1-2:
Closing in favour of #1285, which fixes the same #1267 and additionally takes the before-snapshot of `action_index` under the write lock and ahead of the `DELETE`, so the corrected-rows count measures the index the rebuild replaced rather than one a concurrent writer could shift underneath it. This branch's count is correct only under a quiet store; #1285's is correct always, and it carries the same one-line behaviour change and CHANGELOG entry. Keeping the strictly-stronger variant and folding this branch's docstring point (the store-wide key namespace being why the rebuild is global) into it.

PR #1289, opened 2026-09-18
fix(cli): replay --upto reads the archived prefix on a compacted run (#1172)

### LINKED ISSUE #1172: cli: replay --upto fails on every compacted run and blames the wrong cause

Adhi1-2:
## What

`continuum replay <run> --upto N` failed for **every** value of `N` on any compacted run, and the failure message told the operator to do the one thing that cannot help.

```
$ continuum replay r1 --upto 999     # run head is 13
error: --upto 999 excludes the RUN_STARTED event for run 'r1'; increase --upto or omit it to replay from the beginning
```

`--upto` is documented (`references/cli.md:21`) as the tool for bisecting where a run's state diverged. Compaction is the feature that exists for long-lived runs, so the diagnostic was missing precisely where it is most likely to be needed.

## Root cause

`cmd_replay` read the **live tail only**:

```python
events = storage.read_events(args.run_id, upto=args.upto)
```

`read_events` queries the `events` table, never `events_archive`. After compaction, `RUN_STARTED` lives in the archive, so the guard fired for every `N` — including `999` on a run whose head was 13, which is what proves the message was wrong about the cause.

## The fix

`cmd_replay` now windows the **full** history, archived prefix included, mirroring `cmd_events` (`cli/main.py:701-713`), which reached this contract first under #532:

```python
events = [
    event
    for event in storage.read_all_events(args.run_id)
    if args.upto is None or event.sequence <= args.upto
]
```

`STATE_CHECKPOINTED` and `EVENT_LOG_ANCHORED` are both in `_NON_PROJECTING` (`state/semantic.py:750`), so folding the archived prefix from genesis reaches the same state the anchored path already produced. The existence guard and the `--upto 0` guard are unchanged — the latter still fires when a window genuinely excludes `RUN_STARTED`.

## A second site the fix would have broken

`_verify_against_stored`, one function down, had the same blind spot:

```python
prefix = storage.read_events(run_id, upto=stored.source_sequence)
```

The stored prefix starts at sequence 1, and compaction moves exactly that range into the archive. On a compacted run this reads an **empty** log, so once the primary read was corrected this comparison would have reported every healthy checkpoint as corrupt:

```
$ continuum replay r1 --upto 4
Replayed 4 events -> 2 completed, ...
Verification: DOES NOT match stored version 0 at sequence 13   <-- false alarm
```

Both sites now window `read_all_events`. Each is load-bearing — reverting either one alone makes the new test fail with a distinct symptom (the guard error, or `has no goal` from projecting an empty prefix).

## Why `cmd_replay` and `cmd_events` disagreed

Every other consumer of the event log already reads the archive:

| consumer | read |
|---|---|
| `cmd_events` (`cli/main.py:711`) | `read_all_events` |
| `ActionLedger._replay` | `read_all_events` |
| dashboard HITL (`dashboard/hitl.py:130`) | `read_all_events` |
| `reconcilers.py:262` | `read_all_events` |
| TUI `event_rows` / `action_rows` | `read_all_events` |
| **`cmd_replay`** | ~~`read_events`~~ → `read_all_events` |

## Verification

- New regression tests in `tests/test_cli.py` over a freshly compacted run: `--upto 4 / 13 / 999` all exit 0, fold the prefix their window actually selects (2 / 10 / 10 completed), and report `matches stored version`; `--upto 0` is still refused with the message that can actually help.
- Both fixes proven load-bearing by reverting each independently.
- Full suite: **2293 passed, 26 skipped, 0 failed** (`PYTHONPATH` pinned to the worktree `src/` — `continuum` is an editable install aimed at the main checkout).
- Postgres backend exercised for real against a fresh database: **20 passed** in `test_storage_postgres.py` + `test_provenance_postgres.py`, and end-to-end `replay --upto` on a compacted Postgres run exits 0 with `matches stored version`.
- `ruff check`, `ruff format --check`, `mypy` clean; docs-count guard passes.

Closes #1172.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed replaying compacted runs with the `--upto` option.
  - Archived event history is now included during replay and verification, preserving expected results across compacted and live event segments.
  - The validation error remains available when the requested range genuinely excludes the run start event.

- **Tests**
  - Added coverage for replaying compacted runs with multiple `--upto` values.


Cyrax321 (dismissed):

Ref: #1289 / Issue #1172
Surface: continuum replay --upto, src/continuum/cli/main.py

Bug: cmd_replay read the live tail only, so after compaction the --upto guard fired for every value of N, including values above the run head, and the error message told the operator to increase --upto, the one thing that cannot help. Verified the root cause reading: read_events queries the events table and never events_archive, while RUN_STARTED lives in the archive after compaction.

Fix reviewed: cmd_replay now windows full history, archived prefix included, mirroring the contract cmd_events already established under #532. The bisecting use case named in references/cli.md works again on compacted runs, and the guard's semantics are restored to what the message claims.

Proof:
- Ran tests/test_cli.py at head a895fa14: 110 passed, 6 failed. All 6 are attest tests failing on the missing optional cryptography package in my local environment (same 6 fail on other branches' heads here); the new replay tests pass and CI is green on the head commit.
- No U+2014 in the added lines.

PR #1290, opened 2026-09-18
fix(storage): the compaction anchor records the environment it was validated against (#1049)

### LINKED ISSUE #1049: compact_run mints an environment-blind anchor checkpoint, so a clean run downgrades to REQUEST_HUMAN
## Summary

`compact_run` takes a forced anchor checkpoint before archiving, and it never passes an environment snapshot. `SQLiteStorage.compact_run` (`src/continuum/storage/sqlite.py:511`) and `PostgresStorage.compact_run` (`src/continuum/storage/postgres.py:603`) both call `CheckpointManager(self).checkpoint(run_id, force_version=True)` with no `environment` argument, so the anchor's `StateCheckpoint.environment` is `None`. That anchor becomes the newest checkpoint, so `RecoveryEngine.assess` reads it and hands `None` to the validator.

`diff_environments(None, current)` returns an empty diff by design (`src/continuum/environment/diff.py:139`, "absence of evidence is not evidence of stability"), so `_apply_dependency_status` takes the no-delta branch and marks every pinned dependency `UNKNOWN` with detail "no environment snapshot to compare against" (`src/continuum/state/validator.py:380-388`). `UNKNOWN` is in `_UNUSABLE`, so `safe_to_resume` is false and the mode escalates. A run that resumed cleanly one moment before compaction downgrades to `REQUEST_HUMAN` one moment after, with nothing about the world having changed. Compaction is a storage hygiene operation, and it currently silently downgrades every strict recovery verdict for any run that pins a dependency.

## Affected component

continuum.storage (both engines), continuum.checkpoint.manager, continuum.recovery.engine

## Severity

Major functionality broken, workaround exists. Compaction is a routine, encouraged operation, and the failure is silent: the verdict changes for a reason invisible to the operator and to the run itself. The workaround is to never compact a run you may need to resume, which defeats the feature.

## Evidence

At HEAD `14f8a84`, with a run seeded, checkpointed with a `dataset` snapshot, then compacted:

```
BEFORE compact: resume | safe: True | cp env: True
  reason: all components verified against the current environment
compact archived: {'archived': 8}
AFTER  compact: request_human | safe: False | cp env: False
  reason: external_dependency dataset is unknown
   status: external_dependency unknown - no environment snapshot to compare against
```

The post-compact anchor checkpoint, read directly: `version: 1 | environment: None | trigger: manual`. A `strict_unknown=False` engine still resumes after the same compaction, confirming the environment gap rather than the archive is what flips the verdict.

```python
import tempfile, os
from continuum.environment.snapshot import capture, StaticProvider

env = lambda: capture("run_1", StaticProvider(dataset="v3"))
db = os.path.join(tempfile.mkdtemp(), "a1.db")
store = SQLiteStorage(db)
store.create_run(Run(run_id="run_1", goal="Analyze 100 documents"))
store.append_event("run_1", EventType.RUN_STARTED, {"goal": "Analyze 100 documents", "total": 100})
store.append_event("run_1", EventType.DEPENDENCY_DECLARED, {"resource": "dataset", "version": "v3"})
for i in range(5):
CheckpointManager(store).checkpoint("run_1", environment=env())

before = RecoveryEngine(store).assess("run_1", current_environment=env())
print("BEFORE:", before.mode.value, before.safe)
print("anchor env:", store.latest_checkpoint("run_1").environment)

store.compact_run("run_1")                                # forced anchor, no environment

after = RecoveryEngine(store).assess("run_1", current_environment=env())
print("AFTER:", after.mode.value, after.safe)
print("anchor env:", store.latest_checkpoint("run_1").environment)
for e in after.validation.report.statuses:
    print(" ", e.component.value, e.status.value, e.detail)
```

## Root cause

Introduced with compaction itself in `65bb275` (2026-08-23, "feat(storage): event-log compaction with archive table (schema v5)"). The forced anchor was added to guarantee a projectable boundary, and it reused the plain `checkpoint()` entry point, which has always treated `environment` as a caller-supplied optional.

## Expected behaviour

`continuum compact` must not change a recovery verdict. The anchor checkpoint should carry the environment the run was validated against, so a post-compaction `assess` with an unchanged environment resumes exactly as it did before.

## Suggested fix shape

Thread the environment through the boundary: have `compact_run` accept an optional `environment` (or `EnvironmentProvider`) and forward it to the forced `checkpoint()` call, with the CLI's `cmd_compact` (`src/continuum/cli/main.py:1729`) capturing one the way `continuum validate` does. A cheaper fallback that still fixes the reported symptom: when `environment` is None, carry forward the environment from the run's previous checkpoint rather than storing None, since the anchor represents state at a moment the run had already been validated. Add a compact-then-assess regression test asserting the mode is unchanged.

Related: the liveness, risk and authority scans in `assess` are also archive-blind, tracked separately.

## System information

- Python: 3.14.5

Adhi1-2:
## What

Fixes #1049.

`compact_run` minted a forced anchor checkpoint with no `environment` argument, so the anchor's `StateCheckpoint.environment` was always `None`. That anchor becomes the run's newest checkpoint, and `RecoveryEngine.assess` hands `None` to the validator, which marks every pinned dependency `UNKNOWN` for want of a snapshot to compare against.

A run that resumed cleanly one moment before compaction downgraded to `REQUEST_HUMAN` one moment after, with nothing about the world having changed. Compaction is a storage hygiene operation, and it silently moved every strict recovery verdict for any run pinning a dependency.

## The change

- `Storage.compact_run` accepts an optional `environment` (both engines). `continuum compact` captures one from `--env` the way `continuum validate` does.
- When the caller supplies nothing, the anchor **carries forward** the environment the run's newest checkpoint already recorded — compaction observes the world, it does not change it, so the last verified environment is the one the boundary represents. The helper `_anchor_environment` lives on `Storage` so both engines agree and cannot drift.
- A run with no recorded checkpoint still anchors with `None` rather than inventing a snapshot.
- `EnvironmentSnapshot` was already imported across module boundaries but exported from neither `continuum.environment.snapshot` nor the `continuum.environment` package; it is now in both `__all__`s (this is what made the new type annotations mypy-clean).

This is the issue's recommended design: thread the environment through *and* carry it forward as the fallback, so programmatic and MCP callers that supply nothing are covered too.

## Verification

The issue's reproduction script verbatim, at the anchor both before and after compaction:

```
BEFORE: resume True
anchor env: True
AFTER:  resume True
anchor env: True
```

(At the issue's HEAD this printed `AFTER: request_human | safe: False` with `external_dependency unknown - no environment snapshot to compare against`.)

Regression tests on **both** engines:
- `tests/test_compaction.py` — a pinned run assessed `RESUME` before compaction still assesses `RESUME` after; the anchor's environment is non-None and carries the pinned version; a caller-supplied snapshot wins over the carried-forward one; nothing is invented when there is no checkpoint to inherit from; and the `continuum compact` CLI path holds the verdict too.
- `tests/test_storage_postgres.py` — parity test asserting the Postgres anchor records the environment.

Gates: `ruff check`, `ruff format --check`, `mypy src/continuum` clean; full SQLite suite green. Postgres contract suite run locally against Postgres 16 (20/20, including the new test).

I confirmed the tests pin the behaviour rather than just passing: reverting the carry-forward in `sqlite.py` makes three of the new tests fail with the issue's exact symptom.

## Note

One pre-existing unrelated failure, `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend`: it expects a `RuntimeError` mentioning `psycopg`, but with psycopg installed locally the code path instead makes a real connection attempt to a nonexistent `continuum` database and raises a different error. It fails identically with my changes stashed and is green in CI, where the main suite runs without the driver.


Adhi1-2:
Small convention cleanup in `3b3e486`: the two U+2014 em dashes the #1049 CHANGELOG entry introduced are replaced with a colon and a comma, per AGENTS.md. Nothing else in the diff changed. No blocker had been raised on this one; I swept my open PRs for the same convention after it was flagged elsewhere and found these.

Cyrax321 (changes requested):
Author context: Adhi1-2, external contributor. The storage mechanism is correct in shape: threading an optional environment through compact_run on both engines with carry-forward from the newest checkpoint, and inventing nothing when there is nothing to inherit, matches the claim that compaction observes rather than changes the world. Blocking: the CLI path this PR advertises is dead on this base. The compact subparser at cli/main.py:4046-4049 is built with with_run only, not with_env, so _environment always returns None for compact and continuum compact --env fails with unrecognized arguments. I verified the subparser on this checkout. Wrap compact in with_env the way checkpoint and validate are, and add a CLI test that actually passes --env. The em dash cleanup is acknowledged. Approving the storage direction, requesting changes for the CLI wiring.

src/continuum/cli/main.py
```diff
@@ -1796,7 +1796,7 @@ def cmd_compact(args: argparse.Namespace, storage: Storage, out: Any, err: Any)
             file=err,
         )
         return ExitCode.ERROR
-    report = storage.compact_run(args.run_id)
+    report = storage.compact_run(args.run_id, environment=_environment(args, args.run_id))
```
Cyrax321, line 1799: Blocking: `_environment` reads `getattr(args, "env", None)`, but the compact subparser is registered further down as `compact = with_run(add("compact", cmd_compact, ...))` without `with_env`, so `continuum compact run_1 --env dataset=v3 --force` dies in argparse with `unrecognized arguments`. That makes the explicit path this PR advertises (body: "continuum compact captures one from --env the way continuum validate does", and the CHANGELOG entry repeats it) unreachable: from the CLI, `_environment` can only ever return None, so the carry-forward branch always applies and a caller who wants to pin a specific snapshot cannot. The carry-forward is the right default and covers the common case, which is why the fix is small rather than a redesign: register the flag as `compact = with_env(with_run(add("compact", ...)))` the way validate, resume, and checkpoint do, and add a test that invokes `run("--db", db, "compact", "run_1", "--env", "dataset=v4", "--force")` so the wiring cannot silently drop the flag again. Until that lands, the CHANGELOG entry this PR ships describes a flag that does not exist.

src/continuum/environment/snapshot.py
```diff
@@ -25,7 +25,8 @@
 from pathlib import Path
 from typing import Any
 
-from continuum.models import EnvironmentSnapshot, EnvResource, utcnow
+from continuum.models import EnvironmentSnapshot as EnvironmentSnapshot
```
Cyrax321, line 28: Non-blocking: the PR body says EnvironmentSnapshot "is now in both `__all__`s", but this module's `__all__` just below is unchanged and does not contain it; the `X as X` alias is what actually marks the re-export for mypy and pyright, and direct imports work with or without the list. Either add the name to `__all__` so the module's star-import surface matches the package-level export and the PR body, or reword the body, otherwise the next reader is told the module advertises a name it does not. Functionally nothing breaks, which is why this does not block.

Cyrax321 (changes requested):
This threads an optional environment through compact_run on both engines, mints the anchor with it, and carries forward the newest checkpoint's snapshot when the caller supplies none, which is exactly the design #1049 recommended. The carry-forward mechanics are correct: the inheritance read happens before the new anchor is minted, a caller-supplied snapshot wins, and nothing is invented when the run has no checkpoint, and both engine overrides plus the new SQLite and Postgres tests pin that behaviour.

One blocker: the compact subcommand never registers --env, so the explicit CLI path this PR and the CHANGELOG advertise is unreachable from the shell (details inline). It is a one-line fix using the existing with_env helper, plus one test that actually passes the flag.

A small note on the EnvironmentSnapshot export claim is inline as well. CI is green apart from codecov/patch, which the new anchor and CLI tests address in substance.

Cyrax321 (changes requested):
Hi @Adhi1-2,

This PR has **merge conflicts** in `CHANGELOG.md` and is currently blocked by a missing CLI argument registration.

Please address the following:

1.  **Resolve Merge Conflicts**: Rebase onto the latest `main` to resolve the conflict in `CHANGELOG.md`.
2.  **Fix CLI Wiring**: The `compact` subcommand does not currently register the `--env` flag. As noted in the review, please wrap the subcommand with `with_env` (similar to `validate` and `checkpoint`) so that `continuum compact --env` works as advertised.
3.  **Add CLI Test**: Include a test that explicitly passes `--env` to the compact command to ensure the wiring remains intact.

Once these changes are pushed and CI is green, we can proceed with the merge.

PR #1291, opened 2026-09-18
fix(cli): resume --pinning folds the archived pinning on a compacted run (#1126)

### LINKED ISSUE #1126: resume --pinning drift reads the live tail only, so a compacted run reports every pin as newly pinned

Adhi1-2:
## What

`continuum resume --pinning` reported **every key as newly pinned** on any compacted run — including a run whose recorded pinning was byte-identical to the one passed at resume:

```
BEFORE compact:
  recorded: {'model_id': 'm-2024-09', 'prompt_sha256': 'aaaa…'}
  drift: []
AFTER compact:
  recorded: {}
  drift: ['pinning drift: model_id newly pinned (m-2024-09...)',
          'pinning drift: prompt_sha256 newly pinned (aaaa...)']
```

The display is informational and never gates recovery (issue #241), but it was wrong in the direction of **hiding** drift, which is the opposite of what a drift report is for.

## Root cause

`cmd_resume` folded the live tail only:

```python
recorded = latest_pinning(storage.read_events(run_id))
```

`latest_pinning` scans `ACTION_RECORDED` payloads for the newest non-empty `pinning`. Compaction moves those events into `events_archive`, and the `EVENT_LOG_ANCHORED` marker that replaces the prefix carries only `anchored_through`/`version` — no pinning. So the fold read `{}`, and `pinning_drift({}, current)` has no choice but to call every key newly pinned.

Three directions were masked, all the ones an operator would act on:

| reality | rendered as | what it hides |
|---|---|---|
| unchanged | `newly pinned` (every key) | a false alarm on a run that provably did not change |
| **changed** hash | `newly pinned` | the **previous value** — the one fact that says what moved |
| **unpinned** key | *(nothing)* | the `unpinned (was ...)` line needs the old value the empty record cannot supply |

## The fix

`latest_pinning` now folds the full history:

```python
recorded = latest_pinning(storage.read_all_events(run_id))
```

the same history the assess (#1050) and watch (#1072) pinning folds already read. `latest_pinning` itself is unchanged — it already took an event sequence, and this is the same class as #1050/#1072 at a separate read site, as the issue notes.

## Verification

Three new tests in `tests/test_cli.py` over a freshly compacted run, one per masked direction:

- unchanged pinning → **no** drift section at all
- changed hash → `prompt_sha256 changed (aaaa...` with the archived value, and **not** `newly pinned`
- dropped key → `prompt_sha256 unpinned (was` present

All three fail on unpatched code — the failure output is the bug verbatim (`pinning drift: model_id newly pinned (m-2024-09...)` on a run where nothing drifted) — and pass with the fix.

- Full suite: **2294 passed, 26 skipped, 0 failed** (`PYTHONPATH` pinned to the worktree `src/`; `continuum` is an editable install aimed at the main checkout).
- Postgres backend exercised against a fresh database.
- `ruff check`, `ruff format --check`, `mypy` clean; docs-count guard passes.

## Note on the fixture

The test run declares no external dependency, which is why: compaction mints an **environment-blind** anchor checkpoint, which would gate `resume` on `REQUEST_HUMAN` before the drift display is reached. That is a separate, already-filed issue — **#1049** — and I kept this PR scoped to the pinning fold rather than expanding into it. It is worth flagging that #1049 makes any compacted run with a declared dependency downgrade to `REQUEST_HUMAN` on resume, which is materially more severe than the display bug fixed here.

Closes #1126.


Cyrax321 (approved):

Ref: #1291 / Issue #1126
Surface: continuum resume --pinning, src/continuum/cli/main.py

Bug: cmd_resume folded the live tail only when reconstructing recorded pinning, and the EVENT_LOG_ANCHORED marker that replaces a compacted prefix carries no pinning, so every compacted run reported every key as newly pinned. The report is informational, but it was wrong in the direction of hiding real drift behind fake drift, which defeats the purpose of the report.

Fix reviewed: cmd_resume folds the archived prefix plus the live tail for the pinning reconstruction, matching the full-history contract the other archive-aware callers use. The drift computation itself is untouched, which is correct: the defect was the input, not the comparison.

Proof:
- Ran the pinning and resume selections of tests/test_cli.py at head fc6f685d: 9 passed, including the compacted-run regression cases.
- No U+2014 in the added lines. CI green on the head commit.

Related: #1289 gives replay --upto the same archive-aware treatment. The two are consistent in approach and touch adjacent lines of cmd_resume/cmd_replay; merge in either order and resolve the trivial textual overlap when the second lands.

PR #1293, opened 2026-09-18
fix(interchange): export_evidence builds the primitive models it declares (#1155)

### LINKED ISSUE #1155: interchange: the four exported EvidencePrimitive models are never constructed; export_evidence returns untyped dicts
## Summary

`src/continuum/interchange/evidence.py` declares five pydantic models — `EvidencePrimitive` plus `Transition`, `Observation`, `Relation`, `Checkpoint` — and exports all of them in `__all__`. `export_evidence` (the module's reason for existing) then builds **plain `dict[str, Any]` values by hand** (`:237-321`) and returns `list[dict[str, Any]]`. None of the four subclasses is ever constructed anywhere in `src/`, `tests/`, `docs/`, or `examples/`.

The models are not inert scaffolding waiting for a caller — they are the module's own type system, and the exporter bypasses it. So:

- Every field declared on the subclasses (`event_id`, `event_type`, `observed_at`, `source_id`, `target_id`, `checkpoint_id`, `version`, `trigger`, `integrity_hash`) exists only in a class the producing function never touches. A field can be added, renamed, or dropped from a subclass and nothing breaks, because the dict keys are spelled separately in the function body.
- `EvidencePrimitive.content()` and `.digest()` are unreachable with the models unconstructed.
- `verify_export` (`:327`) accepts `list[dict[str, Any]]` too, so a receiver has no typed shape to work against either; the docstring's own verification example reads `prim["content_hash"]` as a dict subscript.

The only real consumer, `cmd_export_evidence` (`cli/main.py:3502-3516`), imports `export_evidence` alone. This is the same dead-export family as #1093 / #1111, but inside one module rather than across `__all__` omissions.

## Affected component

Other (interchange / evidence export)

## Severity

Minor inconvenience or cosmetic issue — no incorrect behaviour; a declared and exported type layer that the code path it belongs to never uses.

## System information

- Python: 3.13.5

```bash
git grep -n 'EvidencePrimitive' -- src/ tests/ docs/ examples/
# src/continuum/interchange/evidence.py:37   __all__ entry
# src/continuum/interchange/evidence.py:53   class EvidencePrimitive
# ...four subclasses, no construction site anywhere

git grep -n 'Transition(\|Observation(\|Relation(\|Checkpoint(' -- src/ tests/
# (no hits — the models are never instantiated)
```

## What a fix looks like

Two directions, and the choice is the point of the issue:

1. **Wire them.** Construct the appropriate subclass at each of the four `kind` branches in `export_evidence` and return `list[EvidencePrimitive]`. Pydantic's `extra="forbid"` then makes a missing or stray field an immediate error rather than a silent dict-key drift, and `content()` / `digest()` become the chain-verification primitives instead of hand-rolled dicts. Callers serialize with `prim.model_dump()`; `cmd_export_evidence` needs that one-line change, and `verify_export` should take the models (or accept either).
2. **Remove them**, along with `content()` and `digest()`, and document that the export's contract is the dict shape the docstring at `:186-199` already describes.

Option 1 is what the module's own docstring implies — "Each primitive is content-addressed ... and the receiver can detect truncation or tampering" reads as a typed guarantee, not a dict convention. A fixer should also check whether the `Relation` source/target extraction at `:256-263` (which guesses `resource` / `decision_id` / `finding_id`) belongs in the model as a validator, since that is the one place the dict construction has non-trivial logic that a typed constructor would surface.

Adhi1-2:
## What

Fixes #1155.

`src/continuum/interchange/evidence.py` declared `EvidencePrimitive` plus the four subclasses `Transition`, `Observation`, `Relation`, `Checkpoint`, exported all of them — and constructed **none** of them. `export_evidence` hand-rolled `dict[str, Any]` values and returned `list[dict[str, Any]]`.

So every subclass field (`event_id`, `event_type`, `observed_at`, `source_id`, `target_id`, `checkpoint_id`, `version`, `trigger`, `integrity_hash`) existed only in a class the producing function never touched. A field could be added, renamed or dropped and nothing would break, because the dict keys were spelled separately in the function body. `content()` and `digest()` were unreachable, and `verify_export` had no typed shape to work against either.

## The change — Option 1 (wire them)

- `export_evidence` returns `list[EvidencePrimitive]`, constructing the right subclass per event. `extra="forbid"` is now a live check instead of scaffolding: a missing or stray field raises at export rather than silently shipping a different shape.
- `Relation.from_event` hosts the one piece of construction that inspects a payload (a dependency edge's endpoints are spelled `resource` for a declaration, `decision_id` / `finding_id` for those families) next to the fields it fills — the non-trivial logic the issue flagged as belonging in the model.
- `Checkpoint` now declares `event_id` / `event_type`, which the dict it replaced always emitted. Without them, `extra="forbid"` would reject keys the format has shipped since day one. So the model is now truthful about the shape it produces.
- `cmd_export_evidence` serialises with `model_dump(mode="json")`.
- `verify_export` accepts the models **or** the mappings a JSON-lines receiver produces after `json.loads`, so the in-memory and on-disk paths run the same check.

## Wire format: unchanged

I kept the emitted JSON byte-for-byte compatible, because changing it would break receivers comparing against `verify()` output. Verified by driving the CLI over a run with all four kinds:

```
transition   11 keys — same set as before
observation  + observed_at
relation     + source_id, target_id
checkpoint   + checkpoint_id, version, trigger, integrity_hash
```

`content_hash` stays the **event's own hash**, not `digest()`. These are different guarantees that the old code conflated in a comment: `content_hash` is what a receiver compares against `verify()`, `digest()` content-addresses the primitive itself. That distinction is now pinned by a test.

The `bring-your-own-dashboard` guide already documented the typed API (`p.kind, p.sequence, p.content_hash[:12]`, `p.model_dump(mode="json")`) and was describing an interface that did not exist until this change.

## Verification

`ruff check`, `ruff format --check`, `mypy src/continuum` clean; full suite **2301 passed, 27 skipped**. The pre-existing `test_postgres_url_is_routed_to_postgres_backend` failure is unrelated (it expects a `psycopg` `RuntimeError` but with the driver installed locally it makes a real connection attempt instead; fails identically with my changes stashed, green in CI).

New tests pin the issue's actual contract rather than the tautologies they replace — the old suite asserted `"integrity_hash" in p` on a dict, which was always true. Now:
- every exported primitive is an instance of the model it classifies to, and `EvidencePrimitive` is never the concrete type;
- `content()` and `digest()` are reachable and agree with `stable_hash`;
- `content_hash` is the event hash, not the digest (pinning the distinction so a future "simplification" cannot silently redefine the wire format);
- a stray field is rejected, not shipped;
- the models round-trip through the JSON wire format with no key lost, and both the model and `json.loads` paths verify.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved evidence export reliability with structured validation.
  * Preserved evidence export compatibility when serializing and verifying JSON data.
  * Improved handling of relation endpoints and checkpoint metadata.
  * Added safeguards to detect missing or unexpected evidence fields during export.


Cyrax321 (approved):
Author context: Adhi1-2, external contributor. Verified the defect is real in kind: exporting untyped dicts while declaring models leaves the model fields unenforced, and routing construction through the models makes extra=forbid live. The wire format reasoning holds: Checkpoint gains event_id and event_type that the old dicts already emitted, cmd_export_evidence serializes via model_dump(mode=json), and verify_export accepts both models and post-json.loads mappings so both paths run one check. The CodeRabbit signature_inputs thread is not a blocker: verify_export recomputes stable_hash(signature_inputs) against content_hash for event-backed primitives and chains on content_hash, which is the documented receiver check. No em dashes in added lines, all checks green. Approving.

Cyrax321 (dismissed):
Routing `export_evidence` through the declared models closes the drift hole from #1155, and the wiring is faithful to the shape it replaces. I traced each branch against the old dict construction: `_common_fields` is a one-for-one move of the old `base` dict, `Relation.from_event` carries the endpoint extraction verbatim, both checkpoint branches reproduce the old keys and defaults (`version` 0, `trigger` "unknown", `event_id` fallback), and `Checkpoint`'s new `event_id`/`event_type` fields match what the dict always emitted, so the JSON-lines wire format stays byte-stable under `sort_keys=True`.

The sharp edges I probed refute themselves. Payloads are JSON-native by `Event._payload_is_json_native`, so `model_dump(mode="json")` cannot diverge from the old `default=str` output for anything storage hands back, and both backends verify checkpoints on read (`_row_to_checkpoint` raises `CorruptedRecord` on an unsealed record), so `for_checkpoint` can never see a null `integrity_hash` where `content_hash: str` would now reject it. The removed `_cp_by_id` list-plus-map folds into one dict without changing iteration order.

The new tests pin the contracts that actually matter rather than the tautologies they replace: every primitive is its declared subclass, `content_hash` is the event hash and not `digest()`, a stray field raises instead of shipping, and a `json.loads` receiver verifies on the same footing as the in-memory models. Approving.

Cyrax321:
Hi @Adhi1-2,

This PR is approved and all checks are green. The only remaining issue is a **merge conflict** in `CHANGELOG.md` caused by recent updates on `main`.

Please **rebase** onto the latest `main`, resolve the conflict, and push. Once the branch is clean, this is ready to merge.

PR #1295, opened 2026-09-18
test(openai): FakeTC sets the attribute openai-agents 0.22.3 reads (#1294)

### LINKED ISSUE #1294: CI: every Test job fails on openai-agents 0.22.3 — 'FakeTC' object has no attribute '_function_tool_arguments'
## Summary

Every job in the `Test` matrix is red on current PRs, including ones that touch nothing near the adapters. The suite installs the adapter extra with plain `pip install -e ".[dev]"` (`.github/workflows/ci.yml:66`), which resolves `openai-agents>=0.2` to the **newest** PyPI release rather than the `uv.lock` pin. `openai-agents` 0.22.3 (released after the 0.22.1 lock entry) reads a `ToolContext` attribute that the test double in `tests/test_adapters_openai.py` never sets.

## Actual behaviour

```
FAILED tests/test_adapters_openai.py::test_wrap_function_tool_invocation_binds_args_and_intercepts
assert "An error occurred while running the tool. Please try again. Error: 'FakeTC' object has no attribute '_function_tool_arguments'" == {'endpoint': 'https://x', 'method': 'POST'}
```

Confirmed identical on two unrelated PRs (#1289, which touches only `cli/main.py` pinning, and #1290, compaction storage), so it is independent of any PR diff.

## Root cause

`agents/tool.py` in 0.22.3:

```python
async def _on_invoke_tool_impl(ctx, input):
    prepared = ctx._function_tool_arguments
```

`agents/tool_context.py` sets that attribute in `ToolContext.__init__`, but the `FakeTC` double builds its state by hand and overrides `__init__` without calling `super().__init__()`, so the attribute is absent and the invocation raises `AttributeError` — which the SDK catches and turns into its generic error string, so the tool returns an error message where the test expects its result.

## Scope

Only the test double. The adapter itself (`src/continuum/adapters/openai.py`) uses the public `ToolContext` API only — the `ToolContext` type, `tool_name`, `tool_input` — and no internals, so it is unaffected by the change.

## Suggested fix

`FakeTC.__init__` sets `self._function_tool_arguments = None`, exactly what the SDK's own constructor does. The attribute is unused in older releases (verified against 0.22.2: nothing reads or sets it, only an unrelated `parse_function_tool_arguments` function), so the fix is inert below 0.22.3 and required at or above it. Verified passing on both 0.22.2 and 0.22.3.

## Impact

Blocks green CI on every open PR, which is the only signal a fork can produce (review bots cannot approve cross-fork PRs). Any PR that looks red because of this is actually fine.

Adhi1-2:
## What

Every job in the `Test` matrix is red on current PRs — including ones that touch nothing near the adapters:

```
FAILED tests/test_adapters_openai.py::test_wrap_function_tool_invocation_binds_args_and_intercepts
assert "An error occurred while running the tool. Please try again. Error: 'FakeTC' object has no attribute '_function_tool_arguments'" == {'endpoint': 'https://x', 'method': 'POST'}
```

Confirmed identical on two unrelated PRs (#1289 touches only `cli/main.py` pinning; #1290 is compaction storage), so it is independent of any PR diff.

## Root cause

The suite installs the adapter extra with plain `pip install -e ".[dev]"` (`.github/workflows/ci.yml:66`), which resolves `openai-agents>=0.2` to the **newest** PyPI release rather than the `uv.lock` pin (0.22.1). `openai-agents` 0.22.3 reads a `ToolContext` attribute the test double never sets:

```python
# agents/tool.py, 0.22.3
async def _on_invoke_tool_impl(ctx, input):
    prepared = ctx._function_tool_arguments
```

`ToolContext.__init__` sets that attribute, but `FakeTC` builds its state by hand and overrides `__init__` without calling `super().__init__()`. The SDK catches the `AttributeError` and substitutes its generic error string, so the tool returns an error message where the test expects its result.

## The fix

```python
self._function_tool_arguments = None
```

exactly what the SDK's own constructor does. Verified the attribute is **unused in older releases** — nothing in 0.22.2 reads or sets it (only an unrelated `parse_function_tool_arguments` function) — so the change is inert below 0.22.3 and required at or above it. Setting it unconditionally beats version-detection: an extra unused attribute is harmless, and conditional test code is worse.

## Scope

Only the test double. The adapter itself (`src/continuum/adapters/openai.py`) uses the public `ToolContext` API alone — the type, `tool_name`, `tool_input` — no internals, so it is unaffected.

## Verification

Reproduced in an isolated venv with `openai-agents` 0.22.3 (matching what CI resolves), then confirmed the fix passes on **both**:

| `openai-agents` | result |
|---|---|
| 0.22.3 (CI resolves this) | was failing → **1 passed** |
| 0.22.2 (local, pin-adjacent) | **1 passed** (still green, fix inert) |

- Full `tests/test_adapters_openai.py` under 0.22.3: **31 passed, 2 skipped**.
- Full suite under 0.22.2: **2283 passed, 26 skipped, 0 failed** (slow tests excluded locally; they only surface in CI).
- `ruff check`, `ruff format --check` clean; docs-count guard passes.

This unblocks green CI for every open PR, which is the only signal a fork can produce — the review bots cannot approve cross-fork PRs, so a green matrix is as far as the fork goes.

Closes #1294.


Cyrax321:
Thank you for contributing to Continuum. I verified this fixes the same FakeTC failure as #1303. This is superseded by #1303, which covers more for the long term by building FakeTC through ToolContext.__init__ instead of setting a single attribute, so it picks up _function_tool_arguments, _custom_data and any future internal state. Closing this as a duplicate of #1303. Thank you again for contributing and helping unblock CI.

PR #1297, opened 2026-09-18
test(restore): cover every target-resolution and input-validation branch (#1292)

### LINKED ISSUE #1292: recovery: every target-resolution and input-validation branch of approve_restore is untested
## What is untested

`approve_restore` and its helper `_anchor_for` in `src/continuum/recovery/restore.py` resolve which checkpoint a restore rewinds to. The existing tests only ever exercise one way of specifying the anchor:

```python
# tests/test_restore_merge_gate.py, every call
approve_restore(storage, run_id, reason=reason, anchor_sequence=0, carry_forward=carry_forward)
```

So the whole `target=` path — and every validation that guards it — has no test. A coverage run puts `restore.py` at 55%, among the lowest in the repo for a module with no optional dependencies. The untested branches, per `--cov-report=term-missing`, are lines 46-55, 60, 76-78, 93, 114, 116 and 121.

These are not internal-only paths. The CLI passes both straight through:

```python
# src/continuum/cli/main.py, cmd_restore
target = getattr(args, "target", None)
anchor = getattr(args, "anchor", None)
anchor_seq = int(anchor) if anchor is not None else None
result = approve_restore(storage, args.run_id, reason=args.reason or "",
                         target=target, anchor_sequence=anchor_seq, ...)
```

`--target` is an optional positional and `--anchor` is a flag, so a user can supply either or both. Every branch below is reachable from `continuum restore <run> [target]`.

## The six untested branches

Each was driven by hand against a real `SQLiteStorage`; the messages are the current behaviour, not guesses:

| input | result |
|---|---|
| `target` **and** `anchor_sequence` both given | `ValueError: pass exactly one of target or anchor_sequence` |
| `reason="  "` (whitespace) | `ValueError: a restore needs a stated reason; the reason is the audit` |
| `target=None`, run has no checkpoints | resolves to anchor `0` (the empty-history fallback) |
| `target="nope"` (unknown text) | `ValueError: no checkpoint 'nope' for run 'r1'` |
| `target=""` (empty, or whitespace-only) | `ValueError: restore target must be non-empty` |
| `target=999` (unknown int) | `ValueError: no checkpoint with version/source_sequence 999 for run 'r1'` |

The empty-text case is worth pinning explicitly: `_anchor_for` calls `str(target).strip()` before the emptiness check, so both `""` and `"   "` raise rather than falling through to the checkpoint lookup — behaviour that a refactor could easily break. The one rough edge these tests will surface is the unknown-text message, which names neither the run's existing checkpoints nor the accepted target forms, so an operator with a typo gets no help finding the right one.

## Why it matters

`restore` discards the span `(anchor, head]`. Getting the anchor resolution wrong — or silently falling back to anchor `0` when a user mistypes a checkpoint name — rewinds more than intended, and the refusal path is the only thing standing between a typo and data loss. Today every one of these paths is trusted but never exercised, so a refactor of `_anchor_for` could change which checkpoint a mistyped target resolves to and nothing would fail.

The corrupted-checkpoint branch is the one part already covered, by `tests/test_checkpoint_corruption.py::test_restore_resolver_reports_corruption_instead_of_a_missing_checkpoint` (#1059) — that is why it is excluded from the list above.

## Suggested fix

Add parametrised tests to `tests/test_restore_merge_gate.py` (or a new `tests/test_restore_target_resolution.py`) covering the six branches above, asserting on the raised type and message for the five errors and on the resolved anchor for the fallback. The existing suite builds a `SQLiteStorage` on a `tmp_path` fixture, so no new harness is needed. This is a good first issue: the surface is one small pure function, the inputs are plain strings and ints, and the expected behaviour is pinned by the table above.

Adhi1-2:
Closes #1292.

## What

`approve_restore` had one test entry point: `approve_restore(..., anchor_sequence=0)` in `tests/test_restore_merge_gate.py`. Every branch that resolves a **target**, and every input check guarding that resolution, was uncovered — a change to any of them would have shipped green.

`tests/test_restore_target_resolution.py` drives each branch directly and pins the error messages, which were recorded by hand in the issue before the tests were written:

| input | result |
|---|---|
| `target` **and** `anchor_sequence` | `ValueError: pass exactly one of target or anchor_sequence` |
| `reason="  "` | `ValueError: a restore needs a stated reason; the reason is the audit` |
| `target=None`, no checkpoints | anchors at `0` |
| `target=None`, checkpoints present | anchors on the latest checkpoint's `source_sequence` |
| `target=<checkpoint_id>` | resolves to that checkpoint's sequence |
| `target=<version>` / `target=<source_sequence>` | resolves via the numeric ladder |
| `target="7"` (numeric string) | falls through the id rung to the numeric rungs |
| `target="nope"` | `ValueError: no checkpoint 'nope' for run 'r1'` |
| `target=""` / whitespace | `ValueError: restore target must be non-empty` |
| `target=999` | `ValueError: no checkpoint with version/source_sequence 999 for run 'r1'` |
| a checkpoint id from another run | not a valid target for this run |

The "no target" test writes checkpoints out of order on purpose so the highest *version* is not the highest *sequence*: a resolver that returned the version instead of `source_sequence` would report the wrong anchor.

`restore_to_anchor` is exercised too — it is exported in `__all__` but called from neither the CLI nor the MCP server, so nothing else reached it.

## Effect

`src/continuum/recovery/restore.py`: **55% → 90%** coverage, measured against the merge-gate, corruption and new suites together. No source change — `restore.py` is untouched.

Three lines stay uncovered, both for a reason:
- the `CorruptedRecord` rung is already covered by `tests/test_checkpoint_corruption.py:86`;
- the trailing `checkpoint_id` ladder rung (`restore.py:76-78`) is **unreachable**: it searches the run's own checkpoints for a numeric id, but `get_checkpoint` returns any of the run's own ids — and matches on `run_id` — before that loop runs. Worth a look by a maintainer; I left it alone as out of scope for a test-only PR.

## Verification

- `pytest tests/test_restore_target_resolution.py` — 17 passed
- `pytest tests/` — 2278 passed, 24 skipped
- mutation-checked the two input guards: deleting the both-targets check and the blank-target check each fail exactly the tests that claim to cover them, then restored
- ruff check + format clean; markdownlint clean; the docs-count guard passes after re-syncing the documented collected total `2,311 → 2,335` across README.md, the five translated READMEs, CHANGELOG.md, `docs/CONTRIBUTING_ONBOARDING.md`, `references/install.md` and `references/testing.md`

The one known red CI job is pre-existing and unrelated: `tests/test_adapters_openai.py` fails in CI on the `openai 3.16.0` / `openai-agents 0.22.3` resolution (#1286 fixes the pin); it passes locally on this branch.

## Summary by CodeRabbit

- **Tests**
  - Expanded coverage for restore target resolution, input validation, checkpoint selection, and invalid target scenarios, including anchor-based restoration.
  - Documented approximately 2,483 collected tests, with approximately 2,454 passing where reported.

- **Documentation**
  - Updated test-suite totals across localized README files, contributor onboarding materials, installation references, and testing guides.
  - Added an unreleased changelog entry describing the expanded restore-related test coverage.


Cyrax321:

Adhi1-2:
The merge conflicts you flagged are resolved, and the branch is now fully current with `main`.

`main` moved 11 commits since the last refresh, so I merged `upstream/main` (`41f1e71`..`13c2fab`) into this branch. It auto-merged cleanly -- the only file in common was `CHANGELOG.md`, and git resolved it without a conflict. Head is now `fa8773e`.

Verified locally: `test_restore_target_resolution.py` 17 passed, `test_docs_counts.py` 5 passed, `test_cli.py` + `test_action_ledger.py` green (232 passed together); `ruff check`/`ruff format --check` over `src/ tests/ examples/`, `mypy src/continuum`, and the four markdownlint files all clean.

Cyrax321 (changes requested):
Author context: Adhi1-2, external contributor. The test coverage itself is the right shape: branch-direct tests for the _anchor_for ladder and input guards, with error messages recorded from observed behavior, and no production change. Two blockers. 1. The docs-count figures this branch pins (2,415) are stale against current main (2,452 in README.md on this checkout), so the guarded counts this PR updates will fail its own guard on merge. Re-measure on a rebased tree. 2. The new test file tests/test_restore_target_resolution.py does not exist on current main, so I could not run it locally; CI is green on the branch, but the rebase in (1) must keep the file and its assertions intact. Rebase, resync the counts, and this is approvable.

tests/test_restore_target_resolution.py
```diff
@@ -0,0 +1,202 @@
...
+
+    storage = _storage()
+    derivation, carry, summary = restore_to_anchor(storage, RUN_ID, 0, reason="rollback")
+    # An empty prefix crosses nothing outstanding; safety is the caller's call.
+    assert derivation.unsettled_authorizations == frozenset()
+    assert derivation.depended_results == frozenset()
+    assert derivation.uncertain_slots == frozenset()
+    # ``summary`` is the JSON-native projection of ``derivation``: the same
+    # sets, serialised as lists for the event payload.
+    assert summary == derivation.model_dump(mode="json")
```
Cyrax321, line 159: Non-blocking. This equality only holds here because all three derivation sets are empty: `summary_payload` in gate.py emits lists of item dicts sorted by sequence, while `model_dump(mode="json")` serializes the frozensets in hash order, which varies with PYTHONHASHSEED between processes. The assert is fine for this smoke case, but if a non-empty derivation is ever added the same expression would flake; either sorting both sides before comparing or narrowing the comment's claim to the empty case would keep it honest.

tests/test_restore_target_resolution.py
```diff
@@ -0,0 +1,202 @@
...
+# --- the exported precondition helper ------------------------------------- #
+
+
+def test_restore_to_anchor_checks_preconditions_for_a_restore() -> None:
+    """``restore_to_anchor`` is the public precondition-only entry point.
+
+    It is exported in ``__all__`` but called from neither the CLI nor the MCP
+    server, so nothing else exercised it.
+    """
+    from continuum.recovery.restore import restore_to_anchor
```
Cyrax321, line 149: Non-blocking. This import sits inside the test body while the module already imports from `continuum.recovery.restore` at the top. Moving it up keeps the file's dependencies visible at a glance and consistent with the rest of the suite, which matters when later readers scan imports to see which surfaces a test touches.

CHANGELOG.md
```diff
@@ -290,6 +290,27 @@ All notable changes to this project are documented here. The format follows
   provenance labels, quarantine, omission, determinism and the size caps;
   recovery verdicts and safety semantics are unchanged.
 
+### Tests
```
Cyrax321: Non-blocking. `### Tests` is a new subsection type for this changelog; the existing vocabulary is the Keep a Changelog set (Added/Changed/Fixed/Removed) plus the custom `Added, Phase N` forms. Not a defect, just a consistency question worth settling now: should test-only entries accumulate under their own heading, or fold under Changed, so the Unreleased section stays greppable?

Cyrax321 (commented):
This adds a focused suite that drives every target-resolution and input-validation branch of `approve_restore` directly, plus coverage of the exported-but-uncalled `restore_to_anchor`, with the test-count re-sync across the READMEs and references. The pins are non-vacuous: the out-of-order checkpoint writes in the no-target test genuinely catch a resolver that returned version instead of source_sequence, and I ran the new file against `src` on main where all 17 pass, matching the branch-by-branch mapping to restore.py. The claim that the trailing checkpoint_id ladder rung (restore.py:76-78) is unreachable also checks out: real ids come from `make_id("checkpoint")` and are never int-parseable, and `get_checkpoint` resolves any existing id before that loop runs. Nothing here blocks merge; three small non-blocking notes inline.

Adhi1-2:
Re: the stale counts flagged in the last review — the branch pinned 2,415 while README.md on the review checkout said 2,452. Re-measured on the rebased tree: it collects 2,495 now, and all count-bearing docs are synced to that figure. The test file itself is unchanged.

All checks are green.

The rebase dismissed the prior approval, so this needs a fresh review rather than a merge as-is.

Cyrax321 (changes requested):
Hi @Adhi1-2,

This PR currently has **merge conflicts** in multiple documentation files (`README*.md`, `CHANGELOG.md`, etc.) due to recent updates on `main`.

Please **rebase** this branch onto the latest `main` and resolve the conflicts. Once resolved, please ensure that:

1. The **test counts** in all documentation files are re-synced to match the current state of the repository.
2. All CI checks pass, particularly the **documentation count guard**.

After resolving the conflicts and updating the counts, please push the changes so we can proceed with the final review.

PR #1298, opened 2026-09-18
fix(recovery): wire the precondition wrappers in and drop the reason they discarded (#1094)

### LINKED ISSUE #1094: restore_to_anchor and merge_to_anchor are exported with zero callers, and restore_to_anchor silently discards its reason parameter
## Summary

`src/continuum/recovery/restore.py` exports `restore_to_anchor` in its `__all__` (:29), but nothing in the repository calls it — no product code, no script, not even a test. It is a pass-through to `check_preconditions(..., edit_type="restore")` that silently discards its own keyword-only `reason` parameter (declared, never read). The live restore path (`approve_restore`, reached from `cmd_restore`) calls `check_preconditions` directly, and `recovery/__init__.py` re-exports `approve_restore` and `RestorePreconditionError` but deliberately not `restore_to_anchor`.

Its sibling `merge_to_anchor` (`src/continuum/recovery/merge.py:39`) is in the same state minus the test coverage: no caller in `src/`, five references in tests.

Both wrappers date from the #408 precondition-gate work; the gate itself shipped, the convenience wrappers on top of it never got a caller.

## Affected component

Other (recovery API surface)

## Severity

## System information

- Python: 3.13.5

```bash
git grep -w restore_to_anchor
# src/continuum/recovery/restore.py:29:    "restore_to_anchor",   <- the __all__ entry
# src/continuum/recovery/restore.py:80:def restore_to_anchor(   <- the definition

git grep -w merge_to_anchor -- src/
# src/continuum/recovery/merge.py:39:def merge_to_anchor(       <- definition only
```

## What a fix looks like

**Option A: delete both wrappers.** `restore_to_anchor` is ~12 lines with zero test churn; `merge_to_anchor` additionally drops its five test references. The `__all__` entry goes with it.

**Option B: wire them.** If the intent was a one-call "check then approve" convenience, `restore_to_anchor` should call `approve_restore` and actually pass `reason` through — but the live path already does this inline, so removal is the cheaper, honest change.

Adhi1-2:
## What

`restore_to_anchor` (`src/continuum/recovery/restore.py`) was exported in `__all__` with **zero callers** and a keyword-only `reason` parameter it never read. The live restore path (`approve_restore`, reached from `cmd_restore`) spelled `check_preconditions` directly. Its sibling `merge_to_anchor` was in the same state minus the test coverage.

The wrappers are the **read-only half** of their edit — a caller asks "would this be blocked?" against a live database while an agent is mid-run. Two fixes land together because neither half makes sense alone:

- The mutating halves now route through their own wrapper instead of duplicating the call. `approve_restore` → `restore_to_anchor`; `approve_merge`'s target-only branch → `merge_to_anchor`.
- The discarded `reason` parameter is **removed outright**, not retained-but-ignored. A precondition check appends no event to carry a reason on; a caller who passed one was writing an audit string that went nowhere. Silence was the failure mode, so `TypeError` on a stale caller is the correct signal.

`approve_merge`'s source-side branch still calls `check_merge_preconditions` directly: the union form drops the per-side summaries its lineage payload stamps, so delegating through `merge_to_anchor` there would lose them. Documented in the docstring.

This is **Option B** from the issue. Option A (delete both) was rejected because `merge_to_anchor` already has five test references and a legitimate read-only contract against a live DB — deletion would remove a useful API rather than fix the one real defect, the discarded `reason`.

## Falsifiable

Four tests in `tests/test_restore_merge_gate.py` pin the wiring, over a span seeded with a real unsettled authorization so `carry_forward` through the approve path is load-bearing:

- `test_precondition_check_appends_nothing` — the check writes nothing (the live-DB-read promise).
- `test_precondition_check_takes_no_reason` — `reason=` raises `TypeError`.
- `test_approve_restore_stamps_exactly_what_the_check_returned` / `test_approve_merge_stamps_exactly_what_the_check_returned` — the lineage payload stamps exactly the summary the read-only API returned.

Verified by tamper: dropping `carry_forward` on the approve side fails both stamp tests with `EditPreconditionError`. Two earlier falsification attempts were diagnosed as correct-by-design rather than shipped: an off-by-one anchor does not move the summary (the blocking event is always `head`, and the span is `(anchor, head]`), and a changed `edit_type` inside the wrapper passes because the stamp test pins *agreement* between the read-only API and the mutating path, both of which call the same function.

## Verification

- `ruff check src/ tests/` — pass
- `ruff format --check src/ tests/` — pass
- `mypy src/continuum` — pass (127 files)
- `pytest -q` — one failure, `test_postgres_url_is_routed_to_postgres_backend`, **pre-existing and environmental**: psycopg 3.3.5 is installed locally, so the route succeeds rather than raising `RuntimeError`. Reproduces with these changes stashed on plain `upstream/main`.
- `markdownlint-cli2 CHANGELOG.md` — clean

CHANGELOG entry added under `### Changed`.

Closes #1094.


CHANGELOG.md
```diff
@@ -8,6 +8,21 @@ All notable changes to this project are documented here. The format follows
 
 ### Changed
 
+- **The precondition check wrappers are wired to the approve paths they were written for, and no longer take a `reason` they discarded (#1094).**
+  `restore_to_anchor` (`src/continuum/recovery/restore.py`) was exported in
+  `__all__` with zero callers and a keyword-only `reason` parameter it never
+  read, while the live restore path (`approve_restore`) spelled
+  `check_preconditions` directly. Its sibling `merge_to_anchor` was in the
+  same state minus the test coverage. Both wrappers are the read-only half of
+  their edit — a caller asks "would this be blocked?" against a live database
```
Cyrax321: U+2014 em dash on this line, which AGENTS.md bars (no em dashes anywhere in commit messages, code comments, or docs). Replace with a colon or comma, e.g. "the read-only half of their edit: a caller asks ..."

CHANGELOG.md
```diff
@@ -8,6 +8,21 @@ All notable changes to this project are documented here. The format follows
...
+  while an agent is mid-run — so the mutating half now routes through them
```
Cyrax321: Second U+2014 on this line, same AGENTS.md rule. Replace with a period: "...while an agent is mid-run, so the mutating half now routes through them" works as a comma join.

Cyrax321 (changes requested):

Ref: #1298 / Issue #1094
Surface: src/continuum/recovery/restore.py and merge.py precondition wrappers

The change itself is right and I verified it: the read-only wrappers had zero callers while the live path spelled check_preconditions directly, the mutating halves now route through their wrappers, and removing the never-read reason parameter outright (rather than retaining it silently) is the correct call given a precondition check appends no event to carry it. The tests pin that the check appends nothing and takes no reason.

Proof: ran tests/test_restore_merge_gate.py and tests/test_merge_gate_source_494.py at head 45de2fcb: 29 passed. CI green.

One blocker, convention only: two U+2014 em dashes in the new CHANGELOG.md lines (inline comments posted), which AGENTS.md bars. Fix the two lines and this merges; nothing else changes.

Adhi1-2:
The blocker you flagged is fixed in `be0901e`: the two U+2014 em dashes in the #1094 CHANGELOG entry are replaced with a colon and a comma (AGENTS.md). Nothing else in the diff changed.

Cyrax321 (changes requested):
Author context: Adhi1-2, external contributor. Follow-up to the prior review. The product change is still correct as previously verified, but the two flagged em dashes are still present in the added CHANGELOG lines, so the blocker stands. Replace both U+2014 with commas or parentheses per AGENTS.md. Nothing else needs to change.

tests/test_restore_merge_gate.py
```diff
@@ -369,3 +369,123 @@ def test_falsifiable_restore_skipping_unsettled_claim_refuses_like_fork() -> Non
         assert any(e.type is EventType.RUN_MERGED for e in storage.read_events("run_1"))
     finally:
         storage.close()
+
+
+# --- the check-only halves are a live-DB read, not a duplicate (#1094) -------- #
+
+
+@pytest.mark.parametrize("check", ["restore", "merge"])
```
Cyrax321, line 377: Blocking: adding these tests pushes the branch past the docs-count guard. tests/test_docs_counts.py::test_documented_count_matches_suite fails on every platform job (macos/ubuntu/windows, 3.11 through 3.13) with 'suite collects 2483 tests but docs say ~2452: re-sync README.md, the translated READMEs, CHANGELOG.md, docs/CONTRIBUTING_ONBOARDING.md, references/testing.md, and references/install.md'. Main has merged more tests since this branch was cut, so the fix is to update the branch and re-sync the counts in all six locations; a green local pytest -q cannot stand in for this because the guard compares the full CI collection against the documented numbers.

tests/test_restore_merge_gate.py
```diff
@@ -369,3 +369,123 @@ def test_falsifiable_restore_skipping_unsettled_claim_refuses_like_fork() -> Non
...
+    storage = _make_storage()
+    try:
+        fn = restore_to_anchor if check == "restore" else merge_to_anchor
+        with pytest.raises(TypeError):
+            fn(storage, "run_1", 0, reason="goes nowhere")  # type: ignore[call-arg]
+    finally:
+        storage.close()
+
+
+def test_approve_restore_stamps_exactly_what_the_check_returned() -> None:
```
Cyrax321, line 415: Non-blocking: this test and test_approve_merge_stamps_exactly_what_the_check_returned below copy the same ~40-line scenario, differing only in the approve function, the stamped event type, and which read-only wrapper is queried. This file already parametrizes restore and merge for the two checks above, so parametrizing this pair the same way would keep the two scenarios from silently drifting apart as the gate evolves.

tests/test_restore_merge_gate.py
```diff
@@ -369,3 +369,123 @@ def test_falsifiable_restore_skipping_unsettled_claim_refuses_like_fork() -> Non
...
+
+# --- the check-only halves are a live-DB read, not a duplicate (#1094) -------- #
+
+
+@pytest.mark.parametrize("check", ["restore", "merge"])
+def test_precondition_check_appends_nothing(check: str) -> None:
+    """restore_to_anchor / merge_to_anchor exist to be safe against a live
+    database while an agent is mid-run. That promise is void if the check
+    itself writes."""
+    from continuum.recovery.merge import merge_to_anchor
```
Cyrax321, line 382: Non-blocking: everything else this module uses is imported at the top, so merge_to_anchor and restore_to_anchor could sit beside the existing approve_merge/approve_restore imports instead of repeating function-local imports in each of the four new tests.

Cyrax321 (changes requested):
The wiring itself is correct: approve_restore and the target-only branch of approve_merge now route their precondition check through restore_to_anchor/merge_to_anchor with the identical arguments the direct check_preconditions call used, and the read-only wrappers lose the keyword-only reason they never read, with every call site in the tree updated. What blocks merge is CI, not the code: tests/test_docs_counts.py fails in all nine platform jobs because the suite collects 2483 tests while the docs on this branch say ~2452 (drift 31, tolerance 30), so the branch needs an update from main and a re-sync of the documented counts. Two small non-blocking notes inline.

Adhi1-2:
Both em dashes in the #1094 CHANGELOG entry are replaced, and the branch is rebased onto the current main (d2f110f).

Main moved past the base this branched from, so the first CI run failed the docs-count guard on the merged tree: it collects 2,484 tests while the docs still said 2,452, just past the tolerance of 30. I re-measured on the rebased tree and resynced the figure across the READMEs, CHANGELOG, `docs/CONTRIBUTING_ONBOARDING.md`, `references/install.md`, and `references/testing.md`, so the guard is satisfied on the tree CI actually runs.

All 31 checks green including the three Windows jobs, codecov/patch, and anya-review. The precondition wiring and the new `tests/test_restore_merge_gate.py` are unchanged by the rebase. The rebase dismissed the prior approval, so this needs a fresh review rather than a merge as-is.

Cyrax321 (changes requested):
Hi @Adhi1-2,

This PR currently has **merge conflicts** in multiple documentation files (`README*.md`, `CHANGELOG.md`, etc.) and `tests/test_restore_merge_gate.py` due to recent updates on `main`.

Please **rebase** this branch onto the latest `main` and resolve the conflicts. Once resolved, please ensure that:

1. The **test counts** in all documentation files are re-synced to match the current state of the repository.
2. All CI checks pass, particularly the **documentation count guard**.

After resolving the conflicts and updating the counts, please push the changes so we can proceed with the final review.

PR #1300, opened 2026-09-18
test(exports): walk the package for the __all__ guard instead of five modules (#1228)

Adhi1-2:
## What

`tests/test_module_all_exports.py` asserted that every name in `__all__` resolves — by importing **five** modules by name. The issue (#1228) points out that the package declares `__all__` on far more modules than that, so a rename that forgot `__all__` stayed green everywhere it mattered.

The guard now derives its scope from the package itself instead of a hand-rolled list:

- `test_all_symbols_exist_on_modules` enumerates with `pkgutil.walk_packages` and checks **every** module that declares an `__all__` — 122 today, against a floor of 90 so a walk that silently shrank to nothing *fails* rather than passing vacuously.
- `test_star_import_execution` covers the same set instead of 29 hand-listed modules, catching a module whose import has a side effect, or a name that shadows an earlier import.

## The second check, and why it is curated

I also added `__all__ == the public names a module defines itself`, checked by AST. This cannot hold package-wide, so it runs on a curated list of the 29 leaf modules the file already imported:

- **Aggregator modules re-export.** `continuum.actions` lists 15 entries, all defined in submodules. `continuum.cli` does the same.
- **Some modules hold a name back on purpose.** `continuum.budgets` keeps `FALLBACK_MAX_ATTEMPTS` out of `__all__` because it is private to its own defaulting.

Both are per-module policy, not defects. Pinning only modules that define their whole surface keeps the assertion meaningful where it applies. Modules whose import fails on a missing optional extra are skipped rather than erroring the guard; today all 122 import cleanly, so that path is defensive.

## Falsified

Two tamper checks, both of which fail without the fix and pass with it:

1. **Phantom name on an uncovered module** — added `"definitely_not_a_real_symbol"` to `src/continuum/recovery/restore.py`'s `__all__`, a module the old five-item list never covered. Fails: `continuum.recovery.restore missing attribute 'definitely_not_a_real_symbol' listed in __all__`.
2. **Shrunken walk** — restricted `_modules_with_all()` to `continuum.gate` only. Fails: `assert 1 >= 90`.

## Gates

- `pytest -q` — all pass except one **pre-existing, environmental** failure: `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend` fails `DID NOT RAISE RuntimeError` because `psycopg` 3.3.5 is installed locally. Confirmed pre-existing by stashing all changes and re-running on plain `upstream/main` — same failure, unrelated to this PR.

The docs-count guard passes: the small test addition is inside its TOLERANCE, so no doc counts needed re-syncing.

CHANGELOG entry added under `### Changed`.

## Summary by CodeRabbit

* **Tests**
  * Expanded validation of package exports across all discoverable modules.
  * Added checks to ensure star imports execute successfully and exported names remain accurate.
  * Added safeguards against silently reduced module coverage.
  * Added targeted verification for modules whose public interfaces are defined locally.

* **Documentation**
  * Updated the changelog to record the broader export-surface test coverage.


Cyrax321:

Cyrax321 (approved):

Ref: #1300 / Issue #1228
Surface: tests/test_module_all_exports.py, package-wide __all__ guard

The gap was real: the guard imported five modules by name while the package declares __all__ far more widely, so an unexported rename stayed green. The rewrite derives scope from pkgutil.walk_packages instead of a hand list, checks every module declaring __all__ (122 today), and the floor of 90 is the detail that makes this trustworthy: a walk that silently shrinks fails rather than passing vacuously. Star-import execution covers the same set.

Proof:
- Ran tests/test_module_all_exports.py at head 7eff7110: 17 passed.
- Scanned the added lines for U+2014 (the convention blocker I raised on #1299): none present.

PR #1301, opened 2026-09-18
fix(state): the fold books 11 recorded-fact event types as not understood (#1169)

Adhi1-2:
## What

`_dispatch` in `src/continuum/state/semantic.py` has an explicit `match` case for every event type that changes state, and `_NON_PROJECTING` names every type that is a recorded fact. The two hand-written lists did not cover the `EventType` enum between them: **11 of the 51 members were in neither**, so they fell through to `case _: return False` and were booked into `report.ignored_types`.

`ignored_types` is the field `ProjectionReport.complete` defines as *"the fold understood every event type it consumed"*. So any run that restored, merged, settled a review, consumed an authority or sent a webhook reported as **partly unprojectable on events the codebase legitimately emits**.

Reproduced on `upstream/main`, exactly as the issue describes:

```
6 2 {REVIEW_CONFIRMED: 1, RUN_RESTORED: 1, AUTHORITY_CONSUMED: 1, NOTIFICATION_SENT: 1} False
```

After the fix:

```
6 2 {} True
```

Note `applied` stays **2** — the fold's behaviour is unchanged. None of the 11 have an `_Accumulator` handler, so none were ever applied; they are recorded facts, correctly skipped. Only the bookkeeping was wrong.

## Fix

All 11 move into `_NON_PROJECTING`, grouped with the reason each is a recorded fact, matching the set's existing commented style (`# authority (issue #269)`, `# liveness (issue #302)`). I checked each is genuinely non-projecting first — none has a handler in `_Accumulator`, and each is emitted by real code (4–8 sites for the review/restore/merge types).

This also repairs a latent assertion: `tests/test_compaction.py:346` asserts `not report.ignored_types`, so a compaction fixture touching any of these 11 types would have seen that assertion erode.

## Test

Two additions to `tests/test_projection_edges.py`:

1. **`test_every_event_type_is_projected_or_recorded`** — parses `semantic.py` and fails on any `EventType` member in neither the `match` cases nor `_NON_PROJECTING`. This is the family-level guard: a type added to the enum without a home in either list fails here instead of waiting for a fixture to brush against it.
2. **`test_recorded_facts_are_not_reported_as_unprojectable`** — folds a stream containing all 11 and asserts `ignored_types == {}`, `complete is True`, `applied == 1`.

The neighbouring `test_unknown_event_types_are_counted_not_fatal` still holds: `QUANTUM_ENTANGLED` is genuinely not in the enum, so it still lands in `ignored_types` with `complete=False`.

## Falsified

Reverting the source change and keeping the tests, both fail:

```
FAILED tests/test_projection_edges.py::test_every_event_type_is_projected_or_recorded
FAILED tests/test_projection_edges.py::test_recorded_facts_are_not_reported_as_unprojectable
    assert {RUN_RESTORED: 1, ..., AUTHORITY_CONSUMED: 1, ...} == {}
    Left contains 11 more items
```

Restored, both pass.

## Gates

- `pytest -q` — all pass except one **pre-existing, environmental** failure: `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend` (`DID NOT RAISE RuntimeError`) because `psycopg` 3.3.5 is installed locally. Confirmed pre-existing on plain `upstream/main`.

CHANGELOG entry added under `### Fixed`.


Cyrax321:

Adhi1-2:
Closing as a duplicate: issue #1169 was closed by merged PR #1200 "fix(state): account for the eleven event types the fold silently ignored (#1169)" by @abyyxhek, which lands the same change this branch carries.

Verified `gh issue view 1169` is CLOSED and `gh pr view 1200` is MERGED, so this branch has no remaining value and is just queue noise. Deleting the branch to keep the fork tidy.

PR #1302, opened 2026-09-18
fix(cli): write the attest-keygen private key 0600 instead of umask default (#1056)

Adhi1-2:
## What

`continuum attest-keygen --out mykey.pem` writes the **unencrypted** Ed25519 private key with the default file mode, which is **0644** under a standard umask — world-readable on disk — while the command's own output tells the operator to keep it secret. Any local user or backup copy reads raw key material and can produce validly-signed attestations for a tampered chain. The public key is written the same way, which is correct for a public key but makes the two indistinguishable by permission.

Reproduced on `upstream/main` under umask 022:

```
-rw-r--r-- mykey.pem
-rw-r--r-- mykey.pem.pub
$ head -1 mykey.pem
-----BEGIN PRIVATE KEY-----
```

## Fix

`generate_keypair` returns PEM text and leaves custody to the caller, which is the right split, so the missing step is at the write. The private key is now created through `os.open` with mode `0o600` and narrowed by `chmod` afterwards — `os.open`'s mode only applies at **creation**, so a pre-existing 0644 file being overwritten would otherwise keep its wider mode. The public key stays world-readable.

The applied mode is now reported in both the human output and the JSON payload (`private_key_mode`), so an operator on a surprising filesystem can see what they actually got rather than infer it.

A filesystem without permission bits keeps whatever it can express (`contextlib.suppress(OSError)`), the same compromise `save_budgets` already makes for the budget registry — a private key landing **weaker** than a budget file was an inconsistency rather than a convention.

After the fix:

```
Wrote private key mykey.pem (mode 0o600) and public key mykey.pem.pub. Keep the private key secret.
private: 0o600
public : 0o644
```

And the overwrite path, which is the case `os.open` alone cannot fix:

```
$ chmod 644 mykey.pem && continuum attest-keygen --out mykey.pem
after overwrite: 0o600
```

## Test

`tests/test_cli.py` gains four cases: the fresh-file mode, the reported mode appearing in both the human output and JSON, the public key staying readable, and the pre-existing-0644 narrowing path. `tests/test_attestation.py` exercises the signing path but never the on-disk mode, which is why this went uncaught.

## Gates

- `pytest -q` — all pass except one **pre-existing, environmental** failure: `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend` (`DID NOT RAISE RuntimeError`) because `psycopg` 3.3.5 is installed locally. Confirmed pre-existing on plain `upstream/main`.

CHANGELOG entry added under `### Fixed`.

> Note: an older branch `fix/attest-keygen-mode-1056` carrying this same fix was left on a 30-commit-stale base with no PR ever opened against it. This is the same work rebased onto current `upstream/main`, plus the CHANGELOG entry that the older commit lacked.


Adhi1-2:
Closing as a duplicate of #1106, which carries the same fix and has since addressed the review (the POSIX-only guard on the two permission tests) with all CI green — including the three windows-latest jobs that were the original blocker.

I opened this by mistake: an older local branch with this fix had gone stale and a `gh pr list --head` check returned empty for the existing PR, so I rebased the work onto current `upstream/main` rather than advancing the branch #1106 already had.

#1106 is the PR to review. The one thing this branch added that #1106 lacks is the CHANGELOG entry, which I'll push to that branch instead.

Issue #1304, opened 2026-09-19
benchmark: controlled_failures declares expected="RETRY" for two scenarios, but RecoveryMode has no RETRY


abyyxhek:
## Summary

`src/continuum/benchmark/controlled_failures.py` is the ground-truth table for the controlled fault suite: 11 rows, each declaring the `RecoveryMode` an assessment must propose. Two of them name a mode the enum does not define.

`tool_failure` (`controlled_failures.py:58`) and `api_timeout` (`:65`) declare `expected="RETRY"`. `RecoveryMode` (`src/continuum/models.py:122-131`) has seven members — `RESUME`, `REPAIR_AND_RESUME`, `ROLLBACK`, `WAIT`, `REQUEST_HUMAN`, `REPLAN`, `ABORT`. There is no `RETRY`, so those two rows can never match a real `RecoveryDecision.mode`.

## Affected component

Benchmarks

## Severity

Minor inconvenience or cosmetic issue — nothing grades this table today, so the wrong values affect no live output.

## System information

- OS: Windows 11 10.0.26200
- Python: 3.11.15
- continuum-agent: from source, commit `d9a95f2` on `upstream/main`

## Command or code to reproduce

```python
from continuum.models import RecoveryMode
from continuum.benchmark.controlled_failures import SCENARIOS

valid = {m.name for m in RecoveryMode}
print(sorted(valid))
# ['ABORT', 'REPAIR_AND_RESUME', 'REPLAN', 'REQUEST_HUMAN', 'RESUME', 'ROLLBACK', 'WAIT']

print([s.scenario for s in SCENARIOS if s.expected not in valid])
# ['tool_failure', 'api_timeout']
```

## Current behaviour

2 of the 11 rows declare a mode that is not a `RecoveryMode` member, so they are ungradeable.

## Expected behaviour

Every declared ground truth is a real `RecoveryMode`, so the table can be compared against the recovery engine.

## Why this went unnoticed

`ControlledScenario.expected` has exactly one consumer: `tests/test_controlled_failures.py:29-37`, whose allowed-values set is a literal that happens to include `"RETRY"`:

```python
assert scenario.expected in {
    "RESUME", "RETRY", "REPAIR_AND_RESUME", "REQUEST_HUMAN", "REPLAN",
}
```

The test validates the table against itself rather than against the enum, which is how an impossible value passes.

[...]

PR #1306, opened 2026-09-19
fix(bench): controlled-failure ground truth declares modes the engine can propose (#1304)

abyyxhek:
## What's broken

`controlled_failures.py` is the controlled fault suite's ground-truth table, but two of its eleven rows name a mode the enum does not define:

```python
from continuum.benchmark.controlled_failures import SCENARIOS
valid = {m.name for m in RecoveryMode}
print([s.scenario for s in SCENARIOS if s.expected not in valid])
# ['tool_failure', 'api_timeout']
```

`RecoveryMode` has `RESUME`, `REPAIR_AND_RESUME`, `REPLAN`, `WAIT`, `REQUEST_HUMAN`, `ROLLBACK`, `ABORT`. There is no `RETRY`, so those two rows can never match a real `RecoveryDecision.mode` — the moment the table is wired to the engine they fail in a way that reads like a recovery regression but is a typo in the ground truth.

## Why nothing caught it

`ControlledScenario.expected` has exactly one consumer — `tests/test_controlled_failures.py` — and it validated the table against a literal that happened to include `"RETRY"`, i.e. against itself.

## The fix

Each row now names the mode its own description implies, since the engine has no retry concept — the agent retries, the framework only decides whether the run may continue:

- `tool_failure` — "retry is safe" → `RESUME`; a state still verified against the environment is exactly the `validation.safe and not uncertain` arm.
- `api_timeout` — "side effect uncertain" → `REQUEST_HUMAN`, which is what the uncertain-side-effect arm of `_decide` proposes (or `WAIT` in lenient mode).

The test now asserts against the `RecoveryMode` enum and resolves each declared name, so a future impossible value fails instead of being blessed.

Note `adapters/registry.py` also appears in the sibling issue but already carries a complete `__all__` on main.

## Verification

- `tests/test_controlled_failures.py` — 3 passed.
- Full gates green: `ruff check`, `ruff format --check` (both the venv's ruff and CI's pinned 0.16.4), `mypy src/continuum`.

Main had already let its documented suite total drift past the count guard's own tolerance (2393 collected vs 2347 documented), so this branch re-syncs the figure to 2,393 across README.md, the five translated READMEs, CHANGELOG.md, docs/CONTRIBUTING_ONBOARDING.md, references/install.md and references/testing.md — without it the branch inherits a red check it did not cause.

## Summary by CodeRabbit

* **New Features**
  * Added terminal-run protection to prevent resuming completed, failed, aborted, or crashed runs.
  * Added optional terminal animation effects, including spinners, typewriter text, cursor blinking, and shimmer effects.

* **Bug Fixes**
  * Retry and settlement budget usage now remains consistent despite changing request metadata.
  * Webhook deduplication now remains accurate after event-history compaction.
  * Updated controlled-failure recovery expectations for tool failures and API timeouts.

* **Documentation**
  * Updated test totals and recovery scenario counts across project documentation.
  * Documented budget authorization and ledger-based bucket behavior.


Cyrax321 (dismissed):

Cyrax321:
The PR looks good and CI is passing, but there are merge conflicts with main. Please rebase your branch on main or resolve the conflicts. Once resolved, the auto-merge I enabled here will merge this automatically. Thanks!

Cyrax321 (dismissed):
Re-approving after conflict resolution. The merge conflicts were stale test-count figures in CHANGELOG/READMEs/references; resolved toward main, which carries the current counts. The substantive change is untouched: the controlled-failure ground truth now declares modes the recovery engine can actually propose (RETRY is not a RecoveryMode member), and the test asserts against the enum and indexes it so an impossible value fails loudly. Verified: test_controlled_failures.py green, ruff check and ruff format clean, mypy clean over 128 files. The codecov/patch miss is the module-level SCENARIOS tuple, which is import-time data, not a branch.

Cyrax321 (approved):

PR #1307, opened 2026-09-19
fix(core): the last __all__ omissions the package re-exports (#1093)

abyyxhek:
## What's broken

Four modules still define public names that their own `__all__` omits while other modules re-export them, so `from module import *` silently drops them and the completeness guard in `test_module_all_exports.py` never covered them:

| Module | Omitted |
|:--|:--|
| `actions/reconciliation.py` | `unresolved_actions` |
| `environment/snapshot.py` | `process_fingerprint` |
| `checkpoint/policy.py` | `PolicyContext`, `ContextPressurePolicy` |
| `models.py` | `Frozen`, `Origin`, `Provenance`, `TrajectoryReport`, `validate_caused_by`, `PROJECTION_BOOKKEEPING` |

Each is locally defined in the module that omits it and re-exported one level up (`actions/__init__.py`, `environment/__init__.py`, `checkpoint/__init__.py`, and the top-level `continuum/__init__.py` for `Origin` and `Provenance`), which is how the names stayed reachable while being unlisted.

`adapters/registry.py` was on the list too; it already carries a complete `__all__` on main, so only these four remain.

## The fix

Adds the nine names to the four `__all__` lists — one-line-list edits, no behaviour change — and brings the four modules under `test_all_symbols_exist_on_modules`, which pins `__all__` against the module's own AST-level public definitions. That is the part that stops the omissions coming back: the generic check already compares `set(mod.__all__)` against every locally-defined public name, it just wasn't listing these modules. The star-import smoke test imports them too, and there is one assertion per new name.

## Verification

- `tests/test_module_all_exports.py` — 16 passed.

Main had already let its documented suite total drift past the count guard's own tolerance (2393 collected vs 2347 documented), so this branch re-syncs the figure to 2,397 across README.md, the five translated READMEs, CHANGELOG.md, docs/CONTRIBUTING_ONBOARDING.md, references/install.md and references/testing.md — without it the branch inherits a red check it did not cause.


Cyrax321 (approved):

Issue #553, opened 2026-08-31
[#288] N-hop staleness propagation through the causal graph


Cyrax321:
### Affected component

Core (state, checkpoints, recovery)

Parent epic: #288. See execution board #550 for wave context and dependency chain.

### What problem does this solve?

StateValidator propagates staleness only via evidence -> finding -> decision. It does not walk caused_by edges, so invalidating evidence does not mark downstream decisions/actions stale through the provenance graph.

### Proposed solution

Extend StateValidator / RecoveryEngine:
* when evidence is invalidated, BFS over caused_by DAG to mark every downstream node stale
* mark as STALE/CONFLICTED with reason `evidence <id> stale -> decision <id> via caused_by`
* surface in RecoveryContract.invalidated + evidence list (Phase 1 fields)
* preserve: out-of-scope nodes not in graph remain unchanged; cycles raise diagnostic CONFLICTED

Depends on: [#288] DAG projector.

Files: `src/continuum/state/validator.py`, `src/continuum/recovery/engine.py`, `tests/test_provenance_staleness.py` (N=3 hop chain test).

### Alternatives considered

Propagate only one hop. Rejected: field has 6-hop survey requirement.

### Is this a breaking change?

No

### Would you like to work on this?

Yes, I have a plan

PR #1308, opened 2026-09-19
fix(recovery): make assess fully archive-aware (#1050)

abyyxhek:
## What's broken

`RecoveryEngine.assess` is half archive-aware. The confirmation scan, the provenance scan and the validator all fold the archived prefix through a shared `read_all_events` fetch, but three signals computed in the same method still read the live tail only: the liveness breach count, the `RISK_OBSERVED` scan, and the consumed-authorities scan.

`continuum compact` archives the pre-anchor prefix, so all three disappear from the sealed contract the moment a run is compacted. The engine still returns a verdict — just one computed over a deliberately truncated history, and the downgrade is always toward less caution:

```
    contract.triggering_risks: ['event_6c4710e5...']
    contract.liveness: {'breaches': 1, ...}
    rationale: risk meltdown triggers rollback
    contract.triggering_risks: []
    contract.liveness: {'breaches': 0, ...}
```

The contract is the auditable artifact a human or downstream gate acts on, so compaction silently removes the evidence that justified the cautious verdict. Isolating the risk signal with `meltdown`, which policy maps to `ROLLBACK`, makes the downgrade unambiguous: `rollback` before, `request_human` after.

## Root cause

The risk scan and the authority block both landed after #553 established the archive-aware convention and both ignored it; the liveness count came with the same shape. The shared fetch was already sitting right there.

## The fix

All three scans now filter the `archive_aware_events` fetch the method already takes, instead of re-reading the live tail three more times:

- the breach count filters it for `LIVENESS_SILENCE_DETECTED`,
- the risk scan filters it for `RISK_OBSERVED`,
- the authority block passes it to `collect_consumed_authorities`.

The `try/except` fallbacks already in place are untouched, including the authority block's fail-closed degradation.

## Tests

Three regression tests, one per signal, mirroring the existing `test_confirmation_survives_compaction`: each isolates one archived signal, asserts the contract still carries it after a `compact_run`, and asserts the precondition that the event really did leave the live tail. All three fail on unmodified main and pass here:

```
FAILED test_risk_trigger_survives_compaction
FAILED test_liveness_breach_count_survives_compaction
FAILED test_consumed_authority_survives_compaction
```

## Verification

- Full suite: 2355 passed, 41 skipped.

Main had already let its documented suite total drift past the count guard's own tolerance (2393 collected vs 2347 documented), so this branch re-syncs the figure to 2,396 across README.md, the five translated READMEs, CHANGELOG.md, docs/CONTRIBUTING_ONBOARDING.md, references/install.md and references/testing.md — without it the branch inherits a red check it did not cause.


Cyrax321 (dismissed):

Cyrax321:
The PR looks good and CI is passing, but there are merge conflicts with main. Please rebase your branch on main or resolve the conflicts. Once resolved, the auto-merge I enabled here will merge this automatically. Thanks!

Cyrax321 (dismissed):

Cyrax321 (approved):

Issue #1283, opened 2026-09-18
docs: the landing page's meta description states 45 CLI commands and 2,163 tests, and no guard sees either figure


Adhi1-2:
## Summary

`docs/index.html:7` is the site's meta description, the figure set search engines and link previews render:

```html
<meta name="description" content="Framework-agnostic semantic recovery layer for long-running AI agents. 12 MCP tools, 45 CLI commands, 2,163 tests. Deny-by-default, crash-recovery proven.">
```

Two of its three figures are stale:

| Claim in `docs/index.html:7` | Live figure | Source |
|:--|:--|:--|
| 45 CLI commands | **46** | `build_parser()` in `src/continuum/cli/main.py` registers 46 subcommands |
| 2,163 tests | **2,323 collected** | `pytest --collect-only` on `upstream/main` |
| 12 MCP tools | 12 (correct) | 12 `@server.tool` decorators in `src/continuum/mcp/server.py` |

The drift is not in the rendered body. `docs/index.html:232` also says "12 MCP TOOLS", which is correct, and the body's other counts ("29 event types", "7 recovery modes", "6 checkpoint policies") are covered by **#1100** — this issue is only the meta description's CLI-command and test figures, which that issue does not name.

## Why it went unnoticed

`tests/test_docs_counts.py` is the guard that pins documented test figures, and it reads markdown only — `README.md`, the translated READMEs, `CHANGELOG.md`, `docs/CONTRIBUTING_ONBOARDING.md`, and `references/*.md`. `docs/index.html` is HTML, so the guard never opens it. The page's `<meta>` content is invisible to every check.

This is the third time the page's headline has drifted. `fd25b87` wrote "9 MCP tools, 14 CLI commands, 675 tests", **#724** flagged it, and `31747ee` (#927) resynced to "12 MCP tools, 45 CLI commands, 2,163 tests" — the figures that have now gone stale again. Each fix corrected the numbers by hand without anything to catch the next drift, so the next PR moved the test total and the page silently fell behind.

## Affected component

Documentation

## Severity

Minor inconvenience or cosmetic issue.

## System information

- OS: macOS (Darwin 25.0.0)
- Python: 3.13 (`.venv`)
- continuum-agent: from source, at `upstream/main` (`8ef2474`)
- Installed via: `uv pip install -e ".[dev]"`

## Command or code to reproduce

```bash
# CLI subcommands the parser actually builds
python -c "
from continuum.cli.main import build_parser
[...]

PR #1284, opened 2026-09-18
docs(landing): correct the stale meta figures and guard them

Adhi1-2:
Closes #1283.

## The problem

`docs/index.html` is the deployed landing page (`vercel.json` sets `outputDirectory: "docs"`), and it stated stale figures in **two** places:

- the **meta description** — what search engines and link previews render: "12 MCP tools, **45 CLI commands**, **2,163 tests**"
- the **visible metrics card** — what visitors actually read: `2,163+ TESTS PASSING`, `45 CLI COMMANDS`

The parser registers **46** subcommands, the suite collects **2,323** tests, and `server.py` has **12** `@server.tool` decorators, so three of the four figures were stale. This is the **third** drift of this page (fd25b87 → flagged by #724 → resynced by #927 → stale again). Each prior fix was a hand edit, because the docs-count guard scans **markdown only** — `docs/index.html` is HTML, so the guard never opened it.

Correcting only the meta would have left the page contradicting itself: the card would still have shown 2,163 tests and 45 commands to every visitor.

## The fix

Two parts, so the drift is fixed and the next one cannot ship unnoticed.

1. **Corrects both sites** — meta and metrics card — to 46 commands / 2,323 tests. The tool count (12) was already right and is untouched.

2. **Extends the guard** (`tests/test_docs_counts.py`):
   - `test_landing_page_cli_command_count` — the meta's CLI count vs `build_parser()` **in process**. The parser is the ground truth, not `--help` output. Exact match, no tolerance.
   - `test_landing_page_test_count` — the meta's test figure vs a live `pytest --collect-only` within the existing `TOLERANCE = 30`.
   - `test_landing_page_metric_card` (new) — the card's three figures against the exact sources the file's own refresh comment names: parser choices, the `@server.tool` count in `mcp/server.py`, and `pytest --collect-only`. Also checks the hero strip's test figure agrees with the card, so the page cannot argue with itself above the fold.

## Verification

- `pytest tests/test_docs_counts.py -q` → 8 passed.
- **Negative tests**: reverted the file to the stale figures and confirmed the guards fail — the card test reports `the metrics card says 45 CLI commands but build_parser() registers 46`, and the meta test reports `suite collects 2320 tests but the landing page says 2163: assert 157 <= 30`. Then restored.
- The card parser was unit-checked directly: `{"TESTS PASSING": 2323, "CLI COMMANDS": 46, "MCP TOOLS": 12}`, `_mcp_tool_count() == 12`.
[...]

PR #1309, opened 2026-09-19
docs(landing): correct the stale meta figures and guard them (#1283)

abyyxhek:
## What's broken

`docs/index.html`'s meta description is the figure set search engines and link previews render, and two of its three figures are stale:

| Claim | Live |
|:--|:--|
| 45 CLI commands | **46** — `build_parser()` subcommands |
| 2,163 tests | **~2,395 collected** |
| 12 MCP tools | correct |

This is the third time the page's headline drifted (`fd25b87` said "9 MCP tools, 14 CLI commands, 675 tests"; #927 resynced it; it has gone stale again), because every previous fix corrected the numbers by hand while nothing could catch the next drift: `tests/test_docs_counts.py` reads markdown only, so the HTML `<meta>` is invisible to every check.

## The fix

Corrects the two figures, and extends the guard so the resync is not the fourth hand-edit:

- `test_landing_meta_states_the_real_cli_command_count` — reads the meta description and pins its CLI-command figure to the subcommands `build_parser()` actually registers, in-process, no subprocess.
- `test_landing_meta_states_the_real_suite_total` — pins the test figure to a live `pytest --collect-only` under the existing tolerance, reusing `live_total()`.

Both fail if the description stops stating a figure at all, so a future edit cannot quietly opt out of the guard.

## Verification

- `tests/test_docs_counts.py` — 7 passed.

The branch also re-syncs the markdown figures main had already let drift past the count guard's own tolerance (2393 collected vs 2347 documented) to 2,395 across README.md, the five translated READMEs, CHANGELOG.md, docs/CONTRIBUTING_ONBOARDING.md, references/install.md and references/testing.md.


abyyxhek:
Closing in favour of #1284, which was opened a day earlier and is the more complete fix for #1283: it corrects both the meta description **and** the visible metrics card (which still shows `2,163+ TESTS PASSING` / `45 CLI COMMANDS` to every visitor), where this PR only corrected the meta. Deferring to the earlier, broader change rather than splitting the issue across two PRs.

Cyrax321 (approved):

PR #1310, opened 2026-09-19
fix(provenance): wire the derived-provenance module into its producers (#1098)

### LINKED ISSUE #1098: Follow-up to #392: recovery/derived.py is never imported, so both derived-artifact producers inline the stamping and the invariant check has no consumer

Parthipashok04:
## Description

`src/continuum/recovery/derived.py` was built for the #392 non-amplification invariant but nothing in `src/` imported it, so all three of its exports were dead except in one test file (issue #1098). Both producers of derived artifacts inlined the stamping, and each renderer labelled provenance its own way, which meant the invariant was never actually enforced anywhere a reader or an agent could see it.

This wires Option A from the issue:

- **`stamp_derived`** is now the single stamping path. `build_informed_retry` (`recovery/summary.py`) and `build_trajectory_report` (`analysis/trajectory_report.py`) both route through it instead of calling `derived_provenance_for_events` and writing the `derived_origin` key by hand. `recovery/summary.py` also drops its private `_derived_label`, which was a verbatim copy of the module function.
- **`derived_label`** is now the single rendering path. Both renderers use it. This closes a real gap rather than just moving code: `render_trajectory_report` hard-coded `external_agent`/`llm` as the unverified set, so an `imported` origin (which `Origin.self_certified` counts as unverified) rendered as `derived from imported`, and a legacy report with no recorded origin at all rendered as `derived from `. Both now read unverified.
- **`is_derived_unverified`** now marks the trajectory-report section of the curated briefing. Its title read `trajectory reports (sleep-time, system-derived)` even when every source the report was distilled from was self-reported. The caveat has to live in the title because `render_curated_briefing` renders `title` and `lines` but not the section `reason`.

While routing the persistence path through the same field, `record_trajectory_report` drops a redundant `payload["derived_origin"] = str(report.derived_origin)`: `report.model_dump()` already carries the field, so the line re-stamped what was already stamped.

Output for runs whose sources are trusted is unchanged byte for byte: both producers still compute the same origin value, and `derived_label` returns the same string the two removed copies returned for verified origins.

## Related issues

Fixes #1098

## Types of changes

- Type: `refactor`

(The wiring also fixes an incorrect trust label for `imported` and unstamped reports, but the primary intent is consuming the dead module.)

## Breaking changes

No. `_derived_label` was private (underscored, absent from `__all__`) and is the only removed symbol. The user-visible output changes are all in the cautious direction: unverified or legacy reports and the briefing section that carries them now say `unverified` instead of `derived from ...` / `system-derived`. No API, storage format, or safety verdict changes.

## How has this been tested

Windows 11, the repo venv (Python 3.12.10), run against a git worktree of `upstream/main` at 5540cf7 with `PYTHONPATH` pointed at the worktree `src` (the editable install otherwise resolves to the main checkout).

```
pytest tests/test_provenance_non_amplification.py tests/test_trajectory_report.py \
       tests/test_informed_retry.py tests/test_briefing_curation.py   # 32 passed
pytest -q                                                            # 2356 passed, 41 skipped, 1 failed
ruff check src/ tests/ examples/                                     # all checks passed
ruff format --check src/ tests/ examples/                            # 309 files already formatted
mypy src/continuum                                                  # Success: no issues found in 128 source files
```

Four regression tests were added to `tests/test_provenance_non_amplification.py` and each was verified to **fail on the pre-change code path** by restoring the base versions of the three source files and re-running the suite: 3 failed, 7 passed at base; all 10 pass after the change. The fourth is a control that passes both before and after, asserting the unverified title is not applied to reports from trusted sources.

The one remaining full-suite failure is pre-existing and unrelated: `tests/test_docs_counts.py::test_documented_count_matches_suite` reports `suite collects 2398 tests but docs say ~2347` and fails the same way on unmodified `upstream/main` at 5540cf7 (verified by running it with the test additions reverted). It is a documented-count re-sync job, out of scope here.

## Checklist

Tests added or updated for the changed behaviour; the four new tests fail on the old path and pass on the new one. Documentation updated: CHANGELOG `[Unreleased]` gained a `### Changed` entry. CI-green expectation rests on the local run above (ruff, format, mypy, pytest). No security-relevant behaviour changes: the recovery verdict is an input to these paths, never an output, and nothing here gates a mode or a safety result.

## Additional context

I chose the title/severity wiring for the briefing section over the alternative of gating visibility. Quarantining or omitting an unverified trajectory report would be wrong for the same reason the briefing quarantines rather than deletes: the resumed agent needs to know what the previous attempt looked like, including that it was self-reported. Marking the title keeps the information and states its authority.

## Summary by CodeRabbit

* **New Features**
  * Added clearer provenance labels to trajectory reports and informed-retry summaries.
  * Reports based on imported, missing, or otherwise unverified sources now display an “unverified” caveat.
  * Briefing sections now distinguish reports derived from unverified sources from system-derived reports.
  * Informed-retry summaries now consistently include a provenance line.

* **Documentation**
  * Added unreleased changelog entries describing provenance labeling and unverified-source handling.


CHANGELOG.md
```diff
@@ -8,6 +8,23 @@ All notable changes to this project are documented here. The format follows
 
 ### Changed
 
+- **The non-amplification invariant for derived artifacts now has consumers (#1098).**
```
Cyrax321, line 48: BLOCKING at merge time. This entry sits in a CHANGELOG whose count line at 957 still reads ~2,347, and this PR modifies CHANGELOG.md, so the merge puts 2,347 onto main where README.md is at 2,402 and this PR does not touch README.md. test_documented_counts_agree asserts every counted file states one figure, so the merged tree fails it. The branch predates several count bumps (it is behind by roughly 55 tests' worth of history), so this is a rebase job rather than a one-number edit: the four new tests this PR adds will move the collected total again on top of whatever main holds at rebase time.

src/continuum/analysis/trajectory_report.py
```diff
@@ -347,11 +352,12 @@ def health_maybe_generate_trajectory_report(
...
-        "unverified (derived)"
-        if report.derived_origin in ("external_agent", "llm")
-        else f"derived from {report.derived_origin}"
-    )
+    from continuum.recovery.derived import derived_label
+
+    # The stored payload is what the invariant is checked against, so the label
+    # a reader sees comes from the same helper: an IMPORTED or missing origin
+    # reads unverified here too, not just external_agent and llm (#392).
+    label = derived_label(report.model_dump())
```
Cyrax321, line 358: Non-blocking. For an external_agent or llm origin the rendered label changes from "unverified (derived)" to "unverified (derived from external_agent)". That is the intended direction and it is strictly more informative, but the tests assert only the substring "unverified", so no test pins the exact new wording and a later drift in that string would pass. One assertion on the full expected label would lock in the wording this fix chose, which is the part of the output users actually read.

src/continuum/recovery/briefing_curation.py
```diff
@@ -195,13 +195,24 @@ def curate_briefing(storage: Storage, run_id: str, decision: RecoveryDecision) -
         )
     if getattr(state, "trajectory_reports", None):
         from continuum.analysis.trajectory_report import render_trajectory_report
+        from continuum.recovery.derived import is_derived_unverified
 
         trajectory_lines: list[str] = []
+        any_unverified = False
```
Cyrax321: Non-blocking. Confirmed render_curated_briefing emits title and lines and not the section reason, so moving the caveat into the title is the only way it reaches a reader, which is the right call. Trivial: is_derived_unverified(report.model_dump()) runs for every report even after any_unverified is already True, so an ``or`` short-circuit (or ``any(is_derived_unverified(r.model_dump()) for r in ...)``, which also reads better) saves a model_dump per extra report on a path that builds a briefing.

Cyrax321 (changes requested):
This is a tidy wiring job and the refactor claims check out. I read derived_label against the removed private _derived_label and they are identical logic, so render_informed_retry's provenance line is unchanged for every origin the old helper handled; the conditional it was behind was always true, so making it unconditional is equivalent. stamp_derived copies its payload rather than mutating it, and TrajectoryReport.derived_origin defaults to "" so a legacy report fails closed through Origin("") raising, which is the right default for a trust label. The real bug fixed here is the renderer's hardcoded unverified set, and the IMPORTED and unstamped cases both now read unverified.

One thing blocks the merge, and it is the same family I flagged on #1317 and #1312: this branch is based on main when the docs carried 2,347, and main has since moved to 2,402. The PR modifies CHANGELOG.md, so the merge carries 2,347 back onto main while README.md, which this PR does not touch, stays at 2,402. test_documented_counts_agree requires every counted file to state one figure, so CI fails on the merge. The branch is far enough behind that this needs a real rebase, not a count edit.

Two non-blocking notes inline.

Cyrax321 (changes requested):
Author context: Parthipashok04, external contributor. Follow-up to the prior review. The #1098 wiring direction is still correct, but the docs-count merge blocker still stands: the branch pins older figures while the tree now reads near 2,452, and this needs a real rebase out of dirty state, not a count edit. Also coordinate with #1262 on the same issue; only one of the two should land. Rebase, resync every stating file, confirm CI green.

src/continuum/recovery/briefing_curation.py
```diff
@@ -195,13 +195,24 @@ def curate_briefing(storage: Storage, run_id: str, decision: RecoveryDecision) -
...
+            any_unverified = any_unverified or is_derived_unverified(report.model_dump())
         sections.append(
             {
-                "title": "trajectory reports (sleep-time, system-derived)",
+                "title": (
+                    "trajectory reports (sleep-time, derived from unverified sources)"
+                    if any_unverified
+                    else "trajectory reports (sleep-time, system-derived)"
+                ),
                 "provenance": _PROVENANCE_SYSTEM,
```
Cyrax321: Non-blocking: when `any_unverified` is true, this section's title says "derived from unverified sources" while its `provenance` still advertises `_PROVENANCE_SYSTEM`. The module docstring documents the tier as the machine-readable trust signal on every section ("most trusted first") and this dict is what rides out as the hook payload, so agent-authority material leaves here mislabelled as system-tier; nothing in-repo renders the field today, which is the only reason this is not blocking. Since the PR's purpose is making the #392 invariant observable wherever the artifact is surfaced, the tier is the one surface that still contradicts it. Switching to `_PROVENANCE_AGENT` in the unverified branch, as the sibling PR #1262 does, keeps the metadata honest.

src/continuum/analysis/trajectory_report.py
```diff
@@ -180,6 +179,7 @@ def build_trajectory_report(
     now: datetime | None = None,
 ) -> TrajectoryReport:
     """Distill metrics and failure patterns across an event window into a report."""
+    from continuum.recovery.derived import stamp_derived
```
Cyrax321: Non-blocking: this function-local import (and the `derived_label` one in `render_trajectory_report` below) does not need to be deferred. `continuum.recovery.derived` only pulls in `events`, `models` and `provenance_map`, and nothing in the `recovery` package init imports `analysis.trajectory_report` at module level, so importing it at the top of this file creates no cycle; I verified the import both directions. `summary.py` in this same PR imports the same module at top level, so the two placements are inconsistent, and hoisting both would put the dependency where readers look for it. If a cycle was actually hit during development, a one-line comment naming it would be the alternative fix.

src/continuum/recovery/briefing_curation.py
```diff
@@ -195,13 +195,24 @@ def curate_briefing(storage: Storage, run_id: str, decision: RecoveryDecision) -
...
 
         trajectory_lines: list[str] = []
+        any_unverified = False
         for report in state.trajectory_reports:
             trajectory_lines += render_trajectory_report(report)
+            # The distillation is always mechanical, but its *sources* may be
+            # self-reported, and one such source makes the content unverified
+            # no matter who projected it (#392). The title, not the reason, is
+            # what renders, so the caveat has to live there to be seen.
+            any_unverified = any_unverified or is_derived_unverified(report.model_dump())
```
Cyrax321: Non-blocking: `is_derived_unverified` only reads the `derived_origin` key, and `render_trajectory_report` calls `model_dump()` again on the same report for the same purpose, so each report is serialized twice to fetch one string field. `is_derived_unverified({"derived_origin": report.derived_origin})` is enough and states the real dependency. The cost is negligible at these caps, so this is about clarity rather than speed.

Cyrax321 (commented):
This wires the previously dead recovery/derived.py module into both derived-artifact producers and both renderers, drops the redundant re-stamp in record_trajectory_report, and marks the briefing's trajectory section when its sources are self-reported. The removed-behavior audit checks out: the deleted payload line was already carried by model_dump(mode="json") since derived_origin is a str field, _derived_label was a verbatim private copy of derived_label, and the two producers were the only in-src callers of derived_provenance_for_events, so the wiring is complete. The four regression tests cover the imported and unstamped gaps the old renderer had, and CI is green across the matrix. Three non-blocking nits inline.

Heads-up: open PR #1262 addresses the same issue (#1098) with an overlapping diff. This one is the cleaner minimal wiring, with tighter scope and regression tests verified to fail on the old path, but #1262 additionally demotes the section's provenance tier to agent when sources are unverified, and that half is worth carrying over here (first inline comment).

Parthipashok04:
Thanks for the review. Force-pushed the rebase plus all five nits; the branch is now 3 commits on top of current `main` (0 behind).

## The blocker: rebase

Replayed the branch onto current `main` (it was 60 commits behind). `CHANGELOG.md` was the only conflict, resolved by keeping your newer entries and the current `~2,452` count, with the `#1098` entry alongside your `### Changed` block. The three source files this PR touches were not changed upstream since the base, so nothing else conflicted.

`test_documented_counts_agree` now passes: every counted file states `~2,452`, and the slow `test_documented_count_matches_suite` compares that against a live collection of 2,482 (exactly the guard's tolerance of 30). No count edits were needed.

## The five nits

1. **Hoisted the deferred imports** in `analysis/trajectory_report.py`. `stamp_derived` and `derived_label` are now a single top-level import; there is no cycle, matching how `recovery/summary.py` in this same PR already imports the module. `stable_hash` stayed function-local since it is not part of that set.
2. **Short-circuited the unverified scan** in `recovery/briefing_curation.py`: the loop tail is gone, replaced by a single `any(...)` over the reports.
3. **Honest provenance tier**: when any source is unverified the section's `provenance` field is now `agent` rather than `system`, matching the caveat the title already carries. Nothing renders the field today, but it rides out as the hook payload, so it should not advertise system authority for self-reported content. The CHANGELOG entry says so.
4. **Pinned the full label wording** in `tests/test_provenance_non_amplification.py`: exact strings for `external_agent`, `imported`, the unstamped/legacy case, and the trusted case, rather than asserting the `"unverified"` substring.
5. **Stated the real dependency**: the check is now `is_derived_unverified({"derived_origin": report.derived_origin})`, which is the only key it reads.

## One interaction the rebase surfaced

Upstream's `test_render_trajectory_report_labels_a_derived_origin` (from #1235) pinned this PR's old hardcoded renderer output (`"unverified (derived)"`) and used a synthetic `"checkpoint"` origin, which is not a real `Origin` member. I updated it to assert the full new label (`"unverified (derived from external_agent)"`, which is strictly more informative, as you noted) and to use the real trusted origin `"deterministic"` for the verified half, preserving the test's intent.

## Verification

Full suite green from the worktree: `2438 passed, 44 skipped`, 0 failed. `ruff check`, `ruff format --check`, and `mypy src/continuum` clean. I also confirmed the new tier assertion is not a tautology by reverting the tier line and watching `test_briefing_marks_trajectory_section_from_unverified_sources` fail with `assert 'system' == 'agent'` before restoring it.

On coordination with #1262: still happy to defer or rebase onto whichever you prefer to land first.

PR #1312, opened 2026-09-19
ci: bring benchmarks/ inside the ruff lint gates (#1064)

### LINKED ISSUE #1064: benchmarks/ is imported by tests but excluded from every lint gate, carrying 9 findings

soorajsuru2004-sudo:
## Description

`benchmarks/` is imported by the test suite but was excluded from every lint gate in CI and pre-commit configuration, carrying five lint and formatting findings in `benchmarks/fault_injection/runner.py`.

This PR:
- Resolves the findings in `benchmarks/fault_injection/runner.py`: context manager for temporary file (`SIM115`), sorted imports (`I001`), combined nested `if` statements (`SIM102`), `contextlib.suppress` for file unlinking (`SIM105`), and formatting drift.
- Extends the `ruff check` and `ruff format --check` CI steps in `.github/workflows/ci.yml` to include `benchmarks/`.
- Updates `.pre-commit-config.yaml` to include `benchmarks/` in both hook file patterns and updates the comment to reflect that only scratch directories (`demo-run/`, `_bugaudit/`) remain outside.
- Updates `tests/test_precommit_config.py` to assert `benchmarks` is included in CI linted directories and ungated directories remain excluded.
- Synchronizes contributor documentation and verification commands across the repository to include `benchmarks/`.

## Related issues

Fixes #1064

## Types of changes

- Type: `ci`

## Breaking changes

No

## How has this been tested?

Environment: Windows 11, Python 3.12.10

Ran the full lint and format check across all four directories:
```bash
ruff check src/ tests/ examples/ benchmarks/
# All checks passed!

ruff format --check src/ tests/ examples/ benchmarks/
# 344 files already formatted
```

Strict mypy type check:
```bash
mypy src/continuum
# Success: no issues found in 128 source files
```

Executed the test suite across affected tests:
```bash
pytest tests/test_precommit_config.py tests/test_docs_counts.py tests/test_fault_injection.py tests/test_benchmark_counters.py tests/test_bench_unsafe_edit.py tests/test_risk_bench.py tests/test_horizon_benchmark.py tests/test_dropped_constraint_reconstruction.py -q
# 46 passed in 10.32s
```

## Checklist

## Additional context

Verified that no em dashes were introduced across all modified files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Benchmark code is now included in automated linting and formatting checks.
  * Resolved existing lint and formatting issues in the fault-injection benchmarks.

* **Documentation**
  * Updated contributor guides, testing references, release checklists, and localized README instructions to reflect benchmark verification coverage.

* **Tests**
  * Added coverage to verify linting scope and excluded directories remain correctly configured.


README.md
```diff
@@ -100,7 +100,7 @@ Verify:
 continuum --help                 # CLI entrypoint
 continuum-mcp --help             # MCP server entrypoint (needs [mcp] or [dev])
 pytest -q                        # ~2,397 collected, ~2,320 passed, ~27 skipped on a minimal env (exact counts vary)
-ruff check src/ tests/ examples/ && ruff format --check src/ tests/ examples/
+ruff check src/ tests/ examples/ benchmarks/ && ruff format --check src/ tests/ examples/ benchmarks/
```
Cyrax321, line 103: BLOCKING at merge time, not in this branch. This PR edits the ruff line inside the fenced verify block but leaves the count line above it at 2,397, which is what main held when this branched on 2026-09-19. Main has since moved to 2,402, and references/install.md and references/testing.md state 2,402 without being touched here. test_documented_counts_agree asserts every COUNTED_FILES figure is identical, so the merged tree fails it with README at 2,397 against references at 2,402. Same family as #1317, which I flagged there.

Since the PR is already in this block in all six READMEs, the count line is one row away from the edit. Rebase onto current main and re-run the guard rather than re-asserting 2,397.

benchmarks/fault_injection/runner.py
```diff
@@ -411,14 +421,14 @@ def _assess_fresh_key_reissuance() -> tuple[bool, str | None, list[str], bool]:
                     "remaining 0" not in msg
```
Cyrax321, line 421: Non-blocking, and this PR does not change the behaviour: I checked the SIM102 combine against 200k generated messages and the branch is provably equivalent to the two nested ifs it replaces, so this is a safe reformat.

Worth a separate look while the file is under a gate for the first time: the check passes whenever the refusal message contains the digit 0 anywhere, not specifically an indication that remaining is 0. A message like "budget exhausted, 0 attempts used in another bucket" satisfies it. The three phrasings above are the real test and each implies a bare 0, so the last conjunct subsumes them; dropping it and requiring one of the phrasings would make the assertion mean what its failure message says. Pre-existing, so no need to fix it in this PR, but this is the line that would have to change.

Cyrax321 (changes requested):
The change itself is right and complete: I confirmed benchmarks/ carries exactly the five findings this PR fixes in fault_injection/runner.py and nothing else, and nothing under benchmarks/ has landed on main since this branched, so widening the gate breaks no untouched tree. The pre-commit and CI regexes are updated together, which the config comment correctly identifies as the invariant, and test_precommit_config.py now asserts both the new inclusion and the two directories that stay excluded. The docs sync is unusually thorough across six READMEs and five guides.

One thing blocks the merge and it is not the lint work: the branch predates main's test count bump, so the six README blocks this PR edits still carry 2,397 while main's references/install.md and references/testing.md, which this PR does not touch, carry 2,402. test_documented_counts_agree requires every counted file to state one figure, so the merged tree disagrees. The PR's own CI is green because the branch is internally consistent; the failure appears only on merge.

One non-blocking note on an assertion the reformatting touched, inline.

Cyrax321 (changes requested):
Author context: soorajsuru2004-sudo, external contributor. Follow-up to the prior review. The benchmarks lint widening is still right and complete, but the docs-count merge blocker still stands and has grown: the branch pins older figures while the tree now reads near 2,452. Rebase onto current main and resync every stating file the PR touches so the merged tree agrees on one figure.

Cyrax321 (approved):
This widens the ruff lint and format gates to benchmarks/ and makes the tree actually pass them. I verified the premise locally: on main, `ruff check benchmarks/` reports exactly the five findings named in the description, all in `benchmarks/fault_injection/runner.py`, and that file is also the only format offender; at this head it passes both gates under the repo's ruff config, which carries no benchmarks-specific per-file-ignores or excludes that could silently re-narrow the scope.

The runner.py rewrites are behavior-preserving: the `NamedTemporaryFile` with-block still closes the handle before the path is used (`delete=False`, then unlinked in the `finally`), the SIM102 merge of the nested `if "0" not in msg` into the outer condition is logically identical to the old nesting, and `contextlib.suppress(Exception)` matches the old try/except/pass.

The three-way agreement between ci.yml, the pre-commit hook `files` patterns, and tests/test_precommit_config.py holds: `_ci_lint_directories()` parses the new `benchmarks/` token out of both ruff invocations, and the negative assertions now pin demo-run/ and _bugaudit/ as the only ungated directories, matching the updated comment in the pre-commit config. The verification commands were synchronized in every doc that lists them; the remaining old-scope strings in STATUS.md and CHANGELOG.md sit inside dated historical entries, so leaving them untouched is correct. Approving.

Cyrax321 (commented):
Hi @soorajsuru2004-sudo,

This PR is approved and the changes look solid. However, there are currently **merge conflicts** in several files (`.github/workflows/ci.yml`, `.pre-commit-config.yaml`, `CHANGELOG.md`, and the `README*.md` files) due to recent updates on `main`.

Please **rebase** this branch onto the latest `main`, resolve the conflicts, and push the updated branch. Once the conflicts are resolved and CI passes, we can proceed with the merge.

PR #1313, opened 2026-09-19
fix(storage): the Postgres backend keeps a fork's parent_run_id (#1079)

### LINKED ISSUE #1079: Postgres backend never writes or reads runs.parent_run_id, so fork lineage and the family resume block are silently absent there

abyyxhek:
## Description

The Postgres backend declared `runs.parent_run_id TEXT REFERENCES runs(run_id)` but never populated it and never read it back. `create_run` and `create_run_started` both inserted only `(run_id, goal, status, created_at, updated_at, metadata)`, and `_row_to_run` built the `Run` without the column, so a stored value came back as the model default `None` even for a row that somehow had one. SQLite writes and reads the column in all three places.

The fix writes `run.parent_run_id` in both inserts and maps `row["parent_run_id"]` in `_row_to_run`, so the two engines agree on the column. No schema change is needed — the column already exists on both, added by migration `307` for issue #243; on Postgres it was simply unreachable.

Why this matters more than a cosmetic divergence: `children_of` (`recovery/family.py`) filters `list_runs` on `parent_run_id`, so on the Postgres engine it always returned empty. `roll_up_children` therefore never reported a family block, and the rule the TUI and CLI both render — a parent may not RESUME while any non-terminal child is unsafe — could not fire. A Postgres deployment would look healthy and silently never block a parent over an unsafe child. The foreign-key constraint that is supposed to guarantee the referenced parent exists was likewise never exercised, because the column was always null. `fork_run` builds the child as `Run(..., parent_run_id=parent_run_id, ...)`; that value was dropped on write.

`update_run` is untouched and matches SQLite on both engines: parentage is set at creation and is not re-writable.

## Related issues

Fixes #1079

## Types of changes

- Type: `bugfix`

## Breaking changes

No. The column is nullable and already present on both engines; existing rows keep their nulls, which the legacy metadata convention in `children_of` still resolves.

## How has this been tested?

Docker Hub is unreachable from this network, so the CI service container was reproduced locally with `pgserver`'s bundled Postgres 16 binaries against a scratch venv, and the contract suite was pointed at it via `CONTINUUM_TEST_POSTGRES_DSN`:

```
22 passed  (tests/test_storage_postgres.py, tests/test_provenance_postgres.py)
```

The new lineage test was confirmed to fail without the source fix (`assert None == 'pg_par'`) and pass with it, so it guards the actual behaviour rather than the new code shape.

Also run locally (Windows 11, Python 3.11, worktree on `upstream/main`):

```
ruff check src/ tests/              # All checks passed
ruff format --check src/ tests/     # 309 files already formatted
mypy src/continuum/storage          # no issues
pytest -q -k "not slow"             # full suite, exit 0
pytest tests/test_docs_counts.py tests/test_precommit_config.py   # 9 passed
pytest tests/test_storage.py tests/test_family.py tests/test_fork.py
```

## Checklist

Tests added or updated for the changed behaviour — two contract cases in `tests/test_storage_postgres.py`: a fork's lineage round-trips and `children_of` resolves the child, and a parentless run still loads as `None` (the read path must not turn a null column into a `CorruptedRecord`). Documentation updated — `CHANGELOG.md` Unreleased → Fixed. No user-facing behaviour changes beyond the corrected property, so no README/`references` changes were needed; the documented test count stays inside the guard's tolerance (2,400 collected vs ~2,397 documented). Security-relevant limitation: with the column now actually written, the `REFERENCES runs(run_id)` constraint fires for the first time on Postgres and a dangling parent id surfaces as `ConcurrentWriteError`; the CLI's `start --parent` already validates the parent exists first (`test_start_refuses_an_unknown_parent`), so this is a backstop rather than the primary check, and it matches SQLite's declared constraint.

## Additional context

The two new tests only execute when `CONTINUUM_TEST_POSTGRES_DSN` is set and `psycopg` is importable, so they skip cleanly in a plain checkout and run for real in CI's Postgres 16 service container, like the rest of that file.

## Summary by CodeRabbit

* **Bug Fixes**
  * PostgreSQL now preserves parent-child relationships between runs when creating and retrieving runs.
  * Family and child-run queries now correctly resolve run lineage on PostgreSQL, matching SQLite behavior.

* **Tests**
  * Added coverage for parentless runs, child runs, and `children_of` lineage queries.


Cyrax321 (dismissed):

Ref: #1313 / Issue #1079
Surface: Postgres storage backend, run lineage

Bug: runs.parent_run_id was declared with a foreign key and never written or read on the Postgres engine, so every fork silently lost its lineage and, since children_of filters list_runs on that column, the family resume block was vacuous on Postgres: a deployment could never refuse a parent RESUME over an unsafe child, with no error anywhere.

Fix reviewed: both inserts carry run.parent_run_id and _row_to_run maps it back, matching SQLite; no schema change needed since migration 307 already created the column. This is the same conclusion #1124 reached and the diffs agree on every line of production code.

Proof:
- The contract suite gains a case that forks on the engine and asserts children_of resolves the child; the Postgres CI job runs it against a real backend and is green, as is the rest of the matrix at head c8ecc110.
- I ran tests/test_storage_postgres.py locally: 21 skipped (no local Postgres); CI is the evidence here.
- No U+2014 in the committed lines.

Merge coordination: this and #1124 fix the identical defect. Land exactly one. My earlier review of #1124 noted its regression test also asserts the children_of path; this PR's contract-suite case covers the same seam on the engine side, so either is sufficient. Close the loser and reference the winner so #1079's audit trail stays clean.

No blocking issues. Ready to merge (subject to landing only one of the two duplicates).

PR #1315, opened 2026-09-20
fix(recovery): make the remaining assess scans archive-aware (#1050)

soorajsuru2004-sudo:
## What's broken

`RecoveryEngine.assess` was half archive-aware. The confirmation scan, the provenance scan, and the validator all fold the archived prefix through one shared `read_all_events` fetch (`engine.py:274`), but three signals computed in the same method still read the live tail only:

- the liveness breach count (`engine.py:399`)
- the `RISK_OBSERVED` scan (`engine.py:419`)
- the consumed-authorities scan (`engine.py:470`)

`continuum compact` archives the pre-anchor prefix, so all three disappeared from the sealed contract the moment a run was compacted. The engine still returned a verdict — just one computed over a deliberately truncated history, and the downgrade is always toward less caution:

```
    contract.triggering_risks: ['event_6c4710e5...']
    contract.liveness: {'breaches': 1, ...}
    rationale: risk meltdown triggers rollback
    contract.triggering_risks: []
    contract.liveness: {'breaches': 0, ...}
```

The contract is the auditable artifact a human or downstream gate acts on, so compaction silently removed the evidence that justified the cautious verdict. Isolating the risk signal with `meltdown`, which policy maps to `ROLLBACK`, makes the downgrade unambiguous: `rollback` before, `request_human` after.

## The fix

All three scans now filter the `archive_aware_events` fetch the method already takes, instead of re-reading the live tail three more times:

- the breach count filters it for `LIVENESS_SILENCE_DETECTED`,
- the risk scan filters it for `RISK_OBSERVED`,
- the authority block passes it to `collect_consumed_authorities`.

This also retires three redundant live-tail reads. The `try/except` fallbacks already in place are untouched, including the authority block's fail-closed degradation on an unreadable ledger (#1066).

## Tests

Three regression tests, one per signal, mirroring the existing `test_confirmation_survives_compaction`: each isolates one archived signal, asserts the contract still carries it after a `compact_run`, and asserts the precondition that the event really did leave the live tail. All three fail on unmodified main and pass here:

```
FAILED test_risk_trigger_survives_compaction
FAILED test_liveness_breach_count_survives_compaction
FAILED test_consumed_authority_survives_compaction
```

## Verification

- Full suite green (2,400 collected locally; the dashboard socket test is a known Windows full-run flake — it passes in isolation and on pristine main).
- `ruff check` and `ruff format --check` clean on both changed Python files.
- `mypy` not run locally (blocked on this machine); CI carries the type check.

No documented test-count figure is touched, so the count guard stays green (README/CHANGELOG state ~2,397; tolerance is 30).

## Relationship to #1308

This fixes the same issue as #1308, which is approved but blocked by merge conflicts. Those conflicts are entirely in the docs (CHANGELOG, the READMEs, CONTRIBUTING, install/testing) and come from that PR's test-count re-sync, which main has since done itself — its `engine.py` fix does not conflict. This PR carries only the fix and its tests, rebased on current main, so it merges cleanly. Happy to close this if #1308 rebases first.

## Summary by CodeRabbit

- **Bug Fixes**
  - Recovery assessments now continue to account for archived events when evaluating liveness breaches, risk signals, and consumed authorities after compaction.
  - Rollback decisions, breach counts, and human-review requirements remain accurate when relevant events have been archived.

- **Tests**
  - Added regression coverage for recovery behavior involving archived events.


Cyrax321 (changes requested):
Author context: soorajsuru2004-sudo, external contributor. Same verdict as #1327: this fix is already on main. Commit 6fb3ccc0 landed the identical three archive-aware scans for #1050, and origin/main already folds archive_aware_events for the liveness breach count, the RISK_OBSERVED scan, and collect_consumed_authorities. I confirmed all three call sites on current main. This PR is also in dirty merge state with an em dash in the CHANGELOG hunk (AGENTS.md bars U+2014). Please close as superseded, or rebase and show what remains beyond 6fb3ccc0.

src/continuum/recovery/engine.py
```diff
@@ -394,11 +394,14 @@ def assess(
...
+            # Count prior breaches as DETECTED events. Fold the archived prefix
+            # in via the shared fetch: a silence detected before a compaction
+            # still counts, so the breach total cannot reset to zero and
+            # understate how often the run went quiet (same archive-blindness
+            # family as issue #553).
             try:
-                evs = self.storage.read_events(run_id)
                 liveness_breaches = sum(
-                    1 for e in evs if e.type == EventType.LIVENESS_SILENCE_DETECTED
+                    1 for e in archive_aware_events if e.type == EventType.LIVENESS_SILENCE_DETECTED
```
Cyrax321, line 404: Blocking. This line is byte-identical to what is already on main from merged #1308 (squash 6fb3ccc0): current engine.py has exactly `1 for e in archive_aware_events if e.type == EventType.LIVENESS_SILENCE_DETECTED` here, and the same holds for the risk scan (main line 423) and the `collect_consumed_authorities(archive_aware_events)` call (main line 483) in the other two hunks. Because main already contains the change, the `-evs = self.storage.read_events(run_id)` removal no longer exists to remove, which is the conflict GitHub reports as DIRTY. Merging this would either fail or re-apply behavior that is already shipped, so the fix cannot land from this PR.

tests/test_recovery_engine.py
```diff
@@ -577,6 +577,100 @@ def test_confirmation_survives_compaction(store: SQLiteStorage) -> None:
     assert after.safe
 
 
+def test_risk_trigger_survives_compaction(store: SQLiteStorage) -> None:
```
Cyrax321, line 580: Blocking. Main already defines `test_risk_trigger_survives_compaction` at exactly line 580 of this file, plus `test_liveness_breach_count_survives_compaction` at line 612 and `test_consumed_authority_survives_compaction` at line 641, all landed by the same merged #1308 commit 6fb3ccc0 with the same seeding and assertions. Re-applying this hunk would create duplicate function definitions in the same module, so even if the engine.py conflict were resolved the tests cannot merge as written. The regression coverage this PR adds is therefore already in place on main.

Cyrax321 (changes requested):
The fix itself is correct, but it has already shipped: #1308 (same issue #1050) merged on 2026-09-20 as 6fb3ccc0, and current main carries this exact engine.py change (lines 402-403, 423, 483) plus these same three tests under the identical names, which is why the branch reports DIRTY and the hunks no longer apply. The body says #1308 is approved but blocked by merge conflicts; that is stale, it merged two days ago, so the premise for this resubmission no longer holds. On #1050 coverage across open PRs: #1327 changes the exact same three scans (liveness, risk, consumed authority), so it is fully redundant with both this PR and the merged #1308 rather than partitioning the scans. Please close this PR; nothing in it is mergeable against current main.

Cyrax321 (changes requested):
Hi @soorajsuru2004-sudo,

This PR has **merge conflicts** and is effectively redundant. The fix for **#1050** was already merged via **#1308** (commit `6fb3ccc`), which includes the identical changes to `engine.py` and the corresponding regression tests in `test_recovery_engine.py`.

Additionally, there is a formatting violation in `CHANGELOG.md`: please replace the **em dash** with a comma, as per our coding guidelines.

Since the code is already on `main`, please **close this PR**. If you believe there are remaining changes not covered by #1308, please rebase and clarify what is new. Otherwise, no further action is needed here.

PR #1316, opened 2026-09-20
fix(recovery): assess folds the archived prefix for risk, liveness and authority (#1050)

Adhi1-2:
## What

`RecoveryEngine.assess` was half archive-aware. It fetched the merged
archive-aware history (`read_all_events`) and folded it for confirmations and
pinning, but three other signals still scanned the **live tail only**:

| Signal | Scan | Consequence after a compaction |
| --- | --- | --- |
| `RISK_OBSERVED` | trigger → recovery mode | the trigger vanished, so `rollback` downgraded to `request_human` |
| `LIVENESS_SILENCE_DETECTED` | breach count in the contract's `liveness` section | the count reset to `0`; a breached run read as quiet |
| `AUTHORITY_CONSUMED` | `consumed authority blocks resume` rationale | the block dropped, though the gate still refuses to forward |

`compact_run` archives exactly those rows — they are ordinary events in the
pre-anchor prefix. This is the same archive-blindness family as #1186, #1172,
#1128, #1126 and #1072, but these three change the **verdict**, not a display,
and every failure direction is toward *less* caution.

The sharpest case is the risk trigger. A run carrying a `meltdown` risk reads
`rollback` before the compaction and `request_human` after it — with
`triggering_risks` empty, so the contract no longer names the event that
justified the original verdict.

## The fix

All three scans now fold the shared `archive_aware_events` list the rest of
`assess` already fetches (with its existing fall-back to the live log when the
archive read fails). No new storage call, no new code path — the history was
already in hand and these three scans were the ones not reading it.

## Tests

Three regression tests in `tests/test_recovery_engine.py`, one per signal.
Each appends the signal, asserts the pre-compaction verdict, compacts, asserts
the signal has genuinely left the live tail, then asserts the verdict is
unchanged. All three fail on the unpatched engine:

```
FAILED test_a_risk_trigger_survives_compaction
FAILED test_a_liveness_breach_count_survives_compaction
FAILED test_a_consumed_authority_survives_compaction
```

## Verification

- `ruff check src/ tests/ examples/`, `ruff format --check src/ tests/ examples/` — clean
- `mypy src/continuum` — clean (128 files)
- `pytest -m "not slow"` — 2386 passed, 6 skipped
- Postgres suite (`tests/test_storage_postgres.py`, `tests/test_provenance_postgres.py`) — clean


src/continuum/recovery/engine.py
```diff
@@ -394,11 +394,13 @@ def assess(
...
+            # the anchor still counts: the archived prefix is part of the run's
+            # history, and compaction must not reset a silence counter the
+            # contract reports (#1050).
             try:
-                evs = self.storage.read_events(run_id)
                 liveness_breaches = sum(
-                    1 for e in evs if e.type == EventType.LIVENESS_SILENCE_DETECTED
+                    1 for e in archive_aware_events if e.type is EventType.LIVENESS_SILENCE_DETECTED
                 )
             except Exception:
```
Cyrax321, line 405: Non-blocking. Moving the storage read out of these blocks left the exception handling dead. Both this one and the risk_events one at line 425 now wrap a pure iteration over an already-fetched local list: archive_aware_events cannot raise, and neither can an attribute read and an enum identity check. The read_all_events failure that could raise is caught earlier, at the top of the function, with its own live-log fallback.

The cost is that a reader assumes these except clauses tolerate a real failure mode that no longer exists, which matters here because the comments around them are about degraded-log behaviour. collect_consumed_authorities at line 483 is the one that can still raise on a malformed event, so its except stays live. Worth dropping the two unreachable handlers or moving the comments to say the fallback is the one above.

Cyrax321 (approved):
This is the right fix at the right depth. The three scans were the last verdict-affecting readers of the live tail in assess, and read_all_events' own contract in storage/base.py names "authority enforcement" and "cross-run action scans" as callers that must read full history because compaction moves facts without revoking their consequences, so this brings three more callers onto the documented helper rather than adding a special case. I confirmed ledger.all() and ledger.pending() already replay archived-plus-live, so no other verdict-affecting scan in assess is still live-only, and the merged sequence has no duplication because compaction moves rows rather than copying them.

The failure direction matters and is correctly characterised: all three degrade toward less caution, so the silent wrong answer was the safe-looking one. The tests each assert the signal left the live tail before asserting the verdict held, which is what makes them regression tests rather than tautologies.

One non-blocking cleanup inline. Approving.

Adhi1-2:
Closing as superseded: this fix shipped in #1308 (merged 92fe651), which makes `assess` fully archive-aware for the same three signals — the liveness breach count, the `RISK_OBSERVED` scan, and the consumed-authorities scan — all off the shared `read_all_events` fetch that #553 established.

I diffed my `engine.py` against what merged: the only differences are comment wording and `==` vs `is` on the enum comparisons. Behaviour is identical, and the test coverage is the same five compaction-survival tests in both. Nothing of independent value remains, so rebasing would just re-submit a duplicate.

The branch stays available if a maintainer wants the alternate comments.

PR #1317, opened 2026-09-20
fix(docs): resync the landing page metrics and guard them (#1283)

Adhi1-2:
## What

`docs/index.html` is the first thing a visitor sees, and its headline figures were the oldest in the repo:

| Metric | Page said | Actual |
|---|---|---|
| Tests passing | 2,163 | 2,401 collected (README canonical: 2,397) |
| CLI commands | 45 | 46 (`build_parser()` choices) |
| MCP tools | 12 | 12 (correct) |

Nothing noticed, because every existing drift guard reads markdown. Issue #1283 prescribed the fix, and the page's own refresh note (line 225) already names the three canonical sources.

## Changes

- `docs/index.html` — 2,163 → 2,397 in the three places the figure is stated (meta description, hero banner, metrics card), 45 → 46 in both CLI-command sites. MCP tools left alone.
- `tests/test_docs_counts.py` — `docs/index.html` is now a counted file, so its test figure must agree *exactly* with every markdown figure. Three HTML-shaped regexes reach the three sites; the uppercase `TESTS PASSING` pattern is case-sensitive on purpose (`docs/CONTINUUM_MASTER_PLAN.md` says "909 tests passing" about a long-ago snapshot and is not a counted file, but a case-insensitive flag would let it veto the real total). The narrative-forms test gains the three HTML cases.
- `tests/test_cli_docs.py` — new `test_landing_page_command_count_matches_parser` compares the page's two CLI-command sites against `build_parser()` *and* against each other, so a future drift fails loudly instead of silently.

The test figure is 2,397, not the live 2,401: `test_documented_counts_agree` requires every stated figure across all docs to be identical, and the page's refresh note explicitly defers to the README canonical figure.

## Verification

- `ruff check` / `ruff format --check` on `src/ tests/ examples/` — clean.
- `mypy src/continuum` — clean (128 files).
- `markdownlint-cli2 CHANGELOG.md` — clean.
- Full suite: one failure, `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend`. It is pre-existing and environmental — it expects a `RuntimeError` mentioning `psycopg` when the driver is absent, and this machine has psycopg installed. Reproduced on a stashed tree with no diff applied.
- Both new guards proven load-bearing: with `docs/index.html` reverted they fail (`index.html: 2163` disagrees with the markdown consensus of 2397; CLI-command count 45 vs 46).


docs/index.html
```diff
@@ -224,7 +224,7 @@ <h1 class="pixel-headline" aria-label="Agents that survive">
 
     <!-- Metrics refresh note: TESTS from `pytest --collect-only -q` (README canonical figure), MCP TOOLS from the @server.tool count in src/continuum/mcp/server.py, CLI COMMANDS from len(build_parser() choices). -->
     <div class="bottom-copy bottom-left">
-      <div>2,163 TESTS PASSING</div>
+      <div>2,397 TESTS PASSING</div>
```
Cyrax321: BLOCKING only because of the number, not the change. Verified: the guard's regex set finds {2397} in this file and {2402} in main's README.md, and test_documented_counts_agree requires len(set(stated.values())) == 1 across COUNTED_FILES. main moved to 2,402 after this branched (the README count commit that landed since), so the three sites here and the two CLI-command sites need re-measuring on a rebased tree. The refresh note on the line above already names the three sources, so it is one edit plus a test run, which is exactly the maintainability win this PR is after.

tests/test_docs_counts.py
```diff
@@ -67,6 +70,15 @@
         re.IGNORECASE,
     ),
     re.compile(r"^pytest\s+-q\s+#.*?([\d,]+)", re.MULTILINE),
+    # docs/index.html is not markdown, so none of the prose forms above reach
+    # it. It states the figure three times -- the meta description, the hero
+    # banner, and the metrics card (#1283). Uppercase is deliberate: the master
+    # plan says "909 tests passing" about a long-ago snapshot and is not a
+    # counted file, but a case-insensitive flag would let it veto the real
+    # total.
+    re.compile(r"CLI commands,\s*([\d,]+)\s+tests"),
```
Cyrax321, line 79: Non-blocking, but the stated rationale does not hold. docs/CONTINUUM_MASTER_PLAN.md is not in OPTIONAL_FILES or REQUIRED_FILES, so documented_total is never run against it and its "909 tests passing" cannot veto anything regardless of the flag. The case-sensitivity is load-bearing for a different reason: this is the only pattern that reaches the hero banner's `<div>2,397 TESTS PASSING</div>`, and none of the markdown-shaped patterns above would match it at any casing. Worth stating that reason instead, because as written a later reader may "fix" the flag to IGNORECASE on the master-plan argument and not realise they are also widening which forms in counted files get checked.

tests/test_cli_docs.py
```diff
@@ -87,6 +88,35 @@ def test_readme_module_map_command_count_matches_parser() -> None:
...
+    disagrees with either the parser or itself is wrong, so both are compared
+    against the parser rather than pinned.
+    """
+    html = LANDING.read_text(encoding="utf-8")
+    live = len(build_parser()._subparsers._group_actions[0].choices)
+    documented = set()
+    for pattern in (
+        re.compile(r"([\d,]+)\s+CLI commands"),
+        re.compile(
+            r'class="metric-value">([\d,]+)</span>\s*<span class="metric-label">CLI COMMANDS'
```
Cyrax321, line 106: Non-blocking. build_parser()._subparsers._group_actions[0].choices relies on the subparsers being the parser's first action group, which is true today and matches the two existing tests above, so no change is needed here. The concern is only that all three break silently together if a group is ever added ahead of the subparsers; a helper next to build_parser exposing the command names would remove the private-argparse reach for all three at once.

Cyrax321 (changes requested):
The fix is the right shape and the two new guards are well designed: they compare the page against the parser rather than pinning a number, and comparing the page's two CLI-command sites against each other catches a half-updated page. I confirmed both new guards fail with index.html reverted, as the body claims.

One blocking problem: the branch is stale against main. It anchors the page to 2,397, which was the README canonical figure when this branched, but main has since moved to 2,402 across the READMEs. I ran the guard's own regex set against both files: the PR's index.html yields {2397}, main's README.md yields {2402}. test_documented_counts_agree asserts every COUNTED_FILES figure is identical, so merging this as-is fails the guard this PR adds. The number needs re-measuring on a rebased tree, not a re-assertion of 2,397.

Cyrax321 (changes requested):
Author context: Adhi1-2, external contributor. Follow-up to the prior review, and the blocker still stands. The guard design is still the right shape, but the branch is stale: this checkout has README.md at 2,452 and this PR's docs/index.html still at 2,163, so the drift has grown since the 2,397 vs 2,402 mismatch in the last review. Merge state is DIRTY. Rebase onto current main, re-measure the canonical figures from the parser on the rebased tree, and resync the page to those numbers rather than re-asserting the old ones.

Cyrax321 (commented):
Resyncs docs/index.html to 46 CLI commands and 2,467 tests, adds the page to the counted files in test_docs_counts.py, and adds a parser-vs-page guard in test_cli_docs.py. The claims check out: build_parser() yields 46 subcommands on main, the MCP count holds at 12 @server.tool registrations, CI runs the suite unfiltered so test_documented_count_matches_suite verifies 2,467 against live collection within tolerance, and every counted doc now states the same figure. Nothing blocks the merge; specifics inline. One note for the record: the PR body still describes 2,397 as the canonical figure, but the diff lands 2,467, so the body reads stale after the retarget and is worth refreshing for an accurate history.

tests/test_docs_counts.py
```diff
@@ -67,6 +70,15 @@
         re.IGNORECASE,
     ),
     re.compile(r"^pytest\s+-q\s+#.*?([\d,]+)", re.MULTILINE),
+    # docs/index.html is not markdown, so none of the prose forms above reach
+    # it. It states the figure three times -- the meta description, the hero
+    # banner, and the metrics card (#1283). Uppercase is deliberate: the master
```
Cyrax321, line 75: Non-blocking. The stated mechanism cannot fire: CONTINUUM_MASTER_PLAN.md is not in COUNTED_FILES, and documented_total only ever reads counted files, so a case-insensitive flag on this pattern could never let the master plan veto anything. The real reason uppercase matters is that a counted file (a future README paragraph, say) might write 'N tests passing' in lowercase prose with some other number, and case sensitivity keeps the guard anchored to the page's all-caps label. Worth correcting the comment, because the next person who touches these patterns will reason from what it claims.

tests/test_docs_counts.py
```diff
@@ -124,6 +136,10 @@ def test_documented_narrative_count_forms(tmp_path: Path) -> None:
         ("Validado con muertes reales y 1380 tests.", 1380),
         ("Validado con muertes reales y ~2,241 tests.", 2241),
         ("cerca de 2,311 tests", 2311),
+        # docs/index.html: meta description, hero banner, metrics card (#1283).
+        ("12 MCP tools, 46 CLI commands, 2,397 tests.", 2397),
```
Cyrax321, line 140: Non-blocking. These synthetic cases work mechanically with any constant, but 2,397 is a figure that never existed in the repo: the page said 2,163, the markdown corpus said 2,452, and this PR lands 2,467. The PR body still describes 2,397 as the canonical total, so the test example and the body now echo a number with no history behind it. Using either the pre-PR or the landed figure keeps the test consistent with the record for the next resync.

docs/index.html
```diff
@@ -224,7 +224,7 @@ <h1 class="pixel-headline" aria-label="Agents that survive">
 
     <!-- Metrics refresh note: TESTS from `pytest --collect-only -q` (README canonical figure), MCP TOOLS from the @server.tool count in src/continuum/mcp/server.py, CLI COMMANDS from len(build_parser() choices). -->
     <div class="bottom-copy bottom-left">
-      <div>2,163 TESTS PASSING</div>
+      <div>2,467 TESTS PASSING</div>
```
Cyrax321: Non-blocking. The guard this PR adds pins this figure to the docs corpus's collected total (the refresh note above names pytest --collect-only as the source), and the collected count includes the roughly 28 to 41 skipped tests, so the literal reading '2,467 TESTS PASSING' overstates what the number measures. The convention predates this change, but since this PR is specifically about the accuracy of the headline figures, either stating the passing figure or relabeling to 'TESTS COLLECTED' would make the claim literally true.

Adhi1-2:
Rebased onto the current main and re-measured on the rebased tree, as the follow-up review asked: the page's three test-count sites now read 2,452, matching the README canonical figure rather than the 2,163 the branch had drifted to, and its two CLI-command sites read 46, matching the parser.

The guards the PR adds compare the page against the parser and against the README rather than pinning a literal, so a future maintainer resyncing the counts updates the page and the docs together. I verified both guards still fail with `docs/index.html` reverted to the stale figures, so they are load-bearing.

All 31 checks green including the three Windows jobs, codecov/patch, and anya-review. The rebase dismissed the prior approval, so this needs a fresh review rather than a merge as-is.

PR #1318, opened 2026-09-20
fix(docs): sync CITATION.cff with 0.1.2 and guard it (#1120)

Adhi1-2:
## What

`CITATION.cff` pinned `version: 0.1.0` while the package was `0.1.2`. Anyone citing the project got a version two releases stale, and nothing noticed because no guard read the file.

`CONTRIBUTING.md` also misdescribed the release process: it said the version lives in **two** places and that CI "will fail if they diverge **once version-checking is added**". Version-checking already exists and already runs in CI — `tests/test_version_drift.py` (the #838 guard) enforces the `__version__`/pyproject agreement, both README pins, and the release tag. A contributor following the documented paragraph and touching only the two named sites got a red CI for a reason CONTRIBUTING never explained.

## Changes

- `CITATION.cff` — `0.1.0` → `0.1.2`.
- `CONTRIBUTING.md` — the Versioning paragraph now names all **five** sites a bump must move in one PR (pyproject, `__init__.py`, both README pins, CITATION.cff, the release tag) and drops the "once version-checking is added" clause.
- `tests/test_version_drift.py` — new `test_citation_file_matches_pyproject`, modelled on the existing `test_readme_pins_match_pyproject`, so the citation file cannot drift again silently.

## Verification

- `ruff check` / `ruff format --check` on `src/ tests/ examples/` — clean.
- `mypy src/continuum` — clean (128 files).
- `markdownlint-cli2` on CHANGELOG.md and CONTRIBUTING.md — clean.
- New test proven load-bearing: with `CITATION.cff` reverted it fails `CITATION.cff says 0.1.0 but the package is 0.1.2`.
- Full suite: one failure, `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend` — pre-existing and environmental (this machine has `psycopg` installed, so the "driver absent" `RuntimeError` it expects never raises). Reproduced on an unmodified tree.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the software version in citation metadata to `0.1.2`.
  * Added validation to detect mismatched versions across project metadata, citation information, and documentation.

* **Documentation**
  * Updated contributor guidance to identify all locations requiring version updates, including release tags.
  * Clarified how automated checks enforce version consistency during continuous integration and local release verification.


Cyrax321 (dismissed):

Ref: #1318 / Issue #1120
Surface: CITATION.cff, CONTRIBUTING.md release documentation, version guard

Bug: CITATION.cff pinned 0.1.0 while the package shipped 0.1.2, so every citation was two releases stale, and CONTRIBUTING.md told contributors the version lives in two places and that CI would fail on divergence once version-checking is added, when the #838 guard already runs in CI and covers more sites than the two named. A contributor following the documented paragraph got a red build for a reason the docs never mentioned.

Fix verified at head b4f6c727: CITATION.cff now reads 0.1.2, CONTRIBUTING describes the actual release sites including the guard, and the drift test now covers the citation file so it cannot go stale silently again. Ran tests/test_version_drift.py locally: 3 passed, 1 skipped (the tag case, needs a release context).

Proof: no U+2014 in the added lines; CI green on the head commit.

One documentation observation, non-blocking: #1317 (fix/docs landing-page-metrics-guard) also touches version and metrics guarding on the landing page. If both land, give CONTRIBUTING a one-line read to make sure the two descriptions still agree.

PR #1320, opened 2026-09-20
feat(storage): out-of-band blob storage for oversized payloads (#254)

### LINKED ISSUE #254: Large-payload offloading: blob storage for oversized event payloads
### Affected component

Core (events, storage)

Split from #239. Event payloads above a configurable size (gateway evidence bodies, future reasoning summaries, OTel attribute bundles) would bloat both the live log and the archive. Durable-execution platforms solve this with payload codecs: store the bytes out-of-band and keep a digest + reference inline (Temporal's large-payload codec is the reference pattern).

### Proposed solution

1. Threshold constant + config key (`CONTINUUM_PAYLOAD_OFFLOAD_BYTES`, default off).
2. On append, payloads whose JSON serialization exceeds the threshold are written to `<db>.blobs/<sha256>` and replaced inline with `{"__offloaded": sha256, "keys": [...original keys...], "bytes": n}`.
3. `read_events` rehydrates transparently when the blob file exists; missing blob raises CorruptedRecord naming the digest (never silent).
4. Compaction moves blobs alongside their rows (archive keeps working; deletion policy stays operator-owned).
5. verify gains blob-integrity checking in `--deep` mode.

### Alternatives considered

- Compressing inline instead (zlib+base64): helps ~60% but unbounded; does not protect replay latency.
- Refusing oversized payloads outright: pushes the problem onto every adapter.

No

Yes, with guidance

Adhi1-2:
Closes #254.

## What

Event payloads above `CONTINUUM_PAYLOAD_OFFLOAD_BYTES` are written to a
content-addressed `<db>.blobs/<sha256>` beside the database and replaced inline
with a small marker carrying the digest:

```json
{"__offloaded": "<sha256>", "keys": ["gateway", "evidence"], "bytes": 18342}
```

The live log and the archive stop bloating with gateway evidence bodies,
reasoning summaries and OTel attribute bundles, so replay latency stays
bounded no matter how large one payload grows. This is the payload codec
durable-execution platforms use (Temporal's large-payload codec is the
reference): keep a digest inline, store the bytes out of band.

## The five deliverables in #254

1. **Threshold + config key.** `CONTINUUM_PAYLOAD_OFFLOAD_BYTES`, read at
   engine construction. Zero, unset, negative, or unparseable all mean off —
   a typo in the key must not look like the feature working while it does
   nothing.
2. **Offload on append.** `BlobStore.encode` serializes the payload, and over
   the threshold writes the blob **first** and the marker second: a crash
   between the two leaves an orphaned blob (collectable disk waste) rather
   than a row pointing at a blob that was never written, which is unreadable.
   Written via tmp file + `os.replace`, so a reader never observes a
   half-written blob.
3. **Transparent rehydration.** One seam — `_row_to_event` — so `read_events`,
   `read_archived_events`, `read_all_events` and the chain audit all restore
   the original payload with no caller knowing the codec exists. A missing
   blob raises `CorruptedRecord` naming the digest, never a silent empty
   payload: a resume that had lost the event describing what it did would be
   worse than a loud failure.
4. **Compaction.** No bytes move. The digest is content-addressed, so the
   archived row keeps the marker and points at the same file the live row did;
   a blob deleted after compaction is still refused, because the archived
   event still needs it. Nothing here deletes blobs — reclaiming space is
   operator-owned, same as the archive itself, because a blob is the only copy
   of a payload and a reachability scan would be wrong the moment a new event
   with the same content arrives.
5. **`verify --deep`.** Audits every referenced blob across both tables and
   reports `BLOB_MISSING` / `BLOB_DIGEST_MISMATCH` naming the digest, instead
   of the chain walk's generic unreadable-record finding. Reads the raw
   payload column rather than going through rehydration, so the damage is
   attributable before any repair. A no-op on stores the feature never
   touched, so the flag is always safe to pass.

## Off means byte-identical

The feature is off by default, and off means the serialized form a row stores
is exactly what it stored before this PR existed. Two properties make that
safe rather than incidental:

- The chain still verifies, because the event is sealed over the original
  in-memory payload and rehydration restores it — the digest is never part of
  what gets hashed.
- `decode` only interprets a marker while the codec is enabled, and the
  marker is an **exact** key set (`is_marker`), shared by the read path and
  the audit so they cannot drift. A payload that legitimately uses the
  reserved keys round-trips on a store that never opted in, and a near-miss
  (`__offloaded` plus a fourth key) stays inline instead of being reported as
  a blob that never existed.

In-memory databases never offload — there is no file to sit beside, and an
in-memory database is not durable in the first place.

The action index rehydrates too (`_canonical_index_rows`), or an offloaded
`ACTION_*` payload would be the marker dict, carry no `action_type`, and
silently drop from the index while the ledger still folds it.

Postgres keeps payloads inline and accepts `--deep` as a documented no-op:
TOAST already moves oversized values out of the main heap transparently, so
there is no out-of-band file to audit and no digest for a row to lose.

## Tests

`tests/test_payload_offload.py`, 24 tests, covering: config on/off and
malformed values; in-memory no-op; byte-identical rows with the codec off; a
marker-shaped payload round-tripping when off and a near-miss staying inline;
the roundtrip through a blob; deduplication by digest; the hash chain
verifying across an offloaded payload; the `append_sealed` path; missing and
truncated blobs refused by digest; compaction keeping the payload readable
with no bytes moved and the ledger fold intact; deep verify passing, naming a
missing blob, naming a tampered blob, and reaching into the archive; and the
CLI `--deep` surface both ways.

## Quality gates

- `ruff check` / `ruff format --check` — clean (0.16.5, the pinned version)
- `mypy` strict on `src/continuum` — clean, 129 files
- Full suite: **2396 passed, 26 skipped**
- Docs-count guard (`-m slow`) passes; documented total re-synced to ~2,421
  across README, the five translated READMEs, CHANGELOG, and the references,
  preserving the pre-existing offset to the live collect

## Not in scope

`docs/index.html`'s meta description still says 2,163 tests. That is stale by
~250 tests from earlier PRs and is not covered by the docs-count guard, so it
is left alone rather than folded into an unrelated change.

## Summary by CodeRabbit

- **New Features**
  - SQLite can optionally offload oversized payloads to content-addressed blob storage using `CONTINUUM_PAYLOAD_OFFLOAD_BYTES`.
  - Offloaded payloads are transparently restored when read and deduplicated by content.
  - Added `continuum verify --deep` to check referenced blobs and report missing or corrupted data.

- **Documentation**
  - Documented payload offloading and deep verification.
  - Updated documented test-suite counts to approximately 2,421 tests.


Cyrax321 (changes requested):
Author context: Adhi1-2, external contributor. This is a large feature and the core design is sound: content-addressed blobs beside the DB, write-then-rename so a crash orphans rather than dangles, inline format byte-identical while off so the hash chain is untouched, Postgres explicitly out of scope with TOAST as the reason. CI is green. Two blocking defects keep it from merging, both verified in the diff. 1. Disabling the codec after offload silently returns markers. decode returns the raw marker unchanged when not enabled (payload.py decode gate), and _audit_blobs early-returns on not enabled, so unsetting CONTINUUM_PAYLOAD_OFFLOAD_BYTES after offloading turns every offloaded read into the wrong payload with no error and a clean deep audit. The read path must not be gated on the write threshold. 2. An enabled store misreads a legitimate small payload whose keys are exactly {__offloaded, keys, bytes}. is_marker matches on exact key set, so decode treats the inline payload as a blob reference and either raises CorruptedRecord or, on digest collision, returns the wrong payload. The disabled-codec test does not cover the enabled case. Fix both, then this is approvable. Minor: concurrent same-digest writers share one .tmp name, so cross-process offload of identical content can promote a partial blob. Consider pid-unique staging names.

src/continuum/storage/payload.py
```diff
@@ -0,0 +1,220 @@
...
+
+        A payload that is not a marker is returned unchanged, so every reader
+        works on un-offloaded and offloaded stores alike. Off entirely, no
+        marker is ever interpreted: a store that never opted in keeps
+        round-tripping a payload that happens to use the marker keys. A marker
+        whose blob cannot be read raises ``CorruptedRecord`` naming the
+        digest, because returning an empty payload instead would let a run
+        resume having silently lost the event that described what it did.
+        """
+        if not self.enabled or not is_marker(payload):
```
Cyrax321, line 182: Blocking. Gating `decode` on `enabled` ties rehydration to the threshold set in this process's environment at open time. A database written with the codec on and later opened without `CONTINUUM_PAYLOAD_OFFLOAD_BYTES` (a new shell, a cron job, a different operator) has `enabled == False`, so every offloaded row comes back from `read_events`, `read_archived_events` and `read_all_events` with the marker dict `{"__offloaded": ..., "keys": ..., "bytes": ...}` as its payload, silently. That is exactly the failure mode this module's docstring says must never happen: a resume proceeds having lost the event that described what it did, and the ledger fold and the action index rebuild drop those events' actions without a word. A marker sitting in the row is itself proof a blob exists, so `decode` should fire whenever `self._directory is not None`, and `enabled` (threshold > 0) should gate `encode` only.

src/continuum/storage/sqlite.py
```diff
@@ -843,9 +892,62 @@ def verify_events(self, run_id: str) -> IntegrityReport:
...
+        """Check every blob a run's payloads reference (issue #254).
+
+        Reads the payload column of both tables and audits each marker's
+        digest. A run with no offloaded payloads reports nothing, so the deep
+        pass is a no-op on stores the feature never touched. Compaction needs
+        no separate handling: an archived row keeps the marker verbatim, and
+        the blob is content-addressed, so the archive references the same file
+        the live row did.
+        """
+        if not self._blobs.enabled:
```
Cyrax321, line 905: Blocking. Same gating problem one level up: `_audit_blobs` returns early when the current process has no threshold set, so `continuum verify --deep` on a store that plainly has blobs on disk but was opened without `CONTINUUM_PAYLOAD_OFFLOAD_BYTES` audits nothing and reports nothing. The chain walk then flags each offloaded event as a generic hash mismatch, because its payload came back as the marker dict per the `decode` gating, which misdirects the operator toward chain tampering when the real story is an unread blob. The audit should scan for markers whenever the store has a blob directory, independent of the environment.

src/continuum/storage/payload.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def is_marker(payload: Any) -> bool:
+    """Whether ``payload`` is the inline reference ``encode`` writes.
+
+    Exact key set, not a substring test: a real payload that happens to use the
+    marker name alongside other keys must not be read as a blob reference,
+    either on rehydration or in a deep audit. Centralised so the read path and
+    the audit path cannot drift apart on what counts as a marker.
+    """
```
Cyrax321, line 78: Blocking. `is_marker` tests key shape only, so with the codec on, a legitimate under-threshold payload that happens to use exactly the keys `__offloaded`, `keys` and `bytes` (and nothing else) is stored inline by `encode` but read by `decode` as a blob reference, raising `CorruptedRecord` for a blob that was never written. The near-miss test covers a payload with a fourth key and the marker round-trip test covers the codec being off, but nothing covers the exact key set with the codec on, and such a row is unreadable through every path while it is on. Validating the marker's values too (a 64-character lowercase hex digest, an int `bytes`, a list `keys`) would make the format self-describing enough that ordinary content cannot collide with it.

src/continuum/storage/payload.py
```diff
@@ -0,0 +1,220 @@
...
+        works on un-offloaded and offloaded stores alike. Off entirely, no
+        marker is ever interpreted: a store that never opted in keeps
+        round-tripping a payload that happens to use the marker keys. A marker
+        whose blob cannot be read raises ``CorruptedRecord`` naming the
+        digest, because returning an empty payload instead would let a run
+        resume having silently lost the event that described what it did.
+        """
+        if not self.enabled or not is_marker(payload):
+            return payload
+        digest = str(payload[OFFLOAD_MARKER])
```
Cyrax321, line 184: Non-blocking. The digest flows from the row straight into a path join with no validation, so a marker value like `../../other` makes `decode` and `audit` operate on files outside the blob directory: decode would parse and return any JSON file it can reach, and audit would hash any file on disk. Rows are trusted content, so this is hardening rather than a live attack, but pinning the digest shape here (64 lowercase hex chars) is one line and would also make the marker unforgeable by ordinary payloads, per the `is_marker` comment.

src/continuum/cli/main.py
```diff
@@ -3103,10 +3103,15 @@ def cmd_verify(args: argparse.Namespace, storage: Storage, out: Any, err: Any) -
...
-    report = storage.verify_events(args.run_id)
+    report = storage.verify_events(args.run_id, deep=bool(getattr(args, "deep", False)))
     payload = report.model_dump(mode="json")
     if report.ok:
         text = f"Event chain verified: {report.checked} events, no violations."
+        # A deep pass over a store with the codec off reports nothing, so the
+        # line tells an operator the extra check actually ran rather than
+        # letting an unset threshold read as "all blobs present".
+        if getattr(args, "deep", False):
+            text += " Offloaded payloads checked: all present."
```
Cyrax321, line 3114: Non-blocking. This line prints whenever `--deep` is passed and the chain verifies, including when the codec is off in this process and `_audit_blobs` returned without reading a single blob. The comment above says the line exists so an operator knows the extra check actually ran, but in that case it did not, and 'all present' claims a result the audit never produced. Conditioning the message on the store having blobs enabled, or reporting the number of blobs audited, would keep the line honest.

Cyrax321 (changes requested):
This lands the #254 codec at the right seam: one encode at insert, one decode at _row_to_event, content-addressed blobs beside the database, compaction moving no bytes, and a deep audit that names damaged blobs by digest. The matrix is green and the off-by-default path is byte-identical, which the tests pin well.

Two things block. First, the read and audit paths gate on whether the codec is enabled in the current process rather than on whether the store actually contains markers, so a database written with the codec on and reopened without CONTINUUM_PAYLOAD_OFFLOAD_BYTES silently returns marker dicts as payloads and verify --deep audits nothing. Second, the key-shape-only marker test makes a legitimate payload with exactly the reserved keys unreadable while the codec is on.

Two smaller hardening items are inline and non-blocking.

Issue #1322, opened 2026-09-20
action_index_drift reports false positives once a run is compacted in a multi-run store (both engines)


Adhi1-2:
## Summary

`action_index_drift()` compares the `action_index` projection against a canonical fold of the event log. Both engines fold the archive and the live log as two **disjoint segments**: everything in `events_archive` is numbered before everything in `events`. That is correct *within* a run (an archived prefix genuinely predates the live tail), but it is not global insertion order once a second run exists. When run B is compacted, B's action events move to the archive and are renumbered below run A's still-live action events — even though B wrote them later.

The fold therefore stops reproducing the number the incremental writer stored, and a store nothing has tampered with reads as dirty.

## Repro

Two runs, both with a claimed+completed action; compact the *later* one.

```python
store.create_run_started(Run(run_id="early", goal="e"))
a = ActionLedger(store, "early")
k1 = a.claim("send_invoice", {}, key="inv:first", scoped_to_run=False).key
a.complete(k1, external_id="E1")          # early wrote first
store.create_run_started(Run(run_id="late", goal="l"))
b = ActionLedger(store, "late")
k2 = b.claim("send_invoice", {}, key="inv:second", scoped_to_run=False).key
b.complete(k2, external_id="L2")          # late wrote second
store.action_index_drift()                # -> 0
store.compact_run("late")                 # late's rows -> events_archive
store.action_index_drift()                # -> 2  (Postgres) / 1  (SQLite)
```

Postgres, after the compaction — the two keys' orders are transposed:

```
canonical order: {'cb4d512b86': 2, '82573effb8': 4}   # late=2, early=4
stored  updated : {'82573effb8': 2, 'cb4d512b86': 4}   # early=2, late=4
```

SQLite, same shape: the later write takes a negative order (`-1`) while its row stores `rowid` 6.

This is distinct from #1321 (fixed by counting action events only in the fold): #1321 fires with **no** compaction at all, this one fires **only** after a compaction in a store where another run already holds action events.

## Where it lives

- `src/continuum/storage/sqlite.py` `_canonical_index_rows` — `[*archived, *rows]`, live rows keep `rowid`, archived rows get negative positions.
- `src/continuum/storage/postgres.py` `_canonical_index_rows` — same merge, both `ORDER BY ctid`.

## Impact

[...]

PR #1323, opened 2026-09-20
fix(postgres): fold counts action events so the index stops reading dirty (#1321)

### LINKED ISSUE #1321: storage: PostgresStorage.action_index_drift() is a permanent false positive whenever a non-action event precedes an action, so verify reports a dirty index on every healthy Postgres store
## Problem

`PostgresStorage` numbers the action index two incompatible ways.

- The incremental writer (`_maintain_action_index`, `postgres.py:561`) stores `updated_seq = nextval('action_index_ord_seq')`. The sequence advances **once per action event**, so its value is `1 + (number of action events appended so far)`.
- The canonical fold (`_canonical_index_rows`, `postgres.py:726`) assigns `order` as the **dense 0-based position of the row in the merged `events_archive` + `events` stream** — a count that includes every non-action row.

`action_index_drift()` compares those two numbers for equality. They count different things, so they agree only when the number of non-action rows preceding a key's last action event happens to equal the number of action events preceding it. Any ordinary run breaks that: `RUN_STARTED`, `TOOL_CALLED`, `TOOL_COMPLETED`, `EVIDENCE_ADDED`, `FINDING_ADDED` are all non-action rows, and they routinely sit between actions.

SQLite does not have this because both sides share one numbering: the incremental writer stores the event row's `rowid` and the fold reads `rowid` (`sqlite.py:475`, `:651`).

## Root cause (evidence)

Reproduced against a real Postgres 17 with `CONTINUUM_TEST_POSTGRES_DSN`:

```python
store.create_run_started(Run(run_id="run_1", goal="g"))
store.append_event("run_1", EventType.EVIDENCE_ADDED, {"evidence_id": "e1", "summary": "s"})
outcome = ledger.claim("send_invoice", {}, key="invoice:1")
ledger.complete(outcome.key, external_id="INV-1")
print(store.action_index_drift())   # 1 — on a store nothing has tampered with
```

```
healthy store, one extra NON-action event, nothing tampered with
  action_index_drift() = 1 (expected 0)
  stored  : {2d887ace: (2, 'completed')}
  canonical: {2d887ace: (3, 'completed')}
```

`stored.updated_seq = 2` (the second `nextval` consumed: ACTION_RECORDED, then the settle event). The canonical fold puts the same row at merged position 3 (`RUN_STARTED`, `EVIDENCE_ADDED`, `ACTION_RECORDED`, settle). Nothing is wrong with the index; the two sides simply count different populations.

Remove the `EVIDENCE_ADDED` line and `drift()` returns 0 — which is why this survives: the existing guard (`tests/test_storage_postgres.py:296`) asserts `drift() > 0` after a compaction and `== 0` after a rebuild, and both hold for reasons that have nothing to do with the check being correct. The `> 0` step is satisfied by the permanent mismatch, and the `== 0` step passes only because no event is appended between the rebuild and the re-check, so no `nextval` is consumed. The identical SQLite repro returns 0.

## Impact

- **`continuum verify` reports a dirty action index on essentially every real Postgres store.** A run that records any tool call, evidence, or finding between two actions lands on the mismatched side.
- **`verify --repair-index` "repairs" a healthy index and reports spurious corrections, then the drift returns on the next appended action.** This interacts directly with #1267: once rebuild reports the rows it corrected, that count becomes permanently non-zero on Postgres for no reason.
- **`foreign_action` can mis-rank after a rebuild.** It picks the newest writer with `ORDER BY updated_seq DESC`. After a rebuild, live rows carry dense merged positions (bounded by total row count) while the next appended action carries `nextval` (bounded by action-event count). When non-action rows outnumber action events, the fresh event gets a *lower* `updated_seq` than the pre-existing rows it is newer than, so a cross-run `mem:` lookup can rank a stale write as the current one — the same last-write-per-key hazard #1054 fixed on the SQLite side.

Severity is moderate: it does not corrupt the ledger or the log, and it fails loudly rather than silently, but it makes the verify command unusable as a signal on the Postgres backend and trains operators to ignore a check that exists to be trusted.

## Proposed solution

Make the canonical fold reproduce the numbering the incremental writer uses, instead of inventing a second one. Since `nextval` is consumed exactly once per action event, in global insertion order, the fold reaches the same value by counting **action events only**, 1-based, over the archive-then-live stream:

```python
canonical: dict[str, tuple[...]] = {}
order = 0
for row in [*archived, *rows]:          # both already ordered by ctid
    entry = index_entry_from_payload(EventType(row["type"]), payload)
    if entry is None:
        continue                        # non-action rows consume no nextval
    order += 1
    canonical[entry[0]] = (entry, order)
```

This also removes the negative-offset trick: with true global numbering an archived action event's value is necessarily below a later live one's, so last-write-per-key holds without it (the archive-predates-live assumption the current docstring relies on is unchanged). The SQLite engine is already correct and needs no change.

Worth confirming in the same PR: whether `foreign_action`'s `ORDER BY updated_seq DESC` still ranks correctly across a rebuild boundary, or whether the numbering should be made rebuild-proof rather than fold-reproducible.

## Test plan

- Append a non-action event between two action events on Postgres, assert `action_index_drift() == 0` before any rebuild (the case that fails today).
- Assert the same on SQLite as the unaffected regression guard.
- Extend the existing compaction guard so its `> 0` step is produced by *real* drift (tamper with the index row), not by the numbering mismatch — otherwise the fix can pass while both comparisons remain meaningless.
- Assert a fresh append after a rebuild keeps `drift() == 0`.

## Platform coverage

Not platform-specific; requires a live Postgres (`CONTINUUM_TEST_POSTGRES_DSN`), as the existing Postgres suite does.

## Dependencies

None, but it should land with or after #1267: that issue makes rebuild report its corrections, and on an unrepaired Postgres backend that count would be spurious on every store.

## Contributor notes

- `psycopg` is an optional extra; the Postgres suite skips cleanly without `CONTINUUM_TEST_POSTGRES_DSN`.
- A local Postgres on `/tmp:5432` works with `CONTINUUM_TEST_POSTGRES_DSN='dbname=<scratch> host=/tmp'`; drop and recreate the scratch database between runs, the suite shares one database.

Adhi1-2:
Closes #1321.

## The bug

`continuum verify --index` reported a corrupted action index on a Postgres store nothing had tampered with, and `--repair-index` did not help.

`action_index_drift` compares the projection against a canonical fold of the log, and the two sides numbered each row on different scales:

- `_maintain_action_index` stores `nextval('action_index_ord_seq')`, which advances **once per action event**.
- `_canonical_index_rows` numbered a row by its **position in the merged row stream**, which counts `RUN_STARTED`, `TOOL_CALLED`, `EVIDENCE_ADDED` and every other non-action row too.

An ordinary run has non-action rows between its actions, so the two disagreed by one per intervening row. A rebuild was no repair: it rewrote the rows onto the fold's scale, and the next appended action took the next `nextval` and put them straight back out of step.

## The fix

One function. The fold counts **action events only, 1-based** — exactly the number the sequence assigned, because a non-action row consumes no `nextval` and moves no position.

## Why SQLite is immune (and now pinned)

SQLite numbers both sides with the writing event's `rowid`, so intervening rows move both figures together. `test_a_healthy_store_with_non_action_rows_between_actions_reports_no_drift` pins that agreement, so a refactor that switches the SQLite fold to count-only — which is what Postgres needs — is caught rather than silently desynchronising the two scales.

## Verification

- Both new Postgres tests fail with the fix reverted (`drift == 1`) and pass with it.
- Full CI command set green in a fresh shared database: `pytest tests/test_storage_postgres.py tests/test_action_index.py` → 37 passed.
- `ruff check` / `ruff format --check` / `mypy src/continuum` clean; local ruff matches the pinned `0.16.5`.
- Full default suite green except the known pre-existing `test_postgres_url_is_routed_to_postgres_backend`, which expects `psycopg` to be absent and fails on any machine that has it installed.

## Two test-surface notes

**A test bug found along the way.** The `DELETE` in `test_pg_deleted_boundary_event_fails_verify` was not scoped by `run_id`:

```python
"DELETE FROM events WHERE sequence ="
" (SELECT MIN(sequence) FROM events WHERE run_id = 'pg_kb')"
```

`sequence` is per-run, so this deleted that row number from **every** run in the shared database and silently corrupted whichever action event happened to land on it. It has been silently corrupting the shared suite's state; nothing earlier asserted on cross-run state after it, so it went unnoticed. Scoped to `pg_kb`, with a comment on why the restriction is load-bearing.

**Test isolation.** Two of the new tests assert a *store-wide* invariant — the fold's count against the sequence — so they run against a database the test owns exclusively (`isolated_storage`). In the shared suite database another test's compaction moves rows between `events` and `events_archive` and the fold's merged order stops tracking the sequence, which would make a clean store read dirty for reasons these tests do not exercise.

## Not fixed here: #1322

While proving this fix I found a distinct bug in the same family and filed it as #1322 rather than bundling it. Both engines fold the archive and the live log as disjoint segments — everything archived is numbered before everything live — which is correct *within* a run but not across runs. Compacting one run in a multi-run store renumbers its action events below another run's earlier live ones, so the fold again stops reproducing the writer's numbers and drift reports a false positive. It was unreachable before only because the dense scheme was already wrong on every store.

I did not start on it; the issue records the two candidate fixes (preserve a global insertion-order column through the archive, or stop comparing the number now that the projection is one row per key) so the choice is decided before anyone works on it.


tests/test_storage_postgres.py
```diff
@@ -293,10 +302,114 @@ def test_pg_action_index_covers_the_archive_after_rebuild(
...
+    name = f"iso_{os.getpid()}_{next(_iso_counter)}"
+    admin = psycopg.connect(DSN, autocommit=True)
+    try:
+        admin.execute(f'CREATE DATABASE "{name}"')
+    finally:
+        admin.close()
+    params = conninfo_to_dict(DSN)
+    params["dbname"] = name
+    store = PostgresStorage(make_conninfo(**params))
+    yield store
```
Cyrax321, line 378: Non-blocking, and the same fixture appears in #1325, so whichever lands first carries it. The teardown is not in a try/finally, so a failing assertion in either test that uses this skips store.close() and the DROP, and database iso_<pid>_N stays on the Postgres host. Combined with CREATE DATABASE without IF NOT EXISTS and a pid-based name, a later run that recycles the pid then errors out in setup on duplicate_database, on a test that did nothing wrong. A uuid4 suffix, DROP IF EXISTS before CREATE, and a finally that always drops closes it. Harmless in CI's fresh container, accumulates on a persistent dev Postgres.

tests/test_action_index.py
```diff
@@ -175,6 +175,28 @@ def test_spurious_rows_count_as_drift_and_are_removed(store: SQLiteStorage) -> N
...
+    store: SQLiteStorage,
+) -> None:
+    """The canonical fold must number a row as the incremental writer numbered it.
+
+    A run records plenty of non-action events between its actions -- tool
+    calls, evidence, findings. SQLite is immune to #1321 because both the
+    incremental maintenance and the fold use the writing event's ``rowid``,
+    so intervening rows move both figures together. This pins that agreement
+    so a refactor that switches the SQLite fold to counting action events
+    only -- which is what the Postgres engine needs -- is caught here rather
```
Cyrax321, line 188: Non-blocking. This test cannot fail on this PR's change: SQLite's fold is untouched, so both sides still use the writing event's rowid and the assertion holds before and after. It is a guard against a future refactor this PR does not make, which is a legitimate purpose, but the CHANGELOG's "a regression test now pins that agreement" reads as if it were a regression test for the fix. The two Postgres tests in test_storage_postgres.py are the ones that fail with the fix reverted; consider naming this one guard_ or noting it in the body so the coverage story is unambiguous. (Also: "rather than than" at the next line.)

tests/test_storage_postgres.py
```diff
@@ -293,10 +302,114 @@ def test_pg_action_index_covers_the_archive_after_rebuild(
     ledger.complete(outcome.key, external_id="doc:1")
     storage.compact_run("pg_ki")
 
+    # Compaction moves the rows the index describes into events_archive, and
+    # the fold merges that table ahead of every live row, so the projection
+    # is reported dirty until it is rebuilt. That desync is the archive/live
+    # ordering gap, #1322, which is separate from #1321 (a healthy store
+    # reported dirty with no compaction at all).
     assert storage.action_index_drift() > 0
```
Cyrax321, line 345: Non-blocking, but the guarantee behind this assertion changed. Under the old dense scheme the fold counted every row, so any archived action disagreed with its stored nextval structurally and this was always > 0. Under the count-based fold the disagreement depends on other runs' action events sitting earlier in the shared database than pg_ki's archived ones, which is true in this suite but is environmental rather than structural.

Not worth restructuring if this PR is a step towards #1322, but if this test stays asserting > 0 it is the one place the suite depends on the shared database's history rather than on the store it creates.

Cyrax321 (commented):
This is the narrowest of the three takes on this defect family, and the mechanism is right: making the fold count action events only means its number is literally the Nth value the sequence handed out, so drift compares two figures on the same scale and a rebuild writes numbers the next nextval continues rather than contradicts. I verified the sequence starts at the default 1 and that _backfill_action_index already numbers by nextval in ctid order, so the fold now agrees with the other writer path too.

No blocking issues with the product change. The notes below are test hygiene.

The decision worth making before merging any of these is which one you want. #1324 fixes this case and #1322 with one shared helper in actionindex.py, and stops comparing the number altogether rather than reconciling two numbering schemes. This one leaves a compacted store reading dirty until it is rebuilt, because the archive-first merge still assigns an archived key a count that is not the nextval it actually received. That is #1322, honestly disclosed here and not claimed fixed. If you prefer the reconciliation approach over dropping the comparison, this is the correct shape for it and #1324's helper would be the wrong abstraction; if you want one PR for both issues, take #1324 and close this and #1325.

Cyrax321 (dismissed):

Ref: #1323 / Issue #1321
Surface: Postgres action index drift computation, verify --index, repair path

Bug: the two sides of action_index_drift numbered rows on different scales. The incremental maintainer takes nextval on a sequence that advances once per action event, while the canonical fold numbered a row by its position in the merged row stream, counting every non-action event too. An ordinary run therefore drifted by one per intervening row: verify --index reported corruption on a store nothing touched, and --repair-index could not fix it because the rebuild wrote the fold's scale and the next append put them straight back out of step.

Fix reviewed at head 019888f7: the fold now numbers action events only, 1-based, which is exactly the scale the sequence produces, so the comparison is between like units and drift returns to meaning actual divergence. SQLite needs no change since both of its sides use the writing event's rowid, and the reasoning for that asymmetry is stated.

Proof:
- Ran tests/test_action_index.py locally: 16 passed.
- The new Postgres regression tests pin agreement between the incremental scale and the fold's scale across interleaved non-action events; the Postgres CI job is green on the head commit.
- The only failing check is codecov/patch, advisory patch coverage, not a required test.
- No U+2014 in the added lines.

Merge-order note: this builds on the fold-ordering fix in #1261 (same function, _canonical_index_rows). Land #1261 first, then this rebases cleanly; #1324 and #1325 in this same cluster depend on the corrected drift semantics.

Adhi1-2:
Reopening this: I closed it at 09:36 with no explanation, about ninety seconds after an unrelated branch push re-referenced it. There was no reason to close it and I have reverted the close. #1321 is still open and this is still its PR.

State of the branch: rebased onto current main (`13c2fab`), flattened to one commit, `5a57c33`. Gates green locally except the Postgres suite, which needs Docker I do not have on this machine, so the `test-postgres` job here is the real evidence for the fix on the engine it targets.

One thing to be straight about: my 08:11 push invalidated the earlier approval, so this needs a fresh approving review rather than a merge as-is.

Adhi1-2:
Rebased onto current main; all checks green except codecov/patch, which is the structural artifact described above (the new Postgres lines only run under CONTINUUM_TEST_POSTGRES_DSN, set in the test-postgres job, which uploads no coverage). The Test (Postgres backend) job passes, so the fix is exercised on the engine it targets.

Cyrax321 (approved):
Author context: Adhi1-2, external contributor. Verified the mechanism against the base: _maintain_action_index stores nextval from action_index_ord_seq which advances once per action event, while the base fold numbers by merged row-stream position, so every intervening non-action row shifts the comparison by one. The patch counts action events only, 1-based, which is the same scale the sequence produces, and skips non-action rows that consumed no nextval. The SQLite parity test is a useful guard. test_action_index.py passes locally (16 passed). Two notes, neither blocking: merge state is DIRTY so a rebase is required, and the compacted cross-run ordering gap (#1322) is honestly left open. The codecov/patch failure is the known uploading-leg gap fixed by #1330, not missing tests. Approving the product change.

src/continuum/storage/postgres.py
```diff
@@ -749,14 +763,16 @@ def _canonical_index_rows(self) -> dict[str, tuple[tuple[str, str, str, str, str
                 "SELECT type, payload FROM events ORDER BY ctid"
             ).fetchall()
         canonical: dict[str, tuple[tuple[str, str, str, str, str], int]] = {}
-        offset = len(archived)
-        for i, row in enumerate([*archived, *rows]):
+        order = 0
```
Cyrax321, line 769: Non-blocking. The fold now numbers densely (1..N over action events) and rebuild_action_index writes those dense numbers into updated_seq, but nothing ever resyncs action_index_ord_seq. The connection runs in autocommit mode with _write() as only a lock, so in _maintain_action_index the event INSERT, the SELECT nextval, and the upsert are three separately committed statements: a crash between the second and third burns a nextval, and Postgres sequence values are also lost permanently on rollback. On such a store, --repair-index rewrites 1..N while the sequence still hands out N+2 or more, so the very next append reintroduces the symptom this PR fixes (drift returns after repair) and verify --index never converges for the corruption class the repair command exists for. One statement inside rebuild_action_index, after the inserts, closes it: SELECT setval('action_index_ord_seq', COALESCE((SELECT MAX(updated_seq) FROM action_index), 1)).

tests/test_storage_postgres.py
```diff
@@ -328,10 +337,114 @@ def test_pg_action_index_covers_the_archive_after_rebuild(
...
+    name = f"iso_{os.getpid()}_{next(_iso_counter)}"
+    admin = psycopg.connect(DSN, autocommit=True)
+    try:
+        admin.execute(f'CREATE DATABASE "{name}"')
+    finally:
+        admin.close()
+    params = conninfo_to_dict(DSN)
+    params["dbname"] = name
+    store = PostgresStorage(make_conninfo(**params))
+    yield store
```
Cyrax321, line 378: Non-blocking. If PostgresStorage(make_conninfo(**params)) above raises during setup, or store.close() below errors during teardown, the DROP DATABASE never runs and each failed attempt leaves an iso_* database behind on the shared test server. Wrapping the yield in a try and moving both the close and the admin drop into its finally keeps the cleanup reliable.

tests/test_action_index.py
```diff
@@ -175,6 +175,28 @@ def test_spurious_rows_count_as_drift_and_are_removed(store: SQLiteStorage) -> N
...
+    than than silently desynchronising the two numbering scales.
```
Cyrax321, line 189: Non-blocking. Docstring typo: 'rather than than silently desynchronising' duplicates 'than'.

Cyrax321 (commented):
This replaces the fold's row-position numbering with a 1-based count of action events only, which is the number action_index_ord_seq actually assigned, so a healthy store finally folds to the same figures the incremental writer stored. I traced the writer, backfill, rebuild and drift paths against the new counting and they agree, including the archive-first merge within a run, and the unscoped DELETE fix in the boundary test is correct since sequence is per-run. Nothing here blocks the merge; the one robustness gap worth a follow-up is that rebuild never resyncs the sequence, with details inline. Note that codecov/patch shows 0% only because the coverage-producing jobs never run the Postgres backend suite; it is not among the four required checks, which are all green.

anya-research (approved):
Thank you for contributing to Continuum, @Adhi1-2. This fixes the Postgres action index permanent false drift for #1321 by making the canonical fold count action events only, with a scoped DELETE fix, an isolated database fixture, a SQLite guard, and a changelog entry.

The numbering now agrees on one scale. `_maintain_action_index` in `src/continuum/storage/postgres.py` returns early on non-action events and takes `nextval('action_index_ord_seq')` only for actions, and `_canonical_index_rows` skips `None` entries and increments `order` only for actions, 1-based. SQLite stays on `rowid` for both the writer and the fold, so intervening rows move both figures together. The cross-run archive-first ordering gap stays out of scope and is tracked as #1322, with #1337 covering it separately while #1324 and #1325 are closed, so this remains the live fix for #1321.

- Test (Postgres backend): SUCCESS, plus full Test matrix (ubuntu, macos, windows across 3.11 to 3.13): SUCCESS
- Lint and Type-check: SUCCESS, Build wheel artifact: SUCCESS
- codecov/patch: FAILURE, advisory patch coverage only since the Postgres leg uploads no coverage, not a required gate
- `git diff --check`: clean, and the `run_id` scoping on the `pg_kb` DELETE matches the per-run `sequence` semantics

The honest scoping around #1322 and the isolated fixture for the store-wide invariant made the intent easy to trust. Thank you again for the careful diagnosis here.

PR #1324, opened 2026-09-20
fix(storage): action_index_drift stops reporting clean stores as dirty (#1321)

### LINKED ISSUE #1321: storage: PostgresStorage.action_index_drift() is a permanent false positive whenever a non-action event precedes an action, so verify reports a dirty index on every healthy Postgres store

### LINKED ISSUE #1322: action_index_drift reports false positives once a run is compacted in a multi-run store (both engines)

Adhi1-2:
## What

`action_index_drift()` reported a dirty action index on stores that had done nothing wrong, on **both** the SQLite and the Postgres engine. `continuum verify --index` was therefore unusable as a signal: it cried corruption on a healthy store, and the "repair" it offered could not fix it.

## Why the comparison could not be made to agree

The check compared each index row's `updated_seq` against the value a fresh fold of the log would have written. The fold and the incremental writer number rows on **different scales by design**:

1. **Non-action events.** The fold counts every row it walks; the writer consumes one sequence value per action event. So a store whose first action was not also its first event — one `TASK_UPDATED` before the first claim — reads dirty on its very first claim. This is the case issue #1321 reports.
2. **Compaction.** The fold ranks every archived row below every live one (correct *within* a run). Once a run's actions are archived, the fold puts them before another run's live actions that were written earlier, while the sequence values they were assigned say the opposite.
3. **Unrecoverable across compaction.** The `updated_seq` an archived row carries was assigned while it was still live, and nothing in the log carries it across the archive boundary. The fold cannot reproduce it at all without a schema change.

Nor could the store be repaired: `rebuild_action_index` rewrote the same numbers the fold still could not reproduce, so a compacted store read dirty again immediately after a rebuild.

## The fix

Both engines now share one helper, `index_drift_count` in `src/continuum/storage/actionindex.py`, which compares **the key set and each row's status** and deliberately ignores the sequence value.

The omission costs nothing, because the column cannot affect an answer today: `key` is the projection's primary key, so the `ORDER BY updated_seq DESC LIMIT 1` that consumes `updated_seq` in `foreign_action` is scoped to a single row and cannot rank anything. Should the schema later allow one row per write of a key, that ranking needs a numbering both sides can reconstruct — and this helper is where to put it.

Two further changes:

- **Postgres `rebuild_action_index`** hands live keys fresh sequence values in fold order instead of writing the fold's merged position, which was on a *third* scale and re-drifted the store with every action written after a rebuild. Archived keys keep the fold's negative position, which is already below every value the sequence will hand out.
- **Issue #1321 says "the SQLite engine is already correct and needs no change."** That is not so: SQLite's `action_index_drift` compared the same column against the same fold, and a compacted SQLite store read dirty for the same reason. Four of the new tests below fail on `upstream/main` before this change, three of them in the SQLite suites. Both engines share the fix.

## What the guard still catches

The check keeps its teeth. A row whose status the log no longer produces is still drift — a projection row saying `completed` when the log only records `started` is corruption an operator must hear about — and that is the failure the tests now use to prove the `> 0` branch still fires.

## Also resolves #1322

#1322 is the multi-run half of the same defect: it fires *only* after a compaction in a store where another run already holds action events, because the fold numbers every archived row below every live one — correct within a run, false across runs. It lists two candidate fixes and asks that the choice be recorded before anyone works on it.

**This PR takes candidate 2**, "stop comparing the number," and applies it to both issues rather than taking #1321's narrower proposal (count action events only in the fold). That proposal fixes the single-run, no-compaction case but not this one: the two have one root, which is that the fold's numbering is not reconstructible at all. There is a third source beyond the two #1321 and #1322 name — the `updated_seq` an archived row carries was assigned while it was still live, and nothing in the log carries it across the archive boundary, so no fold can reproduce it without a schema change.

#1322 flags the cost of candidate 2: it weakens `test_compaction.py`'s deliberate drift-after-compaction guard. That concern is answered here rather than deferred — the guard's `> 0` step is now produced by genuine drift (an index row whose status the log does not support) instead of by the numbering mismatch, on both engines (`test_action_index_covers_the_archive_after_compaction`, `test_pg_action_index_covers_the_archive_after_compaction`).

#1322's own repro, run verbatim against both engines on this branch, reports `0` before the compaction and `0` after it (it reported `1` on SQLite and `2` on Postgres before).

Candidate 1 — a monotonic global column copied into `events_archive` — remains the right answer if the projection ever becomes per-(key, run), because that is the only shape under which `updated_seq` ranks anything. `index_drift_count` is the single place a numbering comparison would then go back into.

Closes #1321. Closes #1322.

## Tests

New and changed tests (all fail on `upstream/main`, all pass here):

`tests/test_action_index.py`
- `test_renumbering_the_order_column_is_not_drift` — a uniform rescaling of `updated_seq` reads clean.
- `test_a_status_the_log_does_not_support_is_still_drift` — the relaxed comparison still catches a status the log does not produce; rebuild repairs it.
- `test_rebuild_reports_no_corrections_for_an_intact_compacted_store` — a healthy compacted store reports **0** corrections from rebuild, not a dirty count.

`tests/test_compaction.py`
- `test_action_index_covers_the_archive_after_compaction` — renamed from `..._after_rebuild`; the archived completion is still served, drift stays 0, and the drift step is now produced by a tampered status rather than by the numbering mismatch.

`tests/test_storage_postgres.py`
- `test_pg_drift_stays_zero_when_a_non_action_event_precedes_an_action` — issue #1321's reported case, on Postgres.
- `test_pg_drift_stays_zero_after_a_rebuild_and_further_actions` — a rebuilt store stays converged as new actions arrive, and the rebuilt rows still rank this run's writes in the order they happened.
- `test_pg_a_status_the_log_does_not_support_is_still_drift` — the Postgres equivalent of the SQLite guard test.

## Two latent suite bugs fixed along the way

The Postgres suite shares one database with fixed run ids and no per-test cleanup, so a test that mutates shared tables must scope its mutation and restore it. Two did not:

- `test_pg_deleted_boundary_event_fails_verify` ran an unscoped outer `DELETE` — `DELETE FROM events WHERE sequence = (SELECT MIN(sequence) ... WHERE run_id = 'pg_kb')` — which deleted that sequence *number* from **every** run in the shared database. It silently removed unrelated rows from later tests, which showed up as vanishing settle events elsewhere in the suite. The DELETE is now run-scoped and the row is re-inserted in a `finally`.
- `test_pg_archive_tampering_fails_verify` never undid its archive tamper, leaving the shared archive permanently corrupted so every later store-wide check read dirty for a reason that test owned. The original payload is now captured and restored in a `finally`.

## Verification

- `ruff check src tests` / `ruff format --check src tests` — clean (ruff 0.16.5, the pinned version)
- `mypy` on `src/continuum` — clean (128 files)
- `pytest tests/test_action_index.py tests/test_compaction.py` — 41 passed
- `pytest tests/test_storage_postgres.py` — 22 passed, against a real Postgres 16 on a freshly created scratch database
- `pytest` full default suite — one failure, `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend`, which is **pre-existing and environmental**: it asserts a `RuntimeError` mentioning `psycopg` on the assumption the driver is absent, and `psycopg` is installed in this venv. It fails identically on `upstream/main` with these changes stashed.
- `tests/test_docs_counts.py` — the docs-count guard passes; the added tests stay within tolerance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved action-index drift detection to avoid false positives after compaction or sequence renumbering.
  * Correctly detects missing, extra, outdated, or modified index entries.
  * Fixed drift counts so lost entries are counted once.
  * Rebuilding the action index now restores corrupted entries and reports the number of corrections made.
  * Compacted stores now remain clean while preserving access to archived actions.


src/continuum/storage/actionindex.py
```diff
@@ -75,3 +76,48 @@ def index_entry_from_payload(
...
+) -> int:
+    """Count index rows that disagree with the log fold (issue #1321).
+
+    ``expected`` is the fold of the log and ``stored`` is the projection as it
+    stands; both map a ledger key to ``(updated_seq, status)``. A row
+    disagrees when the log no longer produces it, when the log produces a key
+    the projection lacks, or when the row's status is not the folded one.
+
+    ``updated_seq`` is deliberately *not* compared. Two healthy stores make it
+    disagree with the fold no matter how the fold numbers its rows:
```
Cyrax321, line 100: BLOCKING. A key the projection lost is counted twice. `missing` counts it because `key not in stored`, and `changed` counts it again because the `stored.get(key, (0, None))` default yields None, which never equals a real status string.

Verified by direct evaluation of the helper: one lost key returns 2, one lost plus one spurious returns 3 instead of 2. The pre-PR criterion counted a missing key once, through the `changed` term alone (`stored.get(k)` was None, which never equalled the `(seq, status)` tuple). So this PR makes the reported number worse, not better, for that failure mode.

No test covers a lost row: the existing drift test corrupts status on a present row, the spurious test inserts a ghost, and the new tests tamper status. So the double count is unguarded. Fix is to iterate only keys present in both mappings for the `changed` term, or drop the `missing` term and let `changed` own it as the old code did.

tests/test_storage_postgres.py
```diff
@@ -254,49 +255,164 @@ def test_pg_compact_rejects_through_sequence_that_would_eat_the_anchor(
...
+    before = storage._connection.execute(
+        "SELECT payload FROM events_archive WHERE run_id = 'pg_kt' ORDER BY sequence LIMIT 1"
+    ).fetchone()["payload"]
     storage._connection.execute(
         "UPDATE events_archive SET payload = '{\"tampered\": true}' WHERE run_id = 'pg_kt'"
     )
-    report = storage.verify_events("pg_kt")
-    assert report.ok is False
-    assert any(v.kind == "TAMPERED_CONTENT" for v in report.violations)
+    try:
```
Cyrax321, line 317: BLOCKING. `before` captures only the seq-1 payload (`LIMIT 1`) but the restore UPDATE has no sequence filter, so it writes seq-1's payload into every archived row of pg_kt. Verified that make_run plus checkpoint plus compact leaves 2 archived rows for this run, so the seq-2 STATE_CHECKPOINTED row ends up carrying the RUN_STARTED payload and the archive stays corrupted after the test.

That is the exact failure this cleanup was added to prevent: the suite shares one database, and every later store-wide check of pg_kt then reads dirty for a reason this test owns. Scope both the capture and the restore by sequence, or capture and restore all rows.

src/continuum/storage/postgres.py
```diff
@@ -686,19 +686,29 @@ def foreign_action(self, key: str, *, exclude_run: str) -> Action | None:
...
             # psycopg's Connection has no executemany; the cursor does.
             with self._connection.cursor() as cur:
                 cur.executemany(
                     "INSERT INTO action_index(key, run_id, action_id, status, "
                     "updated_seq, action_json) VALUES (%s, %s, %s, %s, %s, %s)",
-                    [
-                        (key, entry[1], entry[2], entry[3], seq, entry[4])
-                        for key, (entry, seq) in canonical.items()
-                    ],
+                    rows,
```
Cyrax321: Non-blocking but worth closing here, since this PR is already rewriting the return semantics on the other engine. SQLite's rebuild now returns index_drift_count(after, before); this one still returns 0. cli/main.py prints that number as "N row(s) corrected", so the same corruption prints "1 row(s) corrected" on SQLite and "0 row(s) corrected" on Postgres. An operator who gates on that number reads the Postgres repair as a no-op. The fix is the same shape as the drift method two functions down: read `before` before the DELETE and compare after.

src/continuum/storage/postgres.py
```diff
@@ -686,19 +686,29 @@ def foreign_action(self, key: str, *, exclude_run: str) -> Action | None:
...
+        fold's merged position: the position counts the non-action events the
+        sequence skips, so a rebuilt row and the next ``nextval`` were on
+        different scales and the store re-drifted with every action written
+        afterwards (issue #1321). Archived keys keep the fold's negative
+        position, which is already below every value the sequence will ever
+        hand out and already in fold order.
+        """
         canonical = self._canonical_index_rows()
+        rows: list[tuple[str, str, str, str, int, str]] = []
+        for key, (entry, order) in canonical.items():
```
Cyrax321: Non-blocking. Two costs in this loop.

It runs outside `with self._write()`, so each live key costs one SELECT nextval round trip on the shared autocommit connection with no lock held. On a store with 50k live keys that is 50k sequential round trips to repair once, and the fold-to-DELETE window widens from a few statements to that whole span. A concurrent writer's row landing in the window is deleted by the DELETE below and never re-folded, so the projection silently loses a live key until someone rebuilds again. _backfill_action_index ten lines from the top of this file already calls nextval server-side inside an INSERT...SELECT; a single setval after assigning values in Python costs one round trip.

Separately, nextval on the shared connection outside the lock can overlap the background checkpoint executor on another thread, and libpq does not support concurrent use of one Connection.

src/continuum/storage/postgres.py
```diff
@@ -686,19 +686,29 @@ def foreign_action(self, key: str, *, exclude_run: str) -> Action | None:
...
     def rebuild_action_index(self) -> int:
-        """Recompute the whole index from the log (global key space)."""
+        """Recompute the whole index from the log (global key space).
+
+        Live keys take fresh sequence values in fold order rather than the
+        fold's merged position: the position counts the non-action events the
+        sequence skips, so a rebuilt row and the next ``nextval`` were on
+        different scales and the store re-drifted with every action written
+        afterwards (issue #1321). Archived keys keep the fold's negative
+        position, which is already below every value the sequence will ever
```
Cyrax321: Non-blocking, altitude. The PR argues in the CHANGELOG and in the helper's docstring that `ORDER BY updated_seq DESC LIMIT 1` in foreign_action ranks nothing because key is the primary key, which is correct. It then spends the round trips above re-minting sequence values specifically to keep that ordering meaningful, and test_pg_drift_stays_zero_after_a_rebuild_and_further_actions asserts directly on that ordering.

Both readings cannot be right in the tree at once. If the clause is dead, delete it and write the fold position for every row, which removes the nextval loop entirely. If it ranks something, then the fold's cross-run ordering is live and needs fixing rather than tolerating. Shipping the cost and the argument-against-the-cost together leaves the next maintainer to guess which was intended.

src/continuum/storage/actionindex.py
```diff
@@ -75,3 +76,48 @@ def index_entry_from_payload(
...
+def index_drift_count(
+    expected: Mapping[str, tuple[int, str]],
+    stored: Mapping[str, tuple[int, str]],
+) -> int:
+    """Count index rows that disagree with the log fold (issue #1321).
+
+    ``expected`` is the fold of the log and ``stored`` is the projection as it
+    stands; both map a ledger key to ``(updated_seq, status)``. A row
+    disagrees when the log no longer produces it, when the log produces a key
+    the projection lacks, or when the row's status is not the folded one.
```
Cyrax321: Non-blocking. The comparison now covers the key set and the status column, so a tampered action_json, run_id or action_id is invisible to verify --index. That blind spot predates this PR, but this PR is the one narrowing the check while the CHANGELOG says the guard "keeps its teeth" and the omission "costs nothing".

action_json is the exact column foreign_action returns to the idempotency guard, so it is arguably the more valuable half to check than status. Worth either comparing it or softening the claims in the CHANGELOG and docstring.

tests/test_storage_postgres.py
```diff
@@ -254,49 +255,164 @@ def test_pg_compact_rejects_through_sequence_that_would_eat_the_anchor(
     assert storage.verify_events("pg_kg").ok is True
 
 
+_EVENT_COLUMNS = (
```
Cyrax321: Non-blocking. Both snapshot-and-restore blocks could be one self._connection.transaction() each, which compact_run already uses on this engine. Rollback is immune to a NOT NULL column added by a later migration, and it also restores when an assertion inside the block fails, which the current `finally` restore does not guarantee for the deleted-row case: the assert on `kinds` raising skips the re-INSERT entirely. _EVENT_COLUMNS, the pre-SELECT and the INSERT would all drop out.

Cyrax321 (changes requested):
This and #1325 are two takes on one defect and both close #1322, so one of them should be chosen rather than both merged. I left #1325 requesting changes; this one is the better-shaped of the two, because it puts the criterion in one shared helper in actionindex.py instead of duplicating the loop per engine per method, which is exactly the structure I asked for over there. It also fixes the #1321 single-run case that #1325 does not touch.

Two things block it. The helper double-counts a row the projection lost, so verify --index reports 2 row(s) of drift for one lost row. And the archive-tamper test's new cleanup restores every archived row to the first row's payload, which leaves the shared suite database corrupted in the way the cleanup set out to prevent.

What this one does not carry from #1325 is the #1267 half: Postgres rebuild_action_index still returns 0 while SQLite now returns a real count, so the two engines print different things for the same repair. Details inline.

src/continuum/storage/actionindex.py
Adhi1-2: @/tmp/replies/r1.md

tests/test_storage_postgres.py
Adhi1-2: @/tmp/replies/r2.md

src/continuum/storage/postgres.py
Adhi1-2: @/tmp/replies/r3.md

src/continuum/storage/postgres.py
Adhi1-2: @/tmp/replies/r4.md

src/continuum/storage/postgres.py
Adhi1-2: @/tmp/replies/r5.md

src/continuum/storage/actionindex.py
Adhi1-2: @/tmp/replies/r6.md

tests/test_storage_postgres.py
Adhi1-2: @/tmp/replies/r7.md

src/continuum/storage/actionindex.py
```diff
@@ -75,3 +76,48 @@ def index_entry_from_payload(
...
+    ``ORDER BY updated_seq DESC LIMIT 1`` that consumes it is scoped to a
+    single row and cannot rank anything. Should the schema later allow one row
+    per write of a key, that ranking needs a numbering both sides can
+    reconstruct, and this comparison is where to put it.
+    """
+    missing = sum(1 for key in expected if key not in stored)
+    extra = sum(1 for key in stored if key not in expected)
+    changed = sum(
+        1 for key, (_, status) in expected.items() if stored.get(key, (0, None))[1] != status
+    )
```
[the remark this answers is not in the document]
Adhi1-2: @/tmp/replies/rgitar.md

Adhi1-2:
Pushed 9ff34ca answering all seven points; replies are threaded on each. Summary:

**Blocking**
- The double-count is fixed. `changed` ranges only over keys present in both mappings, so one lost row is one row of drift and one correction. Covered on both engines.
- The archive-tamper cleanup no longer restores by hand. Both blocks corrupt inside a `transaction()` and raise a sentinel to roll back, so the seq-2 row cannot inherit the seq-1 payload and the row is restored on every exit path.

**Carried over from #1325**
- Postgres `rebuild_action_index` returns the drift that called for the repair, so the same corruption prints the same number on both engines (#1267).

**Resolved your altitude note**
- Both readings could not stand, so the dead one is gone: `ORDER BY updated_seq DESC LIMIT 1` is deleted on both engines, and the Postgres rebuild mints no sequence values at all — it writes the fold position for every row. The `nextval` loop and its widened fold-to-DELETE window go away with it.

**Widened rather than softened**
- `action_json` is now compared. It is the string `foreign_action` returns to the idempotency guard, and it embeds `run_id` and `action_id`, so the check covers a tamper of either column as well.

On #1325: this one now carries everything it did plus the #1321 single-run case and the #1267 half, so there is nothing left for it to add. I will close it and leave the branch up in case you prefer the per-engine-loop shape; say so and I will reopen.

Ran the full gate set on a fresh Postgres database: `ruff check`, `ruff format --check`, `mypy`, and the full suite including `tests/test_storage_postgres.py` are green. One caveat on the coverage report: the standard Test job runs the suite without `CONTINUUM_TEST_POSTGRES_DSN`, so `tests/test_storage_postgres.py` skips and its lines read as uncovered; the `test-postgres` job runs them but uploads no coverage. The new Postgres lines will show as a patch gap regardless of whether they are tested.

Adhi1-2:
Closing in favour of #1325, which subsumes this.

Both PRs fix the same symptom — `action_index_drift` reading dirty on a healthy store — by refusing to compare `updated_seq`. I compared the two criteria directly:

- **This PR** compares `(status, action_json)` via the shared `index_drift_count` helper, with no notion of *which run* owns the row.
- **#1325** compares `(run_id, action_id, status)` and adds a run-agreement guard: a key written by two runs is adjudicated only when the stored row and the fold agree on which run's write won.

That guard is the difference. In the contested-key case — run A completes a key, run B later records the same key as STARTED, then run B is compacted — the fold crowns A's completion while the stored row still belongs to B. This PR's comparison sees two different records and reports drift=1 on a store nobody tampered with. #1325 skips that adjudication and reports 0, because the stored row is the latest write by construction. I have a regression test for exactly that case failing on this PR's criterion and passing on #1325's.

#1325 also carries the #1267 fix (Postgres `rebuild_action_index` returning 0) that this PR leaves alone.

What this PR has that #1325 doesn't: the shared `index_drift_count` helper, which is genuinely nicer factoring than per-engine copies. If a maintainer prefers that shape, the guard should be folded into the helper rather than re-implemented per engine — happy to rework #1325 that way instead. The branch stays available.

Adhi1-2:
Correction to my closing note above, which said "closing in favour of #1325" — that was circular, since I had closed #1325 twenty minutes earlier in favour of this one. Both closures pointed at each other and neither was right.

The maintainer resolved the cluster in the #1323 review thread and approved #1323 on 2026-09-21 ("No blocking issues. Ready to merge"), choosing the reconciliation approach over this PR's drop-the-comparison approach. #1323 is the surviving PR for #1321.

What still stands from this PR, if a maintainer wants it: the shared `index_drift_count` helper in actionindex.py is genuinely nicer factoring than per-engine copies, and the run-agreement guard this PR lacks is the reason it reads a contested key as drift. The branch stays available; the guard would need to be folded into the helper rather than re-added per engine.

PR #1325, opened 2026-09-20
fix(storage): action index reads dirty after a cross-run compaction (#1322)

### LINKED ISSUE #1267: storage: PostgresStorage.rebuild_action_index always reports 0 corrected rows, so verify --repair-index lies on Postgres

### LINKED ISSUE #1322: action_index_drift reports false positives once a run is compacted in a multi-run store (both engines)

Adhi1-2:
## What

`action_index_drift()` reported a corrupted projection on a store nobody
tampered with, the moment one run was compacted while another still held
actions. Both engines are affected.

## Why

Both `_canonical_index_rows` folds merge `events_archive` ahead of `events`
as though the two were disjoint segments. **Within a run** that is true.
**Across runs** it is not: archiving one run renumbers its action events
below another run's still-live ones, so the fold crowns a winner the
incremental writer never stored. Drift then compared that fabricated order
number against the row's `updated_seq` and reported drift=1 on a healthy
store.

## The fix

The comparison now rests on what a row **asserts** — which run owns the key,
which action it points at, the status that action reached — and never on
`updated_seq`. That column has no consumer that needs it exact: `key` is the
projection's primary key, so at most one row per key exists and
`foreign_action`'s `ORDER BY updated_seq DESC` can never choose between
candidates.

A key written by two runs is adjudicated only when the stored row and the
fold agree on which run's write won. The stored row is the latest write by
construction — the writer upserts on every `ACTION_*` event — so a run
disagreement is the fold being wrong, not the projection.

`rebuild_action_index` uses the same criterion, so a repair reports the drift
that called for it.

## Also fixes #1267

`PostgresStorage.rebuild_action_index` rebuilt the projection but
unconditionally `return 0`, so `verify --repair-index` reported "0 row(s)
corrected" while repairing. It now mirrors the SQLite engine's count. This
**supersedes #1285**, which fixed the same symptom; the snapshot is still read
inside the write lock, so #1285's concurrency property is preserved. #1285
should be closed in favour of this.

## Evidence

Four regression tests fail with drift=1 on **both** engines against the
pre-fix code and pass with it:

- `test_a_contested_key_survives_another_run_being_compacted`
- `test_action_index_covers_the_archive_after_rebuild` (rewritten — it
  asserted the false positive)
- `test_pg_action_index_stays_clean_when_another_run_is_compacted`
- `test_pg_action_index_covers_the_archive_after_rebuild` (rewritten)

A fifth, `test_a_key_older_than_a_compacted_run_reads_clean_after_rebuild`,
guards the repair path; the old criterion satisfied it (the rebuild
resynchronizes `updated_seq` from the same fold), so it pins the invariant
rather than proving the change.

The Postgres tests need an isolated database: `action_index_drift` is
store-wide, so the shared suite database — where one test's compaction
renumbers another's rows — would make a clean store read dirty for reasons
the test does not exercise.

While there, fixed a real latent bug the improved test exposed:
`test_pg_deleted_boundary_event_fails_verify` ran an unscoped
`DELETE ... WHERE sequence =`. Sequence is per-run, so that deleted the row
from **every** run in the shared suite database and silently corrupted the
action index of whichever test ran first. The DELETE is now scoped by
`run_id`.

## Gates

`ruff check` / `ruff format --check` / `mypy src/continuum` clean;
`tests/test_docs_counts.py` 5/5 including the slow live-suite check
(collected 2,401; docs re-synced, the earlier figure was carried over from
the branch that also held #1321's tests). Full suite locally against a real
Postgres 17: **2,394 passed, 6 skipped**, with one known environmental
failure — `test_postgres_url_is_routed_to_postgres_backend` asserts a
`RuntimeError` for a missing `psycopg`, which cannot fire on a machine where
the extra is installed; it is green in CI's driverless job.

Closes #1322. Closes #1267.

## Summary by CodeRabbit

* **Bug Fixes**
  * Action indexes now remain accurate after compacting a different run.
  * Rebuilding an action index now reports how many rows were corrected.
  * Missing or inconsistent index entries are detected and repaired more reliably.

* **Tests**
  * Expanded coverage for compaction, index recovery, and PostgreSQL storage scenarios.

* **Documentation**
  * Updated test-suite counts across the README files and testing guides to reflect approximately 2,403 collected and 2,322 passing tests.


src/continuum/storage/sqlite.py
```diff
@@ -588,35 +588,63 @@ def foreign_action(self, key: str, *, exclude_run: str) -> Action | None:
...
+                changed += 1  # an action event the projection lost
+                continue
+            # A key written by two runs is only adjudicated when the row and
+            # the fold agree on which run's write won. The fold rebuilds
+            # global insertion order from table order and compaction breaks
+            # that, so it can crown a write the incremental writer never
+            # stored. The stored row is the latest write by construction --
+            # the writer upserts on every ACTION_* event -- so a run
+            # disagreement is the fold being wrong, not the projection.
+            if have[0] != entry[1]:
```
Cyrax321, line 627: BLOCKING. This skip is wider than the bug it treats. It fires whenever the stored row's run_id differs from the fold's winner, which covers the benign compaction renumbering but also every real divergence on a key two runs both wrote.

Reproduced against the PR's own test setup: after run_1 completes an unscoped key and run_2 records it STARTED and is compacted, tampering with the stored row's status (started -> failed) leaves drift at 0. The pre-PR criterion returned 1. That is the class of corruption verify --index exists to catch, and a tampered status on a contested key is now invisible while the existing single-column tamper test still passes.

The premise that the stored row is always the latest write is load-bearing and not always true. Any out-of-band divergence (inconsistent restore, a hand-corrected row, a future writer that appends ACTION_* events without maintaining the index) is masked here for contested keys. Consider narrowing the tolerance to the specific case being fixed: adjudicate when the row and the fold agree on the winning run, but still flag a row whose action_id or status disagrees with that same run's own log entry.

src/continuum/storage/postgres.py
```diff
@@ -700,27 +716,58 @@ def rebuild_action_index(self) -> int:
...
+                changed += 1  # an action event the projection lost
+                continue
+            # A key written by two runs is only adjudicated when the row and
+            # the fold agree on which run's write won. The fold rebuilds
+            # global insertion order from table order and compaction breaks
+            # that, so it can crown a write the incremental writer never
+            # stored. The stored row is the latest write by construction --
+            # the writer upserts on every ACTION_* event -- so a run
+            # disagreement is the fold being wrong, not the projection.
+            if have[0] != entry[1]:
```
Cyrax321, line 767: BLOCKING, same defect as the SQLite twin, and on this engine the premise fails through an ordinary code path rather than only through tampering.

_insert_event commits the event and then calls _maintain_action_index as a separate statement on an autocommit connection. A crash, timeout, or IntegrityError between the two leaves the event durable in the log while the index row still names the previous owner. The fold crowns the new run, the stored row names the old one, this skip fires, and drift reports a clean store that foreign_action is actively answering wrong. SQLite is immune because _maintain_action_index shares the caller's BEGIN IMMEDIATE transaction, so the two engines do not agree on whether this can happen.

That asymmetry is the strongest argument for fixing the fold rather than tolerating the disagreement: the tolerance has to be tuned per engine to a race only one of them has.

src/continuum/storage/sqlite.py
```diff
@@ -627,12 +655,11 @@ def rebuild_action_index(self) -> int:
...
-        corrections = sum(
-            1
-            for k, val in ((k, (seq, entry[3])) for k, (entry, seq) in canonical.items())
-            if before.get(k) != val
-        )
-        corrections += len(set(before) - set(canonical))
+        corrections = len(set(before) - set(canonical))
+        for key, (entry, _seq) in canonical.items():
+            have = before.get(key)
+            if have is None or (have[0] == entry[1] and have != (entry[1], entry[2], entry[3])):
```
Cyrax321, line 661: BLOCKING. rebuild rewrites a contested key it does not count, which is the #1267 symptom this PR says it fixes, persisting for exactly the contested case.

Reproduced: the stored row was (run_2, started); after rebuild it is (run_1, completed) with a different action_id, and rebuild_action_index() returned 0. verify --repair-index therefore prints "0 row(s) corrected" while changing which run owns the key. That flip is not internal either: foreign_action reads this table, so an unrelated drift in one key makes the repair silently change another run's ledger answer, and because the row now agrees with the fold, drift stays 0 by construction and no later run can detect it.

Either count the contested rewrite as a correction, or have rebuild preserve the incremental writer's row for keys the fold and the writer disagree on (which is what the drift docstring argues is the truth). The current code asserts the stored row is authoritative and then overwrites it.

tests/test_action_index.py
```diff
@@ -299,6 +300,106 @@ def test_a_key_rewritten_by_another_run_is_global_last_write_wins(
...
+            "action": Action(
+                run_id="run_2", action_type="send_invoice", status=ActionStatus.STARTED
+            ).model_dump(mode="json"),
+        },
+    )
+    CheckpointManager(store).checkpoint("run_2")
+    store.compact_run("run_2")
+
+    # The repair rewrites the row with the fold's winner, run_1's completion.
+    assert store.rebuild_action_index() >= 0
```
Cyrax321, line 416: BLOCKING (test hygiene). `>= 0` is a tautology: the method is a set difference plus non-negative increments and cannot return negative; in this test's exact setup it returns 0. The pre-PR `return 0` would satisfy this assertion unchanged.

The CHANGELOG entry for #1267 is a claim that a bug is fixed, and no test asserts the Postgres return value at all, so that fix can regress silently. Assert the actual count, and add a Postgres assertion that a genuinely drifted row is reported as corrected on that engine. AGENTS.md asks for a regression test that would have caught the bug; this one would not.

tests/test_action_index.py
```diff
@@ -299,6 +300,106 @@ def test_a_key_rewritten_by_another_run_is_global_last_write_wins(
...
+    assert store.action_index_drift() == 0
+    found = store.foreign_action(str(shared), exclude_run="nobody")
+    assert found is not None
+    assert found.run_id == "run_1"
```
Cyrax321, line 420: BLOCKING. This and test_a_contested_key_survives_another_run_being_compacted build the identical store (run_1 completes an unscoped key, run_2 records it STARTED, run_2 is compacted) and assert opposite correct answers: line 363 says the owner stays run_2 because compaction is relocation, this line says run_1/completed because the fold says so.

Both are now pinned as intended behaviour for one log, so the suite holds the contradiction in place rather than exposing it. Any later fix of the fold's cross-run ordering fails one of them by construction. That is worth resolving now: if rebuild is not authoritative for contested keys, it should not be rewriting them (see the sqlite.py:661 comment); if it is, then compaction is not purely relocation and the surviving test's premise needs restating.

src/continuum/storage/postgres.py
```diff
@@ -686,9 +686,25 @@ def foreign_action(self, key: str, *, exclude_run: str) -> Action | None:
...
+        Returns the number of rows corrected, mirroring the SQLite engine: a
+        correction is any key whose stored row was missing, spurious, or
+        disagreed with the log under :meth:`action_index_drift`'s criterion.
+        (This used to unconditionally return ``0`` even while repairing --
+        #1267.)
+        """
         canonical = self._canonical_index_rows()
         with self._write():
+            # Snapshot ahead of the DELETE it is measured against, inside the
+            # write lock, so a row a concurrent writer shifted between the two
```
Cyrax321, line 700: Non-blocking, but the comment overclaims. _write() returns self._lock, a per-instance threading.RLock over an autocommit connection, so it grants no isolation from a second process sharing this database, which is this engine's stated multi-process design. The fold at line 697 also runs outside the lock entirely, and the DELETE and the executemany are separate autocommit statements rather than one transaction (compare compact_run, which uses an explicit transaction() block for exactly this reason).

Consequences: a writer in another process landing between the fold and the snapshot has its row deleted and then charged as a spurious correction, and a failure between the DELETE and the insert leaves the index empty with every unscoped claim returning None until the next rebuild. Neither matches "counted rather than silently absorbed".

tests/test_storage_postgres.py
```diff
@@ -283,20 +292,134 @@ def test_pg_deleted_boundary_event_fails_verify(storage: PostgresStorage) -> Non
...
+    name = f"iso_{os.getpid()}_{next(_iso_counter)}"
+    admin = psycopg.connect(DSN, autocommit=True)
+    try:
+        admin.execute(f'CREATE DATABASE "{name}"')
+    finally:
+        admin.close()
+    params = conninfo_to_dict(DSN)
+    params["dbname"] = name
+    store = PostgresStorage(make_conninfo(**params))
+    yield store
```
Cyrax321, line 358: Non-blocking. The teardown is not in a try/finally, so a failing test body or a close() that raises skips the DROP and leaves iso_<pid>_N behind. Combined with CREATE DATABASE without IF NOT EXISTS and a pid-based name, a recycled pid on a shared host then errors out an unrelated test in setup. A uuid4 suffix plus DROP IF EXISTS before CREATE plus a finally that always drops would close it. Scoping the boundary-event DELETE by run_id above is a genuine cross-test corruption fix, nice catch.

CHANGELOG.md
```diff
@@ -937,7 +962,7 @@ All notable changes to this project are documented here. The format follows
   Framework Integration documents the CrewAI/AutoGen/Pydantic-AI thin hooks
   and the gateway/OTel fallback seams; the Roadmap marks the dashboard and
   the enforced-durability work complete; test counts are current
-  (~2,397 collected, ~2,320 passed, ~27 skipped on a minimal env).
+  (~2,401 collected, ~2,322 passed, ~28 skipped on a minimal env).
```
Cyrax321: Non-blocking. This line says ~28 skipped while README.md and docs/CONTRIBUTING_ONBOARDING.md say ~27 skipped for the same ~2,401 collected / ~2,322 passed figures, and both carry the same generated-via marker. The three files should agree on one number. These counts are also already stale against main, which has moved on, so the re-sync will need redoing on merge.

Cyrax321 (changes requested):
The false positive is real and the fix works: I reproduced drift=1 on a healthy SQLite store where one run held a key another run later wrote and then got compacted, and the new criterion reads it clean. The #1267 half (Postgres rebuild returning 0) is also genuinely fixed in the common case.

Two things block the merge, and they share one root cause: the run-agreement guard assumes the stored row is always the latest ACTION_* write, which is only true while every writer goes through the incremental path. When that assumption fails, the guard does not just stop a false positive, it stops drift from detecting the corruption it exists to catch, and rebuild then rewrites the contested key while reporting it as uncorrected.

The deeper question is whether the fold should be fixed instead of tolerated. events_archive stores the original timestamp verbatim, so a timestamp-ordered merge would be genuinely global and invariant under relocation, and the incremental writer already has a compaction-proof order that the archive step throws away. The current design bakes in a disagreement between the fold and the writer and then works around it in the checker, which is why the two new tests end up pinning opposite answers for one log. Worth deciding before this lands rather than after.

Non-blocking notes inline too (docs count mismatch, fixture teardown, a misleading comment). Happy to re-review after a rework of the contested-key path.

Adhi1-2:
#1324 now carries everything this one did: the cross-run fix through the shared `index_drift_count` helper, the #1267 rebuild count, and the #1321 single-run case this one does not touch. The branch stays available if the per-engine loop shape is preferred.

Adhi1-2:
Correction to my closing note above, which pointed at #1324 — that was circular, since I closed #1324 twenty minutes later pointing back here.

The maintainer resolved the cluster in the #1323 review thread (approved 2026-09-21): #1323 is the chosen shape for #1321, with the reconciliation approach preferred over dropping the comparison. So #1323 is the surviving PR for this defect family, and this one and #1324 are both correctly closed in favour of it.

The work this PR did that #1323 does not (the #1322 cross-run-compaction case and the #1267 rebuild-count fix) is not lost — the branch stays available, and #1322 and #1267 remain open issues that can be re-filed separately against #1323 if the maintainer wants them.

PR #1326, opened 2026-09-20
fix(bench): controlled-failure ground truth declares modes the engine can propose (#1304)

### LINKED ISSUE #1304: benchmark: controlled_failures declares expected="RETRY" for two scenarios, but RecoveryMode has no RETRY

Parthipashok04:
## Description

`tool_failure` and `api_timeout` in `src/continuum/benchmark/controlled_failures.py` declared `expected="RETRY"`, a name `RecoveryMode` (`src/continuum/models.py`) never defines. The seven real members are `RESUME, REPAIR_AND_RESUME, ROLLBACK, WAIT, REQUEST_HUMAN, REPLAN, ABORT`, so both rows could never match a `RecoveryDecision.mode` and were ungradeable the moment this table is wired to the recovery engine.

The two scenarios are different failure shapes, so they take different modes rather than one blanket value:

- **`tool_failure` -> `RESUME`.** A returned error settles the action as `FAILED` (`ledger.fail(certain=True)`). `engine.py` builds its uncertain set from `UNKNOWN, STARTED, REQUIRES_REVIEW` only, so a `FAILED` action never lands in it and nothing blocks the run. The description already says "retry is safe".
- **`api_timeout` -> `REQUEST_HUMAN`.** A timeout is not evidence of absence: the request may have landed, so `ledger.fail(certain=False)` records `UNKNOWN` with `side_effect_uncertain` set, and the engine's uncertain-side-effect arm (`engine.py:626-650`) proposes `REQUEST_HUMAN` in strict mode, `WAIT` in lenient. This is the same shape as the existing `external_side_effect` row, which already declares `REQUEST_HUMAN`, and `project.md` is explicit that a system that cannot determine whether a side effect occurred "must NOT blindly retry".

The test now asserts against `RecoveryMode` itself instead of a hand-written literal that happened to include `"RETRY"`. That literal is exactly how an impossible value was blessed by the test meant to catch it. Two further tests were added: one pins the two corrected values by name (membership alone would still accept a wrong-but-valid mode such as `ABORT`), and one asserts the two uncertain-side-effect rows agree, since a drift between them would read as a recovery regression once graded.

No CHANGELOG entry: the issue notes nothing grades this table today, so no live output changes. This corrects the ground truth before it is wired up.

## Related issues

Fixes #1304

## Types of changes

- Type: `bugfix`

## Breaking changes

No. `ControlledScenario.expected` has one consumer (`tests/test_controlled_failures.py`); `by_name` and `SCENARIOS` appear nowhere else in `src/`, `tests/` or `benchmarks/`.

## How has this been tested

Against `upstream/main` (`e64ce99`) in an isolated worktree, Python 3.12.10, pytest 9.1.1.

```bash
# Reproduced the issue first (both ungradeable rows named):
#   ungradeable rows: ['tool_failure', 'api_timeout']

pytest tests/test_controlled_failures.py -v      # 5 passed
pytest tests/ -q -k "benchmark or recovery or controlled"  # 188 passed
ruff check src/continuum/benchmark/controlled_failures.py tests/test_controlled_failures.py   # clean
ruff format --check src/continuum/benchmark/controlled_failures.py tests/test_controlled_failures.py  # clean
mypy src/continuum/benchmark/controlled_failures.py  # no issues
```

The three changed/added tests were verified to **fail on the old table** with the source change reverted, each naming the offending row and value, and to pass once restored. This is the regression test the issue asked for.

## Checklist

Tests added or updated for the changed behaviour. Documentation updated where the change affects user-facing behaviour (n/a, no user-facing behaviour changes). CI passes on the latest commit. Security-relevant changes state known limitations explicitly (n/a, benchmark ground truth only).

## Additional context

This issue already has an open PR, #1306 by the issue reporter, which converges on the same two values and the enum-backed assertion. This PR is independent work that reaches the same fix from the engine's own semantics (the ledger's `FAILED` vs `UNKNOWN` distinction, and `engine.py`'s uncertain set) rather than by inspection, and adds the two extra guards above. Reviewers may prefer either; no offence taken if #1306 is the one merged.

The added comments in `controlled_failures.py` cite the engine and ledger lines that justify each mode so the next reader can trace the reasoning without re-deriving it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated failure recovery guidance so known tool failures resume safely.
  - API timeouts and other scenarios with potentially completed side effects now request human review instead of retrying automatically.
  - Corrected recovery-mode validation to use only supported options.


Cyrax321 (changes requested):
Author context: Parthipashok04, external contributor. The ground truth values this PR proposes are already on origin/main: tool_failure declares RESUME and api_timeout declares REQUEST_HUMAN there, identical to this diff. The enum assertion against RecoveryMode is sound, but there is nothing left to merge. This PR is also in DIRTY state against a stale base. Please close as superseded, or rebase onto current main if you have further changes beyond these two rows.

src/continuum/benchmark/controlled_failures.py
```diff
@@ -55,14 +55,22 @@ class ControlledScenario:
         scenario="tool_failure",
         checkpoint_version="v1",
         environment_version="v1",
-        expected="RETRY",
+        # A returned error is a settled FAILED action, not an uncertain one:
+        # the effect is known not to have landed, so the ledger's uncertain set
+        # (UNKNOWN, STARTED, REQUIRES_REVIEW) never contains it and the run can
+        # continue. RESUME, not a "RETRY" mode the enum does not define.
+        expected="RESUME",
```
Cyrax321, line 63: Blocking: this exact change, tool_failure from RETRY to RESUME, already landed on main in merged #1306, and the api_timeout flip to REQUEST_HUMAN below did too. That is why GitHub reports this PR as conflicting: as it stands the diff re-applies merged work rather than adding to it, so it cannot merge cleanly. Please rebase onto main; after that the reviewable delta is only the two new tests and these rationale comments, which do read as a genuine addition.

tests/test_controlled_failures.py
```diff
@@ -28,11 +36,27 @@ def test_dataset_change_ground_truth() -> None:
...
 def test_each_scenario_has_ground_truth() -> None:
     for scenario in SCENARIOS:
-        assert scenario.expected in {
-            "RESUME",
-            "RETRY",
-            "REPAIR_AND_RESUME",
-            "REQUEST_HUMAN",
-            "REPLAN",
-        }
+        assert scenario.expected in _VALID_MODES, (
```
Cyrax321, line 38: Blocking: main already carries an enum-backed version of this assertion from #1306, including _VALID_MODES and the RecoveryMode[scenario.expected] resolution, so this hunk conflicts textually with what is merged even though the intent is identical. A rebase should reduce this file's delta to just the two new tests below, which are the part worth keeping.

src/continuum/benchmark/controlled_failures.py
```diff
@@ -55,14 +55,22 @@ class ControlledScenario:
...
         description="Tool returns error, retry is safe",
     ),
     ControlledScenario(
         scenario="api_timeout",
         checkpoint_version="v1",
         environment_version="v1",
-        expected="RETRY",
+        # A timeout is not evidence of absence: the request may have landed, so
+        # the ledger records UNKNOWN with side_effect_uncertain set, and the
+        # engine's uncertain arm proposes REQUEST_HUMAN. The same shape as
```
Cyrax321: Non-blocking: the uncertain arm proposes REQUEST_HUMAN only when strict_unknown is on; in lenient mode the same UNKNOWN shape gets WAIT (the lenient branch in engine._decide). The PR body already names that split, so a short parenthetical here, strict mode only, would keep the comment from reading as unconditional to whoever traces this reasoning next.

Cyrax321 (changes requested):
The corrected values themselves are right: RecoveryMode defines no RETRY, ledger.fail(certain=True) settles a tool error as FAILED so nothing uncertain blocks a RESUME, and a timeout lands as UNKNOWN with side_effect_uncertain set so the engine's uncertain arm escalates to REQUEST_HUMAN in strict mode. The blocker is timing, not substance: #1306 merged the same two value flips and the enum-backed assertion on 2026-09-20, so this PR now shows as conflicting and most of its diff re-applies changes that already landed on main. Please rebase onto main so the reviewable delta is only the two new tests and the rationale comments; the strict/lenient nuance noted inline is worth folding in while you are there.

Parthipashok04:
-

PR #1327, opened 2026-09-20
fix(recovery): assess folds archived prefix into liveness, risk and authority scans (#1050)

### LINKED ISSUE #1050: assess is half archive-aware: liveness, risk and consumed-authority scans read the live tail only

Parthipashok04:
## Description

`RecoveryEngine.assess` was half archive-aware. The confirmation scan and the
validator already shared one archive-aware fetch of the run's events
(`read_all_events`, which walks `events_archive` plus the live tail in sequence
order), but three other signals in the same method re-read the live tail only:

- the liveness breach count,
- the `RISK_OBSERVED` scan,
- the consumed-authority scan.

`continuum compact` archives the pre-anchor prefix, so after compaction all
three vanished from the sealed contract: `triggering_risks` emptied, a recorded
`LIVENESS_SILENCE_DETECTED` breach count fell to zero, and a consumed authority
awaiting an external probe dropped out of the rationale. The engine still
returned a verdict, just one computed over a deliberately truncated history,
and every one of those downgrades was toward *less* caution.

The fix follows the issue's suggested shape: all three scans now filter or fold
the existing archive-aware fetch instead of re-reading the live tail. That also
removes three redundant live-tail reads per assessment. The shared fetch already
falls back to the live log when the archive cannot be read, so the guard around
it now covers the fold itself rather than swallowing a read failure.

One adjacent observation, left alone here as out of scope: after compaction, a
run whose declared dependency was captured pre-anchor validates the environment
diff as empty and the dependency reads as `unknown`, which under
`strict_unknown=True` fail-closes to `REQUEST_HUMAN`. That is pre-existing and
cautious (it errs toward a human), so it is not the same class of silent
downgrade this issue is about, but it does mean a compacted run's verdict can
shift for a reason unrelated to the three signals. Happy to file that
separately if it is not already tracked.

## Related issues

Fixes #1050

## Types of changes

- Type: `bugfix`

## Breaking changes

No. `assess` returns strictly more cautious (or equal) verdicts for compacted
runs than before; no signature or contract field changes.

## How has this been tested?

Three regression tests in `tests/test_recovery_engine.py`, one per signal, each
seeding the signal into the pre-anchor prefix, compacting the run, then
asserting the contract still carries it. Each test also pins the precondition
the fix depends on: the event left the live tail and reached the archive, since
a test that compacts nothing proves nothing.

```
D:/pROJ OPEN/CONTINUUM/.venv/Scripts/python.exe -m pytest \
  tests/test_recovery_engine.py \
  tests/test_liveness_events.py tests/test_risk_policy.py \
  tests/test_authority_consumed.py tests/test_authority_probe.py \
  tests/test_gate_resurrection.py tests/test_compaction.py \
  tests/test_provenance_compaction.py
100 passed
```

Failure-on-old-code evidence (base `engine.py` restored via
`git show <base>:src/continuum/recovery/engine.py`, run under the same pytest
config, which resolves `pythonpath = ["src"]` against this worktree):

```
FAILED tests/test_recovery_engine.py::test_triggering_risks_survive_compaction
FAILED tests/test_recovery_engine.py::test_liveness_breach_count_survive_compaction
FAILED tests/test_recovery_engine.py::test_consumed_authority_survive_compaction
3 failed, 41 passed
```

With the fix applied, all 44 tests in the file pass.

```
D:/pROJ OPEN/CONTINUUM/.venv/Scripts/python.exe -m ruff check src/ tests/
D:/pROJ OPEN/CONTINUUM/.venv/Scripts/python.exe -m ruff format --check src/ tests/
D:/pROJ OPEN/CONTINUUM/.venv/Scripts/python.exe -m mypy src/continuum/recovery/engine.py
```
All clean. Python 3.12.10 on Windows 11; pytest 9.1.1.

## Checklist

Tests added or updated for the changed behaviour; a CHANGELOG entry is added
under `[Unreleased] > Fixed`. No user-facing docs change, since this repairs
signals the documentation already describes as present. CI runs on the latest
commit. Security-relevant limitation stated above: the fail-closed
`REQUEST_HUMAN` shift on compacted runs with a captured dependency is noted
rather than changed.

## Additional context

The three new tests use an explicit `_assert_archived` helper because an
earlier draft of the liveness test passed on both the fixed and unfixed engine:
the breach count was being read from an event that had not actually been
archived. Asserting the archive round trip is what makes these tests mean what
they claim.


Cyrax321 (changes requested):
Author context: Parthipashok04, external contributor. This fix is correct in isolation, but it is already on main. Commit 6fb3ccc0 (fix(recovery): make assess fully archive-aware, #1050) makes the identical three changes: liveness breach count, RISK_OBSERVED scan, and consumed-authorities scan all fold archive_aware_events instead of read_events. I verified the hunks match and origin/main already contains the archive-aware scans with passing tests/test_recovery_engine.py. This PR is also in DIRTY merge state, so it cannot merge as-is. Please close this in favor of the landed fix, or rebase onto current main if you intend to add something beyond 6fb3ccc0. Not merging a duplicate.

src/continuum/recovery/engine.py
```diff
@@ -466,8 +471,13 @@ def assess(
...
         # still deny. A later AUTHORITY_RECONCILED with valid true clears the
         # map inside collect_consumed_authorities.
+        # The scan folds the full history: an archived consumption is still a
+        # consumption (issue #1050). The shared fetch above already fell back to
+        # the live log on archive error, so this except guards the fold itself
+        # rather than swallowing a read failure that would silently empty the
+        # map and drop the block.
         try:
-            consumed_authorities = collect_consumed_authorities(self.storage.read_events(run_id))
+            consumed_authorities = collect_consumed_authorities(archive_aware_events)
```
Cyrax321, line 480: Blocking: main already contains this exact line, and the liveness and risk equivalents above, since #1308 merged the archive-aware assess. After a rebase on current main this hunk disappears entirely, which is the clearest sign the branch is superseded rather than in need of conflict resolution. Landing it anyway would churn the merged comments for zero behavior change.

src/continuum/recovery/engine.py
```diff
@@ -466,8 +471,13 @@ def assess(
...
         except Exception:
             consumed_authorities = {}
```
Cyrax321, line 482: Blocking: this branch still carries the pre-#1066 except body, which swallows any failure into {}, the unblocked answer. Main's version of this block degrades closed: the except sets consumed_authorities = None and an elif consumed_authorities is None: branch escalates to REQUEST_HUMAN, because an unreadable authority ledger must never look unblocked when the gate would still deny. If these conflicts are ever resolved by hand, resolving toward this side would silently revert that fail-closed behavior; the None sentinel and the elif must survive any resolution.

Cyrax321 (changes requested):
This folds the archived prefix into the three remaining live-tail scans in assess, which is exactly what #1050 asked for, but main already landed that fix in #1308 (6fb3ccc0, merged 2026-09-20): all three scans there filter the shared archive_aware_events fetch, and the same three regression tests (test_risk_trigger_survives_compaction, test_liveness_breach_count_survives_compaction, test_consumed_authority_survives_compaction) are already in tests/test_recovery_engine.py.

This branch predates that merge, so a rebase on current main leaves no behavioral diff, and GitHub already reports the PR as conflicting (dirty) across all three files. Note that open PR #1315 targets the same three scans with the same approach, so it is equally superseded. Closing this PR is the right move; if you do resolve the conflicts anyway, the authority except block must keep main's #1066 fail-closed sentinel, as flagged inline.

PR #1328, opened 2026-09-21
fix(storage): PostgresStorage.rebuild_action_index reports the rows it corrected (#1267)

### LINKED ISSUE #1267: storage: PostgresStorage.rebuild_action_index always reports 0 corrected rows, so verify --repair-index lies on Postgres

Adhi1-2:
Fixes #1267.

## The bug

`PostgresStorage.rebuild_action_index` ended in an unconditional `return 0`:

```python
        return 0
```

`Storage` declares the method as *"returns corrected rows"* and `SQLiteStorage` honours that (it counts changed-or-added rows plus rows the rebuild removed). This was the only return-value divergence between the two engines, and the value is not internal — `cmd_verify` prints it verbatim:

```python
fixed = int(rebuild())
```

So on Postgres, a `--repair-index` that repaired real drift reported **"0 row(s) corrected"**. The repair itself works; the report about it lied.

## The fix

Mirror the SQLite engine exactly. Snapshot the stored rows under the write lock and ahead of the `DELETE`, then count keys whose stored row was missing, stale, or spurious:

```python
corrections = sum(1 for k, val in expected.items() if before.get(k) != val)
corrections += len(set(before) - set(expected))
return corrections
```

Reading the snapshot under the lock and before the `DELETE` means the count measures the index the rebuild replaced, rather than one a concurrent writer shifted underneath it.

The count is well defined independently of #1323: `action_index_drift` and the rebuild read the same table over the engine's one autocommit connection with nothing writing between them, so the repair reports exactly the drift that called for it, whichever numbering scale the fold uses. This PR does **not** touch that scale — it only makes the report truthful.

## Tests

Three tests for the three kinds of correction, all verified to fail against the buggy `return 0` (`assert 0 == 2`, etc.) and pass with the fix:

- **stale status** — corrupt one row's status, assert the repair reports the drift that called for it and the row comes back `STARTED`;
- **spurious row** — insert a key the log never authorized, assert an exact `+1` over the baseline and that the rebuild removes it;
- **lost row** — delete an index row, assert the repair reports the drift and restores it.

Two of them deliberately do **not** claim an exact `+1`, and the reason is recorded in the test docstrings: until #1323 lands, the fold numbers a row by its position in the row stream while the incremental writer uses `nextval`, so an action row on this engine already reads as drift at baseline — corrupting or dropping one moves the count by zero, not one. The spurious case is the one form that is a clean `+1` regardless of baseline, because a key absent from the fold cannot be in the baseline drift.

Corruption is scoped to each test's own key. The suite shares one database and the key namespace is store-global, so an unscoped `UPDATE action_index` would corrupt every other test's rows too.

## Gates

`ruff check`, `ruff format --check` (pinned 0.16.5), `mypy`, and `markdownlint` all clean. Full suite green except `tests/test_storage.py::test_postgres_url_is_routed_to_postgres_backend`, which fails identically on `upstream/main` in this environment because `psycopg` is installed locally (CI exercises the Postgres service container separately).

## Out of scope

#1321 and #1322 — the numbering-scale and cross-run ordering gaps in the fold — are tracked by #1323 and its follow-ups. This PR changes no comparison criterion, only the return value.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed PostgreSQL index repair reporting so `continuum verify --index --repair-index` now accurately reports the number of corrected or removed rows.
  - Improved consistency between PostgreSQL and SQLite index verification results.


Adhi1-2:
Fixed in `f62c2ad`. The em dash was at line 711 of the previous head `71207c5`; the current head rewrites that sentence with `--` and has no em dashes anywhere in the diff. The two em dashes that remain elsewhere in CHANGELOG.md (lines 197 and 324) are pre-existing on main and outside this PR's hunks.

Cyrax321 (approved):
Verified the fix. The unconditional return 0 on the base contradicts the Storage contract and the SQLite implementation at sqlite.py:614-643, which counts changed-or-added plus removed rows. This patch mirrors that count exactly, snapshots before under the write lock ahead of the DELETE, and the comparison tuple (updated_seq, status) matches what action_index_drift uses. Three regression tests pin stale, spurious, and lost rows with scoped keys. The prior em dash thread is resolved in the current head. CI is green except codecov/patch, which reads 0 percent only because the Postgres suite never ran on the uploading leg; that gap is exactly what #1330 fixes. Approving.

src/continuum/storage/postgres.py
```diff
@@ -707,7 +724,13 @@ def rebuild_action_index(self) -> int:
                         for key, (entry, seq) in canonical.items()
                     ],
```
Cyrax321, line 728: Non-blocking: this counting block re-derives the same comparison that `action_index_drift` performs just below (expected fold against the stored rows, here with `before` in the stored slot), so the counting rule now lives in two places per engine and any future change to it has to be made twice. If it ever needs to evolve, pulling the snapshot-plus-compare into one shared helper would keep drift and rebuild in lockstep. Matching the SQLite reference verbatim is defensible for this fix, so deferring is fine.

Cyrax321 (dismissed):
This closes the one remaining return-value divergence between the two engines: `PostgresStorage.rebuild_action_index` ended in an unconditional `return 0` while `Storage` declares it returns corrected rows and `cmd_verify` prints that number verbatim, so a real Postgres repair reported success with zero work. The implementation is line-for-line identical to the SQLite reference: snapshot the stored rows under the write lock and ahead of the `DELETE`, then count changed-or-missing keys plus rows the rebuild removed. That parity is the right choice here because it keeps both backends reporting identical numbers by construction.

Non-blocking: this overlaps with #1332, which takes the same approach for the same issue and was approved first, so one of the two should close once the other lands. CI is green on all four required checks; codecov/patch is red but is not one of the branch gates, and the new lines are exercised by the Postgres service-container job. Approving, with one defer-able note inline.

PR #1330, opened 2026-09-21
fix(ci): the Postgres leg uploads its coverage so postgres.py reads as covered (#1329)

### LINKED ISSUE #1329: codecov/patch is permanently red for every postgres.py change (coverage uploads skip the Postgres leg)
## What is broken

`codecov/patch` fails on **every** change to `src/continuum/storage/postgres.py`, reporting `0.00% of diff hit` even when the changed lines are fully exercised by the Postgres contract suite.

Concrete example: PR #1328 (fixes #1267, `PostgresStorage.rebuild_action_index`). Its `codecov/patch` check reports:

```
0.00% of diff hit (target 83.89%)
```

Yet running the suite the gate never runs:

```
$ export CONTINUUM_TEST_POSTGRES_DSN="dbname=continuum_test host=localhost user=adhinans"
$ pytest tests/test_storage_postgres.py --cov=continuum.storage.postgres --cov-report=xml
# rebuild_action_index, lines 695-733: 10/10 lines covered
```

## Why

`.github/workflows/ci.yml` uploads coverage from exactly one matrix leg:

```yaml
      - name: Upload coverage report
        uses: codecov/codecov-action@v7.0.0
        if: matrix.python-version == '3.12' && runner.os == 'Linux'
        with:
          files: coverage.xml
```

That leg runs the full suite **without** `CONTINUUM_TEST_POSTGRES_DSN`, so `tests/test_storage_postgres.py` skips wholesale (`pytestmark = pytest.mark.skipif(DSN is None or not _psycopg_available(), ...)`).

The `Test (Postgres backend)` job is the only place the suite actually executes:

```yaml
      - name: Run Postgres contract tests
        run: pytest tests/test_storage_postgres.py tests/test_action_index.py --tb=short -q
```

...and it uploads no coverage at all. So the lines in `postgres.py` are measured by a job that never runs them and exercised by a job that never reports them.

## This is not new

The last `postgres.py` change before #1328 — `9e20e30`, PR #1078, already merged — carries the identical `codecov/patch | failure`. The gate has been permanently red for the backend since it was introduced, which also means the signal it is supposed to provide (a real coverage regression on `postgres.py`) is indistinguishable from the noise.

## Suggested fix

Give the Postgres backend job a coverage upload, e.g. run it with `--cov=continuum --cov-report=xml` and upload `coverage.xml` with `codecov-action`. Codecov merges multiple uploads per commit, so the backend's numbers fold into the same patch/total figures as the Linux/3.12 leg, and `codecov/patch` starts reporting the real coverage of `postgres.py` changes.

Dependabot PR #1278 bumps the action to 7.1.0, but that does not address which leg uploads.

## Impact

Every contributor touching the Postgres backend sees a red `codecov/patch` on an otherwise-green PR, and the check cannot distinguish a genuine coverage regression from the ambient failure.

Adhi1-2:
## What

Fixes #1329.

`codecov/patch` failed on **every** change to `src/continuum/storage/postgres.py`, reporting `0.00% of diff hit` on otherwise-green PRs. The check has been permanently red for the backend since it landed — the last `postgres.py` change before this one (`9e20e30`, PR #1078, merged) carries the identical failure.

## Why

Coverage is uploaded from exactly one matrix leg:

```yaml
      - name: Upload coverage report
        uses: codecov/codecov-action@v7.0.0
        if: matrix.python-version == '3.12' && runner.os == 'Linux'
```

That leg runs the suite **without** `CONTINUUM_TEST_POSTGRES_DSN`, so `tests/test_storage_postgres.py` skips wholesale. The `Test (Postgres backend)` job is the only place the suite actually executes, and it uploaded nothing — so the lines were measured by a job that never ran them and exercised by a job that never reported them.

## The change

The Postgres job now runs under `--cov` and uploads its own report:

```yaml
        run: pytest tests/test_storage_postgres.py tests/test_action_index.py
          --tb=short -q --cov=continuum.storage.postgres --cov-report=xml --cov-report=term-missing

      - name: Upload coverage report
        uses: codecov/codecov-action@v7.0.0
        with:
          files: coverage.xml
          fail_ci_if_error: false
```

Coverage is scoped to `continuum.storage.postgres`, the module this job exercises, rather than the whole package. Codecov merges multiple uploads for the same commit rather than overriding them, so this only adds line data the other leg lacks instead of asserting low coverage on modules these two test files never touch.

`fail_ci_if_error: false` matches the existing upload step, so a Codecov outage cannot turn the backend job red.

## Verification

The exact command runs clean against a local Postgres 16:

```
$ pytest tests/test_storage_postgres.py tests/test_action_index.py \
    --tb=short -q --cov=continuum.storage.postgres --cov-report=xml --cov-report=term-missing
.......................                                            [100%]
Name                                Stmts   Miss Branch BrPart  Cover
src/continuum/storage/postgres.py     423    114    110     14    68%
Coverage XML written to file coverage.xml
```

The XML records the changed lines as hit — on the earlier PR #1328, `rebuild_action_index` (lines 695-733) came back 10/10 covered once the suite actually ran, which is precisely the data the gate was missing.

YAML parses, `markdownlint-cli2` reports 0 issues on the CHANGELOG entry, and the entry adds no em dashes.


Cyrax321 (approved):
Verified that the Postgres backend job is the only leg that executes the Postgres suite while the Linux 3.12 matrix leg that uploads coverage skips it without CONTINUUM_TEST_POSTGRES_DSN. Scoping the new upload to continuum.storage.postgres is correct since Codecov merges per-commit reports, and fail_ci_if_error false matches the existing step. CI is fully green including codecov/patch. Approving.

Cyrax321 (dismissed):
The fix targets the actual gap: the matrix upload runs on the Linux/3.12 leg, which runs without CONTINUUM_TEST_POSTGRES_DSN and therefore skips test_storage_postgres.py, while the Postgres job that exercises postgres.py uploaded nothing. Scoping this job's report to continuum.storage.postgres and relying on Codecov per-commit merging is the right shape, fail_ci_if_error matches the existing step, and the new upload does not need its own if because the job-level docs-only guard already covers it. Live signal agrees: codecov/patch is green on this PR alongside every required check. Nothing blocks the merge.

PR #1331, opened 2026-09-21
docs: adapters.md documents the real constructor signatures for the framework adapters (#1230)

### LINKED ISSUE #1230: docs: adapters.md documents an engine= keyword that the LangGraph, OpenAI and LangChain adapters reject

soorajsuru2004-sudo:
## What

`docs/api/adapters.md` taught one constructor shape for every adapter:

```
LangGraphAgentAdapter(storage, *, engine=None)
OpenAIAgentAdapter(storage, *, engine=None)
LangChainAgentAdapter(storage, *, engine=None)
```

None of the three accepts `engine`. Following the doc raised `TypeError: __init__() got an unexpected keyword argument 'engine'` while the same file's `GenericAgentAdapter` entry promised it worked. The entries also omitted `graph` and `state_to_semantic`, the primary customization points on the LangGraph adapter and the only way to point it at a compiled graph.

## Changes (docs only)

Corrected the three entries to the real signatures and added a sentence to each that it builds its own `RecoveryEngine` internally, so an injected engine is not among its options. That is the fact of the code today; whether it should change is a separate design question and is left alone here.

- `langgraph.py:125` → `(storage, *, graph=None, state_to_semantic=None, auto_file=None, auto_total=None)`
- `openai.py:146` → `(storage, *, state_to_semantic=None, auto_file=None, auto_total=None)`
- `langchain.py:121` → `(storage, *, state_to_semantic=None, auto_file=None, auto_total=None)`

The Generic, Browser, Container, Kubernetes, Filesystem and PythonInProc entries already matched and were left alone.

## Verification

Wrote a throwaway script comparing every documented constructor line in the file against `inspect.signature` on the class (parameter names, kinds and defaults — the doc deliberately omits annotations). All nine adapters now match exactly; the script was not committed, since the issue scopes this as docs-only. `tests/test_mcp_docs.py` still passes.

Closes #1230.

## Summary by CodeRabbit

* **Documentation**
  * Updated framework adapter documentation with revised constructor options, including graph state extraction and automatic file/total customization settings.
  * Clarified that adapters create their own recovery engine rather than accepting an injected engine.
  * Refreshed Horizon benchmark generation timestamps in the README and benchmark reference; reported results remain unchanged.


Cyrax321 (changes requested):
The three constructor signatures are now correct against `src/continuum/adapters/{langgraph,openai,langchain}.py`, and the "no `engine`" note is accurate: `GenericAgentAdapter.__init__` accepts `engine`, but these three forward only `storage`/`auto_file`/`auto_total` to `super().__init__`, so the engine always defaults to `RecoveryEngine(storage)`.

Two blocking issues, both on `docs/api/adapters.md`.

The line at 55-58 says `graph` "is the `StateGraph` the adapter checkpoints against" and calls it a primary customization point. That is not what the code does. `self.graph` is assigned in `LangGraphAgentAdapter.__init__` and then read nowhere: `checkpoint_node(state)` takes the state dict as its argument and never touches `self.graph`. The class docstring is already wrong about this ("Used by `checkpoint_node` to read graph state"), and this PR propagates the mistake into the reference doc. Either drop the `graph` claim to a one-line "accepted for convenience and stored as `self.graph`; the checkpoint node reads the state dict you pass it," or wire `self.graph` into `checkpoint_node`. I would rather have the parameter removed than have the docs promise behavior the engine does not have.

The regenerated benchmark block in `README.md` and `references/bench.md` is unrelated to this PR's stated purpose, which is adapter signatures. It also regenerates to a different value on the next nightly run, so it will re-conflict with main for everyone who rebases after you. Please drop both `<!-- BENCH -->` hunks and leave the generated timestamp alone.

docs/api/adapters.md
```diff
@@ -50,9 +50,14 @@ changing anything. The decision's `mode` is one of the seven `RecoveryMode` valu
...
 ## LangGraphAgentAdapter
 
-`continuum.adapters.LangGraphAgentAdapter(storage, *, engine=None)`
+`continuum.adapters.LangGraphAgentAdapter(storage, *, graph=None, state_to_semantic=None, auto_file=None, auto_total=None)`
 
-Subclass of `GenericAgentAdapter` for LangGraph `StateGraph` workflows. Adds:
+Subclass of `GenericAgentAdapter` for LangGraph `StateGraph` workflows. `graph`
+is the `StateGraph` the adapter checkpoints against, and `state_to_semantic`
+overrides the default extractor that reads `continuum_run_id` and `goal` from a
+graph state dict; the two are the primary customization points on this adapter.
```
Cyrax321, line 58: Blocking. `self.graph` is assigned in `LangGraphAgentAdapter.__init__` and read nowhere in `src/continuum/adapters/langgraph.py`. `checkpoint_node(state)` receives the state dict as its argument and never consults `self.graph`, so the adapter does not checkpoint against the graph object. The class docstring is already wrong about this and this PR carries the same mistake into the reference doc. Either describe the parameter as stored-only, or wire it into `checkpoint_node`.

references/bench.md
```diff
@@ -101,7 +101,7 @@ in every scenario.
 <!-- BENCH:START -->
 ### Latest nightly results (real runs, no invented numbers)
 
-Generated: 2026-09-15T10:43:15.997718  Horizon scenarios: 5  Passed: 4  Failed: 1
+Generated: 2026-09-21T10:35:18.108875  Horizon scenarios: 5  Passed: 4  Failed: 1
```
Cyrax321, line 104: Non-blocking. This regenerated benchmark block is unrelated to the PR's stated purpose, which is adapter constructor signatures. It also re-diffs on every nightly run, so anyone who rebases after you will see it conflict with main again. Please drop both the README.md and references/bench.md BENCH hunks and leave the generated timestamp alone.

README.md
```diff
@@ -240,7 +240,7 @@ CONTINUUM is verified against real LLM agents, live protocol boundaries, and har
 <!-- BENCH:START -->
 ### Horizon-scale benchmark (real runs, no invented numbers)
 
-Generated: 2026-09-15T10:43:15.997718  Horizon scenarios: 5  Passed: 4  Failed: 1
+Generated: 2026-09-21T10:35:18.108875  Horizon scenarios: 5  Passed: 4  Failed: 1
```
Cyrax321, line 243: Non-blocking. Both BENCH tables (this one and references/bench.md) are regenerated automatically by bench-nightly.yml, which commits its own timestamp refresh every night, and this PR is scoped to the adapter constructor docs. Unless the 2026-09-21 timestamp came from an actual `benchmarks/run.py --publish` run before this push, the updated line advertises a benchmark generation that may not have happened, directly under the section's own 'real runs, no invented numbers' banner. Please drop README.md and references/bench.md from this PR, or note in the PR body that the suite was re-run locally.

Cyrax321 (commented):
The three corrected signatures in docs/api/adapters.md match the code on main exactly: langgraph.py `(storage, *, graph=None, state_to_semantic=None, auto_file=None, auto_total=None)`, openai.py and langchain.py `(storage, *, state_to_semantic=None, auto_file=None, auto_total=None)`. The new prose is also accurate: the default extractors do read `continuum_run_id` and `goal` from the state dict, the OpenAI adapter does fall back to `ContinuumContext.to_semantic_state()`, and since the subclasses drop the `engine` parameter the parent constructor builds `RecoveryEngine(storage)` internally when none is injected.

One non-blocking scope issue, noted inline: the PR also bumps the bench table timestamps in README.md and references/bench.md, which are managed by the nightly bench workflow.

Heads-up that #1242 is another open PR fixing the same issue (#1230), so one of the two should be closed once a direction is picked.

PR #1332, opened 2026-09-21
fix(storage): report corrected rows from PostgresStorage.rebuild_action_index (#1267)

### LINKED ISSUE #1267: storage: PostgresStorage.rebuild_action_index always reports 0 corrected rows, so verify --repair-index lies on Postgres

soorajsuru2004-sudo:
## Description

`PostgresStorage.rebuild_action_index` ended in an unconditional `return 0` instead of calculating and returning the number of corrected rows. In contrast, the `Storage` base class specifies that the method "recomputes the index from the log; returns corrected rows", and `SQLiteStorage` honours this contract.

Because `continuum verify --index --repair-index` surfaces this figure directly to the operator (`action index repaired from the log ({fixed} row(s) corrected)`), running verify with index repair on PostgreSQL always reported 0 rows corrected, even after repairing drifted or corrupted index rows.

This change:
- Queries the stored `action_index` before rebuilding to capture pre-rebuild keys, sequences, and statuses.
- Compares the pre-rebuild state against the canonical fold to count changed, missing, and spurious keys, matching the calculation performed by `SQLiteStorage`.
- Returns the computed correction count and updates the docstring to align with the base class contract.
- Updates the existing contract test in `tests/test_storage_postgres.py` to assert that `rebuild_action_index` returns the drift count.
- Adds a unit regression test in `tests/test_action_index.py` verifying that missing, changed, and spurious row corrections (as well as 0-correction clean state) are accurately calculated.

## Related issues

Fixes #1267

## Types of changes

- Type: `bugfix`

## Breaking changes

No. The index rebuild logic and resulting database rows are unchanged; only the return value is corrected to report the number of repaired rows.

## How has this been tested?

1. Regression test verified failing on unpatched code and passing on patched code:
   ```
   tests/test_action_index.py::test_postgres_rebuild_action_index_reports_corrected_count FAILED (assert 0 == 3) -> PASSED
   ```
2. Unit and storage test suites:
   ```
   pytest tests/test_action_index.py tests/test_storage.py -q
   ```
3. Documentation counts and precommit configuration guards:
   ```
   pytest tests/test_docs_counts.py tests/test_precommit_config.py
   ```
4. Full test suite:
   ```
   pytest -q
   ```
5. Code style and type checks:
   ```
   ruff check src/ tests/
   mypy src/continuum
   ```

## Checklist

## Additional context

The regression test in `tests/test_action_index.py` uses a mock connection to verify the row-counting logic in environments without a live PostgreSQL instance, while `test_storage_postgres.py` verifies the contract when `CONTINUUM_TEST_POSTGRES_DSN` and `psycopg` are available in CI.


yuki-fuyutsuki:
**Code Review**

This PR fixes `PostgresStorage.rebuild_action_index` (`src/continuum/storage/postgres.py:698`) which unconditionally returned `0`, breaking the `Storage` contract ("returns corrected rows") and causing `continuum verify --index --repair-index` to always report 0 corrections on Postgres. The change snapshots `before` state inside the write lock, rebuilds from the canonical fold, and counts missing/changed/spurious keys identically to `SQLiteStorage`, with matching docstring and regression tests. Implementation is correct and well-tested; one durability gap is carried over from the previous code.

**medium**
`src/continuum/storage/postgres.py:707`
The rebuild is not atomic. The connection runs `autocommit=True` (`src/continuum/storage/postgres.py:196`), and the method uses `with self._write():` (lock only) for `DELETE` + `executemany INSERT`. Each statement commits separately, so a failure between them leaves `action_index` empty/corrupt and concurrent readers see an empty index. `SQLiteStorage._write` is transactional and `PostgresStorage.compact_run` uses `with self._write(), self._connection.transaction():`. The rebuild should do the same.

```suggestion
        canonical = self._canonical_index_rows()
        with self._write(), self._connection.transaction():
            before = {
```

**low**
`src/continuum/storage/postgres.py:706`
`canonical = self._canonical_index_rows()` is computed before acquiring `_write`, while `before` is snapshotted inside it. A concurrent `ACTION_*` append between the two sees a stale canonical fold, so the returned correction count can be off by one and the rebuild can overwrite a just-inserted row. `SQLiteStorage` has the same TOCTOU, but moving the fold inside the write transaction (or holding the lock across both) would make the count exact.



[github run](/Cyrax321/CONTINUUM/actions/runs/35632770519)

Cyrax321 (approved):
This fixes PostgresStorage.rebuild_action_index returning an unconditional 0 by snapshotting the stored index inside the write lock before the delete and counting changed, missing, and spurious keys against the canonical fold, which is exactly the computation SQLite already performs and the one action_index_drift uses on this same backend. The delta semantics line up with drift, so the verify CLI now prints a count that matches the drift it just measured, and the updated contract test asserts that equality directly. The mock-based unit test covers the arithmetic without a live Postgres while the integration test covers the real path in CI. Nothing blocks the merge; all checks are green.

